PR #3930, opened 2013-01-27
fix #2339: set unicode-bidi property to prevent username to overlay the pubdate [ci skip]

divout:
#2339

before:after
diaspora-rtl-2339


SafaAlfulaij:
Well, That's wrong. The arabic text got flipped. It's "طه" not "هط".
A better workaround is to add the LRM before the string, And that's in LTR layout only.

PR #4142, opened 2013-05-04
Overflow username service to the publisher - Only develop branch

movilla:
I still working... :construction: 
# Before

pantallazo-1
# After

pantallazo


svbergerem:
There is no blueprint publisher anymore.

Issue #4288, opened 2013-07-07
Round user avatar


Flaburgan:
I saw that in the settings, where you can upload a new profile picture, your avatar is displayed rounded (<code>border-radius: 5px; box-shadow: 0px 1px 2px #666;</code>)

I wonder if it's a good idea to display it like that everywhere.

This how it can looks:

### before
![before](https://f.cloud.github.com/assets/930064/758667/4d2f3e6e-e747-11e2-8795-a93a1294eeaa.png)

### after
![after](https://f.cloud.github.com/assets/930064/758668/52dcbca6-e747-11e2-808b-7312d10abb8c.png)

What's your opinion about that?

svbergerem:
Implemented by #5733.

PR #4296, opened 2013-07-11
#4288 - implementation of a round user avatar

juliaguar:
just a quick design change to implement a round user avatar (as mentioned in issue #4288).


svbergerem:
Superseded by #5733. Thank you for your work.

PR #4456, opened 2013-08-28
Changing icons on regular site to use Entypo

ssrihari:
- Changed the sidebar icons and the header Icons
- It looks bearable

Downsides:
- The characters/icons in the font aren't of equal dimensions, so they don't go well together
- Some icons in glyphicons aren't here in entypo


denschub:
Implemented with 0.5.0.0

PR #4497, opened 2013-09-02
Notification bug

manuels:
Here are some tests that trigger bug #4496 and might help debugging it.


denschub:
I'm closing this PR. We refactored almost all of the involved code during the last few releases and there are no new reports.

PR #4539, opened 2013-09-19
[WIP] Add design to 404 not public file

Flaburgan:
There is two 404 files, 404.html and  app/views/errors/not_public.haml which is used to display a 404 when the user is logged but doesn't have access to the post. 404.html was refactored in #4079 

This PR refactor not_public.haml the say way and link directly to 404.css for the design. This direct link is not the best approach but it's the easiest way I found to factorize css. Any opinion anyone?


svbergerem:
@jhass Did #5817 supersede this PR?

jhass:
I'm not sure but I don't think so.

Flaburgan:
Looks like it did.

PR #4596, opened 2013-11-17
Allow additional profile fields (previously private profile) to be set public

jaywink:
Change wording of profile sections to Basic and Additional. Add helper icons and rollover text to hint on visibility.

NOTE! Not ready for merge - some work to do on help pages and I haven't even looked at tests yet. Just here for initial review, whether this is even wanted (needed it myself).

See for example change running on my pod, my now public profile: https://iliketoast.net/u/jaywink

In the below screenshots, I realize now it's confusing that the globe and the padlock are not visible at the same time. The way it goes, you click the padlock in the Additional section and it turns into the globe, and if you click the globe, it turns to a padlock. The other two icons (globe for Basic data and padlock for Settings) are not clickable.

I didn't use a toggle slider as Flaburgan suggested since D\* doesn't have any in use yet and I feel visual makeovers should be done in a separate stage.

Edit profile page - Basic (old Public Profile) and Additional (old Private Profile) sections visible, with additional set to public (the globe, by default it is limited = padlock, for new and existing profiles).

pic

The lower part, didn't have a title before and thus was confusingly mixed with Public Profile - now titled Profile Settings, with padlock to show it's not public data.

pic

Hint rollover in Additional section - wording is not very good at the moment..

pic

### DIFF
--- a/app/views/profiles/_edit.haml
+++ b/app/views/profiles/_edit.haml
@@ -9,10 +9,22 @@
   = render 'profiles/edit_public', :profile => profile, :aspect => aspect, :person => person
 
   %hr
+  :javascript
+    $(document).ready(function () {
+      $('.additional-visibility-icon').click(function() {
+        $('input#profile_is_public').prop('checked', ! $('input#profile_is_public').prop('checked'));
+        $('.additional-visibility-icon').toggle();
+      });
+    });
+  %h3{:class=>"inline"}
+    = t('profiles.edit.additional')
+  = image_tag 'icons/globe.png', class: 'additional-visibility-icon', style: ('display:none;' if ! profile.is_public)
+  = image_tag 'icons/padlock-closed.png', class: 'additional-visibility-icon', style: ('display:none;' if profile.is_public)
+  %span{ :rel => 'facebox', :title => t('profiles.edit.additional_hint') }
+    = image_tag 'icons/monotone_question.png'
+  = check_box_tag 'profile[is_public]', true, profile.is_public
   %br
-  %h3
-    = t('profiles.edit.your_private_profile')
-
+  %br
 
   %h4
     = t('profiles.edit.your_bio')
@@ -40,7 +52,15 @@
 
   %br
   %br
+  
+  %hr
+  %h3{:class=>"inline"}
+    = t('profiles.edit.settings')
+  = image_tag 'icons/padlock-closed.png'
 
+  %br
+  %br
+  
   %h4
     = t('search')
 
--- a/app/assets/stylesheets/profile.css.scss
+++ b/app/assets/stylesheets/profile.css.scss
@@ -123,3 +123,12 @@
     }
   }
 }
+
+#update_profile_form {
+  .additional-visibility-icon {
+    cursor: pointer;
+  }
+  #profile_is_public {
+    display: none;
+  }
+}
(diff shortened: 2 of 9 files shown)


Flaburgan:
The settings page has now be ported to bootstrap, the rebase will probably be painful...

jaywink:
Meh I thought this version of the pull had already been closed :) I don't think a rebase is a good idea - I'll just re-implement the changes on top of develop, as soon as I have time. Prob not in 0.5 sorry :(

I'll close this not-to-be-merged old version for now, this feature is in my own to-do list :)

app/views/profiles/_edit.haml
```diff
@@ -9,10 +9,22 @@
   = render 'profiles/edit_public', :profile => profile, :aspect => aspect, :person => person
 
   %hr
+  :javascript
```
cmrd-senya: What is a proper place for javascript for this page in the actual diaspora codebase? I tried app/assets/javascript/app/pages/profile.js, but it seems that it doesn't get loaded on profile/edit page.

svbergerem: You will have to define that in app/assets/javascripts/app/router.js. It might be ok to add this to the settings page view (app/assets/javascripts/app/pages/settings.js) but you will definitely have to add the correct route in router.js.
```js
...
 "profile/edit": "settings",
...
```

cmrd-senya: I wrote an event in Backbone.View for Settings page for .additional-visibility-icon class, but for some reason it doesn't get called on click on image tag with the class set. Instead I get in the javascript console:
15:34:51.812 An empty string can not be an argument of the method getElementById().1 jquery.js:2823:0


Could somebody please check what is wrong?
Here is my changes: http://pastebin.com/LGCt0Mft

Or at least point me what to check.

theworldbright: What does `this.$('input#profile_is_public')` match?

cmrd-senya: It matches "= check_box_tag 'profile[is_public]', true, profile.is_public", line 64 of paste.

It doesn't really matter, because switchAdditionalVisibility function doesn't get called at all.

theworldbright: If it doesn't matter fair enough. I just thought it was weird that `profile_is_public` doesn't have a matching element. I assume `profile[is_public]` becomes `id: profile_is_public`?

theworldbright: By switchAdditionalVisibility doesn't get called at all, do you mean like if you add an alert statement right after `switchAdditionalVisibility: function() {`, line 14, it won't be called at all?

cmrd-senya: Yes, it becomes:  
<input name="profile[is_public]" id="profile_is_public" value="true" type="checkbox">

Yes, that what I mean.

theworldbright: I'm guessing that `.settings_visibility` does not exist in `profile/edit` and it only exists under `user/edit`

cmrd-senya: I tried to comment it out, but it didn't help.

theworldbright: Hmm, it may be that there is no "el"?

cmrd-senya: Yes, that seems to be the reason. Thanks!

PR #4631, opened 2013-12-18
WIP - Refactor users#update action

oliverbarnes:
This is a partial WIP of the refactor asked for on #3664. After a while sitting on it, it's clear I won't be able to finish it in a timely fashion, as I'm too caught up with my day job. Hopefully this can be a good starting point for somebody else to adopt and continue?

The approach I've taken is to extract logic branches into small discreet controllers. Two so far: User::EmailPreferencesController and User::PasswordController. These also map to separate forms on the user edit page. 

Did the same to their specs. Haven't touched the cucumber features yet.

### DIFF
--- a/app/controllers/users/email_preferences_controller.rb
+++ b/app/controllers/users/email_preferences_controller.rb
@@ -1,0 +1,25 @@
+class Users::EmailPreferencesController < ApplicationController
+  before_filter :authenticate_user!
+
+  def update
+    current_user.update_user_preferences( user_params[:email_preferences] )
+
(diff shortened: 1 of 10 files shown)


denschub:
Closing this PR. This now will result in a massive list of merge conflicts. It has been orphaned for more than a year and old pull requests look kinda sad. Feel free to reopen this if there is a need to.

cmrd-senya:
Does it worth adopting now?

denschub:
Not really. While cleaning up is a nice idea, we should probably build Services now.

cmrd-senya:
Then #3664 should be updated probably.

denschub:
You are very right! Done.

PR #4742, opened 2014-01-25
[wip] attempting to test the 'after_commit' hook

Raven24:
this comes in addition to the fix introduced for #4715, now featuring DatabaseCleaner instead of transactional fixtures in Rspec.
the really annoying thing is, that the `after_commit` hook swallows exceptions silently, so the `stack level too deep` error would not have been noticeable immediately with this either.
I was able to get some other vaguely related tests to fail, though. Also, what would have been noticeable is an extremely long runtime for the `stack level` cases, since it simply takes a while to fill up the stack..

Also, currently there are a few tests that don't seem to like these changes, so I would hope for any help I can get to finish this and get more dependable specs for the AR hooks.


denschub:
Mhm. Almost two years now, I guess it's save to close this. The orphan label remains if someone feels strong about working on old, unfinished changes...

PR #4860, opened 2014-03-17
Added Emoticon support for diaspora

abhineet08:
Hi, this pull request adds support in diaspora for Emoji 
(all emoticons mentioned here :: http://www.emoji-cheat-sheet.com )

And, it also adds a rake task which can be called by
rake emoji:refresh
which updates the list of supported emoticons by adding any extra emoticons which are added to emoji.


dimaursu:
#5530 Fixes the issues with 4 byte characters, that are required to have unicode emojis.

svbergerem:
[On Loomio we decided](https://www.loomio.org/d/vlmyMTNd/emoji-support?proposal=CRoUdzEv) that we don't want to replace emoji characters with images.

PR #4887, opened 2014-03-31
Add box shadow and radius to avatars

goobertron:
I decided to pick up #4296 as it seemed to have been abandoned.

I realise there's quite a lot going on with porting to Bootstrap, and various design changes, so perhaps it's not a good time to merge this. But I wanted to do this work anyway, just as an exercise to help me learn, so here it is.

In #4288 most of us liked rounded corners on avatars, but they didn't fit in with the new profile sidebar in #4374, so that idea was dropped. What I've done on the profile pic on people views is to add a radius of 8px (same as on the bottom of the interaction icons bar) to the top corners only, with no radius on the bottom of the avatar.

See what you think.


svbergerem:
Superseded by #5733. Thank you for your work.

Issue #4269, opened 2013-06-29
Comments are out of order


polsvoice:
I notice that, many times, comments on posts from JoinDiaspora show up out of order on Diasp.org (or are missing!) This makes it hard to follow conversations. Here is an example from [a recent post by Sean Tilley.](https://diasp.org/posts/2065030#)
Here's how the comments look on JoinDiaspora:
![diaspora_screenshot3](https://f.cloud.github.com/assets/941129/726565/68cced52-e10b-11e2-9e0a-ece723c71d21.jpg)
On diasp.org:
![diaspora_screenshot4](https://f.cloud.github.com/assets/941129/726566/68dfc2a6-e10b-11e2-9871-c9a3ec94497a.jpg)
May be related to [this issue?](https://github.com/diaspora/diaspora/issues/3350)
##

Flaburgan:
@JannikStreek any news?

cmrd-senya:
Won't #6406 resolve it? It sorts comments by the time stamps that presented in the DOM generated for every comment model.

cmrd-senya:
Oh, I see, the time stamps are probably different across pods, that's the reason.

SuperTux88:
Fixed with #7436

PR #4935, opened 2014-04-18
Bug 4269: Created_at federation vs ordering issues

JannikStreek:
The comments are now shipping with the created_at field. I had to change the generator to set the field, otherwise the signature verification will fail. Do you want to test the ordering issues first with the comments or shall I start implementing the changes also for posts and other relayables?

refs https://github.com/diaspora/diaspora/issues/4269

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -19,6 +19,9 @@ class Comment < ActiveRecord::Base
   xml_attr :text
   xml_attr :diaspora_handle
 
+  #Don't name it remote_create_at, otherwise it won't work due to some mysterious reasons
+  xml_attr :remote_created
+
   belongs_to :commentable, :touch => true, :polymorphic => true
   alias_attribute :post, :commentable
   belongs_to :author, :class_name => 'Person'
@@ -55,6 +58,14 @@ class Comment < ActiveRecord::Base
 
   def diaspora_handle= nh
     self.author = Webfinger.new(nh).fetch
+  end
+
+  def remote_created
+    Time.zone.parse(self.created_at.to_s).to_i.to_s unless(self.created_at.nil?)
+  end
+
+  def remote_created= date
+    self.created_at = Time.at(date.to_i).utc.to_datetime unless date.blank?
   end
 
   def notification_type(user, person)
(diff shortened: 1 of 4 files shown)


jaywink:
If the signing mechanism really breaks due to fields added or removed then I think the problem is our signing mechanism :( The signing mechanism should ideally just be a wrapper to contain any kind of message, but not being a crypto expert not sure why it isn't so..

Anyway, we shouldn't break backwards compat though we will have to fix this at some point. When we have the federation extracted, what about falling back to several versions? This would cause extra queries first to get the version of the recipient pod though and could make things in general heavier.

jhass:
The problem is simply that our current signing mechanism reconstructs the signable string from its world view instead of the actual attributes retrieved. That is the model/code dictates which attributes in which order are included while the message should dictate that.

jaywink:
OK now I get it. Yeah, well, there is no way we can break something like this even in a major release - it would break the network for a huge amount of pods for quite a while.. We could span this change over several major versions to make it less painful..

I guess the slap itself should specify the schema definition in the possible new improved signing version, [here](https://github.com/Raven24/diaspora-federation/blob/master/lib/diaspora-federation/salmon/slap.rb#L10)ish? That would save always delivering the attributes with the messages - the schema could just be retrieved on demand.

jhass:
Well, one option is to just iterate all attributes received, the other is to add an attribute that defines a list of attributes to include, yeah.

cmrd-senya:
@jhass, am I right, that at [this](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/parser.rb#L13) point all unknown attributes from xml for a model is dropped, so we have to check the signature here?

jhass:
@cmrd-senya I don't know the specifics anymore, but looks like it yes. Whether moving the signature check before it leads to anything sane, I don't know. Other (backwards incompatible) possibilities would be to have the message provide the signature token in plaintext, use a hash over a defined general part of the XML as signature token or define which XML tags go into it in which order explicitly.

cmrd-senya:
I'm ready to help @SuperTux88 in that.

Flaburgan:
@cmrd-senya would be awesome to that solved <3

denschub:
I'll close this PR. Work is to be done in the diaspora_federation gem, so the code in this PR is obsolete. #4269 exists as a tracking bug.

PR #4944, opened 2014-04-25
Missing transaction in the StatusMessage controller

nviennot:
The status message is created and then the aspect visibilities are created. If the server dies in the middle, the database is left in a inconsistent state. We have to execute this controller in a transaction.

The diff looks a little bigger than it should. Just two lines were added.


denschub:
Closing. We did not decide if we actually need a transaction and this PR is more than a year old.

PR #4952, opened 2014-05-06
Fix for #4947

hpetru:
Refactor view and get contacts from aspects if is current user profile
#4947


Flaburgan:
#4947 is now closed because of the port to bootstrap, I guess this can be closed? Are the old files dropped?

jhass:
I guess...

PR #4990, opened 2014-06-07
Feature/4923 spv post deletion

jbelanich:
Fixes issue 4923 by adding a 'delete post' button to the SPV if the author is logged in, and both a 'hide post' and 'ignore user' button if the user is not the author. 'Hide post', 'remove post', and 'ignore user' redirect the user to their stream.

'Ignore user' uses a different icon than the one shown on posts in the stream. I couldn't find the icon of an appropriate size, so I used entypo's mute icon.

Added some tests in spec/javascripts/app/views/single_post_actions_view_spec.js for testing the UI elements.

del-post
hide-post


jaywink:
Thanks for your work but this is superseded by #5547 which was just merged in. Thanks anyway!

PR #5022, opened 2014-06-24
remove html from emails, as it is driving me crazy

maxwell:
This removes (to the best of my knowlage) HTML from email bodies, which seem to be getting passed in raw.


denschub:
Closing. Way too many changes happend in the meantime. If this is still an issue, please start over.

Issue #4661, opened 2014-01-03
Hovercards exceed the right border of the page.


Kalabasa:
Pop-ups exceed the right border of the page, requiring me to scroll horizontally to see.

![screenshot - friday 03 january 2014 - 05 44 30 pht](https://f.cloud.github.com/assets/3705081/1838439/5b812f2a-745c-11e3-9c7d-2c1f1f966de2.png)

PR #5162, opened 2014-08-26
Fixes issue 4661

khall:
https://github.com/diaspora/diaspora/issues/4661

If the right side of the hovercard goes off the screen then pull it to the right. The magic number of 50 is to account for the scroll bar and to give a little padding so the hovercard isn't right against the edge of the screen. Specs pass.


denschub:
Mostly fixed by our bootstrap port and other PRs in the meantime.

Flaburgan:
@denschub so the issue itself should also be closed?

PR #5217, opened 2014-09-13
[WIP] Global hydra

jhass:
Try reducing sidekiq/typhoeus/http_multi memory usage
- Only fetch needed attributes of persons
- Discard person objects as early as possible
- Use only one global hydra instance
- Limit number of cached connections

While being highly experimental, this seems to run stable on my pod for the last two days (in combination with #5209).

Only ever allocating one hydra and limiting the number of connections cached by it drastically reduces and stabilizes memory usage for usage patterns on my pod, I'm stable under 300M RSS now. Though this has the potential to totally wreck outgoing federation performance on different (= higher) load patterns. It may also block Sidekiq workers longer than usual and thus decrease overall Sidekiq throughput. If this goes in, the settings for `sidekiq_concurrency`, `typhoeus_concurrency` and the new `typhoeus_connects` likely need adjusted defaults. Figuring out sensible defaults for these is the hard part of this PR. Reworking the queue priorities in https://github.com/diaspora/diaspora/blob/develop/config/initializers/sidekiq.rb#L20 might also be needed. That said the current defaults are still active on my pod and work good there and I'd still like to target such small instances with the defaults.

Another possible optimization is detecting if one Sidekiq worker is already driving the hydra and not blocking additional ones after they submitted the additional requests. This requires figuring out a reliable way of detecting whether the hydra is already running after the new requests have been queued. That would also requires a complete rework of the retry logic, so I'd like to avoid it if doable.

I'm not sure what more testing could be done from my side since I have no other setups available.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -304,12 +304,10 @@ class Person < ActiveRecord::Base
   end
 
   # Update an array of people given a url, and set it as the new destination_url
-  # @param people [Array<People>]
+  # @param people_ids [Array<Integer>]
   # @param url [String]
-  def self.url_batch_update(people, url)
-    people.each do |person|
-      person.update_url(url)
-    end
+  def self.url_batch_update(people_ids, url)
+    where(id: people_ids).update_all(url: sanitize_url(url))
   end
 
   #gross method pulled out from controller, not exactly sure how it should be used.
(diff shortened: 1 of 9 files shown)


denschub:
So... I'm going to close this. We're in the progress of building a new federation core, so this is something that would have to be done there anyway.

PR #5332, opened 2014-10-11
Using character (&times;) instead of asset for publisher close button

roymckenzie:
Using a character instead of an image asset for the close button on the publisher. The character matches the screen resolution and looks better! 

**Before**
screen shot 2014-10-10 at 8 04 45 pm

**After**
screen shot 2014-10-10 at 8 05 38 pm

For those of us using high resolution screens our eyes will thanks us!

### DIFF
--- a/app/assets/stylesheets/application.css.sass
+++ b/app/assets/stylesheets/application.css.sass
@@ -232,11 +232,15 @@ ul.as-selections
       :width 14px
 
   .delete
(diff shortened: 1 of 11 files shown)


svbergerem:
@roymckenzie Thank you for your work. We don't use the deletelabel icon anymore in the desktop version, so I'm going to close this PR.

PR #5333, opened 2014-10-11
Feature/4646 collapse post button

Faldrian:
Adds an icon in the upper right corner of each opened post, visible on mouseover. Clicking the icon will collapse the post.

No tests for this feature so far included!
I tried several hours to figure jasmine out ... but gave up at some point. If anyone has a hint how to instantiate a fully initialized post (where the content_view.js has been run), I will gladly implement tests for collapsing and expanding posts.


Faldrian:
hm... I think I will no longer pursue this. Wil this PR just stay open or do I close it or what do I do in this case?

jhass:
If you think somebody else should adopt your code, say so and we put the orphan label on it, else just close it.

Faldrian:
Hm... It would be okay if someone else would care to adopt my code (if it is interesting for somebody else). My priorities have shifted, so I probably won't continue coding here.

denschub:
I'm closing this PR as it's nearly unmergable nowadays. Feel free to reopen if necessary.

Faldrian:
That's okay, I don't think this will be revived anymore. It was a nice experiment and I learned something (as always with experiments) but with a bit of distance to the feature I think it does not fit very well to the way diaspora is developing. :)

Issue #5041, opened 2014-06-28
Total user statistic shouldn't include closed accounts


jhass:
Currently it does. One way to determine and exclude closed accounts would be linking them to the AccountDeletions.

Flaburgan:
Looks like https://github.com/diaspora/diaspora/commit/dc444852307cea66884c8b5aad220e4ec37c9207#diff-520128ea12a2d1b32ee237a81379305aR52 fixed that, so this issue and https://github.com/diaspora/diaspora/pull/5351 can be closed.

PR #5351, opened 2014-10-16
5041 total user statistic shouldnt include closed accounts

MarcBurt:
I'm new to this, sorry if I've done anything wrong!
Saw this: https://github.com/diaspora/diaspora/issues/5041 and attempted to fix it, put the logic in the model rather than the presenter. 
The query searches for users with a username (to exclude invitations - mentioned in the thread) and Users with closed accounts.


jhass:
I'd still like this in for the spec, even though the actual fix is already applied. Would you mind rebasing? :)

MarcBurt:
Hi, yes I'll be able to do so.
I won't have access to my dev machine for a while however.

jhass:
I took the liberty to rebase this and refactor it a bit, thanks!

Flaburgan:
I merged that but the total_user count didn't change on https://diaspora-fr.org/statistics.json

jhass:
Yes, the actual change was already included in #5464, I just backported the spec for it from this PR, where the line in #5464 was also taken from.

PR #5384, opened 2014-11-07
Alter suggested bookmarklet to use Markdown

JesseWeinstein:
Switch it to use the content parameter, which doesn't alter the results, and insert some basic Markdown (header for the title, block quote for the selected text (still need to handle the case of no-selection better)) and put the URL at the end.


jhass:
@JesseWeinstein Hey, still want to look into it or shall we put this up for adoption?

JesseWeinstein:
I'm still interested, but I'd be delighted if someone with more knowledge of the codebase could look into it too.

Raven24:
superseded by #5904 
thanks for starting the initiative on this!

PR #5389, opened 2014-11-10
New registration page design

Flaburgan:
The old page still displays screenshot from reverted beta profile and old SPV, so I redesigned the page to be something like that: 

new registration

I think this design can be reused on the forgot password and sign in page but I wanted to show it to you first :)


Flaburgan:
#5391 should be merged soon, let's discuss again about this:
What's the text we want to put on, and on which page? I think that at least the sign in page can also have the text.

goobertron:
Has @pablocubico seen this? He might like to be involved in redesigning this page.

Flaburgan:
https://github.com/diaspora/diaspora/pull/5391 is merged. I'd like to at least change the "Make something" now.

We have different propositions of catchphrases:
* From diasporafoundation.org: The online social world where you are in control
* From diasporafoundation.org: Decentralization. Freedom. Privacy. (I would have ordered that differently)
* An old catchphrase: Share what you want with whom you want
* An old catchphrase: The social network where **you** are in control.

Should I open a loomio discussion?

jaywink:
@Flaburgan loomio vote after hammering four of the top choices sounds good.

svbergerem:
This is how the registration page currently looks like:

registration page

@Flaburgan what do we do with this PR? Is there anything left you would like to change in this PR?

Flaburgan:
Yeah #5743 mainly solved the problem. I liked the idea to have a background image, but this PR is way too old now and https://github.com/diaspora/diaspora/pull/6268 would do that for the home page which is more important. So closing now, if I go back on this one day I would have to restart from scratch anyway. Thank you :)

PR #5391, opened 2014-11-11
Unify disconnected pages

Flaburgan:
After a discussion with @jhass about https://github.com/diaspora/diaspora/pull/5389 he pointed that it would be better to first unify every pages with the current design (white background, black text).

So here we are:
**Registration page**
registration

**Connection page**
connection

**Password recovery 1**
password1

**Password recovery 2**
password2

Old page was like that:
old-pwd

The "diaspora*" on the screenshots is the podname.


Flaburgan:
I remove the green on the sign in button to avoid AugierLe42e's problem. I don't see why this button has another color anyway.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -4,122 +4,102 @@ input:-webkit-autofill{
...
+      @include border-radius(5px, 5px, inherit, inherit);
+    }
+    
+    .form-control:last-of-type {
+      &,
+      &:focus, 
+      &:invalid,
+      &:invalid:focus,
+      &:invalid:required,
+      &:invalid:required:focus {
```
dimaursu: Since we are already using boostrap-sass, maybe tune-up to bootstrap can be done by using Sass variables? You can customize almost the whole framework there, and it will look more consistent, because you can't forget to apply the styling to some element.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -4,122 +4,111 @@ input:-webkit-autofill{
...
-      -webkit-box-shadow : none;
-      -moz-box-shadow : none;
-      box-shadow : none;
-      box-shadow : none;
-
-      border : none;
-      color : $text-dark-grey;
-    }
+    color: $text-dark-grey;
+    position: absolute;
```
dimaursu: What's the relative container of this? Also, maybe it can be done without `position: absolute`? This property takes the element out of the regular flow of the elements, and sometimes it can lead to nasty problems.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -4,122 +4,111 @@ input:-webkit-autofill{
...
-
-      background : transparent;
-
-      padding : 10px;
-
-      margin-bottom : 0;
-      border : none;
+    margin-bottom: 1em;
+    background-color: #fff;
+    position: relative; /* To correctly place the entypo icon */
```
dimaursu: Aha, `fieldset` is also the relative container for `label` That it good, although I would very much like it done without them.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -4,122 +4,111 @@ input:-webkit-autofill{
...
+    input { /* Bootstrap reset */
+      &,
+      &:focus, 
+      &:invalid,
+      &:invalid:focus,
+      &:invalid:required,
+      &:invalid:required:focus {
+        border: 1px solid $border-grey;
+        border-bottom: 0px;
+        border-radius: 0px;
```
dimaursu: from the `bootstrap-sass/template/project/_variables.scss`:
$baseBorderRadius:      4px;
$borderRadiusLarge:     6px;
$borderRadiusSmall:     3px;

You can reset these to 0, and then apply border radius only where it's needed. Dispora* uses rarely any border-radius, is quite squre-ish.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -4,122 +4,111 @@ input:-webkit-autofill{
...
-      border-bottom : 1px solid $border-grey;
-
-      .control-label,
-      input[type=text],
-      input[type=password],
-      .field_with_errors label {
-        padding : 10px;
-        margin : 0;
+    .form-control {
+      font-size: 16px;
```
dimaursu: I think font-size would be better specified in rem's. As a rule of thumb, take this: 
anything width-related - use percentages
anything height-related, +paddinga +margins - use rems

That alone gives you a quite responsive site. 
In this case, as the `<html>` font-size is set to 100% probably (rems are relative units to the html and body font-size), 16px == 1rem

Flaburgan:
The failure is just a random one. I think this is good to go. I'm sorry @dimaursu I didn't apply you style suggestion, the reset of bootstrap is a refactor that we will probably do when porting to BS3, and I don't see how to place the entypo icon without a position: absolute.

I would like to see that merged, because it modifies different file and I just had a difficult merge during the rebase, I think it's waiting for a long time now and I don't want the rebases to become harder.

dimaursu:
As long as absolute positioned elements live in a relative container - fine by me. We just have to be a bit more careful.  I can't wait for the BS3 port to begin, I would enjoy slashing the CSS. I have some crazy ideas for it, including a official way to customize the default theme, SMACSS conventions, and more ^_^

app/assets/stylesheets/colors.css.scss
```diff
@@ -7,7 +7,7 @@ $background-blue: #E7F2F7;
 $grey: #2B2B2B;
 $light-grey: #DDDDDD;
 
-$border-grey: #DDDDDD;
+$border-grey: #CCCCCC;
```
jhass:
```
rgrep "\$border-grey" app/assets | wc -l
51
```

Why is the old color unacceptable?

Flaburgan: I just did some tries on the profile page and the stream and a little darker feels better to me.
If we think that this change should be a separate PR, I can drop it from this one.

jhass:
Since this is about consistency:

The placeholder styling is inconsistent:




Also I'd consider removing the heading on the sign up form and change the button content from "Continue" to "Sign up" instead.

Flaburgan:
Suggestions applied. I also added the footer to the sign up page.

app/assets/stylesheets/colors.css.scss
jhass: Yes, I think it should be separate, it's rather unrelated and quite invasive.

jhass:
Not only random failures:

https://travis-ci.org/diaspora/diaspora/jobs/46850397#L2584
https://travis-ci.org/diaspora/diaspora/jobs/46850398#L2583

Flaburgan:
Hm, that's new. I'll check it.

app/views/passwords/edit.haml
```diff
@@ -1,27 +1,29 @@
...
-    %fieldset
-      = f.hidden_field :reset_password_token
-      .control-group
-        %label{class: "control-label", for: "user_password", style: "width: 45%; text-align: right; padding: 10px 0; overflow: hidden"}
+  .container-fluid
+    .text-center
+      .branding-asterisk
+      %h1
+        = AppConfig.settings.pod_name
+
```
svbergerem: They are still there. Please remove them and also check for trailing whitespaces in other files you changed.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -4,122 +4,102 @@ input:-webkit-autofill{
...
-      box-shadow : none;
-      box-shadow : none;
-
-      border : none;
-      color : $text-dark-grey;
-    }
+    color: $text-dark-grey;
+    position: absolute;
+    top: -9999px;
+    left: -9999px;
```
svbergerem: You moved that part of the code but it is still there. It would be great if you could tell me why
```css
position: absolute;
top: -9999px;
left: -9999px;
```
is better than
```css
display: none;
```

svbergerem:
I'll review this PR again this evening.

Flaburgan:
Tests should be fixed, color set back to the old value and trailing whitespaces removed.

About the position -9999px it's now a common practice used for accessibility: an element with display: none; would be ignored by screen readers.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -4,88 +4,112 @@ input:-webkit-autofill{
...
-
 form.block-form {
-  display : inline-block;
+  margin: 20px auto;
 
   label {
     color : $text-dark-grey;
+    position: absolute;
+    top: -9999px;
+    left: -9999px;
```
dimaursu: So this makes screen readers avoid them? What's the purpose of this? Wouldn't people with disabilities need those labels? Can you provide us with a link to a relevant article? I'm not well-versed is such issues, and I can't find any recommendation on this topic.

Flaburgan: No it is the opposite: we don't want the user to see the label, but we want screen readers to read them, that's why we use this instead of a simple display: block;

svbergerem: Ok, in that case it sounds like a legit intermediate solution. Bootstrap 3 has a `sr-only` class for elements that are only for screen readers: http://getbootstrap.com/css/#forms-inline

Flaburgan:
Can someone relaunch travis? Fails look unrelated.

Flaburgan:
Green travis :)

jhass:
It looks like there are still unaddressed comments by Steffen, please review all comments ;)

Flaburgan:
I don't see any comment left.

svbergerem:
@Flaburgan 

1. The red border on invalid inputs I mentioned in [this comment](https://github.com/diaspora/diaspora/pull/5391#issuecomment-66546565) is still there. You moved the css code that would prevent that and now it only has an effect on inputs that are inside of a fieldset. That is not the case for the comment textarea in the SPV and the Bootstrap streams and maybe some other places.

2. The icon alignment I mentioned in [this comment](https://github.com/diaspora/diaspora/pull/5391#issuecomment-66882489) did not change and I still think that it needs some improvement. I even posted a diff which would fix that.

3. [this comment](https://github.com/diaspora/diaspora/pull/5391#issuecomment-66546565):
> The captcha input doesn't look like the other input fields. It would be great if you could unify those.

4. the same comment:
> I'd still prefer hiding "Hey, make something." on smaller devices.
> > You can do that by adding the class `hidden-phone` to the surrounding element, in this case to `<h1 id="create-something-text">` (see http://getbootstrap.com/2.3.2/scaffolding.html#responsive)

5. There are still a lot of trailing whitespaces. One of those I already [commented on](https://github.com/diaspora/diaspora/pull/5391#discussion_r21648861) but it hasn't been fixed.
```diff
git diff develop --check
app/assets/stylesheets/new_styles/_forms.scss:24: trailing whitespace.
+      &:focus, 
app/assets/stylesheets/new_styles/_forms.scss:59: trailing whitespace.
+      &:focus, 
app/views/passwords/edit.haml:17: trailing whitespace.
+        
app/views/sessions/new.html.haml:14: trailing whitespace.
+      %i.entypo.user      
app/views/sessions/new.html.haml:15: trailing whitespace.
+      = f.text_field :username, 
app/views/sessions/new.html.haml:16: trailing whitespace.
+                     placeholder: t('registrations.new.username'), 
app/views/sessions/new.html.haml:17: trailing whitespace.
+                     class: 'input-block-level form-control', 
app/views/sessions/new.html.haml:18: trailing whitespace.
+                     required: true, 
app/views/sessions/new.html.haml:19: trailing whitespace.
+                     pattern: '[A-Za-z0-9_.@\-]+', 
app/views/sessions/new.html.haml:20: trailing whitespace.
+                     autocapitalize: 'none', 
app/views/sessions/new.html.haml:21: trailing whitespace.
+                     autocorrect: 'off', 
app/views/sessions/new.html.haml:26: trailing whitespace.
+      %i.entypo.lock      
app/views/sessions/new.html.haml:27: trailing whitespace.
+      = f.password_field :password, 
app/views/sessions/new.html.haml:28: trailing whitespace.
+                         placeholder: t('registrations.new.password'), 
app/views/sessions/new.html.haml:30: trailing whitespace.
+                         required: true, 
app/views/sessions/new.html.haml:31: trailing whitespace.
+                         autocapitalize: 'none', 
```

That were the things I mentioned before that haven't been addressed yet. As soon as those are fixed or you tell me why you won't fix them I will take another look at this and do a review of the actual code.

Flaburgan:
Points 4 and 5 are solved, I'll try to work on the others during this week.

Flaburgan:
Here is the current status on focus for the point 3, does anybody know how I can make the input taking the whole available space? display: block doesn't seem to work. Maybe a botstrap technique here?

unify

jaywink:
Tried to help with the captcha - difficult :P Maybe move it outside the form visually, or separate into two rows. The trouble is passing classes to the input seems not possible as it's a third party gem?

app/views/simple_captcha/_simple_captcha.haml
```diff
@@ -1,4 +1,2 @@
-.control-label
-  = simple_captcha_options[:image]
-.controls
-  = simple_captcha_options[:field]
\ No newline at end of file
+= simple_captcha_options[:image]
+= simple_captcha_options[:field]
```
Flaburgan: @jaywink it's possible to put a class on the captcha here. But the problem is not to select it to css, it's which style to apply... I can align it right making it floating, and as the border between the image and the input is on the input (and we will never have focus on the image), this would work. But then we have a floating element, so the top and bottom borders are hidden by the container div...

jaywink: Yeah, that is why I suggested maybe separating the captcha out of the other form parts might make for better and easier to maintain design.

dimaursu:
@Flaburgan try setting the field width to 100%

Flaburgan:
I think this should be good to go.

svbergerem:
1. The implementation of the things @jhass suggested in [this comment](https://github.com/diaspora/diaspora/pull/5391#issuecomment-69729001) got lost.

2. When entering something in an input field the text and the cursor are red until there is some valid input. As soon as there is some valid input the input field has some blue shadow. That's not how it was before and it is inconsistent with other input fields so please fix it. Those input fields also show some weird behavior when they are activated. (some transitions with placeholders moving up and down and resizing borders)

3. The input for the captcha is broken for me on iceweasel 35:
bildschirmfoto von 2015-02-01 23 28 34

Flaburgan:
>Those input fields also show some weird behavior when they are activated. (some transitions with placeholders moving up and down and resizing borders)

I noticed that but I don't success to find hos to remove this :active { box-shadow}, it should already be deactivated, and when I manually put :active with Firefox I don't reproduce...

svbergerem:
@Flaburgan I opened a PR which should fix most issues.

svbergerem:
Oh, I forgot to change the button content from "Continue" to "Sign up" like @jhass suggested before.

svbergerem:
Please change `registrations.new.sign_up` from 'SIGN UP' to 'Sign up'.

Flaburgan:
Awesome, I should have time to look at it tomorrow

svbergerem:
I hope those comments are only there because this PR hasn't been rebased and so [this commit](https://github.com/diaspora/diaspora/commit/d96229fddcad587652d4614ce88ff3ec477d689c) is missing.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -1,91 +1,115 @@
...
 
-
+/* Forms described here are only used on the public pages at the moment */
 form.block-form {
-  display : inline-block;
+  margin: 20px auto;
 
   label {
     color : $text-dark-grey;
+    position: absolute;
```
AugierLe42e: You talk too much @houndci, has someone already told you that ? Oh wait ! No ! You're a bot ! :p

config/locales/diaspora/en.yml
```diff
@@ -825,15 +825,15 @@ en:
...
+
         We’d love you to stay part of diaspora*’s community, and you’re welcome to keep your account live if you want to.
-        
+
         If you want to keep your account live, all you need to do is to sign in to your account before %{remove_after}. When you sign in, take a moment to have a look around diaspora*. It has changed a lot since you last looked in, and we think you’ll like the improvements we’ve made. Follow some #tags to find content you love.
-        
+
         Sign in here: %{login_url}. If you’ve forgotten your sign-in details, you can ask for a reminder on that page.
-        
+
```
svbergerem: In most cases removing trailing whitespaces is a good idea. In this case all those lines belong to one email and I am not sure if those whitespaces are required. Please revert these changes unless you are sure that it doesn't break stuff.

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -1,91 +1,115 @@
...
-    border: 1px solid $border-dark-grey;
-    outline: none;
-    box-shadow: none;
-    color : $text-dark-grey;
+  input[type=submit] {
+    text-align: center;
+    display: block;
+    width: 100%;
+    padding: 8px 0;
   }
```
svbergerem: You can just remove those 6 lines defining the style of submit buttons and use Bootstrap classes for those buttons instead. I'll add a comment for each button so you see which classes are missing.

app/views/passwords/edit.haml
```diff
@@ -1,27 +1,29 @@
...
-    = f.submit t('devise.passwords.edit.change_password'), :class => "new-btn"
+        %i.entypo.lock
+        = f.password_field :password_confirmation, class: "input-block-level form-control", required: true, placeholder: t('devise.passwords.edit.confirm_password'), autocapitalize: "none", autocorrect: "off"
 
-    %br/
-    %br/
-    %br/
-    = link_to t('devise.shared.links.sign_in'), new_session_path(resource_name)
+      = hidden_field(:user, :remember_me, value: 1)
+      = f.submit t('devise.passwords.edit.change_password'), class: "btn"
```
svbergerem: the class should be `btn btn-block`

app/views/passwords/new.haml
```diff
@@ -1,22 +1,24 @@
...
-          = f.text_field :email, :required => true, :autocapitalize => "off", :placeholder => "••••••••", :autocorrect => "off", :autofocus => true, :style => "width: 60%;"
-    = f.submit t('devise.passwords.new.send_password_instructions'), :class => "new-btn"
-
-    %br/
-    %br/
-    %br/
-    = link_to t('devise.shared.links.sign_in'), new_session_path(resource_name)
+        %i.entypo.mail
+        = f.text_field :email, class: "input-block-level form-control", required: true, autocapitalize: "off", placeholder: t('devise.passwords.new.email'), autocorrect: "off", autofocus: true
+      = f.submit t('devise.passwords.new.send_password_instructions'), class: "btn"
```
svbergerem: the class should be `btn btn-block`

app/views/registrations/new.haml
```diff
@@ -0,0 +1,44 @@
...
+
+                - if AppConfig.settings.captcha.enable?
+                  = show_simple_captcha :object => 'user', :code_type => 'numeric'
+
+                = invite_hidden_tag(invite)
+
+              - if AppConfig.settings.terms.enable?
+                %p#terms.text-center
+                  = t('.terms', terms_link: link_to(t('.terms_link'), terms_path, target: "_blank")).html_safe
+              = f.submit t('.sign_up'), class: "btn", data: {disable_with: t('.submitting')}
```
svbergerem: the class should be `btn btn-large btn-block`

app/assets/stylesheets/new_styles/_registration.scss
```diff
@@ -1,48 +1,58 @@
...
     white-space : nowrap;
   }
 
+  #pod-name {
+    text-align: center;
+    margin: 12px;
+    font-size : 35px;
+  }
+
   #sign-up-text {
```
svbergerem: looks like there is no `#sign-up-text` in the app anymore so you can just remove it

app/assets/stylesheets/new_styles/_registration.scss
```diff
@@ -1,48 +1,58 @@
...
     margin-bottom : 0.5em;
   }
 
-  #collage {
-    width : 344px;
-    height : auto;
-    max-width : 95%;
+  #create-something-text,
+  #diaspora-hearts,
+  #sign-up-text {
```
svbergerem: you can remove `#sign-up-text`

config/locales/devise/devise.en.yml
```diff
@@ -30,13 +30,13 @@ en:
...
-        new_password: "New password"
-        confirm_password: "Confirm password"
+        new_password: "NEW PASSWORD"
+        confirm_password: "CONFIRM PASSWORD"
       new:
         forgot_password: "Forgot your password?"
         no_account: 'No account with this email exists'
         reset_password: "Reset password"
-        email: "Email address"
+        email: "EMAIL ADDRESS"
```
svbergerem: Why are those changes necessary? We already have `text-tranform: uppercase` for the placeholders.

app/views/registrations/new.haml
svbergerem: I wrote in a previous comment that you should change the translation of `registrations.new.sign_up`. Instead you could also just use `device.shared.links.sign_up`.

svbergerem:
I just finished my code review and there were some minor remarks. There are also some [tests failing](https://travis-ci.org/diaspora/diaspora/jobs/50702334#L2641) where you will have to change `And I press "Continue"` to `And I press "Sign up"` After resolving the remarks and fixing the tests this PR should be ready to merge.

config/locales/devise/devise.en.yml
Flaburgan: Do you think that input:-moz-placeholder, input::-webkit-input-placeholder, input:-ms-input-placeholder are enough? Not sure it always works, but you're probably right, if some browsers don't display it uppercase, it's not a problem, this is only about design.

jhass: Mmh, https://github.com/diaspora/diaspora/blob/develop/Gemfile#L57

svbergerem: > Add ::selection and ::placeholder selectors support.

They added support for the selector some time ago. :)

app/views/registrations/new.haml
Flaburgan: Hm, is `devise.shared.links.sign_up` used elsewhere? Because it is translated by "Inscription" in French, which is a name. I'd like a verb here.

config/locales/diaspora/en.yml
Flaburgan: There are other email texts in the file and this one is the only one with spaces here, I think it's safe to remove it.

app/views/registrations/new.haml
svbergerem: `grep -Rin "devise.shared.links.sign_up" app/views/`

If that is a problem for some languages you can of course just change registrations.new.sign_up.

config/locales/diaspora/en.yml
svbergerem: Good point.

svbergerem:
Thank you! :)

Flaburgan:
Finally! Sorry for the extra work here @svbergerem now I know where my mistakes came from I'll not do it again ;)

PR #5441, opened 2014-11-30
speed up bundler gem resolution

joshk:
since there is a Gemfile.lock, it's good practice to use --deployment too

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ branches:
     - 'master'
     - 'develop'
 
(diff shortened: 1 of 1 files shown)


jhass:
Closing this for now, thank you anyway.

jhass:
@joshk Small heads up, with Bundler 1.10 introducing optional groups we could get rid of this hack and activate --deployment on Travis, see #6003. Just in case you ran across other projects with a similar issue ;)

PR #5452, opened 2014-12-02
Add currency setting to paypal donation and allow unhosted button setting

Flaburgan:
I didn't find the way to set up easily an "hosted button" on paypal, all I found was the unhosted one, so I modified diaspora\* to allow both type of button.


jhass:
Thanks!

PR #5464, opened 2014-12-08
[Feature] Statistics page

SansPseudoFix:
We create a more readable page for stats. At the moment, it's basic. But we prefer be sure to do it well.
I dont' know for others pods, but framasphere's users like check stats :) 

statistics_preview

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -47,6 +47,7 @@ en:
   _contacts: "Contacts"
   welcome: "Welcome!"
   _terms: "terms"
+  _statistics: "Statistics"
 
   #for reference translation, the real activerecord english transations are actually
   #in en-US, en-GB, and en-AU yml files
@@ -1380,3 +1381,19 @@ en:
       default: "The secret code did not match with the image"
       user: "The secret image and code were different"
       failed: "Human verification failed"
+
+  statistics:
+    name: "Name"
+    network: "Network"
+    services: "Services"
+    total_users: "Total users"
+    active_users_halfyear: "Active users half year"
+    active_users_monthly: "Active users monthly"
+    local_posts: "Local posts"
+    local_comments: "Local comments"
+    version: "Version"
+    registrations: "Registrations"
+    enabled: "Available"
+    disabled: "Not available"
+    open: "Open"
+    closed: "Closed"
--- a/app/assets/stylesheets/new-templates.css.scss
+++ b/app/assets/stylesheets/new-templates.css.scss
@@ -77,3 +77,6 @@
 
 /* code */
 @import 'new_styles/code';
+
+/* statistics */
+@import 'new_styles/statistics'
--- a/app/assets/stylesheets/new_styles/_statistics.scss
+++ b/app/assets/stylesheets/new_styles/_statistics.scss
@@ -1,0 +1,28 @@
+.page-statistics {
+  h1{ text-align: center; }
+
+  h3{
+    margin: 0px;
+    padding: 10px;
+    background-color: $green;
+  }
+
+  .span-3 { 
+    width: 30%;
+    height: 150px;
+    text-align: center;
+    border: 1px solid $border-grey;
+    margin-bottom: 10px;
+    border-radius: 5px;
+    .data{
+      word-wrap: break-word;
+      overflow: hidden;
+      margin-top: 25px;
+      font-size: 25px;
+      line-height: 30px;
+    }
+    .serv-disabled{
+      background-color: $background-grey;
+    }
+  }
+}
--- a/app/controllers/statistics_controller.rb
+++ b/app/controllers/statistics_controller.rb
@@ -3,13 +3,14 @@
 #   the COPYRIGHT file.
 
 class StatisticsController < ApplicationController
+  respond_to :html, :json
+  use_bootstrap_for :statistics
 
-  respond_to :json
-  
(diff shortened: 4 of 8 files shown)


config/locales/diaspora/en.yml
```diff
@@ -1365,3 +1366,20 @@ en:
       default: "The secret code did not match with the image"
       user: "The secret image and code were different"
       failed: "Human verification failed"
+
+  statistics:
+    name: "Name"
+    network: "Network"
+    services: "Available services"
```
jhass: `None` for the disabled state just sounds silly. I still think that this is not explanatory enough for the ordinary user, especially since the page is also targeted at users in search for their first home pod.

Additionally it's currently two pro hiding disabled services and two against it, we need another opinion here.

jhass:
59 commits is a bit much for this feature, please clean up the history with `git fetch upstream; git rebase -i upstream/develop`.

SansPseudoFix:
Well, what about display green header when service is available and grey if not ?

No service available : 
stat_design_final

Twitter available : 
stat_design_final_enabled

I'll clean up commits if you're okay with this proposition :)

jhass:
Then the heading shouldn't say "Available services", but just "Services" and the values should be "Available" / "Not available".

AugierLe42e:
I find "Available" and "Not available" not explicit.

SansPseudoFix:
@AugierLe42e ah ! You see ? ;)

I think "available" is more explicite than "enabled" too.

> Then the heading shouldn't say "Available services", but just "Services" 

True.

goobertron:
This is a stats page for podmins, right? If so, I agree that

> the heading shouldn't say "Available services", but just "Services"

but think the values should be 'Enabled' / 'Not enabled' (rather than 'Available' / 'Not available') as all these services are available to the podmin if the podmin enables them. (They'd be unavailable to users of the pod if the podmin hasn't enabled them, but to say 'Not available' to a podmin suggest that the service can't be connected, so I think 'Not enabled' is better.)

Flaburgan:
I wouldn't say this is specifically for podmins. Podmins have their own stats in the admin panel.

goobertron:
Ah, I thought this was a new design for stats in the admin panel. So this is a public view for the JSON stats, then? In which case, 'Available' / 'Not available' would be appropriate, I agree.

jhass:
Yes, it's a public HTML representation for /statistics.json basically.

SansPseudoFix:
@AugierLe42e cleaned up the code (thx mate).
So, what about compromise : all boxes visibles but colors if service available (or not) ?

AugierLe42e:
@SansPseudoFix : ya welcome buddy !

app/views/statistics/statistics.mobile.haml
```diff
@@ -0,0 +1,24 @@
...
+    = render 'statistics/statistic', name: t('statistics.active_users_monthly'), value: @statistics.monthly_users, activated: "serv-enabled"
+  - if @statistics.expose_posts_counts?
+    = render 'statistics/statistic', name: t('statistics.local_posts'), value: @statistics.local_posts, activated: "serv-enabled"
+  - if @statistics.expose_comment_counts?
+    = render 'statistics/statistic', name: t('statistics.local_comments'), value: @statistics.local_comments, activated: "serv-enabled"
+.row-fluid
+  %h1
+    = t('statistics.services')
+  - Configuration::KNOWN_SERVICES.each do |service|
+    = render 'statistics/statistic', name: "#{service.capitalize}", value: service_status(service, @statistics.available_services), activated: service_class(service, @statistics.available_services)
```
jhass: Did you get an error when requesting the mobile format, or why do you have an exact duplicate here?

If so, instead of duplicating the template, try rendering the same one for both formats:

```ruby
respond_to do |format|
  format.json { }
  format.any(:html, :mobile) { render 'statistics/statistics.html.haml' }
end

AugierLe42e: > Did you get an error when requesting the mobile format, or why do you have an exact duplicate here?

No, just we didn't know how to do that.

jhass: I did neither, just googled it ;)

AugierLe42e:
@jhass : Fixed, you can merge.

jhass:
Thanks!

SansPseudoFix:
Yeah \o/

jaywink:
When logged in on my local it shows like this:

selection_047

Anyone else? When logged out works fine..

SansPseudoFix:
Argh... Last time I tried, no. I'll try today.

SansPseudoFix:
Confirmed. This is ````format.any(:html, :mobile) { render 'statistics/statistics.html.haml' }```` the problem.

AugierLe42e:
Looks like a problem with bootstrap and the rendering line.

Flaburgan:
How the hell do you guys success to mix git that way? This page now shows only two files modified when https://github.com/diaspora/diaspora/pull/5587/files shows a lot...

I'm sorry, but it looks like you break stuff, including the statistics.json route which now invert the monthly and the 6 monthly active users... see https://diaspora-fr.org/statistics.json for example.

jaywink:
Don't sweat @Flaburgan - mistakes happen to everyone, and part of the blame is the people who review the PR's. This is not a critical issue, but of course needs to be fixed, issue here: #5589

PR #5468, opened 2014-12-09
Add followed tags to the mobile menu

Flaburgan:
I had two bad reviews on the mozilla marketplace because the mobile version doesn't allow to access followed tags, so here we go :)


jhass:
Thank you!

Flaburgan:
:D cool, thank you!

PR #5470, opened 2014-12-13
Refactor mobile javascript to get rid of console errors.

jaywink:
Fixes also mobile bookmarklet services.

Basically:
- Went through the whole mobile app checking for console errors
- Extracted mobile file uploader to own file, since it's only needed in status message creation
- Made sure "main" JS does not need to be loaded in status message creation by including I18n JS directly, instead of having to rely on the rest of the JS to run
- Removed unnecessary duplicate loading of mobile.js on a few pages

I would have liked to move the inline JS to JS files and maybe reorganize a bit better so includes could be used- but I think that should be another PR. This fixes actual bugs, ie bookmarklet service icons have not been working for a long time.

### DIFF
--- a/app/views/layouts/application.mobile.haml
+++ b/app/views/layouts/application.mobile.haml
@@ -119,6 +119,6 @@
 
     / javascripts at the bottom
     = jquery_include_tag
-    = javascript_include_tag :mobile
+    = javascript_include_tag "mobile/mobile"
     = load_javascript_locales
     = include_chartbeat
--- a/config/application.rb
+++ b/config/application.rb
@@ -61,7 +61,7 @@ module Diaspora
       mailchimp.js
       main.js
       jsxc.js
-      mobile.js
+      mobile/mobile.js
       profile.js
       people.js
       profile.js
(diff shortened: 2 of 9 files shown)


app/views/layouts/application.mobile.haml
```diff
@@ -112,5 +112,10 @@
 
     / javascripts at the bottom
     = jquery_include_tag
-    = javascript_include_tag :mobile
+    - unless current_page?(:new_user_session)
+      = javascript_include_tag "mobile/mobile"
+    - if current_page?(:new_status_message)
+      = javascript_include_tag "helpers/i18n"
+      = load_javascript_locales
+      = javascript_include_tag "mobile/mobile_file_uploader"
```
jaywink: Yeah I think it's better now? Hopefully cukes will be happy too..

config/application.rb
```diff
@@ -61,7 +62,8 @@ class Application < Rails::Application
       mailchimp.js
       main.js
       jsxc.js
-      mobile.js
+      mobile/mobile.js
+      mobile/mobile_file_uploader.js
```
jhass: Only list files here that you need to have available for `javascript_include_tag` and the like, everything else will get pulled in via the sprockets requires.

jaywink:
"*One more (Travis cukes) time...*" ...

It is possible #5489 fixed the cukes which I broke by rearranging stuff :dancer:

jaywink:
Right, tuned according to comments and tests now pass.

jhass:
Thank you!

PR #5499, opened 2014-12-28
JSON Exporter, part deux

gdpelican:
Thought I'd share this for a little feedback if you'd care to give it a look @jhass @jaywink

This currently:
- Boots the current export off to Sidekiq
- Uploads a json file through carrierwave
- Sends an email notifying the user when the job's complete, which includes a link to the upload
- Should pass tests, although we'll see when Travis gets its mitts on it.

It doesn't:
- Have a good UI solution yet (you just click on 'download my profile' and it takes you to a blank screen with status 200; we'll have to do a flash notice or something to tell the user what's happening)
- Include any of the heavy lifting associations we outscoped from the last PR

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -135,12 +135,14 @@ class UsersController < ApplicationController
     redirect_to stream_path
   end
 
-  def export
-    if export = Diaspora::Exporter.new(current_user).execute
-      send_data export, filename: "#{current_user.username}_diaspora_data.json", type: :json
-    else
-      head :not_acceptable
-    end
+  def export_profile
+    current_user.queue_export
+    flash[:notice] = I18n.t('users.edit.export_in_progress')
+    redirect_to edit_user_path
+  end
+
+  def download_profile
+    send_data File.open(current_user.export.path).read, type: :json, filename: current_user.export.filename
   end
 
   def export_photos
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -291,6 +291,28 @@ class User < ActiveRecord::Base
     end
   end
 
+  ######### Data export ##################
+  mount_uploader :export, ExportedUser
+
+  def queue_export
+    update exporting: true
+    Workers::ExportUser.perform_async(id)
+  end
+
+  def perform_export!
+    export = Tempfile.new([username, '.json.gz'], encoding: 'ascii-8bit')
+    export.write(compressed_export) && export.close
+    if export.present?
+      update exporting: false, export: export, exported_at: Time.zone.now
+    else
+      update exporting: false
+    end
+  end
+
+  def compressed_export
+    ActiveSupport::Gzip.compress Diaspora::Exporter.new(self).execute
+  end
+
   ######### Mailer #######################
   def mail(job, *args)
     pref = job.to_s.gsub('Workers::Mail::', '').underscore
@@ -505,6 +527,6 @@ class User < ActiveRecord::Base
                             "created_at", "updated_at", "locked_at",
                             "serialized_private_key", "getting_started",
                             "disable_mail", "show_community_spotlight_in_stream",
-                            "email", "remove_after"]
+                            "email", "remove_after", "export", "exporting", "exported_at"]
   end
 end
--- a/app/serializers/export/comment_serializer.rb
+++ b/app/serializers/export/comment_serializer.rb
@@ -1,0 +1,10 @@
+module Export
+  class CommentSerializer < ActiveModel::Serializer
+    attributes :text,
+               :post_guid
+
+    def post_guid
+      object.post.guid
+    end
+  end
+end
--- a/app/serializers/export/post_serializer.rb
+++ b/app/serializers/export/post_serializer.rb
@@ -1,0 +1,14 @@
+module Export
+  class PostSerializer < ActiveModel::Serializer
+    attributes :text,
+               :public,
+               :diaspora_handle,
+               :type,
+               :image_url,
+               :image_height,
+               :image_width,
+               :likes_count,
+               :comments_count,
+               :reshares_count
+  end
+end
--- a/app/uploaders/exported_user.rb
+++ b/app/uploaders/exported_user.rb
@@ -1,0 +1,19 @@
+#   Copyright (c) 2010-2011, Diaspora Inc.  This file is
+#   licensed under the Affero General Public License version 3 or later.  See
+#   the COPYRIGHT file.
+
+class ExportedUser < CarrierWave::Uploader::Base
+
+  def store_dir
+    "uploads/users"
+  end
+
+  def extension_white_list
+    %w(gz)
+  end
+
+  def filename
+    "#{model.username}_diaspora_data.json.gz"
+  end
+
+end
--- a/Changelog.md
+++ b/Changelog.md
@@ -137,12 +137,12 @@ diaspora.yml file**. The existing settings from 0.4.x and before will not work a
 * Truncate too long OpenGraph descriptions [#5387](https://github.com/diaspora/diaspora/pull/5387)
 * Make the source code URL configurable [#5410](https://github.com/diaspora/diaspora/pull/5410)
 * Prefill publisher on the tag pages [#5442](https://github.com/diaspora/diaspora/pull/5442)
-* Allows users to export their data in JSON format from their user settings page [#5354](https://github.com/diaspora/diaspora/pull/5354)
 * Don't include the content of non-public posts into notification mails [#5494](https://github.com/diaspora/diaspora/pull/5494)
 * Allow to set unhosted button and currency for paypal donation [#5452](https://github.com/diaspora/diaspora/pull/5452)
 * Add followed tags in the mobile menu [#5468](https://github.com/diaspora/diaspora/pull/5468)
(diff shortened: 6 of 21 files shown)


gdpelican:
Alright, I've changed it so the data is GZipped and added in posts and comments to the serializers (might need some feedback on which fields want to be there)

It'd be cool to get your take on this, @jaywink ?

app/models/user.rb
```diff
@@ -291,6 +291,28 @@ def like_for(target)
...
+  ######### Data export ##################
+  mount_uploader :export, ExportedUser
+
+  def queue_export
+    update exporting: true
+    Workers::ExportUser.perform_async(id)
+  end
+
+  def perform_export!
+    export = Tempfile.new([username, '.json'], encoding: 'ascii-8bit')
```
jaywink: Hmm why is the encoding `ascii-8bit`?

app/controllers/users_controller.rb
```diff
@@ -135,12 +135,14 @@ def getting_started_completed
...
-      head :not_acceptable
-    end
+  def export_profile
+    current_user.queue_export
+    flash[:notice] = I18n.t('users.edit.export_in_progress')
+    redirect_to edit_user_path
+  end
+
+  def download_profile
+    send_data decompressed_export, type: :json, filename: current_user.export.filename
```
jaywink: Is there any reason we can't just send over a compressed file? IMHO as it's text data, this would make sense to decrease downloads. I'm guessing it will be 1 week after this is released that someone will offer an utility to automate exporting of profiles.. ;)

app/serializers/export/comment_serializer.rb
```diff
@@ -0,0 +1,8 @@
+module Export
+  class CommentSerializer < ActiveModel::Serializer
+    attributes :text,
+               :author_id,
```
jaywink: `author_id` and `author_signature` are probably irrelevant for a user, even in a pod migration situation (I think). Could we instead add here author diaspora handle? That is the network wide unique identifier that is actually usable. GUID is of course too but that doesn't say anything to users looking at the export.

app/serializers/export/post_serializer.rb
```diff
@@ -0,0 +1,16 @@
+module Export
+  class PostSerializer < ActiveModel::Serializer
+    attributes :text,
+               :author_id,
```
jaywink: Again, `author_id` is probably irrelevant.

app/serializers/export/post_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+               :public,
+               :diaspora_handle,
+               :type,
+               :image_url,
+               :image_height,
+               :image_width,
+               :likes_count,
+               :comments_count,
+               :reshares_count,
+               :favorite
```
jaywink: Interesting, first time seen this and doesn't to me look like it's used. There is a `toggleFavorite` method in the JS but it's not called :P

Some legacy idea that never got implemented, @jhass ? I'd personally remove it if we don't use the attribute.

app/serializers/export/comment_serializer.rb
jaywink: Heh no wait, this is of course the users own comments, so no need to have a handle. BUT we need to say what post it relates to. How about building an url to the post?

app/serializers/export/comment_serializer.rb
```diff
@@ -0,0 +1,8 @@
...
+               :author_signature,
+               :likes_count
```
jaywink: We don't support `likes_count` at the moment so this could be removed.

jaywink:
Other than the comments relating to the attributes, **super awesome** work :star2: A lot of people will be happy to have this working again, as it's a shame our export is currently worse than many proprietary networks ;)

:sake: for you ;)

jaywink:
Oh also, I'm assuming this will work fine with pods using S3 for storage etc?

app/serializers/export/post_serializer.rb
jhass: Eh, I don't remember that one at all, must be from the time I didn't watch the project. It was introduced in 75d25e947502851e2b79a6f9a101a2a83d2eb363

app/serializers/export/comment_serializer.rb
jhass: More importantly it's just a counter cache.

app/serializers/export/post_serializer.rb
```diff
@@ -0,0 +1,16 @@
...
+               :author_id,
+               :public,
+               :diaspora_handle,
+               :type,
+               :image_url,
+               :image_height,
+               :image_width,
+               :likes_count,
+               :comments_count,
+               :reshares_count,
```
jhass: Counter caches, shouldn't be included.

jaywink: Doesn't it contain the current (when exported) count though? IMHO that is relevant information for the user :)

app/models/user.rb
jhass: ASCII-8bit is Rubys clunky way of saying BINARY (which is an alias for ASCII-8bit in fact). Remember, ASCII is a 7bit encoding ;) Since we write gzipped data here, that's correct.

app/uploaders/exported_user.rb
```diff
@@ -0,0 +1,19 @@
...
+  def store_dir
+    "uploads/users"
+  end
+
+  def extension_white_list
+    %w(json)
+  end
+
+  def filename
+    :"#{model.username}_diaspora_data.json"
```
jhass: Why does this need to be a Symbol?

app/serializers/export/post_serializer.rb
jhass: It should, I don't see much value in that information though.

app/serializers/export/comment_serializer.rb
jhass: I'd just export the post guid.

app/serializers/export/post_serializer.rb
jaywink: I think it could be valuable for some - imagine you want to look at some old export of your data from an old pod for example. Maybe someone will make an app to analyse exports and rank posts with most likes etc.

app/serializers/export/comment_serializer.rb
jaywink: Post guid is good, agreed. It can be used to make a link anyway :)

jaywink:
Also, before merging, the changelog entry for the export should probably be refreshed. I guess the export version does not need to be bumped since the current export version hasn't gone stable yet.

app/uploaders/exported_user.rb
gdpelican: It doesn't, I'll change it.

app/controllers/users_controller.rb
gdpelican: Just send it over as `application/gzip` with a `.gz` file extension, and let the user handle inflating it?

I worry about people on windows machines needing to download a utility to open that file, although I guess we could look at providing a .zip format as well.

jhass: Or maybe make it a tarball, even most Windows users should have a tool for that, I think 7-zip is quite common and can handle those.

gdpelican:
I haven't been able to test using an S3 instance, but I just aped the Carrierwave uploaders for images, so it shouldn't behave any differently.

gdpelican:
Well, I b0rked the rebase a little, but merged in `develop`, updated the serializers, and changed that one symbol that wasn't supposed to be there.

Also made it so we export files in a .json.gz file, which 7zip should handle.

jaywink:
Alright, no more comments to this and looks awesome to me - merging! Nice to get this in to the next release, thanks @gdpelican !

Flaburgan:
Awesome!

Flaburgan:
Currently in test on diaspora-fr :)
sidekiq-export

marienfressinaud:
@Flaburgan > I tried yesterday night but it still shows me "We are currently processing your data. Please check back in a few moments." this morning. I'm pretty sure there is a problem ;).

gdpelican:
Oh noes! I'll have a poke at it tomorrow to see what's up. We're sure
sidekiq is running?

Flaburgan:
Yeah sidekiq is running, Marien and my jobs are listed in "Enqueued". They are the only two jobs in the queue.

jhass:
`export_user` wasn't added to https://github.com/diaspora/diaspora/blob/develop/config/sidekiq.yml#L6

Flaburgan:
Confirmed, adding export_user there immediately made the jobs processed. But the message on the panel is still `We are currently processing your data. Please check back in a few moments.`

marienfressinaud:
Indeed but I have a different message: ` download my profile (Last updated at 2015-01-17 12:36:04 UTC)
refresh my profile data ` so it seems ok for me. Nevertheless there are other bugs, should we open new specific tickets for each bug found?

jhass:
> should we open new specific tickets for each bug found?

Yes, sure.

jaywink:
@gdpelican the sidekiq queue thingy is fixed now, no need to check that :)

margori:
I have a doubt in how sidekiq works.
- I configured sidekiq embed in unicorn in my local dev pod. (embed_sidekiq_worker: true in diaspora.yml)
-when I click over "refresh my profile" link, I expect to have new registers in 'sidekiq.log' file.

Am I wrong?

PR #5500, opened 2014-12-28
Full sized images for lightbox

ThomasLeister:
With the modified code, the lightbox will show the original image file instead of a smaller version. We store the original image on our servers but only use scaled-down versions? That looks strange to me, so i decided to let the lightbox show a hight resolution version of every image.

If an image is too big for the screen, it will be scaled down dynamically by CSS.

Corresponding post on loom.io: https://www.loomio.org/d/95wh18Lt/high-resolution-full-sized-images-for-lightbox

### DIFF
--- a/app/assets/javascripts/widgets/lightbox.js
+++ b/app/assets/javascripts/widgets/lightbox.js
@@ -115,7 +115,7 @@ jQuery.fn.center = (function() {
         imageUrl = selectedImage.attr("data-full-photo"),
         images = selectedImage.parents(self.options.imageParent).find(self.options.imageSelector),
         imageThumb;
-
+      
       if( $.browser.msie ) {
         /* No fancy schmancy lightbox for IE, because it doesn't work in IE */
         window.open(imageUrl);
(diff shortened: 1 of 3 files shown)


jhass:
I still think this is not a good way to do it, but loomio decided to do it.

However, guessing the URL in the Javascript while the server can easily provide it, is simply a road blocker for this implementation to be merged.

Given your Diaspora post, should we put this up for adoption?

ThomasLeister:
Let's merge it to Dev to have this feature in a basic version (the current version). If anyone has time and skills to improve my work and change it according to the suggestions, he/she can make further contributions to it.

jhass:
You didn't even read my comment, did you?

> However, guessing the URL in the Javascript while the server can easily provide it, is simply a road blocker for this implementation to be merged. There's also a broken Jasmine spec: https://travis-ci.org/diaspora/diaspora/jobs/45286822#L572

denschub:
Closing this PR. It starts getting too old and the current implementation is blocked by [Jonnes comment](https://github.com/diaspora/diaspora/pull/5500#issuecomment-70745009). Feel free to reopen.

PR #5506, opened 2014-12-29
'toggle mobile' becomes 'Toggle Mobile'

GPhMorin:
For more capitalization consistency.

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -636,7 +636,7 @@ en:
     application:
       powered_by: "POWERED BY diaspora*"
       whats_new: "what's new?"
-      toggle: "toggle mobile"
+      toggle: "Toggle Mobile"
       public_feed: "Public diaspora* feed for %{name}"
       your_aspects: "your aspects"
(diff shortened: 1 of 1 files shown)


maliktunga:
This pull request is too specific. See https://github.com/diaspora/diaspora/issues/5520

RichKavanagh:
You couldn't wait two more days for the Loomio vote to finish?

maliktunga:
@RichKavanagh Even though the voting period is still running, this proposal is too specific to be acceptable. We shouldn't make 200 commits to apply a uniform capitalization.

jhass:
Please, both of you, we don't need to copy paste a discussion to two places.

maliktunga:
@jhass Thank you for the clarification.

RichKavanagh:
> We shouldn't make 200 commits to apply a uniform capitalization.

How many should we make?

maliktunga:
@RichKavanagh Two or three should be enough in my opinion. That depends of how many text strings are forgotten or mistakenly formatted.

PR #5509, opened 2014-12-30
Replace %{third_party_tools} by the appropriate hyperlink in tags FAQ 

flamontagne:
It's my first contribution to diasopra. It's rather small but it's a start :) Let me know if there is something I forgot to do.

Thank you


svbergerem:
It looks like this breaks https://github.com/diaspora/diaspora/blob/develop/spec/javascripts/app/views/help_view_spec.js#L75

svbergerem:
Apart from the failing spec this PR looks good to me. You just need to change https://github.com/diaspora/diaspora/blob/develop/spec/javascripts/app/views/help_view_spec.js#L77 (compare it to https://github.com/diaspora/diaspora/blob/develop/spec/javascripts/app/views/help_view_spec.js#L72). As soon as you fixed that this PR should be ready to merge.

flamontagne:
Ok, thanks. I'w working on it right now.

flamontagne:
Ok the jasmine spec has been fixed. 

PS: I only executed the rspec specs in the previous PR. That's why I didn't notice this jasmine failing test. Sorry about that!

svbergerem:
@flamontagne No problem. Thank you for your contribution! :)

Issue #779, opened 2011-07-08
Image metadata is retained


diaspora-redmine-github-migration:
Issue 343 from bugs.joindiaspora.com
Created by: **ilya**
On Sun Oct 10 05:43:12 2010

*Priority: Normal*
*Status: Confirmed*

Image metadata should probably be stripped of privacy-sensitive information on photo upload.

 Posted on Github by rsofaer on 2010/10/08 12:38:10 -0700

PR #5510, opened 2014-12-31
#779 Strip EXIF data as user preference

margori:
My solution for issue #779 including 
- Strip EXIF data from uploaded images.
- Privacy user preference to allow not striping.
- DB migration for new user pref.
- Testing both cases.

I need a review of photo_spec file.

Included image from [Flickr](https://www.flickr.com/photos/fscholz/13919459137/in/photostream/)


margori:
> What again were the arguments for keeping it after we're done with doing any processing based on it?

I see this question full of agresivity. So I gonna translate it to a peaceful language, so I can answer it

> I see that you're maintaining EXIF data in processed images. We already have an unpleasant conversation, which I cannot find now, where was decided to strip them. Why did you decide to maintain them?

Since strip EXIF data decision would be in user realm, if he/she decide to maintain them, it's spread to processed images too. If not, exif is striped from original and processed.

jhass:
I didn't mean to be aggressive, I'm sorry if it seemed that way, I'm truly wondering.

We do make privacy (by default) decisions elsewhere, for example by not having a setting for posting public by default, to keep that conscious decision every single time. So, I still do wonder what arguments are there for storing images with EXIF data on the server at all, remember we're not primarily a photo sharing or gallery software.

margori:
>  I still do wonder what arguments are there for storing images with EXIF data on the server at all

Because d* says that [user is in control](https://diasporafoundation.org/). If user decide to make his/her photo EXIF data public, as developer **I'm honoring his/her decision**, I'm at their service.

If he/she doesn't know about EXIF or metadata, let's suggest to strip it, from every uploaded image to the server.

margori:
@Flaburgan Strip_exif included en JSON exporter

> this is changing the default behavior without telling anything to the user.

Default behavior is: leave EXIF data in original image file, **strip in reduced size image files**. Example [public post](https://diaspora.com.ar/people/c66153caaa393a31/photos/8775), [processed image](https://joindiaspora.s3.amazonaws.com/uploads/images/scaled_full_1b8be36639624bcf2b5c.JPG), [original file](https://joindiaspora.s3.amazonaws.com/uploads/images/1b8be36639624bcf2b5c.JPG). Original file has exif, processed doesn't.

@jhass 'Recommended' caption included.

goobertron:
> I still do wonder what arguments are there for storing images with EXIF data on the server at all, remember we're not primarily a photo sharing or gallery software.

@jhass There are some arguments in the discussion on #779, for example retaining copyright information for photographers who want this attached to their photos. I agree Diaspora isn't primarily for sharing or publishing photos; however, unless it's going to cause database, performance or privacy problems, I think it's a good idea to give people the choice whether EXIF data are retained with or stripped from their uploaded photos.

config/locales/diaspora/en.yml
```diff
@@ -1291,8 +1291,10 @@ en:
     privacy_settings:
       title: "Privacy Settings"
       ignored_users: "Ignored Users"
+      strip_exif: "Strip uploaded images metadata (geopos, author, camera, etc.) (Recommended)"
```
goobertron: I'd like to suggest a slight rewording of this sentence:

> "Strip metadata such as location, author, and camera model from uploaded images (recommended)"

margori:
Last week my words were

> ...as developer I'm honoring his/her decision, I'm at their service.

and a read [this thread in Loomio](https://www.loomio.org/d/95wh18Lt/high-resolution-full-sized-images-for-lightbox). It was decided to use full resolution images, but this create a lot of technical issues, use a lot of resources of servers and can put some pods out of business.

So I change my words:

This pull request gives freedom of choice to user without using additional resources.

margori:
Request rebased.

Dear mergers, I really appreciate feedback. It's been 2 weeks of silence.

jhass:
I'm not sure if the privacy tab is the right place for the setting, currently it's more targeted at managing what falls under the term "privacy lists", which is used for blocking parts of communication.

But then I feel like we need a reorganization of the settings pages in the mid-term anyway, especially as more settings are added, so I'm willing to glance over that for now.

Sorry for the silence and thank you for your contribution!

margori:
I agree, privacy tab should be reviewed for UI & UX design. Let's see what alpha testers say.

marienfressinaud:
I think strip EXIF data preference is at the right place (that's totally related to personal privacy). On the other side, I always found "ignored users" preference was not at its place (I may prefer "Contact" page). Preference organization is a hard work and it could be interesting to see how the other social networks handle it.

Issue #5350, opened 2014-10-16
Make it visible which posts I am subscribed to


manwesulimo2004:
The "My Activity" view lists all posts a user has been active in and receives notifications on. I would appreciate a way to see that I am subscribed to these posts in the regular stream view so that I can rest assured I will be notified of any updates. This could be done with a little icon in the top right for instance.

The next step would then be to be able to toggle the subscription state manually. That way I could be notified of activity on posts I did not participate in and also stop being notified on posts I was active on.

PR #5511, opened 2014-12-31
Feature #5350 : allow user to enable / disable notifications for a post from stream

collimarco:


jaywink:
Like the idea, but unfortunately it didn't seem to work on my dev box. I have two users, both sharing, two posts with other clicked participate in the other post (without commenting or liking), then commented as the post author. No notifications were generated.

Also, styling for the default icon seems wrong, it's almost invisible on hover - to cut participation seems fine.
selection_039
selection_040

app/assets/templates/stream-element_tpl.jst.hbs
```diff
@@ -16,6 +16,15 @@
           <a href="#" rel="nofollow" class="block_user" title="{{t "ignore"}}">
             <div class="icons-ignoreuser control_icon"></div>
           </a>
+          {{#if participation}}
+            <a href="#" rel="nofollow" class="destroy_participation" title="{{t ""}}">
```
jaywink: Titles should be filled with a translation. Same in else.

collimarco:
@jaywink Added the missing part 56eb68e. Everything should be ok now.

For the translations I've notice that in general there's no correspondence between the key used in the template and the definitions in `config/locales/diaspora/*.yml`. I mean, the **Rails convention** is that when you find something like `t "stream.hide"` you expect to find `stream: hide: "Hide"` somewhere in the locales. This is not the case. So I cannot add them unless you tell me how you manage them. **Online?** Consider that this is a **new translation to be added** and not an existing one to be translated in other languages.

jhass:
It does work like the rails convention, for the ones used in the client side JS you just have to put them under `javascript`, `{{t "foo.bar"}` becomes `javascript: foo: bar: "quux"` in `config/locales/javascript/javascript.en.yml`.

collimarco:
@jhass Oops… you're right! I've added the missing translations to the *en* locale.

jaywink:
Tested locally and the subscribe/unsubscribe works now perfectly - got notif without having to comment, great!

Only thing, the icons are still not visible properly.

Not subscribed, not hovering over icon:
selection_041

Not subscribed, hovering over icon:
selection_042

Subscribed, not hovering over icon:
selection_043

Subscribed, hovering over icon:
selection_044

Compare to ignore icon, not hovering over:
selection_045

And hovering over:
selection_046

Soooo, the icon needs to follow same guidelines, ie light gray when hovering over post but not icon and darker when hovering over it.

Also, can you squash commits to one for a cleaner commit log since this is all one feature?

Thanks, this feature is very needed! :)

collimarco:
@jaywink I've **improved the icons** so that they are consistent with the others.

I've also done the requested **squash** with rebase -i.

goobertron:
Thanks for this, @collimarco. Are the tooltips styled as per @jaywink's screen shot? If so, they should be styled the same as the tooltip for the ignore icon. Apologies if I'm talking rubbish; I haven't had chance to pull and test locally, so just wanted to ask.

collimarco:
@goobertron Added the tooltip and squashed.

SansPseudoFix:
Yeah, thanks @collimarco! Really usefull feature :)

jhass:
Sorry for letting this hang on for so long, let's see how badly this bloats the participations table further :P

I took the liberty to apply some fixes to this myself, in 073e99992eac4362b9143933c4fe51632ed37d72.

Thanks!

Faldrian:
Yeah! Long missed feature appearing again, thank you! :)

Flaburgan:
thank you @collimarco!

Loelo:
Hi ! Thanks alot for this feature. This is very handy.
Haven't seen that on Loomio, so I jump in there. Do you think this could be easily adapted to have the same button and feature on people profile page : This would then enable notifications for all new (visible by me) posts that this person will make  ?

Flaburgan:
@Loelo hey! This is a nice idea, but it's something completely new, so please open another issue about it.

Loelo:
Hello @Fla, I hope it was the right way (no discussion on loomio first and because I am new  to github), but I opened an issue here : https://github.com/diaspora/diaspora/issues/5860

PR #5513, opened 2015-01-01
'Log out' becomes 'Log Out'

ghost:
More menu-based capitalisation consistency where "Log out" becomes "Log Out".

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -626,7 +626,7 @@ en:
       profile: "Profile"
       settings: "Settings"
       help: "Help"
-      logout: "Log out"
+      logout: "Log Out"
       blog: "blog"
       login: "log in"
       code: "code"


goobertron:
Hmm, well it's good to make it more consistent, but personally I'd rather move away from Capitals At The Start Of Each Word. But that's a future discussion. Thanks for the PR.

jhass:
Shouldn't we make this a one go to consistency everywhere?

I'd also tend to not titlecase at all.

RichKavanagh:
I was basing this PR on #5506

I'm not fussed which way round the casing is, or isn't, so long as it is consistent throughout Diaspora because right now it's all over the place and looks terrible.

goobertron:
I did start, a few months ago, making the capitalisation consistent in the en locale, but I got really bogged down in it and never finished the work. If there is agreement to use sentence case (i.e. normal English capitalisation rules) rather than Caps For Every Word, I'll try to complete this work.

RichKavanagh:
Cheers @goobertron I've fired it over to [this Loomio thread](https://www.loomio.org/d/kY0wClao/capitalisation) for wider discussion.

@jhass will you be merging these recent capitalisation PRs while we wait for a final decision to be reached?

jhass:
I'd rather like a policy and then make it consistent in one go kind of approach, but then I'm not the only one with commit rights ;)

RichKavanagh:
Ha! Cool tag, thanks :+1:

jaywink:
Well, according to the recently finished Loomio vote, sentence capitalization seems to rule the land. So closing this. #5579 can be taken over by anyone wanting to fix these things maybe?

Tnx anyway @RichKavanagh :)

goobertron:
I'm on it.

PR #5516, opened 2015-01-01
Add sidekiq-failures gem.

axilleas:
It has been discussed in #3993. The ability to manually restart failed jobs has been implemented.

As for memory consumption, we could configure it so that it only tracks failures if the job exhausts all its retries ([see readme](https://github.com/mhfs/sidekiq-failures#exhausted)).

We've been using it in out pod for about 2 months and hadn't noticed any glitches so far: 1GB ram, ~300 users, ~170.000 failed jobs so far.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,7 @@ gem 'simple_captcha2', '0.3.2', :require => 'simple_captcha'
 # Background processing
 
 gem 'sidekiq', '3.3.0'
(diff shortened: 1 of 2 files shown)


jhass:
Thank you.

Sidekiq 3 includes a "failed" (they call it "dead") queue by default:

axilleas:
Ah didn't know that! Awesome!

PR #5519, opened 2015-01-03
"download [...]" becomes "Download [...]"

GPhMorin:


maliktunga:
This pull request is too specific. See https://github.com/diaspora/diaspora/issues/5520

RichKavanagh:
You couldn't wait two more days for the Loomio vote to finish?

maliktunga:
@RichKavanagh Even though the voting period is still running, this proposal is too specific to be acceptable. We shouldn't make 200 commits to apply a uniform capitalization.

Issue #5483, opened 2014-12-18
It is possible to open two comment boxes for a single post on the mobile view


Makailija:
On the mobile view, it is possible to open two comment boxes for one and the same post, which is a bit confusing at the least. There are two ways of doing this:

a)
1. Click the "x reactions" button.
2. Click the "Comment" button that appeared below: the first comment box appears (below the possible already-existing comments).
3. Click the pencil icon on the right: the second comment box appears (below the first one).
![comments1](https://cloud.githubusercontent.com/assets/2229403/5489110/c03c9ba4-86ce-11e4-87e7-4c00de58da13.png)

b)
1. Click the pencil icon on the right: the first comment box appears.
2. Click the "x reactions" button. The possible already-existing comments and the "Comment" button appear below the first comment box.
3. Click the "Comment" button that appeared below: the second comment box appears below the possible already existing comments. (So in this case there is one comment box on the top, comments between and another comment box on the bottom.)
![comments2](https://cloud.githubusercontent.com/assets/2229403/5489106/b95232c2-86ce-11e4-919c-19036297a5c9.png)

The expected behaviour is that it would be impossible to open two comment boxes, but I'm not quite sure how this should be attained. One possibility would be greying out the pencil icon after clicking the "Comment" button. (I of course mean the one appearing after clicking the "x reactions" button, not the one next to the comment box.) In this case, however, it should be taken into consideration that in the case (i) that the "x reactions" button is clicked first and then the "Comment" button is clicked, the comment box would appear *below* the comments; by contrast, in the case (ii) that the pencil icon is clicked first and after that the "x reactions" button is clicked, the comment box would appear *above* the comments

Or maybe the problem could be solved just by getting rid of the "Comment" button altogether? (again, I mean the one appearing after clicking the "x reactions" button.) To me it seems that it would be the best solution.

theworldbright:
Here is my plan to address this issue:

On reaction button tap/click
* If comment box is already open
  * Show comment/reactions as before
* If comment box is not already open
  * Show comment/reactions as before
  * Open new comment box

On comment submit
 * Show comment/reactions (with the comment that was just submitted)
 * Open new comment box

The comment box will always appear below the comment/reactions (if it is showing) and the "comment" button will always show with a text area.

PR #5523, opened 2015-01-04
Improvements to the mobile drawer menu 

dimaursu:
I cleaned up the CSS a bit, removing all sort of weird box-shadow prefixed (only -webkit is still used today). If someone's using such and old device, that it requires "-o-", you probably don't need box-shadows for them.

I also added a more natural way to reveal and close the drawer, through swipes.

I removed some of the prefixes for the CSS transformations - none of those removed are not in use anymore. I think -ms-transform is also useless - it's only for IE9 - what mobile device uses IE9? What do you think?


app/assets/stylesheets/mobile/mobile.css.scss
```diff
@@ -49,9 +45,9 @@ h3 {
   text-align: left;
   padding: 10px 0;
   min-height: 34px;
-  
+
```
dimaursu: Sorry for these, my editor automatically removes trailing spaces :-/

app/assets/stylesheets/mobile/header.css.scss
```diff
@@ -169,10 +163,20 @@ header {
 }
 
 /* This class is added when the user open the drawer */
-#app.draw > * {
-          transform: translateX(-80%);
-       -o-transform: translateX(-80%);
-      -ms-transform: translateX(-80%);
-     -moz-transform: translateX(-80%);
-  -webkit-transform: translateX(-80%);
+.draw {
```
dimaursu: The selector here was too specific - due to the way the browser matches selectors, it actually hurts performance.

app/assets/javascripts/mobile.js
```diff
@@ -30,10 +31,18 @@ $(document).ready(function(){
 
   /* Drawer menu */
   $('#menu_badge').bind("tap click", function(evt){
```
dimaursu: There is a lag here on mobile devices - more precisely 300ms; maybe we can use fastclick? The lag is introduced in order to see if the user didn't intend actually to double-tap. With fastclick, pinch-to-zoom would still work, not sure about double-tapping though.

svbergerem:
Please add `gem 'rails-assets-hammer.js'` to the Gemfile instead of adding hammer.js manually.

app/assets/javascripts/mobile.js
dimaursu: #5483 might be related to this issue: tapping twice would fire 2 events (to the user the first event wasn't recognized, because of the said delay). This leads to the "gost click" problem. 
http://stackoverflow.com/questions/20225153/preventing-ghost-click-when-binding-touchstart-and-click

dimaursu: I did some investigations and  my theory proved to be wrong. The fastclick library didn't change the matter. But I think I found the issue. I will comment on the relevant issue.

app/assets/stylesheets/mobile/mobile.css.scss
Flaburgan: That's a good thing ;)

app/assets/stylesheets/mobile/header.css.scss
Flaburgan: Did you test this on every browsers? I remember I had to use this > * to deal with a "bug" in the CSS interpretor which does not react correctly when applying a translate on an element with `position: fixed;` on it. (See http://stackoverflow.com/questions/9364203/position-fixed-div-is-not-fixed-when-parent-rotates-or-translates)

Here you're translating the whole application so I suppose this works (and we will not have the weird reaction of the stream translating at the end of the closing of the drawer) but I want to be sure you tested it: I had a lot of trouble with the menu not opening on iPad, Opera Mini...

app/assets/javascripts/mobile.js
```diff
@@ -30,10 +31,18 @@ $(document).ready(function(){
...
-    evt.preventDefault();
     $("#app").toggleClass('draw');
   });
 
+  var drawerSlide = new Hammer(document.getElementById('app'));
```
Flaburgan: Do you really need to add a new library "just" for this feature? I don't know what's the current policy of the project about adding external libraries but I guess less with use them, less we have things to maintain. @jhass ?

Btw, Firefox OS uses edge gestures to swipe between application. Did you test this with a Firefox OS phone (or the simulator, try the 2.1 version) to see how it reacts?

app/assets/stylesheets/mobile/header.css.scss
dimaursu: I do not have an Ipad unfortunately. I tested in on Firefox Desktop, Chrome for Android on 2 phones and a tablet, and Firefox for Android. I will add in a few moment Opera Mobile to test it there. But I'm afraid that would not help, as Opera Mobile doesn't support transforms, according to caniuse.com

Flaburgan: I was talking about Opera Mini, but I don't know if the two projects are still supported.

app/assets/javascripts/mobile.js
dimaursu: You are right, maybe the whole library isn't needed: but it's not that big of a library, (17k uncompressed, 4k compressed), and the code has to be written: it's better that somebody else deals with all sort of bugs and quirks.

Also, maybe we can add other gestures later, so it may be useful.

I did test with Firefox for Android. Will test in the simulator soon.  In Firefox with the Responsive Design Mode activated it works fine.

dimaursu:
Maybe I should delete this pull-request, and instead merge this? 
https://github.com/dimaursu/diaspora/tree/tablet
I've merged the develop branch, and cleaned up the CSS for autoprefixer.

Flaburgan:
As said, I think that the tablet idea can be improved, I would be in favor of merging that first and then open a new PR to see what we can do for tablets.

dimaursu:
Sure. What do you think is still needed to merge this?

jhass:
For starters a rebase ;)

Flaburgan:
Just tested this PR, the stream is not translated anymore:

test

Btw, you should not develop in the develop branch, it's not easy to test.

Flaburgan:
I think you can also squash the commits together ;)

dimaursu:
That would be rather weird - I only touched a bit of CSS and JS. Sorry about the develop branch, I moved the commits in the meantime. Not sure how I will push the commits again though. Ok, I'll get to the rebase and commit squashing a bit later, MySQL is driving me crazy: I'm working on another pull-request.

jhass:
You can't change the branches of a pull request, that requires opening a new one.

Issue #5379, opened 2014-11-03
Markdown non working properly in a particular case


Airon90:
Give a look to this example: https://diasp.eu/posts/2476271

The bug is on the word "bug" and "developers": the "u" in "bug" should be italics and the "op" in "developers" should be bold.

It seems that italics and bold (do underline and strike exist on Markdown?) don't work properly on Diaspora when applied on a part of a word instead on the entire word, while it seems that Friendica correctly render markup. It seems that symbol substitution such as (r) → ® work properly (I made a misake in copy-paste the "code").

I think that this bug is somehow related to bug #4767 because it is related to the space character

jhass:
Fixed by #5526

Issue #5290, opened 2014-10-06
HTML Tag to obey line breaks over a block of text


balasankarc:
It would be good to have a tag that makes the included text obey line breaks. Like <poem> tag of mediawiki, so that we shouldn't give <br /> at end of each line while posting a poem or similar stuff.

Like
"poem tag "

Twinkle twinkle little star
How I wonder what you are
Up above the world so high
Like a diamond in the sky

"end of poem tag"

should render as the following

Twinkle twinkle little star 
How I wonder what you are
Up above the world so high
Like a diamond in the sky

jhass:
We now use a CommonMark renderer on the desktop site, see #5526.

Issue #3374, opened 2012-06-12
Markdown parsing of URI schemes


adoa:
I would like to make small pictures with links to different services that I use. The URI scheme for mail with 'mailto:' is parsed correctly by markdown. However, Markdown refuses to parse a URI that starts with 'xmpp:'.  On my profile adoa@despora.de you can see what i mean. Currently the XMPP icon does not contain a URI as it does not work.

The mailto scheme however is only parsed on my profile page. In a message for my aspects it does not work either. That puzzles me a bit.

Is that some kind of security rationale that only http, https, ftp and mailto are parsed by markdown? Are there other URI schemes that are supported? 

There is some definition about the xmpp URI scheme, as far as I can see: http://xmpp.org/extensions/xep-0147.html

jhass:
Implemented by #5526.

Issue #5216, opened 2014-09-13
Autolinking inside markdown italic/bold markers fails


dystopiacybernetica:
In my opinion if you post an URL the link should work both in mobile and desktop version, it's really useless and confusing differentiate desktop (in which you have to use markdown code) and mobile (where URL generates a link automatically)

jhass:
Fixed by #5526

PR #5526, opened 2015-01-04
Replace pagedown by markdown-it

### LINKED ISSUE #5457: Update pagedown and integrate pagedown-extra
PageDown is in vendor/assets/javascripts/markdown. Upstream sees slow but steady improvements.

Unfortunately both the bower packages as well as the rubygems that package seem to be updated less than regularly, so it's probably best to keep pulling directly from upstream.

pagedown-extra allows several extensions to markdown, among the important ones are code fences, which allow for a better code styling and maybe even syntaxhighlighting in the future, see the discussion in #5422. Here regularly updated packages seem to exist and should be used.

svbergerem:
Fixes #5457


vendor/assets/javascripts/markdown/Markdown.Converter.js
```diff
@@ -119,6 +143,93 @@ else
...
+                        while (c > 0) {
+                            v = (c % 51) + cp_A;
+                            if (v >= cp_Q)
+                                v++;
+                            if (v > cp_Z)
+                                v += dist_Za;
+                            s = String.fromCharCode(v) + s;
+                            c = c / 51 | 0;
+                        }
+                        return "Q" + s + "Q";
```
Flaburgan: One name variable are not really easy to read

svbergerem: I only updated the code: https://code.google.com/p/pagedown/source/browse/

svbergerem:
Markdown.Converter and Markdown.Sanitizer are the latest versions from https://code.google.com/p/pagedown/source/browse/ (rev. 2a8c75ce3fb5)

The activated extensions for [Markdown.Extra](https://github.com/jmcmanus/pagedown-extra) are Tables, Fenced Code Blocks, SmartyPants, Newlines and Strikethrough. A lot of new users are confused because of the default newline behavior of Markdown. That's why I activated Newlines. I can also activate other extensions if someone thinks that they would help.

Flaburgan:
It's a good idea to activate Newlines ;) Maybe footnotes looks cool too.

svbergerem:
This also fixes (at least) #5216 and #5290

This **doesn't fix** #5379

svbergerem:
Looks like updating pagedown breaks some important stuff. (even if pagedown-extra isn't activated) Hashtags at the beginning of a line are now interpreted as headlines. m(

svbergerem:
I looked at different markdown implementations and read that the reference implementation in perl also misinterprets hashtags and a lot of implementations try to be compatible with the perl implementation. (afaik also pagedown)

We could use [markdown-it](https://github.com/markdown-it/markdown-it) which fixes all mentioned issues and would also fix #5379 and #4201. Hashtags at the beginning of a line aren't misinterpreted anymore. markdown-it has some problems with links which include non-English chars but that is a known problem: https://github.com/markdown-it/markdown-it/issues/2. We would have to wait until they fixed that but the repo is quite active so hopefully it will be fixed soon.

puzrin:
> I looked at different markdown implementations and read that the reference implementation in perl also misinterprets hashtags and a lot of implementations try to be compatible with the perl implementation. (afaik also pagedown)

I'd recommend to forget about "implementstions" and switch to "specification".

http://commonmark.org/

That will solve 99% your issues at once and forever.

> markdown-it has some problems with links which include non-English chars but that is a known problem: markdown-it/markdown-it#2. We would have to wait until they fixed that but the repo is quite active so hopefully it will be fixed soon.

I think ~ 2-3 months, would like to take timeout and see how people accept 3.0.0 release.

Problem is not to fix it fast, problem is to make it work with high speed, and to have nice api for extensions (skype, tel links and so on).

If you can't wait and highest speed is not critical, you can use https://github.com/kevva/url-regex + linkify rule src to quickly create your "right" plugin. That will not be universal & fast as i wish, but can work for you.

PS. When i say fast, i mean > 1mb/sec parse speed. Make sense for server side.

jhass:
Can't we just do the hashtag parsing before the markdown parsing? That should turn the hashtags into links which no longer should be recognized at headings.

puzrin:
> Can't we just do the hashtag parsing before the markdown parsing? That should turn the hashtags into links which no longer should be recognized as headings.

I don't recomment such approach. You need to understand doc structure to not break anything else. That's possible only at parse phase. In other word, you will reinvent your own markdown parser or will to dirty tricks with unsafe output :)

AFAIK, CommonMark spec requires space after ## in headers. You should have no conflicts with hashtag.

svbergerem:
@jhass Calling `markdownify` after `hashtagify` removes the links.

svbergerem:
There are also some other problems with our current approach. By calling `hashtagify` and `mentionify` after `markdownify` we also parse tags and mentions inside of code blocks and I think we shouldn't do that. I guess the cleanest way to do it right would be writing plugins for those two tasks for the parser so they are only called where we really want that.

svbergerem:
I added a temporary fix for markdown-it linkify which we can remove once that upstream function uses another library to find urls. We might also be able to get rid of punycode someday. 

Our specs succeed now. (except for some random failures) I'd say I'll do some cleanup and then this PR should be ready for a review.

app/assets/javascripts/app/helpers/markdown-it-diaspora-linkify.js
```diff
@@ -0,0 +1,180 @@
...
+        }
+      }
+    }
+  }
+
+  md.core.ruler.at('linkify', linkify);
+};
+
+},{"url-regex":1}]},{},[4])(4)
+});
```
jhass: Should probably go into lib/assets

app/assets/javascripts/app/helpers/text_formatter.js
```diff
@@ -1,146 +1,76 @@
...
-    return text.replace(utf8WordCharcters, function(result, linkTag, preceeder, tagText) {
-      if(linkTag)
-        return linkTag;
-      else
-        return preceeder + "<a href='/tags/" + tagText.toLowerCase() +
-                           "' class='tag'>#" + tagText + "</a>";
+    var hashtagPlugin = window.markdownitHashtag;
+    md.use(hashtagPlugin, {
+      // TODO is this [:alnum:] in ruby?
+      hashtagRegExp: '[\\x0080-\\xFFFF\\w\\-]+|<3',
```
svbergerem: Can someone help here? I think it is important that we apply the markup to the same hashtags that `ActsAsTaggableOn` accepts. We set those accepted chars to `[:alnum:]` but I wasn't able to find out what exactly the chars in `[:alnum:]` are so I used the same values we had before.

spec/javascripts/app/helpers/text_formatter_spec.js
```diff
@@ -97,21 +147,21 @@ describe("app.helpers.textFormatter", function(){
           "http://xn--4gbrim.xn----ymcbaaajlc6dj7bxne2c.xn--wgbh1c/",
           "http:///scholar.google.com/citations?view_op=top_venues",
           "http://lyricstranslate.com/en/someone-you-%E0%B4%A8%E0%B4%BF%E0%B4%A8%E0%B5%8D%E0%B4%A8%E0%B5%86-%E0%B4%AA%E0%B5%8B%E0%B4%B2%E0%B5%8A%E0%B4%B0%E0%B4%BE%E0%B4%B3%E0%B5%8D%E2%80%8D.html",
-          "http://de.wikipedia.org/wiki/Liste_der_Abk%C3%BCrzungen_%28Netzjargon%29",
+          "http://de.wikipedia.org/wiki/Liste_der_Abk%C3%BCrzungen_(Netzjargon)",
```
jhass: Why the test case change?

svbergerem: We replaced parentheses because otherwise pagedown had some problems.
```
// markdown doesn't like '(' or ')' anywhere, except where it wants
var workingUrl = unicodeUrl.replace(/\(/, "%28").replace(/\)/, "%29");
```
It looks like markdown-it doesn't have those problems so there was no need to replace them. I changed the test because the behavior changed.

app/assets/javascripts/app/helpers/markdown-it-diaspora-linkify.js
svbergerem: Sure, I'll move that file.

app/assets/javascripts/app/helpers/text_formatter.js
jhass: They're documented at the  [Regexp class](http://ruby-doc.org/core-2.2.0/Regexp.html) in the Ruby documentation:

> POSIX bracket expressions are also similar to character classes. They provide a portable alternative to the above, with the added benefit that they encompass non-ASCII characters. For instance, /\d/ matches only the ASCII decimal digits (0-9); whereas /[[:digit:]]/ matches any character in the Unicode Nd category.

So on [into the standard](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap09.html#tag_09_03_05):

> The set of single-character collating elements whose characters belong to the character class, as defined in the LC_CTYPE category in the current locale.

Oh well not too helpful. Source code diving time then! The engine used by Ruby for regular expressions is called oniguruma. Some digging into its code brings up: https://github.com/ruby/ruby/blob/trunk/enc/unicode/name2ctype.src#L3695

Uhm okay, someone had fun figuring that out there. 566, looks like the number of entries. Well, actually half of it, so we're dealing with pairs here. Talking about codepoints, must be ranges.

Onto some Ruby magic! [Let's crash everyones browser](http://paste.mrzyx.de/pzxah2po4)!

Now that was fun, wasn't it?

spec/javascripts/app/helpers/text_formatter_spec.js
jhass: k, thanks.

app/assets/javascripts/app/helpers/text_formatter.js
svbergerem: Thank you. I searched for "alnum" in the ruby repo but GitHub didn't list that file.

jhass: Yeah, the Github search is still not very trustworthy.

svbergerem:
The definition of `alnum` is now in  lib/assets/javascripts/posix-bracket-expressions.js. (I hope that is ok)
I did some cleanup so the code should be ready to review/merge.

This PR fixes #3374, #4201, #5216 and #5379.
It closes #5290 because we now have GFM style line breaks. It also closes #5457 because it removes pagedown.

I tested this with the following post:

<pre>
#5379
*This* **is** a test to show a b*u*g for devel**op**e(r)s
*This* **is** a test to show a b *u*g for devel**op**e(r)s
*This* **is** a test to show a b*u* g for devel**op**e(r)s

#4201
```
http://www.example.com
&lt;3
```

#5216
*http://example.com* **http://example.com** _http://example.com_

#3374
mailto:test@example.com
xmpp:test@example.com
</pre>

**before**:
markdown-it-before

**after**:
markdown-it-after

jhass:
<3 Awesome! Thank you!

Flaburgan:
@svbergerem this is really awesome. Thank you so much!

Flaburgan:
@svbergerem looks like there is some regressions, users started to use br element to have line-break (probably because they didn't know about double spaces), so now we have posts like that:

https://diaspora-fr.org/posts/935093

(Note that the bold also fails...) Same rendering in the SPV and in the stream btw.

svbergerem:
Yeah I know. markdown-it has an option to allow html input but that would allow too much. I am going to write a plugin similar to Markdown.Sanitizer.js (pagedown).

puzrin:
Bold fails because \<br> follows as text without space or punctuation.

@svbergerem all those hand-made sanitizers are very unsecure. That's why we didn't tryed to filter html at all - no chances to compete with owasp. Consider alternatives:

- plugin for `<br>` only, if you don't need anything else
- convert posts with br to new format (2 spaces or \ in the end of line)

Idea behind markdown-it is not use html at all. I think, historically people mixed html with markdown because it was almost impossible to extend syntax in other parsers.

jhass:
@puzrin then http://spec.commonmark.org/dingus.html?text=an%20**%28:**%3Cbr%3E is wrong?

puzrin:
[markdown-it dingus](https://markdown-it.github.io/#md64=eyJzb3VyY2UiOiJhbiAqKig6Kio8YnI+IiwiZGVmYXVsdHMiOnsiaHRtbCI6ZmFsc2UsInhodG1sT3V0IjpmYWxzZSwiYnJlYWtzIjpmYWxzZSwibGFuZ1ByZWZpeCI6Imxhbmd1YWdlLSIsImxpbmtpZnkiOnRydWUsInR5cG9ncmFwaGVyIjp0cnVlLCJfaGlnaGxpZ2h0Ijp0cnVlLCJfc3RyaWN0Ijp0cnVlLCJfdmlldyI6InNyYyJ9fQ==)

Hm... seems bold is buggy. Will fix soon. Need some time to understand, will it be enougth to add `<>` as punctuation or not.

`<br>` in reference dingus is rendered because it has no option to disable html at all. With all question about seciruty they say "use sanitizer, it's not parser task".

jhass:
><br> in reference dingus is rendered because it has no option to disable html at all.

Sure, that wasn't the point, the markdown-it dingus doesn't render it properly in any case, whether HTML is allowed, not allowed or the strict mode is enabled.

puzrin:
Fixed in 3.0.3. Thanks for report!

jhass:
Thanks for the quick update!

PR #5535, opened 2015-01-09
Use the standard syntax for gradient

dimaursu:


jhass:
Thanks

Flaburgan:
Hm, still not fixed imo. This is how it looks now:
capture du 2015-01-09 16 01 00
This is how it looked before the first PR:
capture du 2015-01-09 16 01 35

PR #5541, opened 2015-01-10
Remove chars from PosixBracketExpressions that js doesn't like

svbergerem:


jhass:
Willing to merge that as a quick workaround, but we should eventually support these scripts.

PR #5547, opened 2015-01-11
Buttons for ignore, hide, delete on SPV

Faldrian:
This is my attempt to add "delete own post", "ignore user" and "hide this post" buttons to the single post view. This will also move the "report post" button to a semantic better place - away from normal post interaction buttons.

There are some open points where I need further feedback on how to do this right:
- Position of new buttons
- If "delete post", "hide post" or "ignore user" action is performed, the browser will load "/stream" on success to lead the user away from the unwanted / deleted post. There may be need for further info messages?

(Also there are no tests yet.)

Screenshot of other users posts:
otehrpost

Screenshot of own post:
ownpost


Flaburgan:
>Position of new buttons

That's a good place in my opinion

>If "delete post", "hide post" or "ignore user" action is performed, the browser will load "/stream" on success to lead the user away from the unwanted / deleted post. There may be need for further info messages?

We certainly need a confirmation message depending of the action performed. And if you're able to scroll back to the place the stream was before going to the SPV, it would be awesome.

Thanks for this very valuable work!

Faldrian:
I know we need a message - I just don't know how to implement. ;)

If I change back to the stream, I need to have a way to show a toast-message when the steam is loaded. It seems to me there is no way to tell the "/stream" page that it should show such a message - by simply navigating to "/stream".
There are examples in current code where a message is displayed after navigation to a different url, but this is (as far as I know) always connected to some action being done by the loaded page (instead of doing the action via ajax and then calling the page normally).

So here I need input how to implement / show a confirmation message when an action was done.

> And if you're able to scroll back to the place the stream was before going to the SPV, it would be awesome.

That won't be possible, because SPV is not necessarily opened from the stream but may be opened by direct link, in a new tab etc.

Flaburgan:
>That won't be possible, because SPV is not necessarily opened from the stream but may be opened by direct link, in a new tab etc.

So we need a "if" ;)

About the message, you can check the "You're already logged in": logout and open diaspora in two tabs, then log in in each tabs. You'll have the "You're already logged in" message in the second one (if I remember correctly). So this is an example of a message displayed in the stream from a different page.

Faldrian:
> So we need a "if" ;)

No, we just don't need that feature.

> About the message, you can check the "You're already logged in": logout and open diaspora in two tabs, then log in in each tabs. You'll have the "You're already logged in" message in the second one (if I remember correctly). So this is an example of a message displayed in the stream from a different page.

When you log in your browser, you sent a POST request stating what you want to do to the server. So the server can device you already have an open session (depending on the cookie) and display that message when the site is loaded.

In my case that will not work, because the server knows nothing where the user came from or what the user has done when "/stream" will be loaded. All requests and action are done before via ajax and only then the "/stream" is navigated to.
I also don't want to use other methods to block, ignore and delete posts / users, because the existing methods are okay and one single point to maintain.

Faldrian:
Maybe @svbergerem may have an idea how to make usable notifications when an action (delete own post / hide post / block author) has gone through and the user will be forwarded to "/stream" in the next seconds or so?

Maybe show a toast message with "This post was hidden. You will be forwarded to your stream in 5s." and have the site forward after the timeout?

jaywink:
I don't see why we need a message - the user is already asked a confirmation and accepts that. It should be enough to "figure out" what happened and why he/she is now in the stream.

Great work :)

Faldrian:
Okay... :)
So I will add some tests this weekend and this PR should be good to go?

app/assets/javascripts/app/views/feedback_view.js
```diff
@@ -40,7 +43,40 @@ app.views.Feedback = app.views.Base.extend({
...
+    $.ajax({
+      url : "/share_visibilities/42",
+      type : "PUT",
+      data : {
+        post_id : this.model.id
+      }
+    });
+
+    // return to stream
+    document.location.href = "/stream";
```
jaywink: Shouldn't this trigger only after ajax comes back as success?

app/assets/javascripts/app/views/feedback_view.js
```diff
@@ -40,7 +43,40 @@ app.views.Feedback = app.views.Base.extend({
...
+  },
+
+  blockUser: function(evt) {
+    if(evt) { evt.preventDefault(); }
+    if(!confirm(Diaspora.I18n.t('ignore_user'))) { return; }
+
+    this.model.blockAuthor()
+      .done(function() {
+        // return to stream
+        document.location.href = "/stream";
```
jaywink: I guess there is no way to not hardcode links into the JS files... tried searching for existing ones, couldn't seem to figure out if we have anything. It would be nice to push all app paths to gon at some point, but that is probably not in scope of this.

jaywink: On second thought, I guess this might be solved by backbonification of the stream.

app/assets/javascripts/app/views/single-post-viewer/single_post_moderation.js
```diff
@@ -0,0 +1,47 @@
+app.views.SinglePostModeration = app.views.Feedback.extend({
```
jaywink: Maybe a screenshot of this view in action? :)

app/assets/javascripts/app/views/feedback_view.js
Faldrian: I implemented this analog to the hidePost in `stream_post_view.js` - maybe it would be good to change that function, too. Remove / Redirect on success, show an error toast if something went wrong?

jaywink: Yeah that sounds safer to me :)

jaywink:
Looks good apart from the comments! :beers:

app/assets/javascripts/app/views/feedback_view.js
Faldrian: Maybe I could introduce some `app.const.locations.stream` and set it to "/stream"... but we I would like to hear other opinions on how to do it... or would do it in another PR for the whole JS-part in one go. We need a proper naming scheme until then.

app/assets/javascripts/app/views/single-post-viewer/single_post_moderation.js
Faldrian: This view is only the little part of the screenshots where the 3 icons [hide, block, report] (or 1 delete icon) is displayed. So: There is already a screenshot. :)

app/assets/javascripts/app/views/feedback_view.js
jhass: We actually have [js-routes](https://github.com/railsware/js-routes) in the Gemfile, someone(tm) should go through the hardcoded routes and update them...

Faldrian: I add that to my list...

Faldrian:
I added tests for all 4 buttons and when the tests go green (seems like some test involving hovercards is failing for several builds...?) it could be merged. :)

jaywink:
Great work, works very well, tested locally. One more request though, could you please squash the 4 commits to one for a cleaner commit log? Can then do the merge.

Thanks! :)

jaywink:
:beers: to you, thanks!

Flaburgan:
Hey! Thank you for working on this, this is a really nice feature.
Looks like there is place for an improvement though:

screen

When a post is reshared, the actions can be made on the reshare, not on the original post. So, like the like, comment and reshare icons, the icons added by this pull request should be displayed in the reshare div, not in the original div.

Faldrian:
Seems logical, thanks for pointing that out. I will add it to my list ... and come back to it when I have time.

Flaburgan:
Sure ;) just point if you don't have time in the coming weeks as this is already merged in develop.

Faldrian:
If you want you can also do this change (and add tests for it) - I'm currently working on another issue.

Flaburgan:
I'm also working on a different issue ^^ but we'll see.

jaywink:
@fla, care to file a bug so we can track that?

On 25.01.2015 22:48, Fla wrote:
>
> I'm also working on a different issue ^^ but we'll see.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/diaspora/diaspora/pull/5547#issuecomment-71392621>.
>

-- 
-----
Br,
Jason Robinson
https://jasonrobinson.me

PR #5556, opened 2015-01-13
Improvements to the mobile drawer

dimaursu:
#5523 from a proper branch


Flaburgan:
Reviewing this, you still does not translate the stream itself when you open the drawer. I find this weird.

app/assets/stylesheets/mobile/header.css.scss
```diff
@@ -169,10 +163,19 @@ header {
...
+    }
+}
+
+#drawer, #main_nav {
+    z-index: 10;
+
+            transition: transform 0.3s ease;
+    -webkit-transition: -webkit-transform 0.3s ease;
+             transform: translateX(0);
+     -webkit-transform: translateX(0);
```
Flaburgan: Why do you need to "translate back"? When the .draw class is removed, the translate(-80%) is removed too, so the drawer should go back to his place.

dimaursu:
I had to reasons not to push the content:
1. it's more common on phones to slide over
http://www.google.com/design/spec/patterns/navigation-drawer.html 
2. it has a performance benefit to it,  it minimizes redraws

app/assets/stylesheets/mobile/header.css.scss
dimaursu: I think I did it for 2 reasons: 
1. to promote the drawer to in a separate rendering layer
2. To have the slide-back animation (not sure about this one, though)

Flaburgan:
Just wanted to point that this is not a simple refactor but there is a change, so want to hear opinion here :) I agree that it's better to not move it performance wise, but it gives me a better feeling to see everything translated.

app/assets/stylesheets/mobile/header.css.scss
Flaburgan: >to promote the drawer to in a separate rendering layer

Not sure what you mean

>To have the slide-back animation (not sure about this one, though)

You'll have the animation without that, see the current version ;)

svbergerem:
I tried this version with the FirefoxOS simulator and I like it. The swipe events don't work but that could be because of the simulator. It would be great if you could rebase this PR and maybe squash the commits to make it easier for others to test this PR. (there were a few conflicts)

Flaburgan:
Yeah I would like to see this merged too, I'm fine with the "no pushing" the content. I still think that we don't need to specifically translate back though (as commented in the code before).

dimaursu:
Ok, I'll do it in a tonight. Swipes work on most real divices I've tested on, including an android tablet, 2 android phones, and an iPhone 4s.

dimaursu:
Done :) Rebased and pushed. Although I would rather support using a fixed-width drawer. I had some experiments listed on loomio, with the site on a tablet

app/assets/javascripts/mobile/mobile.js
```diff
@@ -37,10 +32,18 @@ $(document).ready(function(){
 
   /* Drawer menu */
   $('#menu_badge').bind("tap click", function(evt){
-    evt.preventDefault();
     $("#app").toggleClass('draw');
   });
 
+  var drawerSlide = new Hammer(document.getElementById('app'));
```
svbergerem: Please add "Hammer" in `config/.jshint.json` to `predef`.

app/assets/javascripts/mobile/mobile.js
```diff
@@ -12,12 +12,7 @@
 //= require jquery.autoSuggest.custom
 //= require fileuploader-custom
 //= require rails-timeago
-//= require underscore
-//= require diaspora
-//= require helpers/i18n
-//= require widgets/timeago
-//= require mobile/mobile_file_uploader
-//= require mobile/profile_aspects
```
svbergerem: I think you removed these lines by accident.

svbergerem:
We don't have a SCSS styleguide yet, but it would be great if you could fix the indentation errors.

denschub:
@dimaursu Ping!

jhass:
Superseded by #6233

PR #5559, opened 2015-01-14
Readd aspect sorting

SWW13:
(Re)Added aspect sorting with minimal jquery ui.
- Please have a look at the gem file, i haven't got it working with a version added:
  
  > Could not find gem 'jquery-ui-rails5.0.3 (>= 0) ruby' in the gems available on this machine.
- jQuery AJAX request throws js warning because of the empty response.


denschub:
Thanks for your work. I did took a closer look, but two hints:

- Please specify a version in the `Gemfile` to avoid confusions with gem version upgrades
- Please write tests for any new functionality you implement. You can find some information about our testing workflow [in the wiki](https://wiki.diasporafoundation.org/Testing_workflow), you can also ping us in `#diaspora-dev` on freenode if you need help.

svbergerem:
@SWW13 Oh, you updated this PR. :) Please add a comment when you fix stuff and think that we should review the PR again.

I tested your code on my dev pod and it is working really well. Your js code is only needed for the contacts page so I think you can move it to `app/assets/javascirpts/app/pages/contacts.js`. Maybe add it to a function called `setupAspectSorting` or something similar and call that function from the initialize function.

You added a test for the aspect_order function you wrote. I think it would be great if there would be another test to make sure that the drag and drop in the UI really reorders the aspects and that the order stays the same when you reload the page. I guess that could be a cucumber test. However it looks like the `drag_to` method doesn't work with jQuery UI sortable. There is a [workaround](https://github.com/mattheworiordan/jquery.simulate.drag-sortable.js) but I don't know what would be the best way to integrate that so someone else would have to help.

You added the correct version to the `Gemfile` but in `Gemfile.lock` the version is still missing. I guess calling bundle another time could fix that problem.

The ruby code has to be reviewed by someone else.

app/models/user.rb
```diff
@@ -424,6 +424,14 @@ def guard_unconfirmed_email
     end
   end
 
+  def reorder_aspects(aspect_order)
+    i = 0
+    aspect_order.each do |id|
+      self.aspects.find(id).update_attributes({ :order_id => i })
+      i += 1
+    end
+  end
```
jhass: I'm not sure about the JS part, somebody else has to check that, but I think this should get its own route on `AspectsController` and this method can be nicer written by using `each_with_index`. There's also no need for the `self.` and I'd name the parameter `ordered_aspect_ids`.

config/routes.rb
```diff
@@ -63,6 +63,7 @@
   get "commented" => "streams#commented", :as => "commented_stream"
   get "aspects" => "streams#aspects", :as => "aspects_stream"
 
+  put "aspects/order" => "aspects#update_order"
```
jhass: Please move this into the block below as `put "order" => :update_order`

features/step_definitions/aspects_steps.rb
```diff
@@ -88,6 +88,21 @@ def aspect_dropdown_visible?
...
 end
 
+When /^I drag "([^"]*)" (up|down) (\d+) pixels?$/ do |aspect_name, direction, distance|
+  distance = distance.to_i * -1 if direction == 'up'
+  page.execute_script %{
+    $.when(
+      $.getScript("/assets/jquery-simulate/jquery.simulate.js"),
+      $.getScript("/assets/jquery-simulate-ext/src/jquery.simulate.ext.js"),
+      $.getScript("/assets/jquery-simulate-ext/src/jquery.simulate.drag-n-drop.js"),
+      $.Deferred(function( deferred ){ $(deferred.resolve); })
```
jhass: So what happens if the step is called multiple times without reloading the page?

jhass:
Thank you!

denschub:
:cookie:

Issue #5560, opened 2015-01-15
Hide community spotlight setting if not enabled


goobertron:
The setting 'Show Community Spotlight in stream' on the `/user/edit` view is displayed whether or not the podmin has enabled Community Spotlight for users of that pod.

It should be displayed as a setting only if the pod has enabled Community Spotlight.

Flaburgan:
It is not the only place it appears even if the Community Spotlight are not enabled. If you have an empty stream for example, the message will advice you to add contact and check the community spotlight.

PR #5562, opened 2015-01-16
5560 Hide community spotlight setting if not enabled

sakshi-keepworks:


app/views/aspects/_no_contacts_message.haml
```diff
@@ -7,5 +7,6 @@
   %br
   %br
   = t('.try_adding_some_more_contacts')
-  != t('.or_spotlight', :link => link_to(t(".community_spotlight") , community_spotlight_path))
+  - if current_user.show_community_spotlight_in_stream?
```
Flaburgan: This is not the good condition, we want to hide this message if the podmin didn't activate community spotlight in diaspora.yml. Besides, #5560 is about the setting in the user panel, not this message, but you are right, both have to be hidden.

svbergerem:
You are now hiding the information that one could also add contacts from the community spotlight if someone doesn't want to see the community spotlight in its stream? I don't think that was the behavior #5560 was referring to.

This message and the setting itself should be hidden if the community spotlight are disabled for the whole pod. (See https://github.com/diaspora/diaspora/blob/develop/app/views/contacts/_sidebar.html.haml#L4)

Flaburgan:
@svbergerem I was first! :p

svbergerem:
@sjain1107 Btw. it is great to see you contributing again. :)

sjain1107:
Thanks :) I'm happy that atleast something is correct ;)

sjain1107:
Question: Where are the settings and how to enable or disable community spotlight for the whole pod?

svbergerem:
Settings:
https://github.com/diaspora/diaspora/blob/develop/app/views/users/edit.html.haml#L88
(header dropdown -> settings -> stream settings)

Enable/Disable the community spotlight for the pod:
https://github.com/diaspora/diaspora/blob/develop/config/diaspora.yml.example#L430

Flaburgan:
So the test you want to write is `- if AppConfig.settings.community_spotlight.enable?`

goobertron:
Wow, that was quick! Thanks for your PR, Sakshi, and it's good to see you contributing again.

sjain1107:
@svbergerem @Flaburgan @goobertron Thanks! Will look into it.

jhass:
Two simple controller tests that assert the checkbox is included in the response or not depending on the setting would be nice :)

sjain1107:
@jhass Added tests. Please review. :)

jhass:
Great, thank you!

PR #5574, opened 2015-01-18
Fix markdown-it linkify blocking Firefox

svbergerem:


lib/assets/javascripts/markdown-it-diaspora-linkify.js
```diff
@@ -2,69 +2,11 @@
...
-},{}],4:[function(require,module,exports){
 // Replace link-like texts with link nodes.
 //
 'use strict';
 
-var urlRegex = require('url-regex');
-var LINK_SCAN_RE = /www|@|\:\/\//;
+var urlRegex = /(?:^|\s)(?:(?:https?|ftp):\/\/|mailto:|xmpp:|www[^.\s]*\.)\S+/gi;
```
svbergerem: please review especially this line

jhass: Didn't dig much into this one yet, but how about reusing the old one? https://github.com/diaspora/diaspora/blob/master/app/assets/javascripts/app/helpers/text_formatter.js#L4

svbergerem: That doesn't support `mailto:`, `xmpp:` and `wwwb` etc. It also doesn't support parenthesis inside of URLs.

jhass: I fear this one is a bit too open though: https://www.debuggex.com/r/8859oQ_AME577XSr

We're currently using [twitter-text-rb](https://github.com/twitter/twitter-text) in the backend for URL recognition, it also has a [JS implementation](https://github.com/twitter/twitter-text/tree/master/js), I think it might be worth investigating what they use or maybe even utilizing the library.

svbergerem: I am not sure if this is really much worse than the regex we had before: https://www.debuggex.com/r/4eJ2IFhc7VXnEOKu

I could add a mandatory `@` for `mailto:` and `xmpp:` in the regex. (`mailto:\S+@\S+`)
Which protocols does twitter-text-rb support?

jhass: The matching is generic ([JS](https://github.com/twitter/twitter-text/blob/master/js/twitter-text.js#L424), [JS](https://github.com/twitter/twitter-text/blob/master/js/twitter-text.js#L360), [Rb](https://github.com/twitter/twitter-text/blob/master/rb/lib/twitter-text/regex.rb#L374), [Rb](https://github.com/twitter/twitter-text/blob/master/rb/lib/twitter-text/regex.rb#L302)) but for some reason demands the `://` part and is then later restricted to `https?` ([JS](https://github.com/twitter/twitter-text/blob/master/js/twitter-text.js#L1368), [Rb](https://github.com/twitter/twitter-text/blob/master/rb/lib/twitter-text/validation.rb#L96)) unfortunately, so we would probably need to fork it.

Raven24: I remember we had a browser lock-up issue with the old regex in chrome with some greedy expressions, which was a pain to debug.
so any out-of-the-box solution we can find is preferrable ;)

svbergerem: The last out-of-the-box solution was a disaster. (and is the reason why this PR exists) I found no existing solutions so far that had satisfying results.

svbergerem: Please also note that the regex you commented on is not the current one for this PR.

jhass:
Let's try this one out for some time.

PR #5576, opened 2015-01-18
Add emoji support

dimaursu:
This will convert common stuff like :) or :P into emojis. By default it uses the images from the local assets - they are really small, around 0.61 KB / emoji.

The technical burden will be small for d*, as I used a plugin for markdown-it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,7 +58,7 @@ gem 'autoprefixer-rails', '4.0.2.1'
 
 # Database
 
-ENV['DB'] ||= 'mysql'
+ENV['DB'] ||= 'postgres'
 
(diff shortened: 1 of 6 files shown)


jhass:
This is not what the currently running proposal talks about, and I disagree with it even more than the proposal, I'm sorry.

jhass:
Given there has been no recent discussion around this feature, that you so far haven't addressed the technical problems of this implementation and that we got a [community decision](https://www.loomio.org/d/vlmyMTNd/emoji-support?proposal=CRoUdzEv) not overwhelmingly favoring the general idea, I'm closing this PR for now. Thank you for you effort so far! Also feel free to open a new PR with the above points addressed, once there's consensus in the community about the how.

PR #5584, opened 2015-01-24
Added environment-variable to specify Firefox version for testing

Faldrian:
Implemented this workaround [1] , so we can still use Firefox 34 for testing (since Selenium Driver is currently broken for Firefox 35).

To use, define environment-variable `FIREFOX_BINARY_PATH` pointing to the firefox executable which should be used.

1: http://blog.pixarea.com/2013/02/locking-the-firefox-version-when-testing-rails-with-capybara-and-selenium


jhass:
Thank you!

PR #5587, opened 2015-01-25
Fixes the header problem rendering statistics page

ghost:
Related to [this comment](https://github.com/diaspora/diaspora/pull/5464#issuecomment-71365544)


jaywink:
Great, this fixes that issue. BUT, not sure if this was the case before, but the header is not rendering when *logged out*.

selection_048

jaywink:
Yes it was visible before, I had the old test tab open :)

selection_049

After refreshing it, header disappeared.

SansPseudoFix:
Doesn't works with mobile.

AugierLe42e:
Should be better now.
When logged in :

screen shot 2015-01-25 at 13 27 17

When logged out :

screen shot 2015-01-25 at 13 28 07

Mobile view when logged in :

screen shot 2015-01-25 at 13 29 47


Mobile view when logged out :

screen shot 2015-01-25 at 13 29 03

AugierLe42e:
Mobile view is ugly as boostrap isn't injected.

jaywink:
Travis one build failed due to flaky not related test, so merging. Thanks for fixing this so fast!

jaywink:
> Up to you to write some CSS now :p

LOL, missed this commit. But then, that beautification can be done separately - now it at least works :) IMHO the mobile page looks ok anyway even if the styling is different.

AugierLe42e:
@jaywink : You're welcome, this was my fault. I didn't do the proper tests ;)

Issue #5589, opened 2015-01-26
Wrong information on the new (HTML) statistic page


marienfressinaud:
I can see two types of errors:

- Registrations are shown as opened while they are not
- Services are all "Not available" while they are!

For example, you can compare Diaspora-fr information:

- That's ok on the JSON page: https://diaspora-fr.org/statistics.json
- But not on the HTML page: https://diaspora-fr.org/statistics

Related to https://github.com/diaspora/diaspora/pull/5464

jaywink:
Also ummmm active users *monthly* cannot be higher than active users *half a year* :P Something has changed there too.

Flaburgan:
Yeah @jaywink and this is affecting statistics.json which should not have been changed...

marienfressinaud:
True. Another thing (it may be a feature, I don't know): on the JSON page there is a "services" array AND, at the end of the line, services with a boolean information. I guess services array is excessive…

Flaburgan:
@marienfressinaud the service with a boolean is the old version (0.4), the array the new one. We keep both in 0.5 to be retrocompatible, 0.6 will only have the array.

AugierLe42e:
@Flaburgan : JSON has changed as it include a big refactor by @jhass...

jaywink:
JSON errors fixed by #5590

AugierLe42e:
If nobody solved it before, I'll try to fix this tonight. I have an idea where it can come from, but I'm at work now.

Flaburgan:
I fixed the monthly / halfyear and @AugierLe42e the services but we're still having the issue about registration: they are closed on diaspora-fr but displayed as open.

Flaburgan:
PR from @AugierLe42e is now merged, this can be closed.

PR #5590, opened 2015-01-26
Fix monthly and halfyear active users count

Flaburgan:
Fix regression introduced in https://github.com/diaspora/diaspora/pull/5464 or https://github.com/diaspora/diaspora/pull/5587, see https://github.com/diaspora/diaspora/issues/5589

### DIFF
--- a/app/presenters/statistics_presenter.rb
+++ b/app/presenters/statistics_presenter.rb
@@ -55,11 +55,11 @@ class StatisticsPresenter
   end
 
   def monthly_users
-    @monthly_users ||= User.halfyear_actives.count
+    @monthly_users ||= User.monthly_actives.count
   end
 
   def halfyear_users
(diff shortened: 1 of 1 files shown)


Flaburgan:
I worry because this breaking change doesn't appear in the git diff of the 2 pull requests. That means that other changes like that could have been merged.

Why don't we have tests on that btw, perfect time to write some @AugierLe42 @SansPseudoFix

jaywink:
This is the commit that broke things, little copy paste accident which no one noticed :) https://github.com/diaspora/diaspora/commit/dc444852307cea66884c8b5aad220e4ec37c9207

We have a test for the statistics page, but not for the scope which collects the user counts in the model. Not quite sure how efficient adding tests to those would be to prevent errors like this..

Anyway, thanks @Flaburgan - merging!

PR #5595, opened 2015-01-26
Fix available services not correctly displayed

### LINKED ISSUE #5589: Wrong information on the new (HTML) statistic page

ghost:
This should fix #5589

### DIFF
--- a/app/helpers/statistics_helper.rb
+++ b/app/helpers/statistics_helper.rb
@@ -12,7 +12,7 @@ module StatisticsHelper
   end
 
   def service_status service, available_services
-    if available_services.include? service
+    if available_services.include? service.to_s
       I18n.t('statistics.enabled')
     else
       I18n.t('statistics.disabled')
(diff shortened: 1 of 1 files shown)


jhass:
service is guaranteed to be a symbol, since that array is actually a hardcoded constant. Therefore I'd prefer the helpers to be updated to expect a symbol instead of a string.

AugierLe42e:
Like this ?

jhass:
Yup, thanks!

Btw. your current commit author email is not associated to your github account ;)

AugierLe42e:
This is because I don't want it to be displayed publicly on github.

PR #5596, opened 2015-01-26
Change sign up from button to link

dmorley:
Like many I hit signup every time when I want to login on mobile since the button is a button just like the signin button

### DIFF
--- a/app/views/sessions/new.mobile.haml
+++ b/app/views/sessions/new.mobile.haml
@@ -39,7 +39,7 @@
             .controls
               = f.submit t('devise.sessions.new.sign_in'), :class => 'btn primary'
               - if display_registration_link?
-                = link_to t('devise.shared.links.sign_up'), new_registration_path(resource_name), :class => 'btn primary', :style => "float: right;"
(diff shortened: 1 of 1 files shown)


app/views/sessions/new.mobile.haml
```diff
@@ -39,7 +39,7 @@
             .controls
               = f.submit t('devise.sessions.new.sign_in'), :class => 'btn primary'
               - if display_registration_link?
-                = link_to t('devise.shared.links.sign_up'), new_registration_path(resource_name), :class => 'btn primary', :style => "float: right;"
+                = link_to t('devise.shared.links.sign_up'), new_registration_path(resource_name), :class => 'primary', :style => "float: right;"
```
svbergerem: The `primary` class is useless here. I'd suggest going the Bootstrap way and using `btn btn-link` for the class attribute.

jhass:
@dmorley ping, still interested in getting this in?

dmorley:
@jhass if "primary" is dead but still in the production code then my modification is silly and I am not up to speed with the code base to do more than remove the btn attribute.

PR #5599, opened 2015-01-27
Fix registrations displaying

ghost:
This should fix the registration problem #5589

### DIFF
--- a/app/helpers/statistics_helper.rb
+++ b/app/helpers/statistics_helper.rb
@@ -11,6 +11,14 @@ module StatisticsHelper
     end
   end
 
+  def registrations_status_class statistics
+    if statistics.open_registrations?
+      "serv-enabled"
+    else
+      "serv-disabled"
+    end
+  end
+
   def service_status service, available_services
     if available_services.include? service.to_s
       I18n.t('statistics.enabled')
--- a/app/presenters/statistics_presenter.rb
+++ b/app/presenters/statistics_presenter.rb
@@ -31,7 +31,7 @@ class StatisticsPresenter
   end
 
   def open_registrations?
-    AppConfig.settings.enable_registrations
+    AppConfig.settings.enable_registrations?
   end
 
   def user_counts
--- a/app/views/statistics/_statistics.haml
+++ b/app/views/statistics/_statistics.haml
@@ -7,7 +7,7 @@
     = t('_statistics')
   = render 'statistics/statistic', name: t('statistics.name'), value: @statistics.name, activated: "serv-enabled"
   = render 'statistics/statistic', name: t('statistics.version'), value: @statistics.version, activated: "serv-enabled"
-  = render 'statistics/statistic', name: t('statistics.registrations'), value: registrations_status(@statistics), activated: "serv-enabled"
+  = render 'statistics/statistic', name: t('statistics.registrations'), value: registrations_status(@statistics), activated: registrations_status_class(@statistics)
   - if @statistics.expose_user_counts?
     = render 'statistics/statistic', name: t('statistics.total_users'), value: @statistics.total_users, activated: "serv-enabled"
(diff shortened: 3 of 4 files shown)


app/helpers/statistics_helper.rb
```diff
@@ -4,13 +4,21 @@
 
 module StatisticsHelper
   def registrations_status statistics
-    if statistics.open_registrations?
+    if !!statistics.open_registrations?
```
dimaursu: um? That's double negation, it's the same as the original.

app/helpers/statistics_helper.rb
```diff
@@ -4,13 +4,21 @@
...
       I18n.t('statistics.open')
     else
       I18n.t('statistics.closed')
     end
   end
 
+  def registrations_status_class statistics
+    if !!statistics.open_registrations?
```
dimaursu: Same as before, double negation

app/helpers/statistics_helper.rb
AugierLe42e: Yeah, problem is : function recieve a symbol, so, as is is not a fasly value, it is always considered true, even if registrations are closed. `!!` "operator" forces cast to boolean. Works well ;)

jhass: * If you can call `open_registrations?` on `statistics` that's certainly not a symbol.
* If  `open_registrations?` returns a symbol, `!!symbol` is always `true`.
* if `open_registrations?` returns nil or a symbol, nil is falsey and the symbol is truthy, so the `!!` is unneeded.

AugierLe42e: Well I can't explain but I can assure you that it works with and doesn't work without :/

jhass: Add `require "pry"; binding.pry` in the line before that and trigger the code path.

jhass: Or just add a `?` to https://github.com/diaspora/diaspora/blob/develop/app/presenters/statistics_presenter.rb#L34 to properly fix it.

Also add an RSpec test that triggers the bug.

app/presenters/statistics_presenter.rb
```diff
@@ -31,7 +31,7 @@ def version
   end
 
   def open_registrations?
-    AppConfig.settings.enable_registrations
+    AppConfig.settings.enable_registrations?
```
AugierLe42e: @jhass : Nope, [see by yourself](https://travis-ci.org/diaspora/diaspora/jobs/48826537#L447), it doesn't work at all, plus it makes 2 other tests to fail.

app/presenters/statistics_presenter.rb
```diff
@@ -31,7 +31,7 @@ def version
   end
 
   def open_registrations?
-    AppConfig.settings.enable_registrations
+    AppConfig.settings.enable_registration?
```
AugierLe42e: @jhass : Nope, [see by yourself](https://travis-ci.org/diaspora/diaspora/jobs/48867657#L453). It doesn't work, plus it makes to other tests to fail. :/

jhass: Take a close look at the line you commented on.

AugierLe42e: Ok, are you willin to continue asking me riddles or will you eventually help me ?
[You](https://github.com/diaspora/diaspora/pull/5599#discussion_r23799476) told me tu put a `?` at the end of that line and I did and it doesn't work. So I'm sorry if I'm missing something but right now, I've got better to do than playing puzzles.

jhass: You didn't add a `?`, you replaced the `s` with it.

AugierLe42e: Hmmf... You are right, sorry.

AugierLe42e:
Ok, I think it can be merged now.

jhass:
Thanks!

AugierLe42e:
Thanks to you.

PR #5602, opened 2015-01-28
Add shortcuts for reshare, expand Post and open first link in a post

Faldrian:
This PR will add keyboard shortcuts for:
- r - reshare
- m - expand post (collapsed posts)
- o - open first link in the selected post

The [loomio discussion](https://www.loomio.org/d/DXzZyQGT/add-change-keyboard-shortcuts#comment-83862) said that we should not use direct key strokes ... but I think the way to go would be to have a user option to enable / configure shortcuts / disable keyboard navigation. This could be done in a separate issue, the implementation in this PR is following existing features.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -154,6 +154,7 @@ diaspora.yml file**. The existing settings from 0.4.x and before will not work a
 * Hide user setting if the community spotlight is not enabled on the pod [#5562](https://github.com/diaspora/diaspora/pull/5562)
 * Add HTML view for pod statistics [#5464](https://github.com/diaspora/diaspora/pull/5464)
 * Added/Moved hide, block user, report and delete button in SPV [#5547](https://github.com/diaspora/diaspora/pull/5547)
+* Added keyboard shortcuts r(reshare), m(expand Post), o(open first link in post) [#5602](https://github.com/diaspora/diaspora/pull/5602)
 
 # 0.4.1.2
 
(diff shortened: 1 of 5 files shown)


Faldrian:
Any comments, problems, time to review and merge? :)

svbergerem:
Codewise this looks good to me but I haven't tried it on my dev pod yet.

Does anyone have a problem with the proposed shortcuts? ( @jhass As far as I remember you thought that `r` should reload the stream.)

Faldrian:
> Does anyone have a problem with the proposed shortcuts? ( @jhass As far as I remember you thought that r should reload the stream.)

F5 and Ctrl+r already reload "everything" ... reloading only the stream would be a nice feature, Twitter has it on "." which is at first a bit odd, but when you think of navigating with j/k the "." key is more reachable than r. I would prefer these direct shortcut keys (**r**eshare, **m**ore, **l**ike, **o**pen) for the actions regarding to posts and the other not-so-direct keys for things where you maybe should know what you are doing or what you want to archieve. ;)

svbergerem:
@Faldrian Sure, I agree. I'd just like to check if anyone has any objections because I don't want to merge a (possibly) controversial PR. If everyone is happy with the proposed shortcuts I will test the code on my dev pod and merge it.

Faldrian:
@svbergerem Of couse. ^^ I just wanted to bring some arguments to explain my decision to use these keys. :) Just in case or for better understanding here. Of course this should only merged if there are no objections.

jhass:
I guess it'll be controversial no matter how we map them, so I'd say we'll just try it out ;)

svbergerem:
Thank you!

Issue #5529, opened 2015-01-06
Add direct links to /help


rhaglion:
We already have a great page where many questions of newcomers are answered. But at the moment there is no possibility to link to an specific question.

So if you like to refer to a an single question you have to write something like:
>    Go to help, click on Aspects and open the third question

Links to the specific questions would be an great improvement.

PR #5603, opened 2015-01-28
Added routing system to open a section in help

ghost:
Directly open a given section of help by routing. E.g : /help/mentions will open mentions section of help.

Would be cool to be able to rewrite the route when opening another section.

Implements #5529.

### DIFF
--- a/app/assets/javascripts/app/router.js
+++ b/app/assets/javascripts/app/router.js
@@ -2,6 +2,8 @@
 
 app.Router = Backbone.Router.extend({
   routes: {
+    "help/:topic": "help",
+    "help/": "help",
     "help": "help",
     "contacts": "contacts",
     "conversations": "conversations",
@@ -36,10 +38,10 @@ app.Router = Backbone.Router.extend({
     this.route(/^bookmarklet(?:\?(.*))?/, "bookmarklet");
   },
 
-  help: function() {
+  help: function(topic) {
     app.help = new app.views.Help();
     $("#help").prepend(app.help.el);
-    app.help.render();
+    app.help.render(topic);
   },
 
   contacts: function() {
--- a/app/assets/javascripts/app/views/help_view.js
+++ b/app/assets/javascripts/app/views/help_view.js
@@ -9,7 +9,7 @@ app.views.Help = app.views.StaticContentView.extend({
     "click .faq-link-sharing" : "sharing",
     "click .faq-link-posts-and-posting" : "postsAndPosting",
     "click .faq-link-tags": "tags",
-    "click .faq-link-keyboard-shortcuts" : "keyboardShortcuts",
+    "click .faq-link-keyboard-shortcuts" : "keyboardShortcuts"
   },
 
   initialize : function() {
@@ -19,19 +19,19 @@ app.views.Help = app.views.StaticContentView.extend({
       get_support_a_tutorials: { tutorials: this.linkHtml("https://diasporafoundation.org/tutorials", Diaspora.I18n.t( 'tutorials' ))},
       get_support_a_wiki: { link: this.linkHtml("https://wiki.diasporafoundation.org/Special:Search", Diaspora.I18n.t( 'wiki' ))},
       get_support_a_irc: { irc: this.linkHtml("https://wiki.diasporafoundation.org/How_We_Communicate#IRC", Diaspora.I18n.t( 'irc' ))},
-      get_support_a_hashtag: { question: this.linkHtml("/tags/question", "#question")},
+      get_support_a_hashtag: { question: this.linkHtml("/tags/question", "#question")}
 	};
 
     this.POSTS_AND_POSTING_SUBS = {
       format_text_a: {
         markdown: this.linkHtml("http://diasporafoundation.org/formatting", Diaspora.I18n.t( 'markdown' )),
-        here: this.linkHtml("http://daringfireball.net/projects/markdown/syntax", Diaspora.I18n.t( 'here' )),
+        here: this.linkHtml("http://daringfireball.net/projects/markdown/syntax", Diaspora.I18n.t( 'here' ))
       }
     };
 
     this.TAGS_SUBS = {
       filter_tags_a: {
-        third_party_tools: this.linkHtml("https://wiki.diasporafoundation.org/Tools_to_use_with_Diaspora", Diaspora.I18n.t( 'third_party_tools' )),
+        third_party_tools: this.linkHtml("https://wiki.diasporafoundation.org/Tools_to_use_with_Diaspora", Diaspora.I18n.t( 'third_party_tools' ))
       }
     };
 
@@ -51,18 +51,21 @@ app.views.Help = app.views.StaticContentView.extend({
       title_sharing: Diaspora.I18n.t( 'sharing.title' ),
       title_tags: Diaspora.I18n.t( 'tags.title' ),
       title_keyboard_shortcuts: Diaspora.I18n.t( 'keyboard_shortcuts.title' ),
-      title_miscellaneous: Diaspora.I18n.t( 'miscellaneous.title' ),
+      title_miscellaneous: Diaspora.I18n.t( 'miscellaneous.title' )
     };
 
     return this;
   },
 
-  render: function(){
+  render: function(topic){
     var section = app.views.Base.prototype.render.apply(this, arguments);
 
     // After render actions
     this.resetMenu(true);
     this.renderStaticSection("getting_help", "faq_getting_help", this.GETTING_HELP_SUBS);
+
+    var elTarget = this.findTopic(topic);
+    if(elTarget != null){ $(elTarget).click(); }
 
     return section;
   },
@@ -107,8 +110,12 @@ app.views.Help = app.views.StaticContentView.extend({
 
   menuClicked: function(e) {
     this.resetMenu();
+
     $(e.target).hide();
     $(e.target).next().show();
+
+    var data = $(e.target).data('section');
+    app.router.navigate('help/' + data);
   },
 
   clearItems: function() {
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/app/views/help_view.js
```diff
@@ -133,6 +136,32 @@ app.views.Help = app.views.StaticContentView.extend({
     this.$('#faq').append(help_section.render().el);
   },
 
+  /**
+   * Returns the jQuery <a> element whose data-section property equals the given query
```
jhass: When adding docs, let's do it right. The function does not just return any matching `<a>` on the page, it also has to have the class `faq-link`. How about saying "section title" or so?

I'm also not sure about the functions name, `findSection` or something like that seems more appropriate, a function that's executed for its return value rather than its side-effect should describe what it returns, not what it does.

app/assets/javascripts/app/router.js
```diff
@@ -2,6 +2,8 @@
 
 app.Router = Backbone.Router.extend({
   routes: {
+    "help/:id": "help",
```
jhass: Maybe call `id` `section` or `topic`? Same for all later uses.

app/assets/javascripts/app/views/help_view.js
```diff
@@ -51,19 +51,21 @@ app.views.Help = app.views.StaticContentView.extend({
...
 
-  render: function(){
+  render: function(id){
     var section = app.views.Base.prototype.render.apply(this, arguments);
 
     // After render actions
     this.resetMenu(true);
     this.renderStaticSection("getting_help", "faq_getting_help", this.GETTING_HELP_SUBS);
 
+    if(id != null && typeof  id != 'undefined'){ this.tryLinkRoute(id); }
```
jhass: I think this needs a test whether the view redirects when an ID is given.

app/assets/javascripts/app/views/help_view.js
AugierLe42e: Sure, it needs code review. In fact the most appropriate would be `findElByDataSection` as it is only based on the `data-section` property.
I'm really not comfortable with JS so numerous things might need to be changed.

jhass: Finding good name has nothing to do with the language you're writing in.

> `findElByDataSection`

Please reread my comment, this does still not describe what it does and has the second issue still.

AugierLe42e:
Done. Can be merged ASA reviewed.

app/assets/javascripts/app/views/help_view.js
```diff
@@ -133,6 +140,23 @@ app.views.Help = app.views.StaticContentView.extend({
     this.$('#faq').append(help_section.render().el);
   },
 
+  /**
+   * Returns The section title whose data-section property equals the given query
+   * Returns null if nothing found
+   * @param dataValue Value for the data-section to find
+   * @returns {jQuery}
+   */
+  findSection: function(dataValue){
```
jhass: Now  you can't decide whether to call them "section" or "topic"? ;)

app/assets/javascripts/app/views/help_view.js
```diff
@@ -51,19 +51,22 @@ app.views.Help = app.views.StaticContentView.extend({
...
-  render: function(){
+  render: function(topic){
     var section = app.views.Base.prototype.render.apply(this, arguments);
 
     // After render actions
     this.resetMenu(true);
     this.renderStaticSection("getting_help", "faq_getting_help", this.GETTING_HELP_SUBS);
 
+    var elTarget = this.findSection(topic);
+    if(elTarget != null){ $(elTarget).click(); }
```
jhass: This is behavior, it should have a test or two verifying it.

app/assets/javascripts/app/views/help_view.js
```diff
@@ -107,8 +110,12 @@ app.views.Help = app.views.StaticContentView.extend({
 
   menuClicked: function(e) {
     this.resetMenu();
+
     $(e.target).hide();
     $(e.target).next().show();
+
+    var data = $(e.target).data('section');
+    app.router.navigate('help/' + data);
```
jhass: Same

app/assets/javascripts/app/views/help_view.js
AugierLe42e: Well I started to call them section, bu as you can see, there's already a variable `section` in the `render` function, so I do what I can. But I can still refactor to call it `findTopic` if it really bothers you.

app/assets/javascripts/app/views/help_view.js
AugierLe42e: You mean jasmine test ?

app/assets/javascripts/app/views/help_view.js
jhass: You can easily name it `section_reference` or `section_id` if that's what it really is or rename `section` to `section_html` or even just `html` since that's what it contains iirc.

app/assets/javascripts/app/views/help_view.js
AugierLe42e: Can you tell me how to do ?

jhass: You see if there's already a test file in `spec/javascripts` for this view or create a new one if not.

See https://wiki.diasporafoundation.org/Testing_workflow#Jasmine

AugierLe42e:
I don't understand why [these travis tests fail](https://travis-ci.org/diaspora/diaspora/jobs/48821910)

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -86,5 +86,41 @@ describe("app.views.Help", function(){
...
+      expect(this.view.findTopic('keyboard_shortcuts').html()).toBe(topic.html());
+    });
+
+    it('should find the miscellaneous topic', function(){
+      var topic = this.view.$el.find('a[data-section=tags]');
+      expect(this.view.findTopic('tags').html()).toBe(topic.html());
+    });
+
+    /*
+    it('should rewrite route to help/tags', function(){
```
AugierLe42e: I desabled these tests as I can't find hox to make them work :/

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -86,5 +86,41 @@ describe("app.views.Help", function(){
...
+    it('should rewrite route to help/keyboard_shortcuts', function(){
+      this.view.$el.find('a[data-section=keyboard_shortcuts]').trigger('click');
+      expect(Backbone.history.fragment.endsWith("help/keyboard_shortcuts")).toBeTruthy();
+    });
+
+    it('should rewrite route to help/sharing', function(){
+      this.view.$el.find('a[data-section=sharing]').trigger('click');
+      expect(Backbone.history.fragment.endsWith("help/sharing")).toBeTruthy();
+    });
+    */
```
AugierLe42e: I isabled these tests as I can't manage to make them work. :/

config/routes.rb
```diff
@@ -223,6 +223,7 @@
 
   # Help
   get 'help' => 'help#faq', :as => 'help'
+  get 'help/:topic' => 'help#faq'
```
svbergerem: @AugierLe42e I think you can ignore this comment. ;)

AugierLe42e: Ok, I was wondering was who this was :p

app/assets/javascripts/app/views/help_view.js
```diff
@@ -9,29 +9,29 @@ app.views.Help = app.views.StaticContentView.extend({
     "click .faq-link-sharing" : "sharing",
     "click .faq-link-posts-and-posting" : "postsAndPosting",
     "click .faq-link-tags": "tags",
-    "click .faq-link-keyboard-shortcuts" : "keyboardShortcuts",
+    "click .faq-link-keyboard-shortcuts" : "keyboardShortcuts"
   },
 
-  initialize : function() {
+  initialize : function(options) {
```
AugierLe42e: Don't remember having add that :/

config/routes.rb
salbertson: Are single quotes preferred?

If so you can configure Hound by adding this to `.hound.yml`

```
ruby:
  enabled: true
  config_file: .ruby-style.yml
```

and this to `.ruby-style.yml`

```
StringLiterals:
  Enabled: true
  EnforcedStyle: single_quotes
```

svbergerem: @salbertson We were trying to disable Hound for ruby files because we don't have a code style for ruby yet. First we thought that Hound wouldn't check ruby files when not explicitly enabled but it looks like we were wrong. Now we tried to disable the checks (in [.hound.yml](https://github.com/diaspora/diaspora/blob/develop/.hound.yml#L6)) but that didn't work either.

We just want Hound to check our javascript files and maybe we will add other languages later. Even with the checks in js files we have some problems. We have our own [config file for jshint](https://github.com/diaspora/diaspora/blob/develop/config/.jshint.json) but Hound also checked things like single vs. double quotes in js files although we didn't enable that in our config. About an hour ago I [tried to disable those checks](https://github.com/diaspora/diaspora/commit/f27b9dbf797d02aa18076890241e139619714e50) by explicitly disabling the checks. According to the jshint docs there is no value `false` for `quotmarks` and `maxlen` but I don't know another way to tell Hound to stop those checks and I hope that this works now.

salbertson: https://github.com/diaspora/diaspora/pull/5644

salbertson: @svbergerem, that's frustrating. https://github.com/thoughtbot/hound/pull/617 should help with the disabling of languages. The jshint config change should work, please let me know if it doesn't.

AugierLe42e:
I want to add [this](https://github.com/RyanV/jasmine-fake-window) to jasmine to see if it can fix the tests. Is it ok and how can I do that ?

AugierLe42e:
Ok, I'm tired with this shit. Maybe someone will have more courage than me.

config/routes.rb
salbertson: @svbergerem, The most recent Hound release includes https://github.com/thoughtbot/hound/commit/93be3aa8dbbc29c8477b7a651799896b4c3bca46 which should let you to disable Ruby reviews. The `.hound.yml` file in master looks right.

Were you able to disable `maxlen` and `quotmarks`?

svbergerem: @salbertson Awesome! Thank you for your help. Hound was not complaining about `maxlen` or `quotmarks` recently although we had some PRs with mixed quotes so I think we successfully disabled those checks.

salbertson: No problem!

PR #5612, opened 2015-02-01
Hide sign up button when registration are disabled

Flaburgan:
We should hide the sign up button when the registration are closed, especially with #5391 which adds the header on every not connected pages.


svbergerem:
Thank you! :)

Issue #5618, opened 2015-02-02
Error Message on Profile Update not Shown


sam09:
Steps to reproduce the bug:
 Go to Edit Profile
 Fill in some invalid data, like a huge string for name field.
Click Submit

The page shows profile updated, however the profile is not changed as validation fails.

Flaburgan:
I guess https://github.com/diaspora/diaspora/pull/5619 closed this one.

PR #5619, opened 2015-02-03
Fixed Isssue #5618. The profile update page shows the correct message.

sam09:


jhass:
D'oh, good catch! Thanks!

Issue #5626, opened 2015-02-06
Different rendering with <p> tag


marienfressinaud:
An easy one. I found a small rendering problem with `<p>` tag. If you have a look at my screenshot, you'll see on the first post there is no space between paragraph and list (this is the problem). In the second post, we can see a little margin (that's ok).

![capture d ecran de 2015-02-06 17 19 05](https://cloud.githubusercontent.com/assets/1436309/6082881/c11c1e5a-ae24-11e4-8b48-ad429a0b91f4.png)

Problem comes from the following CSS properties:

```css
.stream_element .collapsible p {
    margin: 0px 0px 0.8em;
}
.stream_element .collapsible p:last-of-type {
    margin: 0px;
}
```

A better way to handle that would be to always have margin at the top and the bottom of both `<p>` and `<ul>` tags.

Flaburgan:
Okay, now you have to stop open issues and start to fix 'em all!

marienfressinaud:
No time for that and I still don't build my Docker image for D* development :p It will come… one day ;).

aka001:
I want to work on the issue. Is the issue already addressed?

svbergerem:
@aka001 There is already a pull request for this issue: #5627

aka001:
okay.

PR #5627, opened 2015-02-06
Fixed Issue #5626

sam09:


svbergerem:
I see two ways to fix #5626:
1. add a margin to all paragraphs in the stream (the solution you chose)
2. reduce the margin-bottom of lists if there is no other list or paragraph below

This is how stream elements look like with your chosen solution (top: before, bottom: after)


The last paragraph in a post intentionally had no margin so I'd like to see what others think about the solution.

jhass:
@sam09 ping, still interested in getting this in?

sam09:
Yes.

jhass:
I think what Steffen meant by outlining the alternative is that he would prefer it over your solution, and I have to agree ;)

sam09:
Another thought is to have margin top set to a non zero value for lists. How about that?

jhass:
Well, I think do would prefer to keep no margin between the top and the bottom of the post and only have margins between those elements if another precedes or follows.

PR #5632, opened 2015-02-07
Fix aspect dropdown on people search page inserted after initial page load

### LINKED ISSUE #5615: Aspect dropdown on people search does not work if inserted after initial page load

svbergerem:
Fixes #5615. I tested the patch on my own pod and it fixed the issue. I don't know how one should test it within our testsuite.


jhass:
Thanks!

Issue #1802, opened 2011-08-23
Photo export is not working


dmorley:
Photo export is broken. pod(s) return a 10k tar file with no images. No errors in log that I can see.

margori:
This bug made irrelevant by PR #5634

svbergerem:
@margori The feature is still in our codebase. ([UsersController#L154](https://github.com/diaspora/diaspora/blob/develop/app/controllers/users_controller.rb#L154)). #5634 only remove a button that said "Photo exporting currently unavailable".

goobertron:
Yes, it's still very important to implement photo exporting, and @margori if you could fix it, that would be fantastic!

PR #5634, opened 2015-02-07
Remove photo export from settings

svbergerem:
It is not working so the link on the settings page is useless. This PR also removes some unused strings that are related to photo/profile export from the locales.

### DIFF
--- a/app/assets/javascripts/view.js
+++ b/app/assets/javascripts/view.js
@@ -58,12 +58,6 @@ var View = {
       $(this).siblings("#tag_following_submit").removeClass('hidden');
     });
 
-    /* photo exporting in the works */
-    $("#photo-export-button").bind("click", function(evt){
-      evt.preventDefault();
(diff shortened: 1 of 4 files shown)


goobertron:
Good idea. Is it worth trying to work on this? I'm wondering how useful a photo export will be, as they will be photos the user uploaded from their computer in the first place. I suppose it will be needed as part of an integrated seed migration tool, eventually.

jhass:
Thanks!

margori:
:-( I was going to implement this feature, fix bug #1802 

I see strange that this PR was merged without discussion, as @goobertron suggested, but whatever...

svbergerem:
@margori I removed the button because it only said "Photo exporting currently unavailable". You can still fix the photo export and add the button once photo export is working again.

margori:
you call, you get it. PR #5685 using zip files.

PR #5635, opened 2015-02-07
Remove unused files, fix syntax in jasmine tests

svbergerem:
I removed some files that looked like they were no longer used.

While doing that I found out that some of the jasmine tests for the publisher were only succeeding when `infinitescroll` has been called before in `app/assets/javascripts/inbox.js` on an not existing element. I also found two other tests that were only "succeeding" because of syntax errors in the tests.


jhass:
Thanks!

Issue #5604, opened 2015-01-29
non existent avatar picture mess up streamcontacts in top right corner


viisauksena:
when some of the mentioned contacts in the stream contact field on the top right corner dont have a avatar-picture the pod-email is shown, 
while this is very long it mess up this list

solution:
![diaspora-issue-stream-avatar](https://cloud.githubusercontent.com/assets/10057308/5951463/478858d2-a768-11e4-917f-9280fadcd2bd.png)
there should be something like a default - none,avatar

jhass:
We had that once, it's broken in some places.

jaywink:
Fixed via #5638

PR #5638, opened 2015-02-08
5604 better avatar error handling

Faldrian:
This is a fix for #5604.

The solution was already in the code, but at the wrong place. When the code was executed at it's original position, the views were not yet rendered - so there were no elements to attach event handlers to.
I moved the code which adds event handler to the general view-render method... so it will be executed after each view is rendered (and only for the newly rendered view, to avoid doubling of event listeners).

I'm not quite happy with adding such a minor code-fixing-thing to THE MAIN VIEW-RENDER method... if you know what I mean  :smirk: . It seems a bit misplaced or can lead to people adding many more general fix things there...

Another solution could be to add the line `views.js:51` to a `postRenderTemplate()` in every view with avatars in its templates. This would mean the code would only be executed when it is needed, making the whole process a bit faster when there are no avatars in a view ... but the implementing person of a new view must know about this and add the method to the new view (making the implementing a bit more troublesome).

### DIFF
--- a/app/assets/javascripts/app/views.js
+++ b/app/assets/javascripts/app/views.js
@@ -46,6 +46,10 @@ app.views.Base = Backbone.View.extend({
     this.$el
       .html(this.template(presenter))
       .attr("data-template", _.last(this.templateName.split("/")));
+
+    // add avatar fallback if it can't be loaded
+    this.$el.find(this.avatars.selector).error(this.avatars.fallback);
+
     this.postRenderTemplate();
   },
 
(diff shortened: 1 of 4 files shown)


jaywink:
IMHO this fix in the main renderer is well justified - we use avatars throughout the site and thus it is actually a good place to put the fix imho. At least for now since we're dealing with regression which has to be fixed for 0.5 to go out.

So :+1: from me at least. Triggered rebuild of random failed test.

jaywink:
Thanks @Faldrian !

jaywink:
One place seems to not work still - the notifications drop down:

Faldrian:
I'll have a look at that later this week, thanks for testing.
(Probably that view does not extend app.views.Base )

Issue #5373, opened 2014-10-25
Technical limitations of how old one can be on D*


RoseTheFlower:
Considering that 1910 is set the oldest year of birth you can choose on Diaspora, it's impossible for people born in 1909 to state their real date of birth.

There [has already been](http://www.kare11.com/story/news/local/land-of-10000-stories/2014/10/09/anna-stoehr-114-plainview-verizon/17011467/) a story about a 113-year old woman facing the same kind of limitations on Facebook, having to lie about her age.

PR #5639, opened 2015-02-08
Technical limitations of how old one can be on D* #5373

aka001:
Please review the pull request.

### DIFF
--- a/app/helpers/profile_helper.rb
+++ b/app/helpers/profile_helper.rb
@@ -1,0 +1,15 @@
+#   Copyright (c) 2010-2011, Diaspora Inc.  This file is
+#   licensed under the Affero General Public License version 3 or later.  See
+#   the COPYRIGHT file.
+
+module ProfileHelper
+  def upper_limit_date_of_birth
+    minimum_year = AppConfig.settings.terms.minimum_age.get
+    minimum_year = minimum_year ?  minimum_year.to_i : 13
+    minimum_year.years.ago.year
+  end
+
+  def lower_limit_date_of_birth
+    125.years.ago.year
+  end
+end
--- a/app/views/profiles/_edit.haml
+++ b/app/views/profiles/_edit.haml
@@ -35,7 +35,7 @@
     = t('profiles.edit.your_birthday')
 
   .row-fluid
-    = select_date profile.birthday, { :prompt => true, :default => true, :order => t('date.order'), :start_year => 2000, :end_year => 1910, :prefix => 'profile[date]' }, { :class => 'span4' }
+    = select_date profile.birthday, { :prompt => true, :default => true, :order => t('date.order'), :start_year => upper_limit_date_of_birth, :end_year => lower_limit_date_of_birth, :prefix => 'profile[date]' }, { :class => 'span4' }
 
   .small-horizontal-spacer
 


app/views/profiles/_edit.haml
```diff
@@ -35,7 +36,7 @@
     = t('profiles.edit.your_birthday')
 
   .row-fluid
-    = select_date profile.birthday, { :prompt => true, :default => true, :order => t('date.order'), :start_year => 2000, :end_year => 1910, :prefix => 'profile[date]' }, { :class => 'span4' }
+    = select_date profile.birthday, { :prompt => true, :default => true, :order => t('date.order'), :start_year => val, :end_year => 1910, :prefix => 'profile[date]' }, { :class => 'span4' }
```
jhass: How about putting just `13.years.ago.year` here directly? See http://api.rubyonrails.org/classes/ActiveSupport/Duration.html#method-i-ago

jhass:
Actually it would be nice to read the minimum age from the config, like it's done in terms page and set the maximum age to 125 years, like I mentioned in #5373

app/views/profiles/_edit.haml
aka001: @jhass Thanks for a quick reply. I have updated the pr. The line ```13.years.ago``` works.

aka001:
@jhass Okay. I am working on it.

aka001:
@jhass I have updated the pr. Kindly look into it. Thank you for all your help.

app/views/profiles/_edit.haml
```diff
@@ -2,6 +2,8 @@
 -#   licensed under the Affero General Public License version 3 or later.  See
 -#   the COPYRIGHT file.
 
+- lower_limit = upper_limit_date_of_birth
+- upper_limit = lower_limit_date_of_birth
```
jhass: There's no need for these assignments, just call the helpers directly below.

app/helpers/profile_helper.rb
```diff
@@ -0,0 +1,22 @@
+#   Copyright (c) 2010-2011, Diaspora Inc.  This file is
+#   licensed under the Affero General Public License version 3 or later.  See
+#   the COPYRIGHT file.
+
+module ProfileHelper
+  def upper_limit_date_of_birth()
+     minimum_year=AppConfig.age.minimum
```
jhass: The key is `AppConfig.settings.terms.minium_age` Also please use spaces around the `=` in assignments.

app/helpers/profile_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+     if AppConfig.age.minimum==nil
```
jhass: With the correct key, you'll see the default value is `false`, not `nil`. Also reuse the variable from above.

app/helpers/profile_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+       minimum_year=13
+     end
+     minimum_year.years.ago.year
+  end
+
+  def lower_limit_date_of_birth()
+     maximum_year=AppConfig.age.maximum
+     if AppConfig.age.maximum==nil
```
jhass: I don't think this needs to be configurable, just hardcode `125.years.ago.year`

aka001:
@jhass  I have updated the pull request according to your comments. Kindly review it.

aka001:
@houndci I have updated the pull request and squashed all of them into one. I hope it's fine now.
Thank you.

app/helpers/profile_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+#   the COPYRIGHT file.
+
+module ProfileHelper
+  def upper_limit_date_of_birth
+    minimum_year = AppConfig.settings.terms.minimum_age.get
+    if !(minimum_year)
+      minimum_year = 13
+    else
+      minimum_year = minimum_year.to_i
+    end
```
jhass: Okay, just one last thing:

Let's prefer `unless` over `if !`:

```ruby
unless minimum_year
  minimum_year = 13
else
  minimum_year = minimum_year.to_i
end
```
Let's prefer `if` over `unless` when we can swap branches:


```ruby
if minimum_year
  minimum_year = minimum_year.to_i
else
  minimum_year = 13
end
```

Let's extract the assignment:

```ruby
 minimum_year = if minimum_year
  minimum_year.to_i
else
  13
end
```

And reduce it to a ternary:
```ruby
 minimum_year = minimum_year ?  minimum_year.to_i : 13
```

aka001: Thank you for the explanation. I wasn't aware of the convention followed in diaspora.
I have updated the pull request and squashed the commits.

jhass:
Thanks!

RoseTheFlower:
Definitely looks better than the way it is now. Thank you.

aka001:
:)

Issue #5625, opened 2015-02-06
Disable sending Mails to User if already deleted


sammy8806:
Actually if a User deactivated his Account and somebody tries to send him a message an mail is sent to f.e. ```deletedaccount_30@example.org```.
This generates junk in the mailqueue.

jhass:
Just something the AccountDeleter needs to do, plus a migration that does it for all already deleted users would be nice.

jhass:
@sammy8806 can you check your DB whether disable_email is set to true for said user?

giniouxe:
Is this still pending ?

jhass:
See #5640

giniouxe:
OK, thanks.

PR #5640, opened 2015-02-09
Fixes issue #5625

zachrab:
@jhass Please let me know if this is all that's needed for this fix.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -489,10 +489,10 @@ class User < ActiveRecord::Base
       self[field] = nil
     end
     [:getting_started,
-     :disable_mail,
      :show_community_spotlight_in_stream].each do |field|
       self[field] = false
     end
+    self[:disable_mail] = true
     self[:strip_exif] = true
     self[:email] = "deletedaccount_#{self[:id]}@example.org"
 
--- a/db/migrate/20150209230946_disable_mail_for_closed_account.rb
+++ b/db/migrate/20150209230946_disable_mail_for_closed_account.rb
@@ -1,0 +1,9 @@
+class DisableMailForClosedAccount < ActiveRecord::Migration
+  def up
+    User.joins(:person).where(people: {closed_account: true}).update_all(disable_mail: true)
+  end
+
+  def down
+    User.joins(:person).where(people: {closed_account: true}).update_all(disable_mail: false)
+  end
+end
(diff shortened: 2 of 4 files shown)


jhass:
Okay, please do two more things please:

Add a spec [here](https://github.com/diaspora/diaspora/blob/develop/spec/models/user_spec.rb#L947)  that starts with a user which has the field set to false and asserts that after calling the method, it's set to true. Maybe also backfill a test for the fields that are set to false there.

Add a migration that sets it to `true` for all already closed accounts. See [here](https://github.com/diaspora/diaspora/blob/develop/app/presenters/statistics_presenter.rb#L50) for how to select these (the blank username is indicating an old invite, not a closed account).

zachrab:
@jhass What do you mean by "backfill a test for the fields that are set to false there"?

app/models/user.rb
```diff
@@ -485,10 +485,10 @@ def clear_account!
       self[field] = nil
     end
     [:getting_started,
-     :disable_mail,
      :show_community_spotlight_in_stream].each do |field|
```
jhass: I can't seem to spot an existing test for these fields, that's what I mean with backfilling ;)

db/migrate/20150209230946_disable_mail_for_closed_account.rb
```diff
@@ -0,0 +1,15 @@
+class DisableMailForClosedAccount < ActiveRecord::Migration
+  def up
+  	closed_users = User.joins(:person).where(people: {closed_account: true})
+  	closed_users.each do |closed_user|
+  		closed_user.update_attribute(:disable_mail, true)
+  	end
```
jhass: Let's use [update_all](http://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-update_all) on the relation here, that should be much faster.

jhass:
Please update `db/schema.rb` to the new migration. Since the migration contains no schema changes, you shouldn't commit them if you have any. I also think it's small enough to be rebased into a single commit:

```bash
git remote add upstream git://github.com/diaspora/diaspora.git
git fetch upstream
git checkout 5625-disable-mail-to-deleted-user
# Leave the first line at pick, choose squash for the others
bin/rake db:migrate
git commit --amend -p db/schema.rb
git push -f origin 5625-disable-mail-to-deleted-user
```

The code itself looking good.

jhass:
@zachrab ping, still interested in getting this merged ;)

zachrab:
@jhass sorry for the delay!

jhass:
Thank you!

Issue #5564, opened 2015-01-16
Lock account


marienfressinaud:
In order to improve admin moderation (in addition of #5324), it could be great to add a "disable account" feature to block spamming/illegal accounts for a certain time. It could be a great compromise between letting users do all what they want and deleting their accounts once for all. Some ideas:

- Set a limit of time (so after x days/weeks/months, the account is available again)
- Give possibility of removing only specific rights (e.g. don't allow to post new comments, don't allow to post new posts)
- A message should appear to the user to explain why their account is disabled or blocked. This message might be edited by the admin.
- Consider to give the possibility to select users by list (e.g. from a list of reported users?). It can facilitate the task on biggest pods.

Flaburgan:
That's a great idea! Our admin panel definitely needs some love...

jhass:
I think the rights idea is overcomplicating this a lot and prone to bugs. Devise already provides a lock account feature that prevents sign in and that we just need to expose to the admin panel.

marienfressinaud:
It was just an idea I found interesting but it's with no doubts complicated! That's fine if the rest is easier to implement.

Flaburgan:
So I dug in that and found that we have a `locked_at` field in the `users` table.

`Update users set locked_at = '2015-01-28 12:19:00' where username = 'test';` locked my account, with a nice message when I try to log in, and then `Update users set locked_at = null where username = 'test';` allowed me to log in again.

So something like `user.lock_account!` and `user.unlock_account!` should work, we just need to expose that method in the admin panel, like did for the `close` method.

aka001:
I have made an attempt to solve this bug. Please review the pr:
https://github.com/diaspora/diaspora/pull/5643

PR #5643, opened 2015-02-09
Lock account #5564

aka001:


app/controllers/admin/users_controller.rb
```diff
@@ -7,6 +7,12 @@ def close_account
     redirect_to user_search_path, notice: t('admins.user_search.account_closing_scheduled', name: u.username)
```
aka001: @houndci Shall I change the single-quotes here into double-quotes ?

svbergerem:
@aka001 We use [houndci](https://houndci.com/) to check our javascript code. Apparently it also checks the ruby code. We were trying to disable checks for ruby but weren't successful so far. As long as those comments don't inform you about obvious errors you can just ignore them. We aren't enforcing a ruby code style right now.

aka001:
@svbergerem Okay. I have updated the pr and squashed the commits.

app/views/admins/_user_entry.haml
```diff
@@ -24,6 +24,9 @@
         - unless user.person.closed_account
           %li= link_to t('admins.user_search.close_account'), admin_close_account_path(user), method: :post, data: { confirm: t('admins.user_search.are_you_sure') }, class: 'btn btn-danger btn-mini'
 
+        - unless user.locked_at
+          %li= link_to t('admins.user_search.lock_account'), admin_close_account_path(user), method: :post, data: { confirm: t('admins.user_search.are_you_sure_lock_account') }, class: 'btn btn-danger btn-mini'
```
aka001: I didn't get the use of "admin_close_account_path" which messed up the "lock account" option.
I used the command in main repository to figured out the use of the same:
> grep -r "admin_close_account_path" .

but didn't get any result which details the same.
Please help me in figuring it out.

jhass: You might want to give http://guides.rubyonrails.org/routing.html a read then.

aka001: @jhass Thank you for the link. It helped a lot. I have taken the liberty to add the unlock feature.
When admin closes an user account it's "locked_at" field doesn't change, which results in "lock account"/"unlock account" option to be visible in admin's page.
Should we change the design to: "lock account/ unlock account" shouldn't be visible if the user account is locked?
I have attached the screenshot which reflects the same. The users "aka" and "aka_002" 's accounts are closed still the "Unlock Account" option is visible. 
screenshot from 2015-02-18 17 17 04

jhass:
@aka001 Looks like the rebase went a bit wrong. Let's try this:

```bash
git remote add upstream git:://github.com/diaspora/diaspora.git
git fetch upstream
git checkout 5564_lock_account
# Drop the line with my commit
# Choose squash for the now second line
# Save & quit
# You'll be in some conflicts, git status shows them, once you cleaned them out use git add to mark them resolved, once all are resolved, use git rebase --continue, that might be needed more than one time
git push -f origin 5564_lock_account

# Later, you can edit the last commit with
git commit --amend
git push -f origin 5564_lock_account
```

Please also add at least one controller test per new route.

app/views/admins/_user_entry.haml
```diff
@@ -24,6 +24,11 @@
         - unless user.person.closed_account
           %li= link_to t('admins.user_search.close_account'), admin_close_account_path(user), method: :post, data: { confirm: t('admins.user_search.are_you_sure') }, class: 'btn btn-danger btn-mini'
 
+        - unless user.locked_at
+          %li= link_to t('admins.user_search.lock_account'), admin_lock_account_path(user), method: :post, data: { confirm: t('admins.user_search.are_you_sure_lock_account') }, class: 'btn btn-danger btn-mini'
+        - if user.locked_at
```
jhass: I think using `else` here would be clearer

app/models/user.rb
```diff
@@ -456,6 +456,16 @@ def close_account!
     AccountDeletion.create(:person => self.person)
   end
 
+  def lock_account!
+    self.locked_at = Time.now
+    self.save
+  end
```
jhass: Devise already adds this: https://github.com/plataformatec/devise/blob/master/lib/devise/models/lockable.rb#L40

app/models/user.rb
```diff
@@ -456,6 +456,16 @@ def close_account!
...
+
+  def unlock_account!
+    self.locked_at = nil
+    self.save
+  end
```
jhass: https://github.com/plataformatec/devise/blob/master/lib/devise/models/lockable.rb#L51

app/views/admins/_user_entry.haml
jhass: You should be able to use https://github.com/plataformatec/devise/blob/master/lib/devise/models/lockable.rb#L59

app/models/user.rb
aka001: @jhass Devise is a separate module. Don't we need to write the code in core repository? Kindly clarify it.

jhass: Devise is an authentication framework, a library that handles user authentication. And we use it, you can see that in this verify file, notice the `devise` at the top.

jhass: Proof that we're already using that functionality: https://github.com/diaspora/diaspora/blob/develop/app/models/user.rb#L479

aka001: okay. Thank you for the details. I will make the required changes and update the pr.

aka001: @jhass 
The code is:
>>unless user.access_locked?
>>  #Lock_account code
>>else
>> #Unlock_account code

So when user account is closed then: 
user.access_locked? will return true, hence else part will get executed and "Unlock Account" will appear. I will have to write the code for user.access_unlocked? in "user.rb". 
Please correct me if I am wrong.

jhass: See this line: https://github.com/diaspora/diaspora/blob/develop/app/views/admins/_user_entry.haml#L43

I'd say make that a bit more pretty by adding `closed_account?` to the User model, you can just delegate that call to the associated Person model, as you can see in the current usage, then replace the call I linked with the it and wrap the entire locking code into an `unless user.closed_account?`

I don't see why you would need an `access_unlocked?` method.

aka001:
@jhass I don't know why my commit is failing tests. Please help me in finding this out.
https://travis-ci.org/diaspora/diaspora/jobs/51268131

config/defaults.yml
```diff
@@ -4,7 +4,11 @@
 
 defaults:
   version:
+<<<<<<< HEAD
+    number: "0.4.1.2" # Do not touch unless doing a release, do not backport the version number that's in master but keep develop to always say "head"
+=======
     number: "head" # Do not touch unless doing a release, do not backport the version number that's in master but keep develop to always say "head"
+>>>>>>> develop
```
svbergerem: here is an unresolved conflict

aka001: @svbergerem Resolved. Thanks.

app/controllers/admin/users_controller.rb
```diff
@@ -1,11 +1,23 @@
 module Admin
   class UsersController < AdminController
 
-    def close_account
-      u = User.find(close_account_params)
-      u.close_account!
-      redirect_to user_search_path, notice: t('admins.user_search.account_closing_scheduled', name: u.username)
-    end
+  def close_account
```
jhass: Please properly indent the code, it's a method in a class in a module, so four spaces of indentation here.

jhass:
There's still a bogus commit in here and this is really fine in one commit, no need for 12. See https://github.com/diaspora/diaspora/pull/5643#issuecomment-74852067

aka001:
I have squashed the commits.

jhass:
Only the last two it seems, there are still 11 commits in this PR

aka001:
I have squashed all the commits.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -19,4 +19,24 @@
     end
   end
 
+  describe '#lock_account' do
+    it 'queues a job to lock the given account' do
```
jhass: Uh, I should have payed closer attention to this earlier, sorry. This doesn't reflect what the method does and neither what the spec tests for. It does not queue a job, it executes that action directly (which is fine, since it's a single `UPDATE` query). So please update the wording here, something like "it locks the given account" should be fine.

Remember that you can edit your commit with `git commit --amend`.

app/controllers/admin/users_controller.rb
```diff
@@ -4,7 +4,19 @@ class UsersController < AdminController
     def close_account
       u = User.find(close_account_params)
       u.close_account!
-      redirect_to user_search_path, notice: t('admins.user_search.account_closing_scheduled', name: u.username)
+      redirect_to user_search_path, notice: t("admins.user_search.account_closing_scheduled", name: u.username)
+    end
+
+    def lock_account
+      u = User.find(close_account_params)
```
jhass: Another detail I missed, let's actually refactor this and get rid of the rather unnecessary `close_account_params` here, passing `params[:id]` to `.find` directly is just fine. Please also update `close_account` for this and drop the method below.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -19,4 +19,24 @@
...
+      other_user = FactoryGirl.create :user
+      expect(other_user).to receive(:lock_account!)
+      allow(User).to receive(:find).and_return(other_user)
+
+      post :lock_account, id: other_user.id
+    end
+  end
```
jhass: Did you actually run your tests? They're failing: https://travis-ci.org/diaspora/diaspora/jobs/51284331#L440

See https://wiki.diasporafoundation.org/Testing_workflow

I'd say don't mock the call out, put an expectation like `.expect(other_user.reload.access_locked?).to be_truthy` after the post.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -19,4 +19,24 @@
     end
   end
 
+  describe '#lock_account' do
+    it 'it locks the given account' do
+      other_user = FactoryGirl.create :user
+      expect(other_user).to receive(:lock_access!)
+      allow(User).to receive(:find).and_return(other_user)
+
+      post other_user.expect(other_user.reload.access_locked?).to be_truthy
```
aka001: @jhass 
This line: 
> other_user.expect(other_user.reload.access_locked?).to be_truthy

returns the error "undefined method `expect' " 
Is this the expected behavior? Kindly explain.

jhass: Yes it is. `expect` is not monkey patched onto all objects, it's only available inside the `it` block. Just drop the `other_user.` before it, I'm not following what that should express anyway ;)

spec/controllers/admin/users_controller_spec.rb
aka001: I understood the use of the same.
I tried doing it the way it's done in "close_account" and it works fine.
https://github.com/diaspora/diaspora/pull/5643/files#diff-3212b7cc0ec103914d1d59733b84f107R38
Should I use expect method instead of this?

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -19,4 +19,24 @@
     end
   end
 
+  describe '#lock_account' do
+    it 'it locks the given account' do
+      other_user = FactoryGirl.create :user
+      expect(other_user).to receive(:lock_access!)
+      allow(User).to receive(:find).and_return(other_user)
+
+      post :lock_account, id: other_user.id
```
jhass: It's not about replacing this call.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -19,4 +19,24 @@
     end
   end
 
+  describe '#lock_account' do
+    it 'it locks the given account' do
+      other_user = FactoryGirl.create :user
+      expect(other_user).to receive(:lock_access!)
+      allow(User).to receive(:find).and_return(other_user)
```
jhass: It's about replacing those two calls.

jhass: I guess I'll just spoil it. What I mean is something like:

```ruby
user = FactoryGirl.create :user

post :lock_account, id: other_user.id

expect(user.reload.access_locked?).to be_truthy
```

aka001: @jhass I didn't get the use of:
```post :lock_account, id: user.id``` initially so I did:
https://github.com/diaspora/diaspora/pull/5643/files#diff-3212b7cc0ec103914d1d59733b84f107R25
I got the whole concept now. I will update the pr.

File: spec/controllers/admin/users_controller_spec.rb [hunk 14]
```diff
@@ -19,4 +19,20 @@
     end
   end
 
+  describe '#lock_account' do
+    it 'it locks the given account' do
+      other_user = FactoryGirl.create :user
```
aka001: @jhass Shall I change the "other_user" to "user" ?

jhass: I used `user` over `other_user` in my example because I found the `other_` rather redundant, I didn't get what it would relate to. But I don't care too much either way.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -19,4 +19,20 @@
...
+      post :lock_account, id: other_user.id
+      expect(other_user.reload.access_locked?).to be_truthy
+    end
+  end
+
+  describe '#unlock_account' do
+    it 'it unlocks the given account' do
+      other_user = FactoryGirl.create :user
+      post :unlock_account, id: other_user.id
+      expect(not(other_user.reload.access_locked?)).to be_truthy
```
jhass: As said before use `!`, not `not` for negation. But here again there's a better way than negation in the first place and that is to use the `be_falsey` matcher.

Besides this spec is not testing much. You start with a user account that's not lock, run some code and then check whether the user account is not locked. So while you ensure that the code you run does not lock the account, you do not ensure that it unlocks the account.

aka001: When I use: 
```expect(other_user.reload.access_locked?).to be_falsify``` instead of
```expect(!(other_user.reload.access_locked?)).to be_truthy```
I get the error:
http://pastebin.com/aSv3nMv9
If I use the later "expect(!(other_user.reload.access_locked?)).to be_truthy" I don't get any error.

Comment (jhass) [inline spec/controllers/admin/users_controller_spec.rb]: `be_falsey` not `be_falsify`

File: spec/controllers/admin/users_controller_spec.rb [hunk 16]
```diff
@@ -19,4 +19,22 @@
...
+    end
+  end
+
+  describe '#unlock_account' do
+    it 'it unlocks the given account' do
+      other_user = FactoryGirl.create :user
+      post :lock_account, id: other_user.id
+      expect(other_user.reload.access_locked?).to be_truthy
+      post :unlock_account, id: other_user.id
+      expect(!(other_user.reload.access_locked?)).to be_truthy
```
aka001: I have first locked the account -> tested it -> unlocked it -> tested it. I think this will cover the test case.

jhass: The

```ruby
post :lock_account, id: other_user.id
expect(other_user.reload.access_locked?).to be_truthy
```

is kind of redundant with the other test, I think a simple `other_user.lock_access!` is enough.

spec/controllers/admin/users_controller_spec.rb
aka001: I have updated the pr. I guess it's correct now.
In which segment of code can I see the implementation of expect method and all it's flags? 
I would like to know more about it.

jhass: The main part is http://www.rubydoc.info/gems/rspec-expectations/RSpec/Matchers but things like `rspec-rails` add more matchers.

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -19,4 +19,22 @@
...
+      other_user.lock_access!
+      expect(other_user.reload.access_locked?).to be_truthy
+    end
+  end
+
+  describe '#unlock_account' do
+    it 'it unlocks the given account' do
+      other_user = FactoryGirl.create :user
+      other_user.lock_access!
+      expect(other_user.reload.access_locked?).to be_truthy
```
jhass: Just drop this line, that expectation is not part of the scope of this test.

spec/controllers/admin/users_controller_spec.rb
aka001: I have updated the pr.

spec/controllers/admin/users_controller_spec.rb
aka001: True that. I have updated the pr.

app/controllers/admin/users_controller.rb
```diff
@@ -2,9 +2,21 @@ module Admin
...
+      redirect_to user_search_path, notice: t("admins.user_search.account_locking_scheduled", name: u.username)
+    end
+
+    def unlock_account
+      u = User.find(params[:id])
+      u.unlock_access!
+      redirect_to user_search_path, notice: t("admins.user_search.account_unlocking_scheduled", name: u.username)
     end
 
     private
```
jhass: Please drop the method below, we made it obsolete earlier.

aka001: Done.

app/controllers/admin/users_controller.rb
```diff
@@ -2,15 +2,24 @@ module Admin
...
 
-    def close_account_params
-      params.require(:id)
+    def unlock_account
+      u = User.find(params[:id])
+      u.unlock_access!
+      redirect_to user_search_path, notice: t("admins.user_search.account_unlocking_scheduled", name: u.username)
     end
+
+    private
```
jhass: No need for the stray `private` either.

aka001: Updated.

jhass:
Thank you!

Flaburgan:
Nice work @aka001 thank you!

aka001:
:)

PR #5561, opened 2015-01-15
Rubocop linter and styleguide

dimaursu:
This should take care of #5531 .
Also, notice that there is a `.rubocop_todo.yml`, to shut rubocop mouth about existing warnings: there are about 4500 of them left (17k in the begining). You can also see the count in that file, per warning.
Eventually we should get rid of that file, but I'd say to take it slow, and do it gradually.

This is the workflow: you remove a `cop` from that file, run rubocop in the rood of `d*`, and fix the warnings, until it's all green again.

This should be an excellent training for someone who wants to learn more about ruby and the best practices.

PR #5644, opened 2015-02-10
Enable and configure Hound for Ruby

salbertson:
I noticed Hound comments related to double comments were being ignored. This change configures Hound to prefer single quotes.

@jhass any other Ruby style preferences?

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -2,5 +2,7 @@ java_script:
   enabled: true
   config_file: config/.jshint.json
   ignore_file: config/.jshint_ignore
+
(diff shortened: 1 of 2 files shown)


jhass:
As @svbergerem explained in https://github.com/diaspora/diaspora/pull/5603#discussion_r24380193 we have not decided on a styleguide for Ruby and and would rather like to disable the Ruby linter for now.

jhass:
See also #5561

salbertson:
Thanks @jhass.

PR #5654, opened 2015-02-12
respect http/https in image paths

fscz:
when running on http, paths to public/uploaded/images... would still be prefixed with https.
make those paths respect the pod's mode of operation

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -99,31 +99,26 @@ class Profile < ActiveRecord::Base
     self.attributes.merge(update_hash){|key, old, new| old.blank? ? new : old}
   end
 
-  def image_url= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
+  def _set_image_url (att, url)
+    if not url.blank?
+      path = url[/(https?:\/\/(.(?!\/))*.)?(.*)/, 3]
+      if not path.match(/^\//)
+        path += '/'
+      end
+      self[att] = path
     end
+  end
+
+  def image_url= url
+    _set_image_url(:image_url, url)
   end
 
   def image_url_small= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
-    end
+    _set_image_url(:image_url_small, url)
   end
 
   def image_url_medium= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
-    end
+    _set_image_url(:image_url_medium, url)
   end
 
   def date= params
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -100,30 +100,18 @@ class Profile < ActiveRecord::Base
   end
 
   def image_url= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
-    end
+    return image_url if url == '' or url.nil?
+    super(url[/(https?:\/\/(.(?!\/))*.)?(.*)/, 3])
   end
 
   def image_url_small= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
-    end
+    return image_url if url == '' or url.nil?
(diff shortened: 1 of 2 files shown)


app/models/profile.rb
```diff
@@ -100,30 +100,18 @@ def from_omniauth_hash(omniauth_user_hash)
   end
 
   def image_url= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
-    end
+    return image_url if url == '' or url.nil?
```
jhass: Just use `blank?` for these

jhass:
Please add at least one spec that fails without your change.

app/models/profile.rb
```diff
@@ -100,30 +100,18 @@ def from_omniauth_hash(omniauth_user_hash)
   end
 
   def image_url= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
```
jhass: Looks like the method is no longer used then, remove it.

app/models/profile.rb
```diff
@@ -100,30 +100,18 @@ def from_omniauth_hash(omniauth_user_hash)
...
 
   def image_url_medium= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
-    end
+    return image_url if url == '' or url.nil?
+    super(url[/(https?:\/\/(.(?!\/))*.)?(.*)/, 3])
```
jhass: Let's take this opportunity and DRY up that code by extracting it into a method. Instead of using `super`, you can use the fields name as a symbol and use the `[]=` method to set the value.

fscz:
please review

svbergerem:
@fscz I am not the right one to review your PR but a few tests are failing and those are not random failures.

https://travis-ci.org/diaspora/diaspora/jobs/50477001

fscz:
I see. From my point of we, those tests should be adjusted. But it's not my decision

app/models/profile.rb
```diff
@@ -99,31 +99,26 @@ def from_omniauth_hash(omniauth_user_hash)
     self.attributes.merge(update_hash){|key, old, new| old.blank? ? new : old}
   end
 
-  def image_url= url
-    return image_url if url == ''
-    if url.nil? || url.match(/^https?:\/\//)
-      super(url)
-    else
-      super(absolutify_local_url(url))
+  def _set_image_url (att, url)
```
jhass: If you want to make a private method, just move it to the private methods, do not use silly names like starting with an underscore. Besides, Ruby naming conventions already use preceding underscores for unused arguments.

app/models/profile.rb
```diff
@@ -99,31 +99,26 @@ def from_omniauth_hash(omniauth_user_hash)
...
+    if not url.blank?
```
jhass: Unless you need to add an `elsif`, prefer `unless`.

app/models/profile.rb
```diff
@@ -99,31 +99,26 @@ def from_omniauth_hash(omniauth_user_hash)
...
+      path = url[/(https?:\/\/(.(?!\/))*.)?(.*)/, 3]
+      if not path.match(/^\//)
+        path += '/'
```
jhass: The string is created inside the method, modifying it with `<<` instead of creating a new one with `+=` is more efficient.

jhass:
> I see. From my point of we, those tests should be adjusted. But it's not my decision

That's a silly statement. If you believe your change is correct and then your change breaks existing tests, of course those needs to be adapted. The decision the reviewer has to make is not whether you need to adapt the tests, but whether your change is correct.

spec/models/profile_spec.rb
```diff
@@ -135,14 +135,19 @@
       expect {@profile.image_url = ""}.not_to change(@profile, :image_url)
     end
 
-    it 'makes relative urls absolute' do
-      @profile.image_url = "/relative/url"
-      expect(@profile.image_url).to eq("#{@pod_url}/relative/url")
+    it 'makes absolute urls relative' do
+      @profile.image_url = "https://host/bar/relative/url"
+      expect(@profile.image_url).to eq("/bar/relative/url")
+    end
```
jhass: This change however, is incorrect. Those profile fields are federated to other pods and used as they are there. Thus making the URLs relative breaks the profile pictures on remote pods.

fscz:
Ok. Let's close this. The bug, I was trying to fix only occurs when running on http. Since this will be dropped soon. It is probably not worth it.

PR #5657, opened 2015-02-12
add proper logging init

fscz:
- the code in lib/federation_logger.rb wasn't working (ActiveSupport::BufferedLogger is deprecated in rails 4 - should just use Logger or ActiveSupport::Logger)
- without this path federation messages are logged nowhere

### DIFF
--- a/config/defaults.yml
+++ b/config/defaults.yml
@@ -7,6 +7,7 @@ defaults:
     number: "head" # Do not touch unless doing a release, do not backport the version number that's in master but keep develop to always say "head"
   heroku: false
   environment:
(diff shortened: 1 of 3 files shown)


Flaburgan:
Could you please use development.log and production.log?

fscz:
instead of logfile?

jhass:
I have no idea why you made the main log configurable, it seems completely unrelated to your PRs intention and description. Fixing the federation logger is uncontroversial, please remove the other changes from this PR and make a separate one if you care about them.

Besides, adding a new configuration entry should update config/diaspora.yml.example too.

PR #5658, opened 2015-02-12
add repeat: false to options

fscz:
It does not make a whole lot of sense to repeat the webfingers over and over again. if they fail they are very likely to fail again in a few seconds. if you want to retry you would have to use exponential backoff

### DIFF
--- a/app/workers/fetch_webfinger.rb
+++ b/app/workers/fetch_webfinger.rb
@@ -4,7 +4,7 @@
 
 module Workers
   class FetchWebfinger < Base
-    sidekiq_options queue: :socket_webfinger
+    sidekiq_options queue: :socket_webfinger, retry: false
 
     def perform(account)
       person = Webfinger.new(account).fetch


jhass:
Sidekiq is using an exponential  backoff, [here](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/middleware/server/retry_jobs.rb#L179) is it's formula.

fscz:
sorry for opening this again. but i still don't think, the webfingers should be repeated. 

if someone searches a user id that cannot be found  - say because of a typo in the subdomain or tld, then this will repeat and it shouldn't. it's better to just display a message 'sorry nothing found' and let the user try again

jhass:
No, it's good for the pod to know as many accounts as possible. First, this enhances usability as that account can be directly found by a later search and also found through tag and display name searches. Second the risk of a targeted MITM is reduced as the pod knows as many public keys as possible. So trying to gather as many profiles as possible is actually a good thing here.

fscz:
But strategy of gathering the profiles seems very questionable. Wouldn't it be better to use a replicated key-value store and have profiles replicated among all pods?

jhass:
I think you're overestimating the projects current development capabilities a bit. We're continuing this project to provide the few thousands users an alive platform, radical design changes still require a lot of work on the codebase to be even thinkable, experimenting with alternate ideas on how to do distributed social networking, should happen in new or at least smaller projects. If you want to have big and lengthy discussions about possible solutions at all, you should talk with or try to join the Social WG at W3.

PR #5659, opened 2015-02-12
Add the ability to link remote user profiles

ghost:
related to diaspora/jsxc#77

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -184,13 +184,21 @@ class PeopleController < ApplicationController
   private
 
   def find_person
-    @person = Person.find_from_guid_or_username({
-      id: params[:id] || params[:person_id],
-      username: params[:username]
-    })
+    username = params[:username]
+    @person = if diaspora_id?(username)
+        Person.where({
+          diaspora_handle: username.downcase
+        }).first
+      else
+        Person.find_from_guid_or_username({
+          id: params[:id] || params[:person_id],
+          username: username
+        })
+      end
 
     # view this profile on the home pod, if you don't want to sign in...
     authenticate_user! if remote_profile_with_no_user_session?
+    raise ActiveRecord::RecordNotFound if @person.nil?
     raise Diaspora::AccountClosed if @person.closed_account?
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/people_controller.rb
```diff
@@ -184,10 +184,15 @@ def aspect_membership_dropdown
...
-    })
+    username = params[:username]
+    if diaspora_id?(username)
+      @person = Person.where(:diaspora_handle => username.downcase).first
+    else
+      @person = Person.find_from_guid_or_username({
+        id: params[:id] || params[:person_id],
+        username: username
+      })
+    end
```
jhass: While we have no Ruby styleguide yet, please at least decide on a hash form to use in a single diff hunk ;)

config/routes.rb
```diff
@@ -175,8 +175,8 @@
       get :tag_index
     end
   end
-  get '/u/:username' => 'people#show', :as => 'user_profile'
-  get '/u/:username/profile_photo' => 'users#user_photo'
+  get '/u/:username' => 'people#show', :as => 'user_profile', :constraints => { :username => /[^\/]+/ }
+  get '/u/:username/profile_photo' => 'users#user_photo', :constraints => { :username => /[^\/]+/ }
```
jhass: I'm not quite following on why you added those tbh.

jhass:
Should have a controller spec for the added functionality.

config/routes.rb
Zauberstuhl: http://guides.rubyonrails.org/routing.html#specifying-constraints

> By default the :id parameter doesn't accept dots - this is because the dot is used as a separator for formatted routes. If you need to use a dot within an :id add a constraint which overrides this - for example id: /[^\/]+/ allows anything except a slash.

Flaburgan: Do we accept dots in username? I don't think so.

svbergerem: @Flaburgan No, but in the diaspora handle. (including @pod.tld)

Zauberstuhl:
done

jhass:
Looking good, rebase into a single commit please.

PR #5661, opened 2015-02-13
Bump markdown-it-sanitizer

### LINKED ISSUE #5569: [Markdown] Bad parsing of <, > and & in code
There is no possibility to type HTML code such as a basic markup like `<div>` in Markdown code (between \`). `<`, `>` and `&` are replaced by their HTML values `&lt;`, `&gt;` and `&amp;`.

Here is a screenshot:
screenshot

There is no problem on mobile so it is probably link to the introduction of markdown-it ( #5526 ). See also #5466

svbergerem:
Fixes #5569 and autolinking of `<https://ddg.com>`.


jhass:
Thanks!

PR #5664, opened 2015-02-13
Remove last_post from PeopleController

svbergerem:
We don't use it anymore and don't have any tests for that. That's why I think we should remove it.


jhass:
Thanks!

PR #5665, opened 2015-02-13
Adding help section for chat

ghost:
Problem with linking :/
final2

### DIFF
--- a/app/assets/javascripts/app/views/help_view.js
+++ b/app/assets/javascripts/app/views/help_view.js
@@ -9,7 +9,8 @@ app.views.Help = app.views.StaticContentView.extend({
     "click .faq-link-sharing" : "sharing",
     "click .faq-link-posts-and-posting" : "postsAndPosting",
     "click .faq-link-tags": "tags",
-    "click .faq-link-keyboard-shortcuts" : "keyboardShortcuts"
+    "click .faq-link-keyboard-shortcuts" : "keyboardShortcuts",
+    "click .faq-link-chat" :  "chat"
   },
 
   initialize : function() {
@@ -35,6 +36,13 @@ app.views.Help = app.views.StaticContentView.extend({
       }
     };
 
+    this.CHAT_SUBS = {
+      add_contact_roster_a: {
+        toggle_privilege: this.getChatIcons(),
+        contacts_page: this.linkHtml(Routes.contacts_path(), Diaspora.I18n.t('chat.contacts_page'))
+      }
+    };
+
     this.data = {
       title_header: Diaspora.I18n.t( 'title_header' ),
       title_getting_help: Diaspora.I18n.t( 'getting_help.title' ),
@@ -51,7 +59,9 @@ app.views.Help = app.views.StaticContentView.extend({
       title_sharing: Diaspora.I18n.t( 'sharing.title' ),
       title_tags: Diaspora.I18n.t( 'tags.title' ),
       title_keyboard_shortcuts: Diaspora.I18n.t( 'keyboard_shortcuts.title' ),
-      title_miscellaneous: Diaspora.I18n.t( 'miscellaneous.title' )
+      title_miscellaneous: Diaspora.I18n.t( 'miscellaneous.title' ),
+      title_chat: Diaspora.I18n.t( 'chat.title' ),
+      chat_enabled: this.chatEnabled()
     };
 
     return this;
@@ -143,7 +153,7 @@ app.views.Help = app.views.StaticContentView.extend({
   /**
    * Returns The section title whose data-section property equals the given query
    * Returns null if nothing found
-   * @param dataValue Value for the data-section to find
+   * @param data Value for the data-section to find
    * @returns {jQuery}
    */
   findSection: function(data){
@@ -187,8 +197,27 @@ app.views.Help = app.views.StaticContentView.extend({
     e.preventDefault();
   },
 
+  chat: function(e){
+    this.renderStaticSection("chat", "faq_chat", this.CHAT_SUBS);
+    this.menuClicked(e);
+
+    e.preventDefault();
+  },
+
   linkHtml: function(url, text) {
     return "<a href=\"" + url + "\" target=\"_blank\">" + text + "</a>";
+  },
+
+  chatEnabled: function(){
+    return gon.chatEnabled;
+  },
+
+  getChatIcons: function(){
+    return '<div class="help-chat-icons">' +
+           '  <i class="entypo lock-open"></i>' +
+           '  <i class="entypo chat"></i>' +
+           '  <i class="entypo trash"></i>' +
+           '</div>';
   }
 });
 // @license-end
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -1,4 +1,8 @@
 class HelpController < ApplicationController
-	before_filter -> { @css_framework = :bootstrap }
-	layout ->(c) { request.format == :mobile ? "application" : "with_header_with_footer" }
+  before_filter -> { @css_framework = :bootstrap }
+  layout -> (c) { request.format == :mobile ? "application" : "with_header_with_footer" }
+
+  def faq
+    gon.chatEnabled = AppConfig.chat.enabled?
+  end
 end
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -449,6 +449,11 @@ en:
       person_multiple_aspects_a: "Yes. Go to your contacts page and click on “My contacts”. For each contact you can use the menu on the right to add them to (or remove them from) as many aspects as you want. Or you can add them to a new aspect (or remove them from an aspect) by clicking the aspect selector button on their profile page. Or you can even just move the pointer over their name where you see it in the stream, and a “hover-card” will appear. You can change the aspects they are in right there."
       delete_aspect_q: "How do I delete an aspect?"
       delete_aspect_a: "In your list of aspects on the left side of the main page, point your mouse at the aspect you want to delete. Click the little “Edit” pencil that appears on the right. Click the “Delete this aspect” button in the box that appears."
+    chat:
+      title: "Chat"
+      contacts_page: "contacts page"
+      add_contact_roster_q: "How do I chat with someone in diaspora*?"
+      add_contact_roster_a: "First, you need to enable chat for one of the aspects that person is in. To do so, go to the %{contacts_page}, select the aspect you want and click on the chat icon to enable chat for the aspect. %{toggle_privilege} You could, if you prefer, create a special aspect called 'Chat' and add the people you want to chat with to that aspect. Once you've done this, open the chat interface and select the person you want to chat with."
     mentions:
       title: "Mentions"
       what_is_a_mention_q: "What is a \"mention\"?"
--- a/spec/javascripts/app/views/help_view_spec.js
+++ b/spec/javascripts/app/views/help_view_spec.js
@@ -134,4 +134,30 @@ describe("app.views.Help", function(){
       });
     });
   });
+
+  describe("chat section", function(){
+    describe("chat enabled", function(){
+      beforeEach(function(){
+        gon.chatEnabled = true;
+        this.view = new app.views.Help();
+        this.view.render();
+      });
+
+      it('should display the chat', function(){
+        expect(this.view.$el.find('a[data-section=chat]').length).toBe(1);
+      });
+    });
+
+    describe("chat disabled", function(){
+      beforeEach(function(){
+        gon.chatEnabled = false;
+        this.view = new app.views.Help();
+        this.view.render();
+      });
+
+      it('should not display the chat', function () {
+        expect(this.view.$el.find('a[data-section=chat]').length).toBe(0);
+      });
+    });
+  });
 });
--- a/app/assets/stylesheets/help.css.scss
+++ b/app/assets/stylesheets/help.css.scss
@@ -91,6 +91,18 @@ ul#help_nav {
       border-radius: 0px 0px 4px 4px;
       background-color: white;
       padding: 10px 20px;
+
+      div.help-chat-icons{
+        text-align: center;
+        font-size: 50px;
+        line-height: 70px;
+
+        i.entypo{
+          color: #bfbfbf;
+
+          &.chat{ color: #000000; }
+        }
+      }
     }
   }
 }
--- a/app/assets/templates/faq_chat_tpl.jst.hbs
+++ b/app/assets/templates/faq_chat_tpl.jst.hbs
@@ -1,0 +1,6 @@
+<div class='question opened collapsible'>
+    <a class='toggle' href='#'>
+        <h4>{{ add_contact_roster_q }}</h4>
+    </a>
+    <div class='answer hideable'>{{{ add_contact_roster_a }}}</div>
+</div>
--- a/app/assets/templates/help_tpl.jst.hbs
+++ b/app/assets/templates/help_tpl.jst.hbs
@@ -68,6 +68,12 @@
             <a href="#" class="section-unselected faq-link" data-section="miscellaneous" data-items="back_to_top photo_albums subscribe_feed diaspora_app">{{ title_miscellaneous }}</a>
             <span class="section-selected">{{ title_miscellaneous }}</span>
           </li>
+          {{#if chat_enabled }}
+          <li>
+            <a href="#" class="section-unselected faq-link-chat" data-section="chat" data-items="add_contact_roster i_m_a_podmin">{{ title_chat }}</a>
+            <span class="section-selected">{{ title_chat }}</span>
+          </li>
+          {{/if}}
         </ul>
       </div>
     </div>


jhass:
The help section is clearly user oriented, we shouldn't include items for the podmin, in fact we shouldn't even display this section if the chat is disabled.

AugierLe42e:
Does client side has access to the configuration of the pod ?

jhass:
Not directly, that would be dangerous. I'm sure @zauberstuhl knows a way to detect it from the client side though.

Zauberstuhl:
The first thing which pops into my mind is that you could check whether `jsxc` variable exists.

AugierLe42e:
If not, I can modify the help controller to pass whether tha chat is active or not in a data-section, for instance ?

jhass:
Yes, that would be the alternative if there's no general way of detecting it from the client side. You would use gon for that.

AugierLe42e:
How about the linking problem ? Seems like the HTML is escaped :/

svbergerem:
@AugierLe42e Your images include English text. We decided some time ago that we don't want to add images to the help section which would need a translation.

AugierLe42e:
@svbergerem : That's right. Let's see if I can include it in HTML.

Zauberstuhl:
Checked it and it is not possible. You should go with the controller

AugierLe42e:
Thank you.

AugierLe42e:
Works but still having troubles with Jasmine tests and link injections.

app/controllers/help_controller.rb
```diff
@@ -1,4 +1,10 @@
 class HelpController < ApplicationController
-	before_filter -> { @css_framework = :bootstrap }
-	layout ->(c) { request.format == :mobile ? "application" : "with_header_with_footer" }
-end
+
+  before_filter -> { @css_framework = :bootstrap }
+  layout -> (c) { request.format == :mobile ? "application" : "with_header_with_footer" }
+
+  def faq
+    @data_chat = AppConfig.chat.enabled ? "enabled" : "disabled"
```
jhass: Let's use [gon](https://github.com/gazay/gon) here.

jhass:
Could also use a rebase.

app/controllers/help_controller.rb
AugierLe42e: Is it a usual use-case in d* ?
It seems a bit over-engineered to use a another just for that, right ?

jhass: It's already available and solves the problem better than your current solution. I don't see why that would be over-engineered, in fact it's less code you have to add in this PR.

AugierLe42e: > in fact it's less code you have to add in this PR.

In fact it's more code I have to remove. But if you think it's better, then let's do this.

jhass: The changeset counts, not your work towards it ;P

app/assets/javascripts/app/views/help_view.js
```diff
@@ -187,8 +197,23 @@ app.views.Help = app.views.StaticContentView.extend({
...
   linkHtml: function(url, text) {
     return "<a href=\"" + url + "\" target=\"_blank\">" + text + "</a>";
+  },
+
+  linkImage: function(src, alt){
+    return '<div class="help-image"><img src="' + src + '" alt="' + alt + '"/></div>';
+  },
+
+  chatEnabled: function(){
+    return gon.chat_enabled === 'enabled';
```
AugierLe42e: This doesn't work at all.

app/controllers/help_controller.rb
```diff
@@ -1,4 +1,8 @@
 class HelpController < ApplicationController
-	before_filter -> { @css_framework = :bootstrap }
-	layout ->(c) { request.format == :mobile ? "application" : "with_header_with_footer" }
-end
+  before_filter -> { @css_framework = :bootstrap }
+  layout -> (c) { request.format == :mobile ? "application" : "with_header_with_footer" }
+
+  def faq
+    gon.preloads[:chat_enabled] = AppConfig.chat.enabled ? 'enabled' : 'disabled'
```
jhass: Here you use the key `gon.preloads[:chat_enabled]` but in the JS you use the key `gon.chat_enabled`. I find the key `gon.chat_enabled` better, so chance this instance.

jhass: Also passing `true`/`false` to to it as returned by `AppConfig.chat.enabled?` should be just fine.

app/controllers/help_controller.rb
```diff
@@ -1,4 +1,8 @@
 class HelpController < ApplicationController
-	before_filter -> { @css_framework = :bootstrap }
-	layout ->(c) { request.format == :mobile ? "application" : "with_header_with_footer" }
-end
+  before_filter -> { @css_framework = :bootstrap }
+  layout -> (c) { request.format == :mobile ? "application" : "with_header_with_footer" }
+
+  def faq
+    gon.chat_enabled = AppConfig.chat.enabled?
```
AugierLe42e: Doesn't work either. JS variable is `undefined` which make me think the variable is not corretly passed to the JS script.

jhass: Looks like gon converts that to headlessCamlCase.

AugierLe42e: Weird... Thanks for the tip. Works no. What about the tags not being injected correctly ?

injection

jhass: Handlebars feature. `{{}}` escapes, `{{{}}}` does not.

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -134,4 +141,28 @@ describe("app.views.Help", function(){
       });
     });
   });
+
+describe("chat section", function(){
```
AugierLe42e: You are right, milou !

AugierLe42e:
Two last things and we are ok. [Images are not correctly included](https://github.com/diaspora/diaspora/pull/5665/files#diff-7425515578ef82f0d52e7bf0b0bacbf9R211) and [this Jasmin test still fails](https://github.com/diaspora/diaspora/pull/5665/files#diff-41cc85e93f2018f87904ee1daccc7280R146)

jasmin

app/assets/javascripts/app/views/help_view.js
```diff
@@ -187,8 +197,23 @@ app.views.Help = app.views.StaticContentView.extend({
...
+
+    e.preventDefault();
+  },
+
   linkHtml: function(url, text) {
     return "<a href=\"" + url + "\" target=\"_blank\">" + text + "</a>";
+  },
+
+  linkImage: function(src, alt){
+    return '<div class="help-image"><img src="images/' + src + '" alt="' + alt + '"/></div>';
```
jhass: We got [js_image_paths](https://github.com/jhass/js_image_paths#usage), use it here.

AugierLe42e: Is there something special to do for an image to be taken in account ?
Because :
injection1
and though :
injection2

jhass: Does `public/assets/` exist? Try removing it. Also try `bin/rake tmp:cache:clear` and restarting the app server.

AugierLe42e: Yay !
help

svbergerem: In most cases (when there are less than 20 contacts in the aspect) there will be also the mail icon in the control bar. The really important thing is the chat icon. I am still not convinced that we should add an image here. Why don't we describe where you will find the button (Go to the contacts page, select an aspect and click on the chat icon in the icon bar under the header to enable chat for the aspect) and add the original entypo icon?

AugierLe42e:
@svbergerem : Why do you want to make me do complicated things where an image is sufficient ? I followed your advice : I dropped the images with english text and I added your description. I don't think this is that important, really.

svbergerem:
@AugierLe42e You added a screenshot with 3 icons. As I already said in most cases there will also another icon (mail) next to those. We might add/remove/change icons in the future and then we will have to change the screenshot. The 3 icons from the screenshot are already in the browser's cache.

Instead of loading them another time as an image I suggested to add them as an icon. I think it would be sufficient if we would only include the chat icon but we could also display all 3 or 4 icons. For the code it doesn't really make a difference.

Instead of adding an image 
`<div class="help-image">' + alt + '</div>`
you would have to add an icon
`<i class="entypo chat"></i>`.
Why is that more complicated?

Adding those icons via entypo icons instead of an image makes it also easier to change the appearance of the icon later. We want to make the icon red? Just add `color: red`. We want to change the icon size? Just change `font-size`. If we would like to do the same with the screenshot we would either have to scale or edit the image. Both won't be as efficient.

AugierLe42e:
You may be right.

AugierLe42e:
@svbergerem : done. Will you help me to solve jASMIN TEST

app/assets/javascripts/app/views/help_view.js
```diff
@@ -187,8 +197,31 @@ app.views.Help = app.views.StaticContentView.extend({
...
+  chatEnabled: function(){
+    return gon.chatEnabled;
+  },
+
+  getChatIcons: function(){
+    return '<div class="help-chat-icons"> \
+              <i class="entypo lock-open"></i> \
+              <i class="entypo chat"></i> \
+              <i class="entypo trash"></i> \
+            </div>'
```
AugierLe42e: Oh dear...

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -134,4 +141,28 @@ describe("app.views.Help", function(){
...
+
+  describe("chat section", function(){
+    describe("chat enabled", function(){
+      beforeEach(function(){
+        gon.chatEnabled = true;
+        this.view.render();
+      });
+
+      it('should display the chat', function(){
+        expect(this.view.$el.find('div[data-section=chat]').length).toBe(1);
```
svbergerem: use `a[data-section=chat]` instead.

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -134,4 +141,28 @@ describe("app.views.Help", function(){
...
+    });
+
+    describe("chat enabled", function(){
+      beforeEach(function(){
+        gon.chatEnabled = false;
+        this.view.render();
+      });
+
+      it('should not display the chat', function () {
+        expect(this.view.$el.find('div[data-section=chat]').length).toBe(0);
```
svbergerem: use `a[data-section=chat]` instead.

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -134,4 +141,28 @@ describe("app.views.Help", function(){
       });
     });
   });
+
+  describe("chat section", function(){
+    describe("chat enabled", function(){
+      beforeEach(function(){
+        gon.chatEnabled = true;
```
svbergerem: In the help view you set `chat_enabled` on initialization. Because of that it doesn't change when you change `gon.chatEnabled` afterwards.

Add `this.view = new app.views.Help();` after this line to fix that.

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -134,4 +141,28 @@ describe("app.views.Help", function(){
...
+      });
+
+      it('should display the chat', function(){
+        expect(this.view.$el.find('div[data-section=chat]').length).toBe(1);
+      });
+    });
+
+    describe("chat enabled", function(){
+      beforeEach(function(){
+        gon.chatEnabled = false;
```
svbergerem: Add `this.view = new app.views.Help();` after this line.

spec/javascripts/app/views/help_view_spec.js
AugierLe42e: Yes it changes because `gon` changes snake_case to headlessCamelCase. This is awkward but meh... I should change the help view instead.

svbergerem: I don't think you should. There is no need to change `gon.chatEnabled` on a normal setup after loading the page so `this.chat_enabled` doesn't have to change either.

`this.chat_enabled` doesn't change after loading the page because you copy the value of `gon.chatEnabled` on initialization and don't change `this.chat_enabled` afterwards. (which is fine as I already wrote) You just have to make sure that in this particular case `this.chat_enabled` changes which you can do by adding `this.view = new app.views.Help();` here.

AugierLe42e: Ok. I thought chat `this.chat_enabled` was just an alias to the function `chatEnabled()` and was called each time the page is rendered.

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -134,4 +141,30 @@ describe("app.views.Help", function(){
...
+      it('should display the chat', function(){
+        expect(this.view.$el.find('a[data-section=chat]').length).toBe(1);
+      });
+    });
+
+    describe("chat enabled", function(){
+      beforeEach(function(){
+        this.chat_enabled = false;
+        this.view = new app.views.Help();
+        this.view.render();
```
AugierLe42e: @svbergerem : Unless I've missed something, [this does not work](https://travis-ci.org/diaspora/diaspora/jobs/51999707#L570)

AugierLe42e:
\o/ @svbergerem : thanks, works now !
@jhass : you can merge this.

app/assets/javascripts/app/views/help_view.js
```diff
@@ -187,8 +197,31 @@ app.views.Help = app.views.StaticContentView.extend({
...
+    e.preventDefault();
+  },
+
   linkHtml: function(url, text) {
     return "<a href=\"" + url + "\" target=\"_blank\">" + text + "</a>";
+  },
+
+  linkImage: function(src, alt){
+    return '<div class="help-image"><img src="' + ImagePaths.get(src) + '" alt="' + alt + '"/></div>';
+  },
```
svbergerem: We don't use that function anymore. Should we still add it?

AugierLe42e: It may be useful later. Now it is written, why should we remove it ?

app/assets/javascripts/app/views/help_view.js
```diff
@@ -187,8 +197,31 @@ app.views.Help = app.views.StaticContentView.extend({
...
+  chatEnabled: function(){
+    return gon.chatEnabled;
+  },
+
+  getChatIcons: function(){
+    return '<div class="help-chat-icons"> \
+              <i class="entypo lock-open"></i> \
+              <i class="entypo chat"></i> \
+              <i class="entypo trash"></i> \
+            </div>';
```
svbergerem: I agree with Hound here. I'd prefer
```js
return '<div class="help-chat-icons">' +
       '  <i class="entypo lock-open"></i>' +
       '  <i class="entypo chat"></i>' +
       '  <i class="entypo trash"></i>' +
       '</div>';
```

app/assets/javascripts/app/views/help_view.js
jhass: Because with that argument we would still maintain about 6-10kloc more code in this repository. Please remove it.

app/views/help/faq.html.haml
```diff
@@ -3,9 +3,9 @@
...
 
 - content_for :head do
-  = load_javascript_locales("help")
+  = load_javascript_locales('help')
 
 - content_for :page_title do
   = t('_help')
 
-#help
+#help
```
svbergerem: These are no important changes in this file so please don't commit them.

config/locales/diaspora/en.yml
```diff
@@ -449,6 +450,10 @@ en:
       person_multiple_aspects_a: "Yes. Go to your contacts page and click on “My contacts”. For each contact you can use the menu on the right to add them to (or remove them from) as many aspects as you want. Or you can add them to a new aspect (or remove them from an aspect) by clicking the aspect selector button on their profile page. Or you can even just move the pointer over their name where you see it in the stream, and a “hover-card” will appear. You can change the aspects they are in right there."
       delete_aspect_q: "How do I delete an aspect?"
       delete_aspect_a: "In your list of aspects on the left side of the main page, point your mouse at the aspect you want to delete. Click the little “Edit” pencil that appears on the right. Click the “Delete this aspect” button in the box that appears."
+    chat:
+      title: "Chat"
+      add_contact_roster_q: "How do I add contacts to my roster ?"
+      add_contact_roster_a: "Go to the %{contacts}, select an aspect and click on the chat icon to enable chat for the aspect. %{toggle_privilege} Or when you create a new aspect we will also ask you if you would like to enable chat for that aspect"
```
svbergerem: Maybe 'Go to the contacts page, ...'? In that case I'd remove `help.contacts` and add `help.chat.contacts_page` instead.

spec/javascripts/app/views/help_view_spec.js
```diff
@@ -8,6 +8,13 @@ describe("app.views.Help", function(){
     Diaspora.Page = "HelpFaq";
   });
 
+  describe("Fonctions", function(){
```
svbergerem: typo

AugierLe42e:
Ok now ?

config/locales/diaspora/en.yml
```diff
@@ -449,6 +449,11 @@ en:
...
       delete_aspect_q: "How do I delete an aspect?"
       delete_aspect_a: "In your list of aspects on the left side of the main page, point your mouse at the aspect you want to delete. Click the little “Edit” pencil that appears on the right. Click the “Delete this aspect” button in the box that appears."
+    chat:
+      title: "Chat"
+      contacts_page: "contacts"
+      add_contact_roster_q: "How do I add contacts to my roster ?"
+      add_contact_roster_a: "Go to the %{contacts_page} page, select an aspect and click on the chat icon to enable chat for the aspect. %{toggle_privilege} Or when you create a new aspect we will also ask you if you would like to enable chat for that aspect"
```
svbergerem: I think this should be 'Go to the %{contacts_page}, select ...' and `contacts_page` should be 'contacts page'. Otherwise there might be languages where the current solution doesn't work.

svbergerem:
Apart from the comment I just added this looks good. I have no time right now to test this PR and merge it afterwards so either you will have to wait a few days or someone else merges this PR after you fixed the final remark.

config/locales/diaspora/en.yml
```diff
@@ -449,6 +449,11 @@ en:
       person_multiple_aspects_a: "Yes. Go to your contacts page and click on “My contacts”. For each contact you can use the menu on the right to add them to (or remove them from) as many aspects as you want. Or you can add them to a new aspect (or remove them from an aspect) by clicking the aspect selector button on their profile page. Or you can even just move the pointer over their name where you see it in the stream, and a “hover-card” will appear. You can change the aspects they are in right there."
       delete_aspect_q: "How do I delete an aspect?"
       delete_aspect_a: "In your list of aspects on the left side of the main page, point your mouse at the aspect you want to delete. Click the little “Edit” pencil that appears on the right. Click the “Delete this aspect” button in the box that appears."
+    chat:
+      title: "Chat"
+      contacts_page: "contacts page"
+      add_contact_roster_q: "How do I add contacts to my roster ?"
```
jhass: No [Plenken](http://en.wikipedia.org/wiki/Plenken) please.

config/locales/diaspora/en.yml
```diff
@@ -449,6 +449,11 @@ en:
...
+      add_contact_roster_a: "Go to the %{contacts_page}, select an aspect and click on the chat icon to enable chat for the aspect. %{toggle_privilege} Or when you create a new aspect we will also ask you if you would like to enable chat for that aspect"
```
jhass: Also missing period at the end. How about "for this aspect", or "make contacts of this aspect in the chat available"? I'd drop the "Or " from the next sentence and say "you can choose to" instead of "we will ask you".

AugierLe42e:
Corrected.

config/locales/diaspora/en.yml
```diff
@@ -449,6 +449,11 @@ en:
       person_multiple_aspects_a: "Yes. Go to your contacts page and click on “My contacts”. For each contact you can use the menu on the right to add them to (or remove them from) as many aspects as you want. Or you can add them to a new aspect (or remove them from an aspect) by clicking the aspect selector button on their profile page. Or you can even just move the pointer over their name where you see it in the stream, and a “hover-card” will appear. You can change the aspects they are in right there."
       delete_aspect_q: "How do I delete an aspect?"
       delete_aspect_a: "In your list of aspects on the left side of the main page, point your mouse at the aspect you want to delete. Click the little “Edit” pencil that appears on the right. Click the “Delete this aspect” button in the box that appears."
+    chat:
+      title: "Chat"
+      contacts_page: "contacts page"
+      add_contact_roster_q: "How do I add contacts to my roster?"
+      add_contact_roster_a: "Go to the %{contacts_page}, select an aspect and click on the chat icon to enable chat for the aspect. %{toggle_privilege} When you create a new aspect you can choose to enable chat for that aspect"
```
jhass: Still missing period for the second sentence. Looking at it again, the second "the aspect" in the first sentence seems a bit redundant maybe just replace it with "it". Also maybe the second sentence might better end in something like "for it too."

AugierLe42e: Maybe :

*When you create a new one you can choose to enable chat for it* ?

jhass: Guess we need @goobertron's help here ;)

AugierLe42e:
@Zauberstuhl : whil working on this, maybe you want to add something ? Like how to connect with an external client ?

Zauberstuhl:
we can add items if it is merged. I have no access to your branch :)

config/locales/diaspora/en.yml
goobertron: Hi! I'm here.

Is it just the one question and answer being added, or is there more text elsewhere? This one seems to cover only how to enable chat for an entire aspect, not how to actually add an individual contact to your chat roster. Will there be, or should there be, help available for other aspects of the chat feature?

Is the list of people in your chat interface officially called a 'roster', or is that your term, Augier? If the term doesn't appear in the chat interface, I'll probably choose another one as 'roster' isn't the most common term.

AugierLe42e: > Will there be, or should there be, help available for other aspects of the chat feature?

Theorically, yes. The first that come to my mind is how to access the chat feature from an external client. And then, maybe, how to use video/audio.

> Is the list of people in your chat interface officially called a 'roster', or is that your term, Augier?

Nope, this is the official JSXC terminology and this the term that appears on the wiki. I may not be the best term, though.

The question may not be the right one, actually. What about : "*How do I fill my friends list ?*"

goobertron: How about something like:

> ### How do I chat with someone in diaspora*?
> 
> - First you need to enable chat for one of the aspects *[or all of them?]* that person is in.
> - Next, add that person to your 'roster' (the list of people available to chat to)
> - Finally, click their name in the chat interface and start a chat!

Note: I haven't been able to test the chat feature yet, so I've not really an idea about the actual steps. I'm just proposing a more general question with a series of answers for the individual steps required to start chatting with someone; you'll have to write text which is accurate to the actual use case, if what I've said above isn't accurate.

If chat is not automatically activated for each user when a podmin enables it in the pod, the first step might be:

> - If you haven't already enabled the chat feature, go to your user settings page to do this.

Does this sound like a useful approach?

Flaburgan: No need to add the contact to the roster, once you activated the chat for an aspect, every contacts in this aspect are immediately available to chat with.

AugierLe42e: > Finally, click their name in the chat interface and start a chat!

I think they'll that point by themselves ^_^

So, I'll propose :

> ### How do I chat with someone in diaspora*?
You need to enable chat for one of the aspects that person is in. To do so, go to the contacts page, select an aspect and click on the chat icon to enable chat for the aspect. Or, you can create a new one. Then you can choose to enable chat for it by enabling the option [here, get the text displayed in the Facebox]

What do you think about it ?

goobertron: We're nearly there! How about:

> ### How do I chat with someone in diaspora*?
> 
> First, you need to enable chat for one of the aspects that person is in. To do so, go to the %{contacts_page}, select the aspect you want and click on the chat icon to enable chat for the aspect. You could, if you prefer, create a special aspect called 'Chat' and add the people you want to chat with to that aspect. Once you've done this, open the chat interface and select the person you want to chat with.

I don't think any of the rest of the in-app help section contains images, so I wouldn't recommend adding them here. However, it would also be good to add a section about chat to the tutorials on the project site. To do that, make a PR in [github](https://github.com/diaspora/diaspora-project-site/tree/master/app/views/pages). I'll work with you on that for the English, if you like.

AugierLe42e: It's not images it's just entypo icons ;)
Ok to me, I update the text.

goobertron: You forgot to update the question as well as the answer...

config/locales/diaspora/en.yml
```diff
@@ -449,6 +449,11 @@ en:
       person_multiple_aspects_a: "Yes. Go to your contacts page and click on “My contacts”. For each contact you can use the menu on the right to add them to (or remove them from) as many aspects as you want. Or you can add them to a new aspect (or remove them from an aspect) by clicking the aspect selector button on their profile page. Or you can even just move the pointer over their name where you see it in the stream, and a “hover-card” will appear. You can change the aspects they are in right there."
       delete_aspect_q: "How do I delete an aspect?"
       delete_aspect_a: "In your list of aspects on the left side of the main page, point your mouse at the aspect you want to delete. Click the little “Edit” pencil that appears on the right. Click the “Delete this aspect” button in the box that appears."
+    chat:
+      title: "Chat"
+      contacts_page: "contacts page"
+      add_contact_roster_q: "How do I add contacts to my roster?"
```
goobertron: Should be "How do I chat with someone in diaspora*?"

app/controllers/help_controller.rb
```diff
@@ -1,4 +1,8 @@
...
-	before_filter -> { @css_framework = :bootstrap }
-	layout ->(c) { request.format == :mobile ? "application" : "with_header_with_footer" }
-end
+  before_filter -> { @css_framework = :bootstrap }
+  layout -> (c) { request.format == :mobile ? "application" : "with_header_with_footer" }
+
+  def faq
+    gon.chatEnabled = AppConfig.chat.enabled?
+  end
+end
```
AugierLe42e: You are not very kind to me, Milou... Always criticizing me :',(

AugierLe42e:
Looks ok to me now.

jhass:
Please squash commits. Tipp: You can edit a commit with `git commit --amend`

app/assets/javascripts/app/views/help_view.js
```diff
@@ -35,6 +36,13 @@ app.views.Help = app.views.StaticContentView.extend({
       }
     };
 
+    this.CHAT_SUBS = {
+      add_contact_roster_a: {
+        toggle_privilege: this.getChatIcons(),
+        contacts_page: this.linkHtml('/contacts', Diaspora.I18n.t('chat.contacts_page'))
```
jhass: Oh, actually we wanted to start using `js-routes` for these.

jhass:
Thanks!

PR #5666, opened 2015-02-14
Editorconfig

jhass:
http://editorconfig.org/

Settings are up to debate of course, I tried to match what's
used currently. Also please hint if I missed any file extensions in use.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


svbergerem:
Seems legit. Let's see if anyone has any objections. Otherwise I'll merge this.

svbergerem:
Thank you!

PR #5669, opened 2015-02-14
5638 added avatar fallback to notification dropdown

Faldrian:
This is a fix for the missing avatar fallback replacement which @jaywink noticed. This is an addition to #5638

### DIFF
--- a/app/assets/javascripts/widgets/notifications-badge.js
+++ b/app/assets/javascripts/widgets/notifications-badge.js
@@ -108,6 +108,9 @@
             self.getMoreNotifications();
         }
       });
+
+      // add avatar fallback if it can't be loaded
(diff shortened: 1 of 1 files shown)


svbergerem:
We have the same problem in the conversations view.
1. in the messages list on the left (participants list on hover)
2. in the list of participants for opened conversations
3. avatars for each message (not tested but I don't think that we implemented a fallback for those)

Faldrian:
I will add those, thanks. I don't like implementing the same code on many different places... but I don't think these views extend from a common class where the rendering is called, so we can just implement in one place... hm.

svbergerem:
I guess you would have to add that line at least to those views where we don't render a handlebars template.

PR #5247, opened 2014-09-24
[WIP] #4466 Port streams to bootstrap

pablocubico:
Hi!

This is a work in progress, some thing should be discussed, as you can see, we are using container-fluid which makes D\* go all the way to the screen width. I have a 1920x1080 monitor, so you can see how huge it gets. 

I don't know what's the "official" position about this, should we limit the width to some reasonable size? 

Also, we have to port the header.

Searching for "blueprint" on controllers and views gives no result. Not a scientific measurement, but I think we are really close to finally drop blueprint.

image

PR #5672, opened 2015-02-16
Bootstrap stream: add spinner and improve comment box

svbergerem:
This fixes two issues mentioned in #5247. It adds the ajax-spinner to the stream and improves the comment box. The height of the comment box is now set to 30px (just like in the blueprint stream) and only expands when someone enters some text that is too big. (done by js like for the blueprint stream) The submit button will be hidden unless the comment box is focused or there is some text in the comment box.

### DIFF
--- a/app/assets/stylesheets/comments.css.scss
+++ b/app/assets/stylesheets/comments.css.scss
@@ -36,7 +36,13 @@
     }
     padding-left: 12px;
     width: 95%;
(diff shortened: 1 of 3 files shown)


svbergerem:
The last commit also fixed the report and block buttons for Bootstrap streams.

jhass:
Thank you!

Issue #5673, opened 2015-02-16
Notifications with the same date but a different year grouped together on the Notifications page


Makailija:
I have two notifications with the same date, February 11th, but a different year, 2015 and 2014, grouped together on the Notifications page. I don't believe this is intended behavior, because it does not make sense to me why they should be grouped. Can someone confirm this behavior? Here is a screenshot:

![notifications3](https://cloud.githubusercontent.com/assets/2229403/6207638/bf6a7b6e-b5af-11e4-923b-8a4b2e16aa3f.png)

#5361 is kinda related.

goobertron:
Hmm, that's odd. And definitely not desired behaviour.

Have checked against two of my accounts: my main account (on a pod running 0.4.1.2) doesn't have this problem; my account on a pod running develop unfortunately doesn't have frequent enough notifications for there to be any two notifications on the same date from different years. We'll need someone else on a pod running develop to check that.

svbergerem:
I don't know much about ruby but [this](https://github.com/diaspora/diaspora/blob/develop/app/controllers/notifications_controller.rb#L47) looks to me like we are only comparing month and day. This would have been an issue for quite some time. @goobertron The code looks like the two notifications have to be on the same page. Only in that case, and if they were created on the same day but in a different year, this issue could occur. I guess on your main account those notifications weren't on the same page.

goobertron:
Hmm, interesting. That's plausible. If you're right, it would be a pretty rare use case (hardly any notifications during the year but two falling on the same date one year apart), but still a loophole worth fixing!

You're right, it's several dozen pages back to a year ago on my main account. @Makailija, can you confirm whether that notification from a year ago would have appeared on the same page even if it had been shown in its correct place?

goobertron:
Fixed by #5676 so can be closed.

Makailija:
@goobertron  Yes, the notification would have been on the same page, as you supposed.

Thanks for the great work, guys, you're awesome!

PR #5676, opened 2015-02-16
Add year to notifications page

### LINKED ISSUE #5361: Feature: Mark the year of the notifications on the Notifications page (desktop/mobile)
At the moment, the year of the notifications is not explicitly stated on the Notifications page. This makes the page a bit confusing, especially in the case that you have just a few notifications from a longer period of time. At the moment, you could have a Notifications page which looks like this:

**24 September** XXX started to share with you *(a month ago)*
**15 February** YYY commented your post *(8 months ago)*
**4 July** ZZZ has mentioned you *(a year ago)*
**3 December** WWW has liked your post *(2 years ago)*

(This is of course a very crude mock-up, but it illustrates well what I am after.)

First you are probably confused by the odd chronological order: December, July, February, September. Then you figure out that the notifications are from different years. Then, if you want to know from which year a notifications is are, you have to *count* it backwards from its timeago comment. So, after some brain-twisting, you have found out that WWW liked your post in 2012. (Or if you are really lucky, you notice that a tooltip with the full date is shown if you hoover over the timeago comment. I only noticed that after reading about it on another issue discussion.)

Be that as it may, adding the year of the notifications would make the page much more intuitive. I think the best way to do it would be to separate the notifications of a certain year by a heading. So instead of the confusing list above, you would have a clearer list like this one:

### 2014 ###
**24 September** XXX started to share with you *(a month ago)*
**15 February** YYY commented your post *(8 months ago)*
### 2013 ###
**4 July** ZZZ has mentioned you *(a year ago)*
### 2012 ###
**3 December** WWW has liked your post *(2 years ago)*

No confusion, brain-teasing, or searching for hidden tooltips.

This applies both to desktop and mobile versions of the Notifications page. I didn't find a similar issue, but I think #4866  might be referenced.

svbergerem:
Fixes #5361 and #5673.

### DIFF
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -55,12 +55,27 @@ module NotificationsHelper
     object_link(note, notification_people_link(note))
   end
 
-  def the_day(i18n)
-    i18n[0].match(/\d/) ? i18n[0].gsub('.', '') : i18n[1].gsub('.', '')
+  def the_day(date)
+    date.split('-')[2].to_i
   end
 
-  def the_month(i18n)
-    i18n[0].match(/\d/) ? i18n[1] : i18n[0]
+  def the_month(date)
+    I18n.l(Date.strptime(date, '%Y-%m-%d'), :format => '%B')
+  end
+
+  def the_year(date)
+    date.split('-')[0].to_i
+  end
+
+  def locale_date(date)
+    I18n.l(Date.strptime(date, '%Y-%m-%d'), :format => I18n.t('date.formats.fullmonth_day'))
+  end
+
+  def display_year?(year, date)
+    unless year
+      Date.current.year != the_year(date)
+    else
+      year != the_year(date)
+    end
   end
 end
-
--- a/spec/helpers/notifications_helper_spec.rb
+++ b/spec/helpers/notifications_helper_spec.rb
@@ -93,4 +93,23 @@ describe NotificationsHelper, :type => :helper do
       end
     end
   end
+
+  describe '#display_year?' do
+    it 'returns false if year is nil and the date includes the current year' do
+      expect(display_year?(nil,Date.current.strftime('%Y-%m-%d'))).to be_falsey
+    end
+
+    it 'returns true if year is nil and the date does not include the current year' do
+      expect(display_year?(nil,'1900-12-31')).to be_truthy
+    end
+
+    it 'returns false if the date includes the given year' do
+      expect(display_year?(2015,'2015-12-31')).to be_falsey
+    end
+
+    it 'returns true if the date does not include the given year' do
+      expect(display_year?(2015,'2014-12-31')).to be_truthy
+      expect(display_year?(2015,'2016-12-31')).to be_truthy
+    end
+  end
 end
--- a/app/assets/stylesheets/notifications.css.scss
+++ b/app/assets/stylesheets/notifications.css.scss
@@ -15,11 +15,26 @@
       margin-bottom: 10px;
     }
 
+    .year_container { margin-top: 75px; }
+    .header + .year_container { margin-top: 50px; }
+    .year {
+      background-color: $white;
+      color: $light-grey;
+      font-size: 20px;
+      line-height: 20px;
+      margin-bottom: -20px;
+      text-align: center;
+    }
+
+    .year_container + .day_group {
+      border-top: 1px solid $border-grey;
+      padding-top: 10px;
+    }
+
     .day_group {
       margin-bottom: 10px;
       padding-bottom: 10px;
       border-bottom: 1px dashed $border-grey;
-      &:last-child { border-bottom: none; }
       .date {
         text-align: center;
         color: $light-grey;
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -44,7 +44,7 @@ class NotificationsController < ApplicationController
     @notifications.each do |n|
       n.note_html = render_to_string( :partial => 'notify_popup_item', :locals => { :n => n } )
     end
-    @group_days = @notifications.group_by{|note| I18n.l(note.created_at, :format => I18n.t('date.formats.fullmonth_day')) }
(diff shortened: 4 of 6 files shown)


app/helpers/notifications_helper.rb
```diff
@@ -55,12 +55,27 @@ def notification_message_for(note)
...
+  end
+
+  def locale_date(date)
+    I18n.l(Date.strptime(date, '%Y-%m-%d'), :format => I18n.t('date.formats.fullmonth_day'))
   end
-end
 
+  def display_year(year, date)
+    if year.nil?
+      not Date.current().year === the_year(date)
```
jhass: `Date.current.year != the_year(date)`

app/helpers/notifications_helper.rb
```diff
@@ -55,12 +55,27 @@ def notification_message_for(note)
...
+    else
+      not year === the_year(date)
```
jhass: See above.

app/helpers/notifications_helper.rb
```diff
@@ -55,12 +55,27 @@ def notification_message_for(note)
...
+  def the_year(date)
+    date.split('-')[0].to_i
+  end
+
+  def locale_date(date)
+    I18n.l(Date.strptime(date, '%Y-%m-%d'), :format => I18n.t('date.formats.fullmonth_day'))
   end
-end
 
+  def display_year(year, date)
```
jhass: `display_year?`

app/helpers/notifications_helper.rb
```diff
@@ -55,12 +55,27 @@ def notification_message_for(note)
...
+    if year.nil?
```
jhass: `unless year`

spec/helpers/notifications_helper_spec.rb
```diff
@@ -93,4 +93,23 @@
       end
     end
   end
+
+  describe '#display_year' do
+    it 'returns false if year is nil and the date includes the current year' do
+      expect(display_year(nil,Date.current().strftime('%Y-%m-%d'))).to be false
```
jhass: No empty parens, `be_false` or better `be_falsey`

spec/helpers/notifications_helper_spec.rb
```diff
@@ -93,4 +93,23 @@
...
+    end
+
+    it 'returns true if year is nil and the date does not include the current year' do
+      expect(display_year(nil,'1900-12-31')).to be true
```
jhass: `be_true`/``be_truthy`

app/helpers/notifications_helper.rb
svbergerem: Oh, i didn't know that `unless` also works with `else`. I will change that.

svbergerem:
@jhass Thank you for comments. All of those should be fixed now.

app/helpers/notifications_helper.rb
jhass: It does, there's just no `elsunless` and it doesn't work with `elsif`.

goobertron:
Purely from a point of view of personal taste, I'd suggest less space in between the two years in the first screenshot, and possibly the year aligned on the left with the dates rather than centred over the notifications.

To make it more visible, the year could be in the dark red of unread notifications; it could even be placed to the right of the date to save vertical space. Something like this:

_notif2
Just some thoughts: I don't dislike what you've done, but I think my suggestions might help to make it a bit neater.

jhass:
Thanks!

jhass:
Ups, looks like I was a bit quick :P

goobertron:
Or I was too slow!

I was out for a few hours, so only just came back to the computer.

svbergerem:
@goobertron I think adding the year in red looks too distracting. What I tried to do with the big margin and the centered year was getting the users attention while going through his notifications without distracting him and to keep the focus on the content.

goobertron:
Sure, I understand that, and I'm also not convinced by the red. I'm only thinking of how the design might be tweaked to keep the attention-grabbing aspect without so much white space.

How about two solid lines, with the year in between them - either with the year centred over the whole div or aligned with the dates? Like one of these:

_notif3

_notif3a
(For some reason the upper of the two solid lines doesn't display at the magnification shown on the page on my screen, but zoom in and you'll see it.)

The year (and space between the rules) could even be a bit larger to grab attention more.

I hope you don't mind me suggestion changes, especially as your PR has already been merged!

svbergerem:
> I hope you don't mind me suggestion changes, especially as your PR has already been merged!

Definitely not. Your opinion is always much appreciated. :) The PR has been merged but that doesn't mean that we can't improve it in further PRs.

Those screenshots look much better but I'm still not really convinced. I think in both screenshots there should be some margin to the top. (And then we are basically back to the solution I implemented in this PR) Maybe we need someone else's opinion here.

jhass:
I like the version that I merged. If you want to make it even more obvious I'd make the year at least as big as the day in month, if not bigger. And maybe vertically center it, that is add the same amount of whitespace below and above, not necessarily increasing the total amount of whitespace though.

goobertron:
I should say that I don't object to the amount of white space in the version merged, and given that it will only appear between two years, it won't be an intrusive feature in most people's notification views (it would normally appear only every few pages). But if we can make the year numerals more obvious with less *blank* vertical space, it would look a bit nicer, I think. What I suggested above (and which Jonne confirmed) could work, like this:

_notif4

Anyway, I'm signing off for tonight - see you both tomorrow. Thanks, both, for your continued excellent work for Diaspora.

svbergerem:
Ok, now I get what you are talking about. :-P In that screenshot I guess there isn't really less white space but it doesn't feel like unused white space anymore. There is one problem, though. We will have to remove the border of the last notification before the year (you already did that in the screenshot) but that is not that easy. Given you have the following code
```haml
.day_group
.day_group
.year_container
.day_group
.day_group
```
You can define the style of the first `.day_group` after `year_container` with
```scss
.year_container + .day_group { ... }
```
but as far as I know there is now way to define the style of the last `.day_group` before `.year_container`.

Does anyone have a solution that doesn't need a big refactoring?

svbergerem:
Forget it, I solved the problem. :-P

Issue #4301, opened 2013-07-15
hashtags are not rendered after invisible characters


svbergerem:
See [this post](https://pod.geraspora.de/posts/1404123)

PR #5680, opened 2015-02-17
Strip Unicode format characters prior post processing

jhass:
Since they break hashtag parsing and some weird clients seem to insert them around `#`.


svbergerem:
I guess this would fix #4301.

jhass:
Oh, yeah. I didn't notice.

svbergerem:
Thank you.

PR #5685, opened 2015-02-19
Export user photos

margori:
My solution for bug #1802 

Implemented "Export photos" button in user settings, currently disabled.

I'm using 'zip' files instead of 'tar', 'cause it's widely known and easily handled.


jhass:
I fear there's a lot more to consider to this. Let's start with some assorted points that immediately pop into my mind.

* We shouldn't do such an expensive (and thus potentially lengthy) operation directly in the action, it should be a background job, else it will fail and run into timeouts.
  * This then raises the question of how we inform the user that the download is ready, when mail is enabled, we could just sent a mail with the download link.
  * However if mail is not available, we need a flow for that. One imaginable scenario is to pop a message to the user that the export will be prepared and they should check back in a few minutes. This then means we should keep the last export for a configurable amount of time.
    * That means we would need a cleanup job for the exports.
    * An existing export would need to be cleaned out if the user requests a new one.
    * A download link for an existing export would need to be displayed.
 * The download URL would need to contain a token, so they're not guessable.
  * We would need to store the URL somewhere in the DB
* For the unlikely event that I can be convinced to not put this into a background job, we should consider and investigate X-Sendfile as delivery mechanism.
* We probably should investigate and verify whether the chosen approach works on a PaaS platform such as Heroku and whether we want to support storage locations like Amazon S3.

Flaburgan:
You can have a look at how this is handled by https://github.com/diaspora/diaspora/pull/5499

jhass:
#5499 at least backgrounds the work and sends a mail, it doesn't have an optimal UI flow either yet nor any cleanup.

goobertron:
> However if mail is not available, we need a flow for that. One imaginable scenario is to pop a message to the user that the export will be prepared and they should check back in a few minutes. This then means we should keep the last export for a configurable amount of time.

How about: an alert that the job may take some time, and then the user gets a notification in the usual way when the export is ready, with a download link. The export could be kept until downloaded (perhaps up to a maximum time).

margori:
I'm gonna try to create a output stream.

If can't, I'm gonna use #5499 method so user see same behavior for both exports.

margori:
Best approach to stream a zip file is using https://github.com/fringd/zipline but it's not ported to rails 4 yet. :-(

margori:
Done: used #5499 way of export data.

Comments are very welcome.

app/models/user.rb
```diff
@@ -313,6 +313,33 @@ def compressed_export
...
+  end
+
+  def perform_export_photos!
+    export = Tempfile.new([username, '_photos.zip'])
+    
+    Zip::OutputStream.open(export) { |zos| }
+
+    Zip::ZipFile.open(export.path, Zip::File::CREATE) do |zipfile|
+      photos.each do |photo|
+          photo_path = "#{Rails.root}/public/uploads/images/#{photo.remote_photo_name}"
```
jhass: We can't assume that, it might be hosted on S3.

app/models/user.rb
```diff
@@ -313,6 +313,33 @@ def compressed_export
...
+
+  def queue_export_photos
+    update exporting_photos: true
+    Workers::ExportPhotos.perform_async(id)
+  end
+
+  def perform_export_photos!
+    export = Tempfile.new([username, '_photos.zip'])
+    
+    Zip::OutputStream.open(export) { |zos| }
```
jhass: Huh?

config/locales/diaspora/en.yml
```diff
@@ -812,6 +812,27 @@ en:
...
+        The diaspora* email robot!
+    export_photos_failure_email:
+      subject: "We’re sorry, there was an issue with your photos, %{name}"
+      body: |-
+        Hello %{name}
+
+        We’ve encountered an issue while processing your photos for download.
+        Please try again!
+
+        Cheers,
```
goobertron: A purely style comment: I think saying 'cheers' here might seem a bit dismissive, because it's giving bad news, and 'cheers' is more of an upbeat/happy sign-off. Perhaps remove 'we're sorry' from the email title and change this line from 'Cheers,' to 'Sorry,'.

config/locales/diaspora/en.yml
```diff
@@ -812,6 +812,27 @@ en:
...
+      body: |-
+        Hello %{name},
+
+        Your photos have been processed and are ready for download by following [this link](%{url}).
+
+        Cheers,
+
+        The diaspora* email robot!
+    export_photos_failure_email:
+      subject: "We’re sorry, there was an issue with your photos, %{name}"
```
goobertron: I'd suggest rewording this title to 'There was a problem exporting your photos', move the 'sorry' bit to the end of the email and remove the user name from the title.

app/models/user.rb
margori: If we can't then PR #5499 requires review as well.

jhass: How so?

app/models/user.rb
margori: Strange code seen [here](http://thinkingeek.com/2013/11/15/create-temporary-zip-file-send-response-rails/)

jhass: Strange indeed.

app/models/user.rb
margori: Oh! You're talking about line 331. I thought you were talking about line 317.

I assume that files are stored in same machine as web server. :-(

margori:
New try. In this opportunity I'm assuming files could not be in local file system.

app/models/user.rb
```diff
@@ -313,6 +313,45 @@ def compressed_export
...
+  end
+
+  def perform_export_photos!
+    export = Tempfile.new([username, '_photos.zip'])
+    
+    Zip::ZipOutputStream.open(export.path) do |zos|
+      photos.each do |photo|
+        temp_photo = Tempfile.new(photo.remote_photo_name, encoding: 'ascii-8bit')
+        begin
+          photo_path = "#{Rails.root}/public/uploads/images/#{photo.remote_photo_name}"
```
jhass: Sorry, but `Rails.root` does not give you the upload location or even the sites domain or anything, it gives you the absolute path on the filesystem where the application is located, so this still doesn't work for stuff stored on S3.

I would give you more pointers, but I'm myself very unsure about what the best way to handle that case would be.

app/models/user.rb
```diff
@@ -313,6 +313,45 @@ def compressed_export
...
+          if File.file?(photo_path)
+            temp_photo.write(IO.read(photo_path))
+          else
+            temp_photo.write(URI.parse(photo.url).read)
```
jhass: Your tests never ran into this, there's no URI#read afaik

margori: You're right. Some photos in test fixture are coming...

margori:
New try.

In this opportunity, photo file is gotten through unprocessed photo uploader.

margori:
This PR is related to #5343

app/models/user.rb
```diff
@@ -313,6 +313,35 @@ def compressed_export
...
+  end
+
+  def perform_export_photos!
+    temp_zip = Tempfile.new([username, '_photos.zip'])
+    Zip::ZipOutputStream.open(temp_zip.path) do |zos|
+      photos.each do |photo|
+        begin
+          zos.put_next_entry(photo.remote_photo_name)
+          zos.print(photo.unprocessed_image.file.read)
+        rescue
```
jhass: Do you have a more specific exception that you expect? Or even a better, a way to guard the code raising to from running in that case?

app/controllers/users_controller.rb
```diff
@@ -152,10 +152,15 @@ def download_profile
...
   def export_photos
-    tar_path = PhotoMover::move_photos(current_user)
-    send_data( File.open(tar_path).read, :filename => "#{current_user.id}.tar" )
+    current_user.queue_export_photos
+    flash[:notice] = I18n.t('users.edit.export_photos_in_progress')
+    redirect_to edit_user_path
   end
 
+  def download_photos
+    send_data File.open(current_user.exported_photos_file.path).read, filename: current_user.exported_photos_file.filename
```
jhass: `File.open.read` is very bad, it doesn't ensure the file is closed and thus leaks FDs. `File.read` would ensure that, however it reads the entire file into memory, which is bad too. This should probably use `X-Sendfile` or a streaming API.

margori: I thought about using send_file but works for local files, fails on remote.

If remote then it's better to redirect.

What do you think?

app/models/user.rb
margori: If file doesn't exists it returns Errno::ENOENT.

app/controllers/users_controller.rb
jhass: Actually, why not redirect in both cases?

margori: Good idea! Let's try it.

margori:
Done: redirecting when download_photos is called.

In this case, filename should be unguessable. Then new request are stored with random filename.

app/uploaders/exported_photos.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def filename
+    "#{model.username}_photos_#{secure_token(12)}.zip" if original_filename.present?
+  end
+
+  protected
+  def secure_token(length=16)
+    var = :"@#{mounted_as}_secure_token"
+    model.instance_variable_get(var) or model.instance_variable_set(var, SecureRandom.hex(length/2))
+  end
```
jhass: That's just `16**6` combinations, I see no reason to limit it that much.

margori:
Filename token increased to 128bits.

margori:
Re-based and code updated for new gems.

Need review. Thanks!

app/models/user.rb
```diff
@@ -314,6 +314,36 @@ def compressed_export
...
+    update exporting_photos: true
+    Workers::ExportPhotos.perform_async(id)
+  end
+
+  def perform_export_photos!
+    temp_zip = Tempfile.new([username, '_photos.zip'])
+    Zip::ZipOutputStream.open(temp_zip.path) do |zos|
+      photos.each do |photo|
+        begin
+          photo_data = photo.unprocessed_image.file.read
```
jhass: Looking good so far, I guess we just have to see how it behaves on S3 (+ Heroku) setup. I still don't understand how this call pulls the photo from S3 tbh. If you're able to clarify that... :)

Another last small remark I have is moving the `temp_zip.close` below to an `ensure` block.

margori: This is what a thought:
- If I delete a post with a photo in it, photo file is deleted from my disc.
- I don't found code that delete (unlink) the file. So there's an object that knows what to do.
- The uploader field may know where to upload and to delete it.
- If so, it may also know where to read it.
- I tried it and worked, at least in my local file system.
- I found that admin can choose where to store files by setting [carrierwave config file](https://github.com/diaspora/diaspora/blob/develop/config/initializers/carrierwave.rb). I didn't try over S3. Amazon wants my credit card number for a free account and... no, darling.

What I really want to know is how to create a test to simulate a S3 storage. Do you have a sample code?

jhass: No, I don't have an example test for that, I'm actually not sure whether it's testable, I guess one could enable hosting on S3 in the test and then webmock should catch the network call.

Hit me up on IRC if you want credentials to a S3 bucket for your tests.

jhass: I think I fixed the S3 access ;)

margori: Yes, you did. :)

Downloads work as expected, but having problem with upload. CarrierWave use same file name of downloaded image for the new uploaded zip. So upload fails saying "You can upload a jpg. A zip expected" :(

jhass: Mmh, it did upload a zip though:

margori: Good. This capture confirms that one export file is maintained.

margori:
New changes
- Removed "zip" extension validation in 'exported_photos.rb'.
- Ensured "update exporting_photos: false" in case of errors.

These changed were successfully tested in AWS S3.

jhass:
Alright, thank you!

Flaburgan:
Nice, thank you!

margori:
UHU! Thank you too!

goobertron:
Hey, thanks very much, @margori!

PR #5686, opened 2015-02-19
Fix notification for new conversations

### LINKED ISSUE #5330: Private messages do not make notification for new message when new thread is open
Steps to reproduce : send a new private message to someone (not a new answer to a previous PM), he will receive no notification.

svbergerem:
Fixes #5330, maybe introduced by [this commit](https://github.com/diaspora/diaspora/commit/33c3b38f2f8c6ede39b1357eb3cb49f18543cdce). The spec I posted in #5330 wasn't working but I wrote a cucumber feature.

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -20,7 +20,7 @@ class Conversation < ActiveRecord::Base
   def max_participants
     errors.add(:max_participants, "too many participants") if participants.count > 20
   end
-  
+
   def local_recipients
     recipients.each do |recipient|
       if recipient.local?
@@ -86,7 +86,7 @@ class Conversation < ActiveRecord::Base
   end
 
   def receive(user, person)
-    cnv = Conversation.find_or_create_by!(self.attributes)
+    cnv = Conversation.create_with(self.attributes).find_or_create_by!(guid: guid)
 
     self.participants.each do |participant|
       ConversationVisibility.find_or_create_by(conversation_id: cnv.id, person_id: participant.id)
--- a/features/desktop/conversations.feature
+++ b/features/desktop/conversations.feature
@@ -20,5 +20,7 @@ Feature: private conversations
     And "Alice Awesome" should be part of active conversation
     And I should see "hello, alice!" within ".stream_container"
     When I sign in as "alice@alice.alice"
-    And I reply with "hey, how you doing?"
+    Then I should have 1 unread private message
+    And I should have 1 email delivery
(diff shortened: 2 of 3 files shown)


Flaburgan:
@svbergerem hm, if the commit you linked is the problem, maybe we have other regression elsewhere.

jhass:
Mmh, you mean that self.attributes creates something that makes it always unique or never matching?

svbergerem:
@jhass Here is an example of `self.attributes`:
`{"id"=>nil, "subject"=>"test9", "guid"=>"30cbe7809a5e0132c4c9047d7b8eb256", "author_id"=>1, "created_at"=>Thu, 19 Feb 2015 12:10:17 UTC +00:00, "updated_at"=>nil}`

I guess there is no object with the same attributes in the db so we try to create a new one which fails because there is already a conversation with the same guid in the db.

app/models/conversation.rb
```diff
@@ -86,7 +86,7 @@ def subscribers(user)
   end
 
   def receive(user, person)
-    cnv = Conversation.find_or_create_by!(self.attributes)
+    cnv = Conversation.create_with(self.attributes).find_or_create_by!(self.attributes[:guid])
```
jhass: Doesn't this have to be .`.find_or_create_by!(guid: guid)` then? This just seems to insert a `WHERE 123456abcdef` into the query if I'm following correctly.

svbergerem: Already tried that. In that case we create a second conversation in the cukes which makes them fail:

jhass: Just checked, it generates the following query:

```sql
SELECT  "conversations".* FROM "conversations"  LIMIT 1
```

That just happens to match in the test.

jhass: `Conversation.create_with(attributes).find_or_create_by!(guid: guid)` generates: `SELECT  "conversations".* FROM "conversations"  WHERE "conversations"."guid" = 'cf7cd2e09a6001329fdd14feb5c40cd3' LIMIT 1` and doesn't fail the test for me.

svbergerem:
@Flaburgan Of course we have other regressions somewhere else but I don't think that the linked commit caused them. :-P

All other changes in the commit seem legit.

jhass:
Alright, good catch! Thank you!

Flaburgan:
Thank you Steffen! This bug was so annoying!

marienfressinaud:
I strongly agree, thanks a lot!

PR #5689, opened 2015-02-19
5366 commenting when ignored

WebDevFromScratch:
I hope the commit names are quite self-explanatory. I am not sure how to implement tests:
- in case of userLike, I just simply don't know how to test this...
- in case of flash alert for failing like, I didn't see a corresponding backbone view test and am not sure if I should create a file to test just this case?

Hope the code is not too bad, it's my first time working on an open source project :) Feedback would be most appreciated!

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -134,6 +134,7 @@ diaspora.yml file**. The existing settings from 0.4.x and before will not work a
 * Fix missing translation on privacy settings page [#5671](https://github.com/diaspora/diaspora/pull/5671)
 * Fix code overflow for the mobile website [#5675](https://github.com/diaspora/diaspora/pull/5675)
 * Strip Unicode format characters prior post processing [#5680](https://github.com/diaspora/diaspora/pull/5680)
+<<<<<<< HEAD
 * Disable email notifications for closed user accounts [#5640](https://github.com/diaspora/diaspora/pull/5640)
 * Made failure warning more explicit when attempting to interact with a post of someone who's ignoring you
 [#5689](https://github.com/diaspora/diaspora/pull/5689)
--- a/Changelog.md
+++ b/Changelog.md
@@ -135,6 +135,8 @@ diaspora.yml file**. The existing settings from 0.4.x and before will not work a
 * Fix code overflow for the mobile website [#5675](https://github.com/diaspora/diaspora/pull/5675)
 * Strip Unicode format characters prior post processing [#5680](https://github.com/diaspora/diaspora/pull/5680)
 * Disable email notifications for closed user accounts [#5640](https://github.com/diaspora/diaspora/pull/5640)
+* Made failure warning more explicit when attempting to interact with a post of someone who's ignoring you
+[#5689](https://github.com/diaspora/diaspora/pull/5689)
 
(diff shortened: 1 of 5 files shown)


Changelog.md
```diff
@@ -134,6 +134,8 @@ diaspora.yml file**. The existing settings from 0.4.x and before will not work a
 * Fix missing translation on privacy settings page [#5671](https://github.com/diaspora/diaspora/pull/5671)
 * Fix code overflow for the mobile website [#5675](https://github.com/diaspora/diaspora/pull/5675)
 * Strip Unicode format characters prior post processing [#5680](https://github.com/diaspora/diaspora/pull/5680)
+* Commenting (+ liking) on a post of someone who ignores you to show the reason
+[#5366](https://github.com/diaspora/diaspora/pull/5366)
```
Flaburgan: I think you can write "Fix ... ignores you **doesn't** show the reason" and put the link on the same line.

app/assets/javascripts/app/models/post/interactions.js
```diff
@@ -44,12 +44,16 @@ app.models.Post.Interactions = Backbone.Model.extend({
   },
 
   userLike : function(){
-    return this.likes.select(function(like){ return like.get("author").guid === app.currentUser.get("guid")})[0];
+    return this.likes.select(function(like){
+      if(like.get("author") !== undefined) {
```
Flaburgan: In which case the author can be undefined? That looks like a bug in the creation of the like object.

Changelog.md
goobertron: I'd say 'Make failure warning more explicit when attempting to interact with a post of someone who's ignoring you'.

app/assets/javascripts/app/models/post/interactions.js
WebDevFromScratch: Okay, so here is how I found this:
* hit 'like' for the first time on a post by someone who's ignoring you -> 422 error
* then hit the 'like' button again -> see what happens

svbergerem: @Flaburgan We do the same for reshares:

```javascript
return reshare.get("author") && reshare.get("author").guid === app.currentUser.get("guid")
```

@WebDevFromScratch Could you unify this function and the `userReshare` function? I'd prefer the version of the `userReshare` function because here you don't return anything if the author is undefined. In `userReshare` we return false. Please also fix the issues Hound mentioned.

Changelog.md
```diff
@@ -134,7 +134,10 @@ diaspora.yml file**. The existing settings from 0.4.x and before will not work a
 * Fix missing translation on privacy settings page [#5671](https://github.com/diaspora/diaspora/pull/5671)
 * Fix code overflow for the mobile website [#5675](https://github.com/diaspora/diaspora/pull/5675)
 * Strip Unicode format characters prior post processing [#5680](https://github.com/diaspora/diaspora/pull/5680)
+<<<<<<< HEAD
```
svbergerem: This line should be removed.

svbergerem:
Superseded by #6646. Thank you.

PR #5691, opened 2015-02-19
Contacts page: 'Start a conversation' button should depend on number of contacts in aspect

### LINKED ISSUE #5235: Contacts page: "Start a conversation" button depends on total number of contacts
There is a bug in the new contacts page: The button to start a conversation should be visible if there are less than 20 contacts in your aspect. Right now that depends on your total number of contacts.

To fix this one would need to change the contacts controller ([line 42](https://github.com/diaspora/diaspora/blob/develop/app/controllers/contacts_controller.rb#L42)): <code>@contacts_size</code> should be the number of contacts in your aspect.

svbergerem:
Fixes #5235.

### DIFF
--- a/app/views/contacts/_header.html.haml
+++ b/app/views/contacts/_header.html.haml
@@ -1,8 +1,8 @@
 .header
   - if @aspect
     #aspect_controls.pull-right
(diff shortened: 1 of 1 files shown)


jhass:
Thanks!

Issue #5315, opened 2014-10-09
Indicates in the setting that the email is not visibile to the other users


Flaburgan:
I have a user who asked me about that, we can make that more clear. I'm probably going to open a PR soon, just open this issue as a reminder.

aka001:
I want to work on this issue. 
I have a query, please clarify this.
By design, the email is not visible to other users.
So, is the issue about specifying in the settings page that the email isn't public?

Flaburgan:
Yes, just add a explanation about it, like "Your email will never be seen by other users".
Maybe this should be indicated on the registration page too.

aka001:
Okay. I will do that and send a pr soon.

aka001:
I will continue on this issue later. Meanwhile if someone wishes to fix this issue, feel free to do so.

Flaburgan:
https://github.com/diaspora/diaspora/pull/5956 added this but looks like the port to bootstrap broke it. Should I close this and report it on https://github.com/diaspora/diaspora/issues/6063 ?

PR #5692, opened 2015-02-19
Indicates in the setting that the email is not visibile to the other users #5315

aka001:
The screenshot shows the changes done. Kindly review this pr. 
screenshot from 2015-02-20 03 03 05
screenshot from 2015-02-20 03 02 54

### DIFF
--- a/app/views/registrations/new.haml
+++ b/app/views/registrations/new.haml
@@ -41,4 +41,5 @@
               - if AppConfig.settings.terms.enable?
                 %p#terms.text-center
                   = t('.terms', terms_link: link_to(t('.terms_link'), terms_path, target: "_blank")).html_safe
+              %small= t('users.edit.your_email_private')
               = f.submit t('.sign_up'), class: "btn btn-block btn-large", data: {disable_with: t('.submitting')}
--- a/app/views/users/edit.html.haml
+++ b/app/views/users/edit.html.haml
@@ -27,6 +27,7 @@
         .span6
           %h3
             = t('.your_email')
+          %small= t('.your_email_private')
           = form_for 'user', :url => user_path, :html => { :method => :put } do |f|
             = f.error_messages
             .form-inline
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -1280,6 +1280,7 @@ en:
       change: "Change"
       your_handle: "Your diaspora* ID"
       your_email: "Your email"
+      your_email_private: "Your email will never be seen by other users"
       change_email: "Change email"
       email_awaiting_confirmation: "We have sent you an activation link to %{unconfirmed_email}. Until you follow this link and activate the new address, we will continue to use your original address %{email}."
       change_password: "Change password"


app/views/registrations/new.html.erb
```diff
@@ -0,0 +1,82 @@
...
+              <div id="terms" class="text-center">
+                <%= t('.terms', terms_link: link_to(t('.terms_link'), terms_path, target: "_blank")).html_safe %>
+              </div>
+            <% end %>
+            <%= f.submit t('.continue'), class: "new-btn", data: {disable_with: t('.submitting')} %>
+        <% end %>
+      </div>
+    </div>
+  </div>
+</div>
```
aka001: Does this file need to be dropped?

jhass: I guess you started working without updating from upstream first, then rebased before opening the PR? #5391 was merged today and that ported this template to HAML

aka001: In that case, since there are minor changes, I will create a new pr.

svbergerem: There is no need for a new PR. You can just modify the changes in this PR.

aka001: I have updated the pr. Kindly look into it.
The latest changes in the registration page:
screenshot from 2015-02-20 05 41 01

svbergerem:
Looks like you removed the wrong changes from this PR. Let's clean up this PR before talking about the code itself.
1. Remove the right file:
`git rm app/views/registrations/new.html.erb`
2. Change `app/views/users/edit.html.haml` like you did that before. (You removed those changes)
3. Revert the changes in `config/locales/diaspora/en.yml` except for the strings you added. You only need to add the string once and you can use it in both places. For example when you keep the string for `users.edit.your_email_private` you can use `t('.your_email_private')` in `app/views/users/edit.html.haml` and `t('users.edit.your_email_private')` everywhere else.

aka001:
@svbergerem Thank you for the detailed explanation. I have updated the pr and squashed the commits. Kindly review them.

goobertron:
I'm not sure why the heading 'Your email' appears a line below 'Your diaspora* ID'; the two should be aligned horizontally. It doesn't look as though your PR could have changed this. Could it be your browser?

aka001:
@goobertron It was indeed my browser. In firefox the view appears as:
screenshot from 2015-02-21 03 36 50

svbergerem:
@aka001 This is much better. For the settings page I'd prefer a lock icon with a tooltip instead of the text. That would be similar to what we do in the single post view or what @jaywink started to do in #4596. The lock icon for the single post view is added [here](https://github.com/diaspora/diaspora/blob/develop/app/assets/templates/single-post-viewer/single-post-content_tpl.jst.hbs#L35). We are using enypo icons, in this case the lock icon which you get with
```haml
%i.entypo.lock.gray
```
Just like in the single post view you could put that in a `%span` where you add the title for the tooltip.

You would have to write a small backbone view (similar to the [tags view](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/tags_view.js)) where you [activate the tooltip](http://getbootstrap.com/2.3.2/javascript.html#tooltips) and call it from the [router](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/router.js). That could look similar to this:


For the registration page I don't think that this is the right position for the text. I don't have a better idea but maybe @goobertron or @Flaburgan could help here. (perhaps they are even happy with the current solution)

Flaburgan:
Yeah the place of the indication on the registration page is not good imo, because:
1. This is where we display the link to the terms
2. The user needs to have the information before entering his email, here, it's too far after.

svbergerem:
How about a [popover](http://getbootstrap.com/2.3.2/javascript.html#popovers) on focus? (We could also use a tooltip but to me popovers looked more appropriate in this case.)


You can add the text with the `data-content` attribute to the input. The popover will automatically use the value of the title attribute (“Enter an email”) as the header of the popover. You would have to initialize the popover in a backbone view with code similar to this one:
```javascript
$('input#user_email').popover({
  placement: 'left',
  trigger: 'focus'
});
```

Flaburgan:
yeah that's better, though I don't know what's better between popover and tooltip

svbergerem:
This is the sign up page with a tooltip:
sign up tooltip

I'd prefer the tooltip for the settings page and the popover for the sign up page.

goobertron:
Thanks fr supplying the ideas and screenshots, Steffen.

> I'd prefer the tooltip for the settings page and the popover for the sign up page.

I agree.

It would be nice to get some indication on the sign-up page that there *is* more information to be found, in a popover or whatever. I was wondering about using the lock icon inside the email text entry field, perhaps at the right-hand side. However, the lock icon is already used for the password fields, so that might be confusing.

svbergerem:
@goobertron We could activate autofocus for the email input like @Flaburgan did this for the username on the login page [here](https://github.com/diaspora/diaspora/blob/develop/app/views/sessions/new.html.haml#L22). Then the popover would be visible immediately.

goobertron:
> We could activate autofocus for the email input like @Flaburgan did this for the username on the login page here. Then the popover would be visible immediately.

Perfect!

Flaburgan:
if the tooltip is displayed on focus, it's already fine imo.

aka001:
Thank you everyone for your inputs. I am running a bit busy with my college projects. I will complete it by tomorrow night IST.

svbergerem:
@aka001 What is the status? Do you need any help?

aka001:
@svbergerem Thank you for the reply. I do need help but I will continue this work after some days and mention the details here. I am having exams in 4 days. Sorry for the delay.

svbergerem:
Sure, no problem. Good luck!

svbergerem:
@aka001 ping.

aka001:
@svbergerem I have a project deadline tomorrow and will be travelling after that. I will start on this issue after 30th April. Sorry for the delay.

svbergerem:
Superseded by #5956. Thank you for your work!

aka001:
@svbergerem Awesome.

Issue #5690, opened 2015-02-19
Re-sharing a post doesn't update the stream


WebDevFromScratch:
So, if you post a new message or comment, like, etc. the stream view updates. But when you re-share someone elses post it does not. Is it supposed to stay this way?

I hope I'm not repeating an issue here, I couldn't find a similar one (although it would be really weird if noone noticed this behaviour before).

jhass:
Sure somebody noticed, but looks nobody cared to open an issue yet ;)

WebDevFromScratch:
Okay ;) I can actually look into it, too.

PR #5699, opened 2015-02-23
5690 update stream on reshare

WebDevFromScratch:
This is supposed to fix issue #5690.

While the immediate adding new reshares to the stream now works, the re-shares count is not updated for this newly reshared post. I tried to fix that, but unsuccessfully. I think `app.views.Reshare` in `content_view.js` is responsible for doing this, but I couldn't make it work.

### DIFF
--- a/app/assets/javascripts/app/models/post/interactions.js
+++ b/app/assets/javascripts/app/models/post/interactions.js
@@ -103,23 +103,21 @@ app.models.Post.Interactions = Backbone.Model.extend({
       , reshare = this.post.reshare()
       , flash = new Diaspora.Widgets.FlashMessages();
 
-    reshare.save({}, {
-      success : function(){
+    reshare.save()
+      .done(function(){
         flash.render({
           success: true,
           notice: Diaspora.I18n.t("reshares.successful")
         });
-      },
-      error: function(){
+        interactions.reshares.add(reshare);
+        if (app.stream) {app.stream.addNow(reshare)}
+        interactions.trigger("change");
+      })
+      .fail(function(){
         flash.render({
           success: false,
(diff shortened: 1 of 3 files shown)


jhass:
Looking good to me, if you could just fix the style issue and rebase into a clean commit using `git rebase -i` to squash them?

app/assets/javascripts/app/models/post/interactions.js
```diff
@@ -103,23 +103,21 @@ app.models.Post.Interactions = Backbone.Model.extend({
...
+    reshare.save()
+      .done(function(){
         flash.render({
           success: true,
           notice: Diaspora.I18n.t("reshares.successful")
         });
-      },
-      error: function(){
+        interactions.reshares.add(reshare);
+        if (app.stream) {app.stream.addNow(reshare)};
```
WebDevFromScratch: From one error to another, meh... @jhass Could you tell me if squashed the commits now? I thought I did this before, but apparently I must have been doing something wrong..

jhass: No you didn't. `git rebase -i upstream/develop` should open an editor where you can choose `pick`, `squash`, `fix` and so on. You keep the first commit at `pick` and choose `squash` for the others.

WebDevFromScratch: Oh, okay, now I remember that I was having some problems with atom editor and those editing `rebase` before... I will fix this.

jhass: Cool, tip for the future: You can edit a commit with `git commit --amend`, which is fine to do as long as you're in feature branch.

jhass:
Thank you!

Issue #5469, opened 2014-12-09
Add links to /commented and /liked streams in sidebar


goobertron:
These views are available, but currently, as far as I know, it is only possible to access them by adding the relevant path manually to the pod domain in the URL bar.

It would be good to provide links to them under 'My Activity' in the sidebar menu, so that when 'My Activity' is clicked (or when /activity is loaded by some other method), the two options appear under 'My Activity' in the sidebar menu. As so:

![streams](https://cloud.githubusercontent.com/assets/2530168/5359434/47acd408-7fb5-11e4-9b87-5437e286b7f8.png)

I'd like to try implementing this myself, so I'm not just someone who says 'You should do this!', but on looking at the code it looks too complex for me, so I won't volunteer!
##

Fensterbank:
Hm, what happened with this commit? This feature never came...

Flaburgan:
I still have to fix the legend of the avatars in the right column. Waiting for #6535 to know if we keep it or not.

noplanman:
I just came across this now and wanted to point out a userscript that I put together when I joined diaspora*.
You can find it [here](https://github.com/noplanman/DiasPlus) and it adds the 2 mentioned links in the top menu (and a few other features).

Flaburgan:
I still plan to rebase my PR on develop, but that's not targeted for 0.6 so we have more urgent things to do.

mahonrig:
Looks like this was completed, close this issue?

SuperTux88:
You are right, closing 👍

noplanman:
In which version were they added?
Will have to update my userscript to remove them to prevent any duplicate links being added.

SuperTux88:
0.6.5.0 (I marked the issue with the milestone)

noplanman:
Aaah right, just saw that they get added as submenu items, all clear!
Thanks @SuperTux88

PR #5502, opened 2014-12-29
Add links to liked and commented pages

Flaburgan:
Closes #5469

I still have some css and JS to write but it already works.

PR #5703, opened 2015-02-23
Implemented #5469

sam09:
Liked and Commented posts can be accessed directly from sidebar.


Flaburgan:
I already opened a pull request to add those links: https://github.com/diaspora/diaspora/pull/5502

sam09:
Okay

denschub:
So, this is actually a dupe of #5502 then. Sorry @sam09, but thanks for your work! Check out the other open issues as well :)

Issue #5698, opened 2015-02-21
Update the help section


Flaburgan:
A lot of things changed in 0.5 so we need to update the help section. This task can be done by non-technical persons so don't hesitate to ask here if you want to help even if you know nothing about the codebase, we can help you.

AugierLe42e:
Maybe worth to make a list of changes @Flaburgan ?

aka001:
I would like to help with this issue. Please help me in getting started.

Flaburgan:
@aka001 this is not a technical issue but only a check that actual help texts match the new version of diaspora*. So, simply open diaspora* running the last develop code, go to the help section, read and change it if needed to reflect the new diaspora* interface.

aka001:
@Flaburgan Okay. I will create a pr for the same.

aka001:
I have started with this pull request https://github.com/diaspora/diaspora/pull/5706 . Kindly have a look.

Flaburgan:
I guess this should have the "next-major" milestone.

goobertron:
If someone can list which functionality has changed and therefore which parts of the help section need updating, I'll write some new text. I don't have the energy right now to go through the whole UI testing for changes, I'm afraid.

Flaburgan:
I have most of the changes in mind I think, I'll see if I can find some time to open a PR.

goobertron:
Adding a location to the post: we should say where this information is obtained, and who your data are shared with. It looks to be Mozilla geolocator, which from their FAQ appears to send user data to Google Location Services: see https://www.mozilla.org/en-GB/firefox/geolocation/

Is this correct?

goobertron:
Ah, got it from /app/assets/javascripts/osmlocator.js - it's OpenStreetMap.

goobertron:
PR submitted: #5859

Please check the changes and make sure all the updates required have been made, and let me know of anything else I need to do.

PR #5706, opened 2015-02-27
Update the help section #5698

aka001:


aka001:
Is giving link to loomio fine under "What if my question is not answered in this FAQ? Where else can I get support?"

Flaburgan:
No, we don't want support request on Loomio. You can redirect to the FAQs on the wiki though.

config/locales/diaspora/en.yml
```diff
@@ -412,13 +413,14 @@ en:
...
       get_support_a_irc: "Join us on %{irc} (live chat)"
+      get_support_a_faq: "Read our %{faq} page on wiki"
       get_support_a_hashtag: "Ask in a public post on diaspora* using the %{question} hashtag"
     account_and_data_management:
       title: "Account and data management"
       move_pods_q: "How do I move my seed (account) from one pod to another?"
       move_pods_a: "In the future you will be able to export your seed from a pod and import it on another, but this is not currently possible. You could always open a new account and add your contacts to aspects on that new seed, and ask them to add your new seed to their aspects."
       download_data_q: "Can I download a copy of all of my data contained in my seed (account)?"
-      download_data_a: "Yes. At the bottom of the Account tab of your settings page there are two buttons for downloading your data."
+      download_data_a: "Yes. At the bottom of the Account tab of your settings page there is a button for downloading your data."
```
aka001: I found only one button at end of account tab page "Request my profile data".

aka001:
I have updated the pr. Kindly look at it.

config/locales/diaspora/en.yml
jhass: Mmmh, #5685 is going to readd the second one though.

aka001: I have updated the pr.

jhass:
I don't think we live up to the promise of submitting messages with `Ctrl+enter` everywhere yet, but we should so that only adds more demand to add it :P

Thanks!

aka001:
Got it. :P

aka001:
@jhass Shouldn't we have a point in help section about "Getting started as a developer" or "Want to contribute to diaspora?" to help new developers?

jhass:
I don't think so, the help section is user oriented, not developer oriented.

As for login & password, that would first need us to think about how the user can find it while not being logged in.

aka001:
IMHO, The help section must be visible to every user(irrespective of whether user is logged-in or not).

PR #5480, opened 2014-12-17
[WIP] Port of header to bootstrap

ghost:
Problems remaining:
- [x] Hovercard are not created in in the dropdown.
- [x] The tooltip is broken, probably CSS.
- [x] The "Mark all as read" link is still clickable after having been disabled.
- [x] The badge style must be improved.

PR #5707, opened 2015-02-27
Port of notifications.js and notification-badge.js to Backbone part.1

ghost:
Needed for #5480.

This PR tries to unify [header.js](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/widgets/header.js), [notifications.js](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/widgets/notifications.js) and [notifications-badge.js](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/widgets/notifications-badge.js) with [notifications-view.js](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/notifications_view.js) and port everything to Backbone.

Almost done. Steps left to go :
- [X] have a beer,
- [X] have another beer,
- [X] code the darn thing and write some tests,
- [X] fix the [failing Cucumber test](https://travis-ci.org/diaspora/diaspora/jobs/52569805#L1366),
- [X] fix the [failing Jasmine tests](https://travis-ci.org/diaspora/diaspora/jobs/52569803#L1750),
- [X] find a proper way to deal with [inline JS](https://github.com/AugierLe42e/diaspora/blob/port-notifications-backbone/app/assets/templates/header_tpl.jst.hbs#L111),
- [X] find someone to suffer on reviewing that mountain of crap.


app/assets/templates/header_tpl.jst.hbs
```diff
@@ -107,3 +107,9 @@
   <div id="lightbox-backdrop"></div>
 
 </div>
+
+<script type="application/javascript">
+    $(document).ready(function(){
+        new app.views.Notifications({ el: '#notification_dropdown' });
+    });
+</script>
```
jhass: No new inline JS please.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -7,6 +7,7 @@ app.views.Notifications = Backbone.View.extend({
   },
 
   initialize: function() {
+    alert('This should actually work for notifications dropdown');
     Diaspora.page.header.notifications.setUpNotificationPage(this);
     $(".unread-toggle .entypo").tooltip();
     app.helpers.timeago($(document));
```
AugierLe42e: @svbergerem : I don't understand why the event  `"click .unread-toggle" : "toggleUnread"` is not binded when I had the script to the header template. Can you help me ?

app/assets/templates/header_tpl.jst.hbs
AugierLe42e: It is just for testing purposes, don't mind about this. ;)

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -56,41 +50,37 @@ app.views.Notifications = Backbone.View.extend({
...
+      var t = 'unread';
       note.removeClass("unread").addClass("read");
-      $(".unread-toggle .entypo", note)
+    }
+
+    $(".unread-toggle .entypo", note)
         .tooltip('destroy')
         .removeAttr("data-original-title")
-        .attr('title',Diaspora.I18n.t('notifications.mark_unread'))
+        .attr('title',Diaspora.I18n.t(t))
```
AugierLe42e: wut

AugierLe42e:
It starts to be big. Should I try to only solve the problems and seperate further modifications in a separate PR ?

app/assets/javascripts/widgets/notifications-badge.js
```diff
@@ -79,31 +79,25 @@
...
+      self.dropdownNotifications.scroll(function() {
+        var bottom = self.dropdownNotifications.prop('scrollHeight') - self.dropdownNotifications.height();
+        var currentPosition = self.dropdownNotifications.scrollTop();
         isLoading = ($('.loading').length === 1);
         if (currentPosition + 50 >= bottom && notificationsLoaded <= self.notifications.length && !isLoading) {
-            $('.notifications').addClass("loading");
+            self.dropdownNotifications.addClass("loading");
             ++currentPage;
             self.getMoreNotifications();
         }
```
AugierLe42e: I think this boiler plate can be replaced with [this jQuery plugin](https://github.com/jeffchao/simple-infinite-scroll)

jhass: Maybe we can use the infinite scrolling plugin we already have instead? Or at least drop the existing one and use the new one everywhere? I would like to not add two plugins for essentially the same job.

AugierLe42e: I was not aware we already had one plugin. Thanls fot the tip.

app/assets/templates/header_tpl.jst.hbs
```diff
@@ -107,3 +107,9 @@
   <div id="lightbox-backdrop"></div>
 
 </div>
+
+<script type="application/javascript">
+    $(document).ready(function(){
+        new app.views.Notifications({ el: '#notification_dropdown' });
+     });
+</script>
```
AugierLe42e: I need to find a proper way to deal with this.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -3,41 +3,36 @@
...
+    "a#mark_all_read_link": "onMarkAllRead" // This doesn't work :(
   },
 
   initialize: function() {
     Diaspora.page.header.notifications.setUpNotificationPage(this);
     $(".unread-toggle .entypo").tooltip();
     app.helpers.timeago($(document));
+    var self = this;
+    $('a#mark_all_read_link').on('click', function(ev){ self.onMarkAllRead(); });
   },
```
AugierLe42e: Need help here.

app/controllers/notifications_controller.rb
```diff
@@ -41,8 +41,8 @@ def index
 
       pager.replace(result)
     end
-    @notifications.each do |n|
-      n.note_html = render_to_string( :partial => 'notify_popup_item', :locals => { :n => n } )
+    @notifications.each do |note|
+      note.note_html = render_to_string( :partial => 'notification.html', :locals => { :note => note } )
```
jhass: Shouldn't we just respond with JSON here and render them client side while porting this to backbone?

AugierLe42e: Maybe. But this PR is definitely too big. I think the port to Handlebars should be made in another PR.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -49,47 +41,55 @@ app.views.Notifications = Backbone.View.extend({
...
+  },
+
   updateView: function(guid, type, unread) {
     var change = unread ? 1 : -1,
         all_notes = $('ul.nav > li:eq(0) .badge'),
         type_notes = $('ul.nav > li[data-type=' + type + '] .badge'),
         header_badge = $('#notification_badge .badge_count'),
-        note = $('.stream_element[data-guid=' + guid + ']');
+        note = $('.stream_element[data-guid=' + guid + ']'),
+        markAllRead= $('a#mark_all_read_link');
```
denschub: missing space.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -49,47 +41,55 @@ app.views.Notifications = Backbone.View.extend({
...
 
-    if(unread) {
-      note.removeClass("read").addClass("unread");
-      $(".unread-toggle .entypo", note)
-        .tooltip('destroy')
-        .removeAttr("data-original-title")
-        .attr('title',Diaspora.I18n.t('notifications.mark_read'))
-        .tooltip();
-    }
+    var t = 'read';
```
denschub: you can merge this declaration with the var-block above... also `t` is a very bad name.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -16,28 +16,20 @@ app.views.Notifications = Backbone.View.extend({
...
-  setRead: function(guid) {
-    $.ajax({
-      url: "/notifications/" + guid,
-      data: { set_unread: false },
-      type: "PUT",
-      context: this,
-      success: this.clickSuccess
-    });
-  },
+  getAllUnread: function(){ return $('.media.stream_element.unread'); },
```
AugierLe42e: We didn't decided anything about quotes, did we ?

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -49,47 +41,57 @@ app.views.Notifications = Backbone.View.extend({
...
-    } else {
-      all_notes.removeClass('badge-important').addClass('badge-default');
-    }
-    if(type_notes.text()>0){
-      type_notes.addClass('badge-important').removeClass('badge-default');
-    } else {
-      type_notes.removeClass('badge-important').addClass('badge-default');
-    }
-    if(header_badge.text()>0){
+    [all_notes, type_notes, header_badge].forEach(function(element){
```
AugierLe42e: Same thing about case ? I remember I've told I prefer snake_case for local variables.

app/assets/javascripts/app/views/notifications_view.js
svbergerem: Neither did we about camel case and 'Diaspora' is defined in the .jshint.json. I have no idea what @houndci is doing here.

AugierLe42e: Ok, so it ain't just me.

app/assets/stylesheets/bootstrap-headerfix.scss
```diff
@@ -41,11 +41,12 @@ header {
...
-    .notification_element {
+
+    .media.stream_element{
       font-size: 13px;
-      .timeago {
-        border: medium none;
-        cursor: text;
+      img.avatar {
+        width: 35px;
+        height: 35px;
```
svbergerem: Please keep that in header.scss where it has been before and to make sure that it doesn't change the style of hovercards it might be a good idea to use something like `> .avatar` what we also had before. While touching this file I guess you could also remove `font-size: 13px;`. We already set the default font-size to 13px.

AugierLe42e: I put it here because Boostrap's rule aren't overriden otherwise.

capture d ecran 2015-03-10 a 22 02 54

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -16,28 +16,20 @@ app.views.Notifications = Backbone.View.extend({
...
-    });
-  },
+  getAllUnread: function(){ return $('.media.stream_element.unread'); },
+
+  setRead: function(guid) { this.setReadStatus(guid, false); },
+
+  setUnread: function(guid){ this.setReadStatus(guid, true); },
 
-  setUnread: function(guid) {
+  setReadStatus: function(guid, state){
```
svbergerem: When called with `guid, true` it sets the notification as unread so I'd call this function `setUnreadStatus`.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -49,47 +41,57 @@ app.views.Notifications = Backbone.View.extend({
...
+  },
+
   updateView: function(guid, type, unread) {
     var change = unread ? 1 : -1,
         all_notes = $('ul.nav > li:eq(0) .badge'),
         type_notes = $('ul.nav > li[data-type=' + type + '] .badge'),
         header_badge = $('#notification_badge .badge_count'),
-        note = $('.stream_element[data-guid=' + guid + ']');
+        note = $('.stream_element[data-guid=' + guid + ']'),
+        markAllRead = $('a#mark_all_read_link'),
```
svbergerem: Let's call it `markAllReadLink` so one really knows what this is about.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -49,47 +41,57 @@ app.views.Notifications = Backbone.View.extend({
...
+        translationKey = 'read';
```
svbergerem: Before it was `notifications.mark_read`, now it is just `read`. I don't think that will work. Btw you could also write
`translationKey = unread ? 'notifications.mark_read' : 'notifications.mark_unread'`

app/assets/javascripts/app/views/notifications_view.js
AugierLe42e: Quite right.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -49,47 +41,56 @@ app.views.Notifications = Backbone.View.extend({
...
-    }
-    if(header_badge.text()>0){
+    [all_notes, type_notes, header_badge].forEach(function(element){
+      element.text(function(i, text){
+        return parseInt(text) + change });
+    });
+
+    var bi = 'badge-important';
+    var bd = 'badge-default';
+    var setBadgeClass = function(el, klass1, klass2){ el.addClass(klass1).removeClass(klass2); };
```
svbergerem: To improve readability you could do something like this:
```javascript
[all_notes, type_notes].forEach(function(badge) {
  if(badge.text() > 0) {
    badge.addClass('badge-important').removeClass('badge-default');
  } else {
    badge.removeClass('badge-important').addClass('badge-default');
  }
});
```
(I think bi and bd don't improve readability)

AugierLe42e: Even more sexy !

app/assets/stylesheets/header.scss
```diff
@@ -162,45 +163,43 @@ body > header {
...
           .entypo {
             cursor: pointer;
             color: lighten($black,75%);
             font-size: 17px;
             line-height: 17px;
+            padding: 10px;
           }
-          .tooltip {
-            position: fixed;
-          }
```
svbergerem: You removed this and only added it for unread notifications. We need the same for read notifications so please revert this.

app/assets/stylesheets/header.scss
```diff
@@ -162,45 +163,46 @@ body > header {
...
-        }
+          img.avatar {
+            width: 35px;
+            height: 35px;
+          }
+
+          .pull-right{
+            margin-top: 20px;
+            .aspect_membership_dropdown{ display: none; }
+          }
```
svbergerem: This looks weird for notifications that only need 2 lines. In that case the eye has too much margin to the top. I'd propose the following change:
```diff
diff --git a/app/assets/stylesheets/header.scss b/app/assets/stylesheets/header.scss
index 26a6190..d49f484 100644
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -176,17 +176,14 @@ body > header {
             height: 35px;
           }
 
-          .pull-right{
-            margin-top: 20px;
-            .aspect_membership_dropdown{ display: none; }
-          }
+          .pull-right > .aspect_membership_dropdown { display: none; }
+          .unread-toggle { margin: 10px; }
 
           .entypo {
             cursor: pointer;
             color: lighten($black,75%);
             font-size: 17px;
             line-height: 17px;
-            padding: 10px;
           }
 
           &.unread {
```

svbergerem:
In the notification dropdown the tooltips are only working after you clicked on the entypo icon.

AugierLe42e:
How is it possible ? Let me check that.

app/assets/javascripts/widgets/notifications-badge.js
```diff
@@ -76,36 +83,34 @@
...
-          if($.inArray(notification, notifications) === -1)
+          if($.inArray(notification, notifications) === -1){
             self.dropdownNotifications.append(notification.note_html);
+          }
         });
       });
 
+      $(".unread-toggle .entypo").each(function(){
+        $(this).tooltip('destroy').tooltip();
+      });
```
AugierLe42e: I'm not really sure about my code here, I suspect there's a more efficient way to achieve this in the above loop.

svbergerem:
Thank you.

Issue #3824, opened 2013-01-09
Environment sanity checks in script/server


jhass:
Git:

* [x] <del>Check if available at all before attempting to create the tar</del>
* [ ] Check if in detached head state and if so print a warning and don't let the tar code fail
* [ ] Check if in merge conflict and abort if so

Other:

* [ ] Check if RVM available and if correct gemset if so, warning, not fatal
* [ ] Check if bundle is complete
* [ ] Check if database is migrated

aka001:
@jhass The changes I make in script/server doesn't reflect when I run the command "rails server". The command "rails server" is also to the file "config/load_config.rb".
I don't know where am I going wrong. Please help me.

jhass:
To run `script/server` you run, well, `script/server`. `bin/rails server` does not run the script and never will.

aka001:
@jhass Got that. Thanks. I have created a pr https://github.com/diaspora/diaspora/pull/5708 .

mrbrdo:
should be fixed by https://github.com/diaspora/diaspora/pull/5891
checking migrations not worth it as mentioned in the commit message

PR #5708, opened 2015-02-27
Environment sanity checks in script/server #3824

aka001:


script/server
```diff
@@ -61,6 +61,22 @@ then
   fatal "config/diaspora.yml is missing! Copy over config/diaspora.yml.example to config/diaspora.yml and edit it properly!"
 fi
 
+# Check if detached head state
+git status | grep 'Changes not staged for commit' &> /dev/null
+if [ $? ]
+then
+  warning "Your changes are not staged for commit. Kindly stash or commit them"
+fi
```
jhass: You can't detect detached head status by seeing if there are unstaged changes. Detached head status happens when you do for example `git checkout v0.3.0.0`, that is checkout anything other than a branch.

Also the main point was to ensure creating the tarball below doesn't fail in that state.

script/server
```diff
@@ -61,6 +61,22 @@ then
...
+
+# Check if bundle is complete
+bundle check | grep 'Install missing gems' > /dev/null
+if [ $? ]
+then
+  warning "Your bundle is not up to date, run the command \"bundle install\""
+else
+  echo "Your bundle is up to date"
```
jhass: No need for this message.

script/server
```diff
@@ -61,6 +61,22 @@ then
...
+if [ $? ]
+then
+  warning "Your changes are not staged for commit. Kindly stash or commit them"
+fi
+
+# Check if bundle is complete
+bundle check | grep 'Install missing gems' > /dev/null
+if [ $? ]
+then
+  warning "Your bundle is not up to date, run the command \"bundle install\""
```
jhass: This should be more than a warning, it should fail.

script/server
```diff
@@ -61,6 +61,22 @@ then
...
 
+# Check if detached head state
+git status | grep 'Changes not staged for commit' &> /dev/null
+if [ $? ]
+then
+  warning "Your changes are not staged for commit. Kindly stash or commit them"
+fi
+
+# Check if bundle is complete
+bundle check | grep 'Install missing gems' > /dev/null
```
jhass: > Description:
>  Check searches the local machine for each of the gems requested in the Gemfile. If all gems are found, Bundler prints a success message and exits with a status of 0. If 
  not, the first missing gem is listed and Bundler exits status 1.

There's no need for the grep, just checking the return code is enough.

jhass:
The current script works even when git is not available or if the current directory is not a git repository, please retain those properties.

aka001:
@jhass Without 'git' how we can determine whether the repository is in detached head state?
I have fixed the bundle check issue.

jhass:
That point is really mainly for not letting the tar generation fail in that case, that is add it to the conditions that skip it.

script/server
```diff
@@ -61,6 +61,18 @@ then
...
 fi
 
+# Check if detached head state
+if git status | grep 'Changes not staged for commit' > /dev/null ;
+then
+  warning "Your changes are not staged for commit. Kindly stash or commit them"
+fi
+
+# Check if bundle is complete
+if !( bundle check ) ;
```
jhass: No need for the parens or the semicolon. you also should suppress its output IMO.

aka001:
I have updated the pr. Kindly review it.

script/server
jhass: The above comment still holds.

script/server
```diff
@@ -61,6 +61,33 @@ then
...
+fi
+
+# Check if bundle is complete
+if ! bundle check > /dev/null
+then
+  fatal "Your bundle is not up to date, run the command \"bundle install\""
+fi
+
+# Check if database is migrated and correct gemset
+if rake db:migrate:status | grep 'down' > /dev/null ;
```
jhass: In the original issue I say

> but if we can't find a way that takes less than 3 or 5 seconds its probably not worth it.

On my server:
```
$ time bin/rake db:migrate:status
...
real    0m7.318s
user    0m5.760s
sys     0m1.200s
```

script/server
```diff
@@ -61,6 +61,33 @@ then
...
+then
+  warning "Your database is not migrated, run the command \"rake db:migrate\""
+elif rake db:migrate:status | grep 'rake aborted' > /dev/null ;
+then
+  warning "The gemset is not correct"
```
jhass: This really meant to check if RVM is available that the currently activated gemset is named diaspora, not whether "sokmething went wrong while running a rake command".

script/server
```diff
@@ -61,6 +61,33 @@ then
...
+# Check if detached head state
+if git status | grep 'Changes not staged for commit' > /dev/null ;
+then
+  warning "Your changes are not staged for commit. Kindly stash or commit them"
+fi
+
+# Check if RVM available
+if rvm -v | grep 'No command '"'"'rvm'"'"' found' > /dev/null ;
+then
+  warning "RVM is not installed on the system. Please install it"
```
jhass: We recommend, not require RVM.

jhass:
Superseded by #3824 

Thank you!

Flaburgan:
#3824 is the issue, this PR is superseded by https://github.com/diaspora/diaspora/pull/5891 ;)

jhass:
Ups, thanks!

PR #5710, opened 2015-02-28
Remove .css extension from SCSS files

jhass:
sass-rails 5.0, pulled in via #5709,  deprecated those.


denschub:
Sorry in advance for maybe breaking other PRs... ;)

Issue #5631, opened 2015-02-06
People's preview broken


sedrubal:
The preview for people appeares and dissappears sometimes, when I move the cursor on the avatar of a person.

S-Ercan:
Fixed and created PR https://github.com/diaspora/diaspora/pull/5714 for this.

The issue was the following:
If, on hover, your mouse pointer happened to be just off the border of the appearing hovercard, this would register as a `mouseleave` event in the hovercard and the hovercard would then get hidden again.
As a result, the `mouseenter` event in the original element that you hovered on was triggered again.
This then kept repeating itself, leading to the flickering behaviour described by sedrubal.

I fixed this by changing the top and left coordinates of the hovercard to match those of the parent element (that was hovered on originally). This way the mouse pointer is guaranteed to be inside the hovercard, preventing the mentioned scenario from occurring.

PR #5714, opened 2015-03-01
Fix for issue #5631

ercanse:
Fixes the flickering hovercard issue reported in https://github.com/diaspora/diaspora/issues/5631.

### DIFF
--- a/app/assets/javascripts/app/views/hovercard_view.js
+++ b/app/assets/javascripts/app/views/hovercard_view.js
@@ -16,7 +16,7 @@ app.views.Hovercard = app.views.Base.extend({
       .on('mouseleave', '.hovercardable', _.bind(this._mouseleaveHandler, this));
 
     this.show_me = false;
-    this.parent = null;  // current 'hovercarable' element that caused HC to appear
+    this.parent = null;  // current 'hovercardable' element that caused HC to appear
 
     // cache some element references
     this.avatar = this.$('.avatar');
@@ -61,7 +61,10 @@ app.views.Hovercard = app.views.Base.extend({
 
   _mouseleaveHandler: function(event) {
     if( this.active === false ||
-        $.contains(this.el, event.relatedTarget) ) { return false; }
+      $.contains(this.el, event.relatedTarget) ) { return false; }
+
+    if( this.mouseIsOverElement(this.parent, event) ||
+      this.mouseIsOverElement(this.$el, event) ) { return false; }
 
     this.show_me = false;
     if( this.$el.is(':visible') ) {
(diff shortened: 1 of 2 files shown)


AugierLe42e:
Thx, this one annoyed me >.<

jhass:
Sorry, but not an improvement in all cases for me: http://cloud.aeshna.de/u/mrzyx/hovercard.webm

I think it also changes look & feel quite a bit in the other places. Isn't there a way to debounce the additional events?

S-Ercan:
Ok, I've reverted the changes regarding hovercard positioning and used a different approach now to solve the flickering issue.

I've modified the `_mouseleaveHandler` logic so that, while `mouseleave` events can still be fired from both the parent and the hovercard, they are processed only if we have left both.
If you leave the hovercard but are still in the parent, or you leave the parent but are still in the hovercard, the handler returns false.
The absence of the first restriction was what caused the flickering problem.

app/assets/javascripts/app/views/hovercard_view.js
```diff
@@ -140,6 +143,14 @@ app.views.Hovercard = app.views.Base.extend({
       top: p_pos.top + p_height - 25,
       left: p_pos.left
     });
-  }
+  },
+  
+  mouseIsOverElement: function(element, event) {
+  	var el_pos = element.offset();
```
jhass: We use two spaces for indentation I think, @svbergerem do you know why Hound isn't complaining about this?

svbergerem: Could be related to https://github.com/thoughtbot/hound/issues/418.

app/assets/javascripts/app/views/hovercard_view.js
```diff
@@ -140,6 +143,14 @@ app.views.Hovercard = app.views.Base.extend({
...
       left: p_pos.left
     });
-  }
+  },
+  
+  mouseIsOverElement: function(element, event) {
+    var el_pos = element.offset();
+    return event.pageX >= el_pos.left && 
+      event.pageX <= el_pos.left + element.width() &&
+  	  event.pageY >= el_pos.top &&
```
jhass: Please make sure to only use spaces to indent.

app/assets/javascripts/app/views/hovercard_view.js
```diff
@@ -140,6 +143,14 @@ app.views.Hovercard = app.views.Base.extend({
...
     });
-  }
+  },
+  
+  mouseIsOverElement: function(element, event) {
+    var el_pos = element.offset();
+    return event.pageX >= el_pos.left && 
+      event.pageX <= el_pos.left + element.width() &&
+  	  event.pageY >= el_pos.top && 
+  	  event.pageY <= el_pos.top + element.height();
```
jhass: This line has a hard tab too.

jhass:
Awesome this works great. Please just make sure to resolve the above comments and squash your commits into one:

```bash
git remote add upstream git://github.com/diaspora/diaspora.git
git fetch upstream
git checkout 5631-broken-people-preview
# Choose pick for the first line, squash for the other ones, save & quit
git push -f origin 5631-broken-people-preview
```

As a tip for the future: You can edit a commit with `git commit --amend`, which is, given you're in a feature branch, is fine to do even after you pushed it.

S-Ercan:
Here goes! Indentation fixed, commits squashed.
Thanks for your feedback!

jhass:
Thank you!

Flaburgan:
Thanks!

PR #5719, opened 2015-03-02
Try to make Travis faster again

jhass:
Let's see if this works :)


Flaburgan:
Why is it better than the previous one?

jhass:
Allegedly faster, and at least on my machine it is. Looks like no difference on Travis though :/

PR #5722, opened 2015-03-02
Add a notification subscribtion to the single post view

ghost:
notifsubscribtion

### DIFF
--- a/app/assets/javascripts/app/views/single-post-viewer/single_post_moderation.js
+++ b/app/assets/javascripts/app/views/single-post-viewer/single_post_moderation.js
@@ -6,12 +6,14 @@ app.views.SinglePostModeration = app.views.Feedback.extend({
   events: function() {
     return _.defaults({
       "click .remove_post": "destroyModel",
+      "click .create_participation": "createParticipation",
+      "click .destroy_participation": "destroyParticipation"
     }, app.views.Feedback.prototype.events);
   },
 
   presenter: function() {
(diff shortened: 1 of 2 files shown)


Flaburgan:
Thanks to deal with that! I agree that it's more important to have it in the SPV than in the stream, because when you get a notification and think "Damn, I don't want to know about this stuff anymore", you have a link to the post, but not to the stream. So, it's hard to find it again in your stream, and to be able to shut off (or on) the notification in the SPV is a nice improvement. I would like to see the same icon used on both places though, so if you change it here, you would have to do it in the stream and in the header too.

AugierLe42e:
This part may need a refactor, but too complex for me.

jhass:
> Is it worth to fix Milou's comments ?

Always, not fixing them requires a config change instead.

AugierLe42e:
Doing it right now so.Someone to help me with [cucumber](https://travis-ci.org/diaspora/diaspora/jobs/53472264#L2108) ?

jhass:
Looks like one of the "random" failures.

AugierLe42e:
Two times the same ? Weird.

AugierLe42e:
Ok it was. So I guess it's ready to merge.

AugierLe42e:
To be continued after #5762 is merged;

AugierLe42e:
Can be merged now.

svbergerem:
Thank you!

PR #5728, opened 2015-03-04
Fix #5716 focus header link color

### LINKED ISSUE #5716: :active link color in bootstrap header is blue
Or whatever the pseudo class is there (happens during drag for example):

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9099258-active-link-color-in-bootstrap-header-is-blue?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SansPseudoFix:
5716_link_header_focus

### DIFF
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -39,14 +39,13 @@ body > header {
     color: #CCC;
     color: rgb(147,147,147);
 
-    &:hover {
+    &:hover, &:focus {
       background: none;
       color: $highlight-white;
     }
     &:focus {
-        color: $highlight-white;
-        outline: thin dotted $border-dark-grey;
-        text-decoration: none;
+      outline: thin dotted $border-dark-grey;
+      text-decoration: none;
     }
   }
 
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -43,6 +43,11 @@ body > header {
       background: none;
       color: $highlight-white;
     }
+    &:focus {
+        color: $black;
+        outline: thin dotted;
+        text-decoration: none;
+    }
   }
 
   &.landing {
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -44,8 +44,8 @@ body > header {
       color: $highlight-white;
     }
     &:focus {
-        color: $black;
-        outline: thin dotted;
+        color: $highlight-white;
+        outline: thin dotted #CCC;
         text-decoration: none;
     }
   }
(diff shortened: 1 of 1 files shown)


jhass:
Mmh, I thought about keeping the hover cover. Why did you choose black?

SansPseudoFix:
I don't like black too. You think ````#FAFAFA```` ($highlight-white). It will be better, yeah. But if we matter about accessibility, I think we must keep dots too (but, if nobody agree, it's ok too).

jhass:
I think I wouldn't mind the dots much, maybe play around with giving them a lighter or darker grey than the link.

SansPseudoFix:
Something like that ?

5716_link_header_focus_grey

SansPseudoFix:
Well, as usuall, I read too quick. So : 
- a) grey link - black dots 
5716_link_header_focus_black-dots_grey

- b) black link - grey dots
blakc-link_grey-dots

- c) previous message

jhass:
Neither really, The whitish with the black doesn't mix well, the whiteish on the border stings out way too much for my taste. How about the normal (unfocused) link color for the border on focus?

SansPseudoFix:
I dont really see any differences, but I'm colorblind :P 
````
&:focus {
        color: $highlight-white;
        outline: thin dotted #CCC;
        text-decoration: none;
    }
````
link-grey-grey

**Comment by jhass** [comment]
How about `$border-dark-grey`?

**Comment by goobertron** [comment]
Would it be OK to keep it the same colour as when unfocused? I don't particularly see a need to have it differentiated.

**Comment by jhass** [comment]
@goobertron focus is basically the same as hover, hover is for the mouse, focus is for the keyboard.

**Comment by goobertron** [comment]
Then use the same colour as for hover?

**Comment by SansPseudoFix** [comment]
I definitly don't see difference. But, here with the same colour as for hover with ````$border-dark-grey```` border.

focus-link_grey_dark_grey

**Comment by jhass** [comment]
@goobertron that's what the PR is doing, @SansPseudoFix is in the opinion that an outline is a must have for focus too though, so we're discussing the best color for it.

File: app/assets/stylesheets/header.scss [hunk 1]
```diff
@@ -43,6 +43,11 @@ body > header {
       background: none;
       color: $highlight-white;
     }
+    &:focus {
+        color: $black;
+        outline: thin dotted;
+        text-decoration: none;
+    }
```
jhass: Why not merge the common attributes with the &:hover above?

SansPseudoFix: Hm, because it's WIP...(No no, I don't try to find excuses ;) ). I'll do that soon.

Raven24:
+1

jhass:
Thanks!

PR #5731, opened 2015-03-04
Global overflow fix

SansPseudoFix:
Before (overflow is hidden) : 
overflow-bs-stream-before

After : 
overflow-bs-stream-after

### DIFF
--- a/app/assets/stylesheets/new_styles/_base.scss
+++ b/app/assets/stylesheets/new_styles/_base.scss
@@ -14,6 +14,14 @@ body {
   }
 }
 
+/* Overflow */
+h1, h2, h3, h4, h5, h6,
+p,
+blockquote,
+code,
+pre { word-wrap: break-word; }
+a.tag { word-break: break-all; }
+
 /* new link color */
 $link-blue : rgb(42,156,235);
 a { color : $link-blue  }
--- a/app/assets/stylesheets/conversations.scss
+++ b/app/assets/stylesheets/conversations.scss
@@ -14,7 +14,6 @@
 
     p {
       margin: 0 0 1em 0;
-      word-wrap: break-word;
       &:last-child { margin-bottom: 0; }
     }
     &.new_message { border-bottom: none; }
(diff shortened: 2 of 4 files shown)


jhass:
This breaks words even if they could be wrapped properly. How about `word-wrap: break-word`?

SansPseudoFix:
Indeed :)

svbergerem:
We have the same problem at least for paragraphs and headings. In the single post view we fixed it with `p { word-wrap: break-word; }` but headings are a problem in the SPV, too. I think we should have a more general solution by adding `word-wrap: break-word` to `new_styles/_base.scss` for `p`, `blockquote`, `code`, `h1`, `h2`, … and maybe also some other tags and removing the line afterwards from the different places where we added it as a partial fix.

SansPseudoFix:
I agree. It's just a quick fix. But I will try to refactor every overflow.

SansPseudoFix:
Still WIP. I try to fix tag button, but it doesn't cooperate.
issue_overflow_button_tag

Please, tell me if my code is not clean. ````{ word-wrap: break-word;}```` does not work for a.tag, so I use ````{ word-break: break-all; }````.

Stay tuned!

app/assets/stylesheets/new_styles/_base.scss
```diff
@@ -13,6 +13,16 @@ body {
...
 
+/* Overflow */
+h1, h2, h3, h4, h5, h6,
+p,
+blockquote,
+code,
+pre,
+input,
+.button, .btn{ word-wrap: break-word;}
+a.tag, { word-break: break-all; }
```
SansPseudoFix: Ok... Sneaky comma...

svbergerem:
I'm not sure about `input, .button, .btn`. I don't think we should have line breaks inside of buttons. (And it doesn't sound like a common problem)

Everything else looks reasonable.

SansPseudoFix:
Yes, I screwed up. I push to quickly. Sorry about that, I saw it too late. I was talking about grouping code, but it seems ok, so, cool :)

svbergerem:
Thank you! :)

PR #5733, opened 2015-03-06
Avatar aspect update and DRY

margori:
Adopting #4887 , related to #4288 

I started with @goobertron work and added DRY concept, Don't Repeat Yourself, moving lines to '_base.scss' file.

Have doubt in 'conversation.scss', 'stream_element.scss', 'getting_started.scss', etc. It seems that several sizes are required, but placing shadow and radius corner in 'base' is an improve.

Screenshots coming soon. Feedback welcome.

### DIFF
--- a/app/assets/stylesheets/new_styles/_base.scss
+++ b/app/assets/stylesheets/new_styles/_base.scss
@@ -34,11 +34,19 @@ a { color : $link-blue  }
   &.micro {
     height: 20px;
     width: 20px;
+    box-shadow: 0px 1px 2px 0px #666;
+    -webkit-box-shadow: 0px 1px 2px 0px #666;
+    -moz-box-shadow: 0px 1px 2px 0px #666;
+    border-radius: 3px;
   }
 
   &.small {
     height: 35px;
     width: 35px;
+    box-shadow: 0px 1px 2px 0px #666;
+    -webkit-box-shadow: 0px 1px 2px 0px #666;
+    -moz-box-shadow: 0px 1px 2px 0px #666;
+    border-radius: 4px;
   }
 }
 
--- a/app/assets/stylesheets/new_styles/_base.scss
+++ b/app/assets/stylesheets/new_styles/_base.scss
@@ -31,22 +31,22 @@ a { color : $link-blue  }
 .icon-blue  { background-image: image_url("img/glyphicons-halflings-blue.png");  }
 
 .avatar {
+  border-radius: 4px;
+
   &.micro {
     height: 20px;
     width: 20px;
-    box-shadow: 0px 1px 2px 0px #666;
-    -webkit-box-shadow: 0px 1px 2px 0px #666;
-    -moz-box-shadow: 0px 1px 2px 0px #666;
-    border-radius: 3px;
   }
 
   &.small {
     height: 35px;
     width: 35px;
-    box-shadow: 0px 1px 2px 0px #666;
-    -webkit-box-shadow: 0px 1px 2px 0px #666;
-    -moz-box-shadow: 0px 1px 2px 0px #666;
-    border-radius: 4px;
+  }
+
+  &.medium {
+    height: auto;
+    width: auto;
+    max-width: 75px;
   }
 }
 
--- a/app/assets/stylesheets/notifications.scss
+++ b/app/assets/stylesheets/notifications.scss
@@ -78,6 +78,10 @@
       .avatar {
         width: 35px;
         height: 35px;
+        box-shadow: 0px 1px 2px 0px #666;
+        -webkit-box-shadow: 0px 1px 2px 0px #666;
+        -moz-box-shadow: 0px 1px 2px 0px #666;
+        border-radius: 3px;
       }
 
       .unread-toggle {
--- a/app/assets/stylesheets/notifications.scss
+++ b/app/assets/stylesheets/notifications.scss
@@ -78,10 +78,6 @@
       .avatar {
         width: 35px;
         height: 35px;
-        box-shadow: 0px 1px 2px 0px #666;
-        -webkit-box-shadow: 0px 1px 2px 0px #666;
-        -moz-box-shadow: 0px 1px 2px 0px #666;
-        border-radius: 3px;
       }
 
       .unread-toggle {
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -339,7 +339,6 @@ body > header {
     left: 8px;
     top: 8px;
     display: block;
-    border-radius: 3px;
   }
 
   .header-nav {
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -339,6 +339,7 @@ body > header {
     left: 8px;
     top: 8px;
     display: block;
+    border-radius: 3px;
   }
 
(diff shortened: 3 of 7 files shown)


margori:
screenshot - 050315 - 22 10 07
screenshot - 050315 - 22 11 01
screenshot - 050315 - 22 11 11
screenshot - 050315 - 22 11 33
screenshot - 050315 - 22 12 09
screenshot - 050315 - 22 12 40

SansPseudoFix:
Nice!

goobertron:
Thanks for picking this up from me. I *think* that at the time the majority feeling was that the shadow should be *inset* rather than external, although I didn't realise this so implemented an external one. Let's see what people say to your screenshots.

svbergerem:
For the small avatars (reshare / conversation participants) the shadow looks too intense to me.

Raven24:
I agree with @svbergerem about the tiny images, else I find the very discreet and unobstrusive.
Not sure if the border radiuses should be different for the big profile image.

Could you maybe post a screenshot with all size versions lined up next to each other?

Flaburgan:
Yeah the border radius is too big imo. What about 2%?

margori:
> For the small avatars (reshare / conversation participants) the shadow looks too intense to me.

I agree.

> Yeah the border radius is too big imo. What about 2%?

It makes "eliptical" corner for rectangular images. I don't see it as a good solution.

---------------------

I aim to a middle point solution: no shadow, circular round corner (4px, same as buttons), aspect values set in '_base.scss' file.

margori:
screenshot - 080315 - 23 56 53
screenshot - 080315 - 23 57 16
screenshot - 080315 - 23 57 24
screenshot - 080316

svbergerem:
Nice! :) Does anyone have any further remarks?

marienfressinaud:
I'm just a bit worried for avatars on a white background such as the Framasoft one (framasoft@framasphere.org). How does the `box-shadow` render?

svbergerem:
@marienfressinaud

marienfressinaud:
Ok thanks! I prefer the old version for white-background avatars but the new one is definitely better for other avatars :). A solution could be having an option to choose if we want the box-shadow in the profile administration but I don't think it's interesting to add such an option.

jhass:
If I'm really honest, I never saw the need for this and still don't see it as an improvements, not in either variant that was discussed in this or the two previous PRs. But that's just my personal style preference.

svbergerem:
I like the introduced border-radius. As @marienfressinaud mentioned the shadow looks bad on avatars with a white background so I think I'd prefer to remove the shadow from all avatars.

app/assets/templates/reshare_tpl.jst.hbs
```diff
@@ -6,7 +6,7 @@
 
       {{#with root}}
         <a href="/people/{{author.guid}}" class="img {{{hovercardable this}}}">
-          <img src="{{author.avatar.small}}" class="avatar" />
+          <img src="{{author.avatar.small}}" class="avatar micro" />
```
svbergerem: Why did you add `micro` here? We set the height and width of those avatars to `30px` [here](https://github.com/diaspora/diaspora/blob/develop/app/assets/stylesheets/stream_element.scss#L129) so adding `micro` won't have any effect.

margori:
No shadow at all.
screenshot - 100315 - 12 55 54
screenshot - 100315 - 12 56 13

----------------

I found rounded corner less aggressive, smoother or more handleable. My hand won't be wounded 'cause there's no spikes.

app/helpers/people_helper.rb
```diff
@@ -32,7 +32,7 @@ def person_link(person, opts={})
   end
 
   def person_image_tag(person, size = :thumb_small)
-    image_tag(person.profile.image_url(size), :alt => person.name, :class => 'avatar', :title => person.name, 'data-person_id' => person.id)
+    image_tag(person.profile.image_url(size), :alt => person.name, :class => 'avatar small', :title => person.name, 'data-person_id' => person.id)
```
svbergerem: We use `person_image_link` with different sizes so always adding `small` doesn't sound sensible.

margori:
I'm seeing no improve to use DRY concept in CSS :( or I'm doing it poorly... Defining avatar size in code doesn't let designer to work.

But still I want to merge round corners.

SansPseudoFix:
What about a tiny grey border (only visible for white background avatar)?

avatar_border_grey_ddd

````
border: 1px solid $border-grey;
````

File: app/assets/stylesheets/new_styles/_base.scss [hunk 3]
```diff
@@ -40,6 +42,15 @@ a { color : $link-blue  }
...
   }
+
+  &.medium {
+    height: auto;
+    width: auto;
+    max-width: 75px;
+  }
+
+  &.large {
+  }
```
svbergerem: I guess you can remove this. ;)

app/assets/stylesheets/notifications.scss
```diff
@@ -75,11 +75,6 @@
         .unread-toggle { opacity: 1 !important; }
       }
 
-      .avatar {
-        width: 35px;
-        height: 35px;
-      }
```
svbergerem: we still need those lines for the notifications page.

svbergerem:
@jhass 
> I never saw the need for this and still don't see it as an improvement, not in either variant that was discussed in this or the two previous PRs.

Do you think the avatars in this PR (border-radius: 4px, no shadow) are worse than the avatars on the develop branch? (no border radius, no shadow)

In that case we would have to decide that on loomio, otherwise I would merge this PR as soon as the last issues are fixed.

jhass:
I guess I don't mind either way, didn't mean that to block anyway, just voiced my opinion.

margori:
> @SansPseudoFix  What about a tiny grey border (only visible for white background avatar)?

I see same problem as shadow: controversial decision.

------------

Re-based. Feedback welcome. Your eyes can see what I cannot 'cause I'm to close to the code.

app/assets/stylesheets/new_styles/_base.scss
svbergerem: Oh, I guess that was a misunderstanding. I was just asking to remove the `.large` part but not the part about `.medium` which you added here because you removed it in single-post-view.scss.

SansPseudoFix:
@margori border is more discrete:

* white backgroud, you have border to distinguish
* dark: you don't see border cause really light

svbergerem:
@SansPseudoFix I don't like the border. 

It looks like we can all more or less agree on the border-radius so let's merge that with this PR and discuss the border/shadow thing afterwards.

SansPseudoFix:
It was just an alternative to ````box-shadow````, but I'm ok without :P

app/assets/stylesheets/new_styles/_base.scss
margori: Yes, I misunderstood. I think it's a visual design problem of Github. Above the comment saying "this" are several lines, so I don't know what are you referring to.

Work around is to avoid saying "this" and say "line 52 and 53" or "two lines above this comment".

Root solution is to "paint" lines and comment. Then it refers to those painted lines and only those shown.

svbergerem:
Thank you! :)

margori:
Thank you too! :D

PR #5734, opened 2015-03-06
Remove unused (vendor) css

svbergerem:
I wasn't able to find any place where we use `new-btn` or one of the deleted files.


jhass:
<3

Flaburgan:
Damn, I like red even more than I like green. Thank you!

PR #5735, opened 2015-03-06
Improve Bootstrap stream style

svbergerem:
- remove  unused `new_styles/_variables.scss`
- add variables for font-size and line-height
- decrease font-size (We had 13px in blueprint views. For the Bootstrap views it was 14px. Now I set it to 13px again.)
- modify margin-bottom for stream-faces
- fix avatar alignment


svbergerem:
This also changes the style of the comments in the SPV: (which now look like comments in the stream)

before:


after:

jhass:
I knew something was odd, thanks!

PR #5738, opened 2015-03-07
add middleware to redirect undigested assets

### LINKED ISSUE #5280: Old naming of assets (without hash code) used in public/default.html
Some static assets requested in `public/default.html` do not exist, such as `assets/icons/{cog,pen_write,smiley_laughing}.png`. These files seem to be using the "dash-hash" naming scheme now.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5090323-old-naming-of-assets-without-hash-code-used-in-public-default-html?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Raven24:
pretty simple, just redirect the request to the digested version determined by the asset manifest.
don't know if this can be tested easily, since it's really production-specific.

should fix #5280

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -71,4 +71,6 @@ Diaspora::Application.configure do
   if AppConfig.environment.assets.host.present?
     config.action_controller.asset_host = AppConfig.environment.assets.host.get
   end
+
+  config.middleware.use UndigestedAssets
 end
--- a/lib/undigested_assets.rb
+++ b/lib/undigested_assets.rb
@@ -1,0 +1,36 @@
+
+class UndigestedAssets
+  def initialize(app)
+    @app = app
+    @asset_prefix = Rails.configuration.assets.prefix
+
+    load_manifest
+  end
+
+  def call(env)
+    if m = (env['PATH_INFO'] || '').match(/#{@asset_prefix}\/(.+)/i)
+      file = m[1]
+
+      if manifest_available? && undigested_file?(file)
+        return [307, { "Location" => File.join('', @asset_prefix, @manifest[file])}, ["See other"]]
+      end
+    end
+
+    @app.call(env)
+  end
+
+  private
+
+  def load_manifest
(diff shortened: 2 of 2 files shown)


jhass:
To be honest I'm not sure if this the correct way though and I'd like to limit it to things like default.png and at most the ones used by default.html, though even there I wonder if we can't just make that page a normal view instead and use asset helpers.

Rails 4 didn't just introduce such a major change out of fun, they have proper reasoning for it and that has to do with properly invalidating browser caches in not so optimal setups, which diaspora with it's big diversity in deployment is a prime example for.

Raven24:
you're right, 
also I totally based this on the assumption that the rails app handles requests to the `/assets` path, which we delegate to the reverse proxy in all our config examples, though.
So, this is a nice little rack middleware, but does nothing to actually mitigate the situation ... *facepalm*

jhass:
Well, our example configs operate on the mode of "if not found proxy to app", which means it would work for the regular setup. And I think this code could be useful to redirect default.png, since I didn't find a solution to properly adjust [this problematic line](https://github.com/diaspora/diaspora/blob/develop/app/models/profile.rb#L80) yet, or moreover the broken default pictures when somebody wipes their assets.

Raven24:
huh ... I guess I didn't base my config off of one of the gists we have linked. :confused: 
I have stuff like

```  
location ~ ^/(images|javascripts|assets|stylesheets|uploads)/  {
    root /srv/diaspora/public;
    expires 30d;
    add_header Cache-Control public;
    access_log off;
  }
```

weird ...

anyway, I also think this shouldn't just wildcard-redirect every asset we have, so maybe we can define a set of images and just whitelist them for redirection or something.

jhass:
Probably the old evil if nginx config.

PR #5743, opened 2015-03-08
Replace Make Something text by the diaspora ball logo on registration page

Flaburgan:
Waiting for https://www.loomio.org/d/ijXVz764/diaspora-catchphrase to be chosen, I propose to already remove the "Hey make something" text to not be included in the next release. At the moment, I replaced it by the diaspora ball logo:

screen

What do you think?


jhass:
Mmmh, I think I like it, more than any catchphrase actually. What about pushing the form down so that it's vertical center is on the same line than the vertical center of the ball?

AugierLe42e:
Looks better to me. Though it removes the possibility to have a background. And I find the white background a bit dpressing.

Flaburgan:
@AugierLe42e it's definitely not a definitive version to me, I would love to see something like https://diasporasf.org/ or https://diaspote.org/ with a dandelion picture as the default register page, that's why I started https://github.com/diaspora/diaspora/pull/5389. So this is a "quickfix" for the next release, and we will have time to think about it again when 0.5 will be released.

@jhass agree, I'll see what I can do.

Flaburgan:
@jhass Done ;)

jhass:
Screenshot? :)

Flaburgan:
screen

AugierLe42e:
@Flaburgan : Ok then. Nice to read that ;)

svbergerem:
@Flaburgan How does that look on a screen with a width of 1024px?

Flaburgan:
capture d ecran 2015-03-08 a 22 34 57
I guess I could reduce the left margin

svbergerem:
AFAIK we also have a svg version somewhere. We could use that with max-width: ...

goobertron:
I like the approach. Perhaps you could use a percentage for the left margin and the width of the image, to cope with the smaller screen(s)?

Link to download SVG: http://naesk.deviantart.com/art/Diaspora-Artwork-Toolkit-254178705

Flaburgan:
So what do we want to do? I don't think that there are still living beings on earth who use 1024x768 screen and everything is fine with 1280.

AugierLe42e:
> I don't think that there are still living beings on earth who use 1024x768 screen and everything is fine with 1280.

You could be amazed...

goobertron:
> I don't think that there are still living beings on earth who use 1024x768 screen

Possibly not in the affluent West, but in the whole world?

svbergerem:
I uploaded a SVG of the ball logo in our [wiki](https://wiki.diasporafoundation.org/File:Seed-ball-logo.svg).

Flaburgan:
Thanks, I should be able to finish that tomorrow.

Flaburgan:
Now using the svg image but the vertical align of the form doesn't work anymore. Unfortunately, time to sleep.

Flaburgan:
I don't success to correctly vertical align the form now that the image width is %. Any tips @svbergerem @SansPseudoFix ?

denschub:
You won't be able to do a vertical centering without further hacks, since no of the images parent elements is as high as the browser. you would end up adding `min-height: 100%;` to all of the parent elements, which is (imho) not worth the effort.

AugierLe42e:
I suggest you put the image as a background of a div placed directly into the `<body>`. The way you can center an image as background is far more flexible.

Flaburgan:
Okay I succeeded to make the background responsive using background-size: contain. I had to directly link the image and not use stripe though. Once travis is green, this should be ready to be merged.

Flaburgan:
I kept the svg image, I think it's nice to have it in the repository.

jhass:
> I kept the svg image, I think it's nice to have it in the repository.

Agree, move it to /graphics/ though

svbergerem:
Thank you.

@jhass and I found out that `url("branding/ball.png")` has to be `image-url("branding/ball.png")` instead. See https://github.com/rails/sass-rails#asset-helpers. Otherwise this will only work on dev pods. I changed that in a separate commit.

PR #5751, opened 2015-03-10
Fix photo count in the profile view

### LINKED ISSUE #5402: Number of photos on the profile page is always <=15
The photos controller uses `current_user.photos_from(..)` [here](https://github.com/diaspora/diaspora/blob/develop/app/controllers/photos_controller.rb#L28) which is defined in [/app/models/user/querying.rb](https://github.com/diaspora/diaspora/blob/develop/app/models/user/querying.rb#L152) and uses `prep_opts` which is defined [here](https://github.com/diaspora/diaspora/blob/develop/app/models/user/querying.rb#L160) and sets the limit of returned photos to 15.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6159371-number-of-photos-on-the-profile-page-is-always-15?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #5402.

### DIFF
--- a/app/controllers/people_controller.rb
+++ b/app/controllers/people_controller.rb
@@ -86,7 +86,7 @@ class PeopleController < ApplicationController
         end
         gon.preloads[:person] = @person_json
         gon.preloads[:photos] = {
-          count: photos_from(@person).count(:all),
+          count: photos_from(@person, :all).count(:all)
         }
         gon.preloads[:contacts] = {
           count: Contact.contact_contacts_for(current_user, @person).count(:all),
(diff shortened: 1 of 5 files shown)


svbergerem:
Should be ready to merge.

jhass:
Thanks!

PR #5752, opened 2015-03-10
fix tag button overflow

SansPseudoFix:
Before: 
before

After:
after

"Normal behavior" (small tag):
small_tag

### DIFF
--- a/app/assets/stylesheets/tag.scss
+++ b/app/assets/stylesheets/tag.scss
@@ -35,4 +35,14 @@ h1.tag {
       }
     }
   }
+  .span7 {
+    .tag-following-action {
+      max-width: 100%;
+      input[type="submit"] {
+        overflow-x: hidden;
+        text-overflow: ellipsis;
+        max-width: 100%;
+      }    
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


Flaburgan:
imo a max-width of 150px looks already way big enough.

svbergerem:
@Flaburgan that would even cut off tags like #diasporafoundation. The current solution looks much better to me.

I'd just like to keep the margin/padding inside of the button even if the text is too long. @SansPseudoFix It would be great if you could fix that.

app/assets/stylesheets/tag.scss
```diff
@@ -35,4 +35,15 @@ h1.tag {
       }
     }
   }
+  .span7 {
+    .pull-right {
+      &:first-child { max-width: 100%; }  
+      input[type="submit"] {
+        display: block;
```
svbergerem: it also works without `display: block`

app/assets/stylesheets/tag.scss
```diff
@@ -35,4 +35,15 @@ h1.tag {
       }
     }
   }
+  .span7 {
+    .pull-right {
+      &:first-child { max-width: 100%; }
```
svbergerem: Instead of this `first-child` "hack" I'd prefer adding another class to [this line](https://github.com/diaspora/diaspora/blob/develop/app/assets/templates/tag_following_action_tpl.jst.hbs#L1) (something like `tag-following-action`?) and instead of applying the style to `.pull-right:first-child` you could apply everything to `.tag-following-action`.

SansPseudoFix:
@svbergerem Done. But still not this ````padding````...

svbergerem:
Thank you! I also had problems setting the `padding` for that `input` and this PR already improves the current design.

PR #5755, opened 2015-03-11
Refactor vendor javascript

svbergerem:

PR #5758, opened 2015-03-11
Fix missing assets bug

svbergerem:
#5755 broke some cukes. Let's see if this PR can fix that.

### DIFF
--- a/app/views/profiles/_edit_public.haml
+++ b/app/views/profiles/_edit_public.haml
@@ -1,10 +1,4 @@
--#   Copyright (c) 2010-2011, Diaspora Inc.  This file is
--#   licensed under the Affero General Public License version 3 or later.  See
--#   the COPYRIGHT file.
-
 - content_for :head do
-  = javascript_include_tag :profile
-
   :javascript
     $(document).ready(function () {
       var data = $.parseJSON( '#{@tags_array.to_json.gsub("'", "\\\\'")}' ),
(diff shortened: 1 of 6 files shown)


app/views/profiles/_edit_public.haml
```diff
@@ -1,10 +1,4 @@
--#   Copyright (c) 2010-2011, Diaspora Inc.  This file is
--#   licensed under the Affero General Public License version 3 or later.  See
--#   the COPYRIGHT file.
-
 - content_for :head do
-  = javascript_include_tag :profile
```
svbergerem: `jquery.autoSuggest.custom` was before in that file which we already load in `main.js` and `mobile/mobile.js`.

app/views/status_messages/bookmarklet.html.haml
```diff
@@ -1,9 +1,3 @@
--#   Copyright (c) 2010-2012, Diaspora Inc.  This file is
--#   licensed under the Affero General Public License version 3 or later.  See
--#   the COPYRIGHT file.
-
-= javascript_include_tag :home
```
svbergerem: `fileuploader-custom` was before in that file which we already load in `main.js`.

jhass:
Sorry for merging too early, and thanks again!

svbergerem:
@jhass No problem. Travis is behaving really weird today.

PR #5762, opened 2015-03-12
Use entypo for control icons

svbergerem:
We already use them in the SPV and this PR unifies the icons.
entypo icons
These icons are:
- cross: for hiding posts, photos or conversations
- trash: for deleting posts, photos, comments or conversations (if you are the last participant)
- bell: notifications / participation
- block: ignore user
- megaphone: report post, photo or comment

Here is a post with the new post control icons:
post with icons

Bootstrap already has a `.controls` class so I called this one `.control-icons`.

I removed `_landing.scss` because I wasn't able to find `landing` anywhere else in our codebase. (And I found it because of the `.controls` in line 24)


Flaburgan:
Not sure about the megaphone, the exclamation mark is quite good imo.

svbergerem:
I don't like the current situation with the exclamation mark. It is too thin and because of that hard to recognize. Another possible icon would be `warning`:
warning

AugierLe42e:
I like the warning icon.

SansPseudoFix:
I think exclamation triangle is more explicit.
But not sure about blocking icon. What about silhouette with cross?

AugierLe42e:
@SansPseudoFix : The icon doesn't exist crossed.

svbergerem:
@SansPseudoFix We already use the block icon on the profile page.

Flaburgan:
The triangle is way better :)

SansPseudoFix:
@svbergerem I know, but imo, a user-crossed icon is more explicit (but it's not the point here :) ).

svbergerem:
This PR now uses the `warning` icon.

svbergerem:
Green Travis. :)

goobertron:
Exclamation mark in triangle works for me too. Thanks, Steffen.

AugierLe42e:
@svbergerem : Don't you think that [this block](https://github.com/svbergerem/diaspora/blob/entypo-controls/app/assets/templates/stream-element_tpl.jst.hbs) should just be a partial render of [this template](https://github.com/svbergerem/diaspora/blob/entypo-controls/app/assets/templates/single-post-viewer/single-post-moderation_tpl.jst.hbs) ?

svbergerem:
@AugierLe42e Not without a proper refactor which won't be part of this PR.

AugierLe42e:
Ok, I'm willing to do it if you can guide me.

jhass:
Well, let's try this I guess.

Thanks!

PR #5763, opened 2015-03-12
Use handlebars helpers wherever possible

svbergerem:
and remove the unused stream frame template


AugierLe42e:
You should take a rest ! :p

jhass:
Thanks!

PR #5767, opened 2015-03-12
Change overflow-x to overflow for long tag button

SansPseudoFix:
My bad  (on [this PR](https://github.com/diaspora/diaspora/pull/5752)).

Before (horizontal scroll):
fix-before

After:
fix-after

### DIFF
--- a/app/assets/stylesheets/tag.scss
+++ b/app/assets/stylesheets/tag.scss
@@ -39,7 +39,7 @@ h1.tag {
     .tag-following-action {
       max-width: 100%;
       input[type="submit"] {
-        overflow-x: hidden;
+        overflow: hidden;
         text-overflow: ellipsis;
         max-width: 100%;
       }    


svbergerem:
I am not able to reproduce the horizontal scroll. `overflow-x: hidden` should prevent the overflow and I don't see why `overflow: hidden` should be better in this situation. It would be great if you could give some more information. Which browser are you using? Do you have an explanation for the horizontal scroll bar?

SansPseudoFix:
@svbergerem No explanation at all. I don't understand too. After diaspora-fr update, I tried this tag ````#ceciestuntagsuperlongpourfaireplaisiraspfparcequequandmemeilenachiedoncillabienmeritedepouvoirvoirleresultatdesontravailetducoupjemedemandeaussicommentcavarendreenvrai````, and horizontal scroll bars appeared. Tested with iceweasel / Debian Jessie.

svbergerem:
@SansPseudoFix Oh, you are right. The tag I tried was too long for the button but not long enough to overflow the page. Thank you for the quick fix, I'll merge this.

SansPseudoFix:
Same for me during my tests. If you understand what happen, tell me :P

svbergerem:
[MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x) says about `overflow-x`
> This is an experimental technology

@denschub asked someone on the IRC (someone from mozilla I think) who said that `overflow-x` doesn't work properly on all kinds of elements.

SansPseudoFix:
Ok! Thx!

PR #5769, opened 2015-03-12
Remove unused css and reduce blockquote font-size

svbergerem:


app/assets/stylesheets/admin.scss
```diff
@@ -29,7 +29,7 @@ body > div.container {
 
 .users {
   li.user {
-    border-bottom: 1px solid $light-grey;
+    border-bottom: 1px solid $text-grey;
```
svbergerem: `new_styles/new_mixins` overwrote `$light-grey` on the desktop version with the same value `$text-grey` has. Because that was the only important thing `new_styles/new_mixins` did I removed the file and changed this variable.

jhass: Using a variable named text-gray for a border seems odd though, maybe just define `$text-gray` to `$light-gray` and `$light-gray` to the actual color in `colors.scss`?

svbergerem: Oh, this is actually the only place where I should have kept `$light-grey` because we import `colors.scss` at the beginning of the file. Will change this back to `$light-grey`. I didn't change the value of `$light-grey` because we also use it in the mobile version.

app/assets/stylesheets/application.scss
```diff
@@ -5,16 +5,23 @@
 @import "colors";
 @import 'sizes';
 @import 'mixins';
-@import 'new_styles/new_mixins';
 
 /* core */
+@import 'media-box';
+@import 'autocomplete';
```
svbergerem: @salbertson SCSS linting is [turned off](https://github.com/diaspora/diaspora/blob/develop/.hound.yml#L8) but as you can see Hound commented on SCSS code after a forced push.

svbergerem:
Should be ready to merge.

jhass:
Thanks!

PR #5775, opened 2015-03-14
[WIP] Comment submission canceled after timeout

### LINKED ISSUE #5566: [Mobile] Sending PM, posts or comments when no network should fail
Currently, when sending a long message with unstable network, the message stays blocked infinitly. There is no way to copy it or edit it again and it is never transmitted. This requires to reload the page and write the comment again.

On mobile version, there should be an uptime for each transmission. If the message cannot be send until then, the UI should display a warning and come back to the edition of the message.

ghost:
Seriously, why do I keep hurting myself this way... >.<
(Tries to) fix #5566.

### DIFF
--- a/app/assets/javascripts/mobile/mobile.js
+++ b/app/assets/javascripts/mobile/mobile.js
@@ -234,35 +234,54 @@ $(document).ready(function(){
 
   $(document).on("submit", ".new_comment", function(evt){
     evt.preventDefault();
-    var form = $(this);
+    var form = $(this),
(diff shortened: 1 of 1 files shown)


jhass:
I don't follow... What does this fix?

AugierLe42e:
It tries to fix #5566.

jhass:
ah, 1 is a bit short though, no?

AugierLe42e:
It's for test purposes but it fails. Need help (like always, it seems).

PR #5781, opened 2015-03-15
Various style fixes

svbergerem:
- Supersedes #5332 
- Fixes the close button for popovers
- Removes `bootstrap_header_padding` which no longer had any effect
- Removes the second `#hide_location` button from the publisher. We only had it for blueprint compatibility and didn't display it in the bootstrap publisher.
- Improves style of getting started message
### old:

getting_started_old
### new:

getting_started_new


jhass:
Thank you!

PR #5786, opened 2015-03-16
Refactor and improve sidebar css

svbergerem:
## before - after

leftnavbar
rightbar

### DIFF
--- a/app/assets/stylesheets/getting-started.scss
+++ b/app/assets/stylesheets/getting-started.scss
@@ -38,6 +38,10 @@
       }
     }
   }
+
+  .as-results .as-list {
+    box-shadow: 0px 1px 1px #666;
+  }
 }
 
 .popover h3 {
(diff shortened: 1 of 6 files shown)


jhass:
Any chance to have some padding to the placeholder in the followed tags input?

svbergerem:
@jhass 
padding tags input

jhass:
Better, but if I'm honest that it's not vertically centered still bugs me.

svbergerem:
tags input

Flaburgan:
Nice! Why did you remove the gray background on the invitation link? This shows that the field can't be modify. What do you think?

svbergerem:
The background was white in the blueprint version and for me the gray background was too distracting. For me input fields with gray backgrounds are part of forms and inform you about a value that will be sent. In this case it is an input field you should interact with by selecting the invitation link.

Maybe an input field isn't even the right element for the invitation link but that shouldn't be part of this PR.

jhass:
Thank you!

Flaburgan:
Hey @svbergerem as you changed the invitation link field design, you should change it in the contact page and the search page too to be consistent.

svbergerem:
@Flaburgan done: #5799.

PR #5791, opened 2015-03-16
Fix for 5745 - conversation pagination design is broken

### LINKED ISSUE #5745: Conversation pagination design is broken
screen

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9333521-conversation-pagination-design-is-broken?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SansPseudoFix:
Fix #5745 

Before:
5745-10-before
(oh, sorry for "En attente de...")

After:

Default value for `:inner_window` was `4`. I set `1`. Now it display **one** item around current page.
Default value for `:outer_window` was `1`, it displayed one item before the last one.

Page 1:

Page 3:

### DIFF
--- a/app/assets/stylesheets/conversations.scss
+++ b/app/assets/stylesheets/conversations.scss
@@ -180,6 +180,21 @@
     a:hover {
       text-decoration: none;
     }
+    .pagination {
+      margin-left: auto;
+      margin-right: auto;
+      text-align: center;
+      
+      .disabled a {
+        background: $background-grey;
+      }
+    }
+  }
+}
+
+@media (max-width: 1354px) {
+  #left_pane #conversation_inbox .pagination ul > li > a {
+    padding: 4px 7px;
   }
 }
 
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -847,6 +847,22 @@ form#new_user.new_user input.btn {
   text-shadow: 1px 1px 20px rgb(126, 240, 77);
 }
 
+#conversation_inbox {
+  .pagination {
+    margin-left: auto;
+    margin-right: auto;
+    text-align: center;
+    
+    ul > li > a {
+      padding: 4px 8px;
+    }
+    
+    .disabled a{
+      background: $background-grey;
+    }
+  }
+}
+
 input#user_password, #user_username, #user_password_confirmation, #user_email {
   height: 30px;
 }
--- a/app/views/conversations/index.haml
+++ b/app/views/conversations/index.haml
@@ -21,7 +21,7 @@
             - else
               #no_conversations
                 = t('.no_messages')
-            = will_paginate @conversations, :renderer => WillPaginate::ActionView::BootstrapLinkRenderer
+            = will_paginate @conversations, :previous_label => '&laquo;', :next_label => '&raquo;', :inner_window => 1, :renderer => WillPaginate::ActionView::BootstrapLinkRenderer
 
     .span8
       - if @conversation
--- a/app/views/conversations/index.mobile.haml
+++ b/app/views/conversations/index.mobile.haml
@@ -27,4 +27,4 @@
         %i
           = t('.no_messages')
 
-  = will_paginate @conversations, :renderer => WillPaginate::ActionView::BootstrapLinkRenderer
+  = will_paginate @conversations, :previous_label => '&laquo;', :next_label => '&raquo;', :inner_window => 1, :outer_window => 0, :renderer => WillPaginate::ActionView::BootstrapLinkRenderer


jhass:
Mh, I'm not sure if that's the proper fix, since it depends on the locale. See Romanian:



I think we have no choice other than to remove the labels and use a symbol for the previous and next buttons, in order to get a predictable width. And then check for three digit (100+) page numbers.

goobertron:
I think &laquo; and &raquo; would be enough on their own to denote 'previous' and 'next'.

app/assets/stylesheets/conversations.scss
```diff
@@ -180,6 +180,11 @@
     a:hover {
       text-decoration: none;
     }
+    .pagination {
```
SansPseudoFix: Argh! I will do.

SansPseudoFix:
Ok, now:


And:
5745-3-bis

Flaburgan:
Nice :)

jhass:
Let's not have to go back to this, please test with three digit numbers.

SansPseudoFix:
@jhass I was affraid you ask me that. You wanna kill my server, admit!

SansPseudoFix:
My server is still alive :D 


---

---


Let's go for four digit numbers! (I'm kidding)

app/assets/stylesheets/conversations.scss
svbergerem: @SansPseudoFix We have no SCSS styleguide yet. @houndci seems to ignore our hound config sometimes. I think that might be after a forced push. You don't have to change this if you don't want to.

svbergerem:
@SansPseudoFix You could just change the numbers wit the Firefox dev tools (or chromium or whatever) instead of creating more than a thousand conversations on your server. :-P

Since the width of the conversation list depends on the width of the window: for which sizes are we trying to fix this? What is the width of the window in the screenshots?

SansPseudoFix:
> @SansPseudoFix You could just change the numbers wit the Firefox dev tools (or chromium or whatever) instead of creating more than a thousand conversations on your server. :-P
No, it was so beautiful to see my terminal blinking like a christmas tree :P 

> Since the width of the conversation list depends on the width of the window: for which sizes are we trying to fix this? What is the width of the window in the screenshots?
Yeah, you're right. Screenshots have 1345px width. And right now, I can't test other width cause... my server passed away :D

SansPseudoFix:
Ok, for 1024px width:


Under 1345px with 3 digit numbers, pagination is broken...

SansPseudoFix:
What about media queries? 

````CSS
@media screen and (max-width: 1345px) {
.pagination ul > li > a {
    padding: 4px 7px;
  }
}
````
Looks like that:
5745-1024-bis

**Comment by SansPseudoFix** [comment]
Now, pagination is responsive: between 1024px and 1353px, ````padding: 4px 7px;````. Between 1354px and up to infinity (and beyond) ````padding: 4px 12px;````.

1024px:



1278px:



1354px:

File: app/assets/stylesheets/conversations.scss [hunk 2]
```diff
@@ -180,6 +180,17 @@
...
+      margin-left: auto;
+      margin-right: auto;
+      text-align: center;
+    }
+  }
+}
+
+@media (min-width: 1024px) and (max-width: 1354px) {
```
svbergerem: I think the `min-width` is useless here. That would make the padding bigger if the window width is smaller than 1024px. Just define a max-width.

SansPseudoFix:
I removed ````min-width```` and fix pagination for mobile version:

Before:
5745-mobile-before

After:
5745-mobile-after

svbergerem:
@SansPseudoFix Would `inner_window: 1` work on the mobile website when you reduce the padding?

SansPseudoFix:
@svbergerem Perhaps, but I'm afraid that, by reducing padding, make mobile version unusable.

svbergerem:
Removing the numbered buttons for the last and next page could be a UX issue. I know that you can still use the arrow buttons but nevertheless that could make the UX worse for some users.

SansPseudoFix:
Worse than really too small buttons? And first and last buttons are still there. I just removed those between current, because, indeed, we can use arrows.

SansPseudoFix:
I change padding to ````4px 8px```` (upper, design is broken) and set ````:inner_window```` to 1. Here the results (sorry about big screeshots):

5745-mobile-padding-first
---
5745-mobile-padding-middle
---
5745-mobile-padding-end

Flaburgan:
Any reason you don't put a white background on the "..."?

AugierLe42e:
I think it's because it is not clickable, which makes sense.

Flaburgan:
It's white on the desktop version. We can maybe turn it to light-grey, but no background at all feels weird.

SansPseudoFix:
@Flaburgan: because it didn't have background before, so I left how it was. I was suprised at the begining too.

svbergerem:
@SansPseudoFix @Flaburgan 
> Perhaps, but I'm afraid that, by reducing padding, make mobile version unusable.

Do you think the current version in this PR is still usable? I don't have a mobile device to test this but if you say this is ok I will merge this PR after @SansPseudoFix squashed the commits.

svbergerem:
Oh, and add a background for `...` in the mobile version. I agree that this currently looks weird.

SansPseudoFix:
@svbergerem Well, finally, it's usable :P 

I unified mobile and desktop version too.

5745-bg-desktop
5745-disabled-bg

svbergerem:
Thank you.

Issue #4020, opened 2013-03-01
Get all pending specs green


jhass:
...or remove them if they aren't right any more. Partial contributions welcome!

* [ ] Services::Facebook#post swallows exception raised by facebook always being down
      *spec/models/services/facebook\_spec.rb:19*
* [x] InvitationCodesHelper add some examples to (or delete)      
      *spec/helpers/invitation\_codes_helper_spec.rb*
* [x] Role add some examples to (or delete) 
      *spec/models/role_spec.rb*
* [x] Stream::Tag with a user displays a post with a comment containing the tag search
      *spec/lib/stream/tag_spec.rb:36*
* [ ] Services::Twitter#post swallows exception raised by twitter always being down
      *spec/models/services/twitter_spec.rb:18*
* [x]  EvilQuery::Participation should include your statusMessages
       *spec/lib/evil\_query_spec.rb:27*
* [x] it is relayable it should behave like it is relayable validations on :author_id the author is on the parent object author's ignore list when object is created sends a retraction for the object
      *spec/shared_behaviors/relayable.rb:35*
* [x] Stream::Person returns the most recent posts -
      randomly failing on postgres
      *spec/lib/stream/person_spec.rb:13*

PR #5792, opened 2015-03-17
added specs and validations for Role model

MothOnMars:
- added shoulda-matchers gem for one-line specs
- added FactoryGirl syntax methods

Partial fix for https://github.com/diaspora/diaspora/issues/4020

I also added a few validations that seemed to make sense for roles, and tweaked the `Role.is_admin?` method to return a boolean. Hopefully CI will catch any issues that these changes may cause, but please let me know if I should back out any of those changes to the model itself.


jhass:
Mh, I'm not sure I like shoulda-matchers, it feels inconsistent with the expect syntax.

MothOnMars:
@jhass, the one-line syntax used by shoulda-matchers is supported by RSpec, even when using the expect syntax: https://github.com/rspec/rspec-expectations/blob/master/Should.md#one-liners

I feel that the inconsistency is a small trade-off for being able to write a one-liner instead of multiple lines. It's a solid Thoughtbot gem with great support & [documentation](http://matchers.shoulda.io/). It's probably saved me hundreds of hours and thousands of lines of spec code over the years. Check it out...let me know what you think...and in the meantime, I'll start making Hound happier... :smile:

jhass:
Mh, okay deal, we can try it for a while.

app/models/role.rb
```diff
@@ -1,18 +1,23 @@
...
-  scope :admins, -> { where(name: 'admin') }
+  validates :person, presence: true
+  validates :name, uniqueness: {scope: :person_id}
+  validates :name, inclusion: {in: %w(admin spotlight)}
+
+  scope :admins, -> { where(name: "admin") }
 
   def self.is_admin?(person)
-    find_by_person_id_and_name(person.id, 'admin')
+    self.exists?(person_id: person.id, name: "admin")
```
jhass: Not sure why hound isn't complaining here, but the `self.` is unnecessary.

MothOnMars: thanks - fixing

jhass:
Thank you!

PR #5794, opened 2015-03-18
Use jshintrb and guard-jshintrb instead of jshint

svbergerem:


svbergerem:
Should be ready to merge.

jhass:
Thanks!

PR #5797, opened 2015-03-18
[WIP] Markdown editor

ghost:
Proposition to use [bootstrap-markdown](http://www.codingdrama.com/bootstrap-markdown/) on the publisher.

Here is how it looks: 

md-editor

Keep in mind that you are only looking a PoC. Design is not definitive.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -98,6 +98,7 @@ gem "rails-assets-markdown-it-sanitizer",               "0.2.2"
 gem "rails-assets-markdown-it--markdown-it-for-inline", "0.1.0"
 gem "rails-assets-markdown-it-sub",                     "0.1.0"
 gem "rails-assets-markdown-it-sup",                     "0.1.0"
+gem "rails-assets-bootstrap-markdown",                  "~> 1"
 
 # jQuery plugins
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -487,6 +487,10 @@ GEM
       bundler (>= 1.3.0, < 2.0)
       railties (= 4.2.0)
       sprockets-rails
+    rails-assets-bootstrap (2.3.2)
+      rails-assets-jquery (>= 1.8.0, < 2.1.0)
+    rails-assets-bootstrap-markdown (1.1.4)
+      rails-assets-bootstrap (~> 2.3.0)
     rails-assets-diaspora_jsxc (0.1.1)
       rails-assets-jquery (~> 1.11.1)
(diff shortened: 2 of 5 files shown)


Flaburgan:
That looks cool! How does the black preview button work? Does it update the actual preview post? I would love to see a write / preview tabs system like here in github, this would be way easier to port to comment and conversation.

SansPseudoFix:
It could be really useful for many users. Many of them doesn't understand how to use markdown!

Flaburgan:
I can see one downside here: the picture button. We have to find a way to avoid confusion between the "Upload from your computer to diaspora*" action and the "Link an image from the web" action.

svbergerem:
Two preview buttons? In the demo the preview button displays the post inside of the publisher. That doesn't work here because we need to extract the mentions from the user input before we could create a preview. We would also have to make sure that the post in the publisher looks exactly like the post in the stream so we would have to add all styles to the preview inside of the publisher that currently apply to posts in the stream.

Does the markdown editor work with our mention input system?

AugierLe42e:
> That looks cool! How does the black preview button work? Does it update the actual preview post?

I think the black button can be removed as we already have our preview button. This plugin embed its own MD parser. But actually, the preview button did nothing when I tested it.

Let me point that what you are looking is the 1.x version of the plugin, compatible with BS2 only. What is showed on the website is 1.x version compatible with BS3 only. So, I'm not sure that all the features on the website will be available here.

svbergerem:
The website says
> Glyphicons (glyph) and Font Awesome (fa) are supported.

We will need Entypo support.

AugierLe42e:
As sources are open, we can adapt the plugin if needed, of course ;)

AugierLe42e:
Also, maybe this PR could be a good excuse to separate the publisher in a Handlebars template to use it with comments and conversations also ? It will let the user mention someone and upload photos in the comment.

jhass:
> Let me point that what you are looking is the 1.x version of the plugin, compatible with BS2 only. What is showed on the website is 1.x version compatible with BS3 only. So, I'm not sure that all the features on the website will be available here.

Sounds like the perfect reason to revisit this once we ported the site to Bootstrap 3. In the meantime we can discuss this and alternate solutions and how to implement them in a regular issue and you can rather focus on your existing PRs and short  term feasible contributions ;) Therefore I would like to close this for now, I don't think it's realistic currently.

svbergerem:
@AugierLe42e 

> It will let the user mention someone and upload photos in the comment.

No, you will have to extend the comment model to do that.

Issue #5804, opened 2015-03-20
Order of comments on the mobile version is reversed


Flaburgan:
The most recent comment is now displayed first.
This regression is probably caused by https://github.com/diaspora/diaspora/pull/5790

jhass:
No, I don't see how #5790 could have caused that.

Flaburgan:
Well, last pull was https://github.com/diaspora/diaspora/commit/0856eb8f86b6e2d8fba53b5defbd2268d1668106 and everything worked fine, I updated this morning and now the order is inversed.

jhass:
Oh you're right, that one was subtle. He switched from `@comments` to `@post.comments`, I didn't notice :/

PR #5805, opened 2015-03-21
order-of-comments-on-the-mobile-version-is-reversed #5804

visualsayed:
I'm sorry for causing this issue.

### DIFF
--- a/app/views/shared/_post_stats.mobile.haml
+++ b/app/views/shared/_post_stats.mobile.haml
@@ -11,7 +11,7 @@
       = @post.likes.size
 
   %ul.comments
-    = render partial: "comments/comment", collection: @post.comments, locals: {post: @post}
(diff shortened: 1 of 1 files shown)


jhass:
No worries. I wonder though if we can't reuse the `@comments` from the controller as we did earlier? You use `@post`, so the partial is dependent on the controller setting a variable already anyway.

Another nicer solution would be to use a [scope](http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many) on the association I guess.

visualsayed:
let me check this again.

visualsayed:
The Comment model already define for_a_stream scope that returns comments ordered by created_at ASC

jhass:
Thanks!

PR #5809, opened 2015-03-22
Refactor post content css, fix margin-bottom for lists in comments

### LINKED ISSUE #5626: Different rendering with <p> tag

svbergerem:
- fixes #5626 for comments
- sets the image max-width to 100% in the stream:
  image width stream
- improves oembed style in combination with polls:
  oembed polls

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -79,6 +79,9 @@
 @import 'comments';
 @import 'diaspora_jsxc';
 @import 'chat';
+@import 'markdown-content';
+@import 'oembed';
+@import 'post-content';
 
 /* right bar */
 @import 'sidebar';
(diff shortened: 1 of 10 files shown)


jhass:
Thanks!

PR #5817, opened 2015-03-26
Generate error pages on assets:precompile

jhass:
This allows us to reuse any CSS we have, unify
their look and unify their look with the regular
page design.

This works by instantiating ActionView and rendering
templates in a rake task.

Inspired by the errgent gem.

I tried to introduce no design changes, so they still look ugly. I'll leave prettifying to somebody else, this is about architecture ;)

Oh, this also makes sure the correct assets are referenced with Rails 4.


lislis:
thanks for tacking this :)

denschub:
Allright, let's try this. ;)

PR #5818, opened 2015-03-26
Fix frontend regressions

svbergerem:
- Fixes a oembed style regression introduced in #5809. The oembed div in posts had some margin even if there was no content.
- Adds missing `.container` divs to admin pages
- Adds a note to the changelog to inform podmins about changes for their custom splash pages.


denschub:
:cookie:

PR #5822, opened 2015-03-28
Replace default.html with a regular view

### LINKED ISSUE #5280: Old naming of assets (without hash code) used in public/default.html

jhass:
So we can correctly reference the images, fixes #5280 

This is solely to close #5280, it explicitly does not:
- cleanup the CSS
- update the content
- make the content translatable
- change the design so it integrates well into the rest of the application

These are all tasks that should be done in the mid-term, but not here.


denschub:
Thank you very much!

Flaburgan:
I think I should be able to work on this later in the spring.

PR #5824, opened 2015-03-30
Fix tag overflow mobile page

SansPseudoFix:
Another overflow...

Before:
overflow-tag-mobile-before

After:
overflow-tag-mobile-after

Maybe it's a little bit huge. What you think?

I'm using _EditorConfig_, so it must ok for indent, but if not, tell me :)

### DIFF
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -1204,3 +1204,8 @@ select#aspect_ids_ {
     background-color: $light-green;
   }
 }
+
+/* --- Tag page --- */
+#app #main h1 {
+  word-wrap: break-word;
+}


app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -1204,3 +1204,9 @@ select#aspect_ids_ {
     background-color: $light-green;
   }
 }
+
+/* --- Tag page --- */
+#app #main h1 {
+  overflow: hidden;
```
svbergerem: Please remove this line. Your fix also works without it and this line breaks the style of tags that contain letters like 'g':
broken tag style

SansPseudoFix:
Done :)

svbergerem:
Thank you.

PR #5837, opened 2015-04-01
Move default_url_options to application.rb

jhass:
Apparently the subclasses got loaded before the initializers
were run, thus not inheriting the config.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -102,5 +102,12 @@ module Diaspora
 
     # Will be default with Rails 5
     config.active_record.raise_in_transactional_callbacks = true
+
+    # Setup action mailer early
+    config.action_mailer.default_url_options = {
+      protocol: AppConfig.pod_uri.scheme,
+      host:     AppConfig.pod_uri.authority
+    }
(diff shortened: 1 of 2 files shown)


config/application.rb
```diff
@@ -102,5 +102,12 @@ class Application < Rails::Application
 
     # Will be default with Rails 5
     config.active_record.raise_in_transactional_callbacks = true
+
+    # Setup action mailer early
+    config.action_mailer.default_url_options = {
+      protocol: AppConfig.pod_uri.scheme,
+      host:     AppConfig.pod_uri.authority
```
jhass: wat

svbergerem: https://github.com/bbatsov/rubocop/blob/ca377951bb2ba17184d5ea9fb1bc8842097de1c9/spec/rubocop/cop/style/extra_spacing_spec.rb#L23

https://github.com/thoughtbot/hound/blob/master/config/style_guides/ruby.yml#L605

jhass: But https://github.com/diaspora/diaspora/blob/develop/.rubocop.yml#L36

PR #5843, opened 2015-04-02
Limited post content excluded from subject of comment email

margori:
Currently, when a user comments a limited post, a email is sent to all participants of that post.

Email subject contains first 70 chars of post text. I expect no part of limited text to be disclosed in email notification.

What subject do you expect for a comment to a limited post?

### DIFF
--- a/app/mailers/notification_mailers/comment_on_post.rb
+++ b/app/mailers/notification_mailers/comment_on_post.rb
@@ -6,7 +6,11 @@ module NotificationMailers
       @comment = Comment.find(comment_id)
 
       @headers[:from] = "\"#{@comment.author_name} (diaspora*)\" <#{AppConfig.mail.sender_address}>"
-      @headers[:subject] = "Re: #{@comment.comment_email_subject}"
+      if @comment.public?
+        @headers[:subject] = "Re: #{@comment.comment_email_subject}"
+      else
+        @headers[:subject] = I18n.t("notifier.comment_on_post.limited_subject")
+      end
     end
   end
 end
--- a/spec/mailers/notifier_spec.rb
+++ b/spec/mailers/notifier_spec.rb
@@ -318,6 +318,26 @@ describe Notifier, :type => :mailer do
     end
   end
 
+  context "limited comments" do
+    let(:commented_limited_post) {
+      bob.post(:status_message, to: :all, public: false,
+        text: "### Limited headline \r\n It's **really** sunny outside today")
+    }
+    let(:limited_comment) { eve.comment!(commented_limited_post, "Totally is") }
+
+    describe ".comment_on_limited_post" do
+      let(:limited_comment_mail) { Notifier.comment_on_post(bob.id, person.id, limited_comment.id).deliver_now }
+
+      it "SUBJECT: does not show limited message" do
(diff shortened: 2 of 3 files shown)


margori:
I give up. Houndci guy too damn picky.

jhass:
Actually these are all reasonable comments :)

spec/mailers/notifier_spec.rb
```diff
@@ -318,6 +318,26 @@
     end
   end
 
+  context "limited comments" do
+    let(:commented_limited_post) {
+      bob.post(:status_message, to: :all, public: false,
```
jhass: This one probably happens because you aligned the argument in the next line to the call instead of indenting it.

app/models/status_message.rb
```diff
@@ -128,7 +128,11 @@ def update_and_dispatch_attached_photos(sender)
   end
 
   def comment_email_subject
-    message.title length: 70
+    if self.public?
```
jhass: The `self.` here is actually redundant.

spec/mailers/notifier_spec.rb
```diff
@@ -318,6 +318,25 @@
...
+
+      it "SUBJECT: does not show limited message" do
+        expect(limited_comment_mail.subject).not_to include("Limited headline")
+      end
+
+      it "BODY: does not show limited message" do
+        expect(limited_comment_mail.body.encoded).not_to include("Limited headline")
+      end
+    end
+  end
```
jhass: This one just shows up from a previous run but didn't get invalidated in the diff. You can ignore it.

spec/mailers/notifier_spec.rb
```diff
@@ -318,6 +318,26 @@
     end
   end
 
+  context "limited comments" do
+    let(:commented_limited_post) {
+      bob.post(:status_message, to: :all, public: false,
+        text: "### Limited headline \r\n It's **really** sunny outside today")
+      }
```
jhass: Obvious enough, isn't it? :)

app/models/status_message.rb
```diff
@@ -128,7 +128,11 @@ def update_and_dispatch_attached_photos(sender)
...
+      message.title length: 70
+    else
+      I18n.t("status_messages.limited_subject")
```
jhass: I think `notifier.comment_on_post` would make a better location :)

Also how about moving the "Re: " added at app/mailers/notifications/comment_on_post.rb to a translation and leaving it off here, having the subject something like "There's a new comment on a post you commented" instead of "Re: Limited post"?

app/mailers/notification_mailers/comment_on_post.rb
```diff
@@ -6,7 +6,11 @@ def set_headers(comment_id)
       @comment = Comment.find(comment_id)
 
       @headers[:from] = "\"#{@comment.author_name} (diaspora*)\" <#{AppConfig.mail.sender_address}>"
-      @headers[:subject] = "Re: #{@comment.comment_email_subject}"
+      if (@comment.public?)
```
margori: Holy cow!!! C#, can't you leave me alone for a second?!

jhass:
Thanks!

margori:
Welcome! :)

PR #5846, opened 2015-04-03
Disable auto follow back on aspect deletion

### LINKED ISSUE #5845: Auto-Follow-Back broken, when the selected aspect doesn't exist
When a user activates the auto-follow-back, and then deletes the selected aspect, and then someone starts sharing with him ... the job fails with "ActiveRecord::AssociationTypeMismatch: Aspect(#75260260) expected, got NilClass(#4716400)" but sends a "xyz started sharing with you." notification. This ends in a huge federation-spam with share-requests.

The auto-follow-back should be disabled, when the aspect doesn't exist anymore.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10392065-auto-follow-back-broken-when-the-selected-aspect-doesn-t-exist?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #5845. 
- When deleting the auto follow back aspect it displays a flash message "%{name} was successfully removed. You used this aspect to automatically follow back users. Check your user settings to select a new auto follow back aspect." and disables the auto follow back feature.
- Changes the flash message `aspects.destroy.failure` from "%{name} is not empty and could not be removed." to "%{name} could not be removed." because we also remove aspects that are not empty.

### DIFF
--- a/app/controllers/aspects_controller.rb
+++ b/app/controllers/aspects_controller.rb
@@ -44,14 +44,20 @@ class AspectsController < ApplicationController
   end
 
   def destroy
-    @aspect = current_user.aspects.where(:id => params[:id]).first
+    @aspect = current_user.aspects.where(id: params[:id]).first
 
     begin
+      if current_user.auto_follow_back && @aspect.id == current_user.auto_follow_back_aspect.id
+        current_user.update(auto_follow_back: false, auto_follow_back_aspect: nil)
+        flash[:notice] = I18n.t "aspects.destroy.success_auto_follow_back", name: @aspect.name
+      else
+        flash[:notice] = I18n.t "aspects.destroy.success", name: @aspect.name
+      end
       @aspect.destroy
-      flash[:notice] = I18n.t 'aspects.destroy.success', :name => @aspect.name
     rescue ActiveRecord::StatementInvalid => e
-      flash[:error] = I18n.t 'aspects.destroy.failure', :name => @aspect.name
+      flash[:error] = I18n.t "aspects.destroy.failure", name: @aspect.name
     end
+
     if request.referer.include?('contacts')
       redirect_to contacts_path
     else
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -215,7 +215,8 @@ en:
       failure: "Aspect creation failed."
     destroy:
       success: "%{name} was successfully removed."
-      failure: "%{name} is not empty and could not be removed."
+      success_auto_follow_back: "%{name} was successfully removed. You used this aspect to automatically follow back users. Check your user settings to select a new auto follow back aspect."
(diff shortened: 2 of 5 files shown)


app/controllers/aspects_controller.rb
```diff
@@ -44,14 +44,23 @@ def new
...
-      @aspect.destroy
-      flash[:notice] = I18n.t 'aspects.destroy.success', :name => @aspect.name
+      if current_user.auto_follow_back && @aspect.id == current_user.auto_follow_back_aspect.id
+        @aspect.destroy
+        current_user.auto_follow_back = false
+        current_user.auto_follow_back_aspect = nil
+        current_user.save
+        flash[:notice] = I18n.t "aspects.destroy.success_auto_follow_back", name: @aspect.name
+      else
+        @aspect.destroy
```
jhass: You can move the destroy behind this if/else.

app/controllers/aspects_controller.rb
```diff
@@ -44,14 +44,23 @@ def new
...
+    @aspect = current_user.aspects.where(id: params[:id]).first
 
     begin
-      @aspect.destroy
-      flash[:notice] = I18n.t 'aspects.destroy.success', :name => @aspect.name
+      if current_user.auto_follow_back && @aspect.id == current_user.auto_follow_back_aspect.id
+        @aspect.destroy
+        current_user.auto_follow_back = false
+        current_user.auto_follow_back_aspect = nil
+        current_user.save
```
jhass: We can combine these into one statement: `current_user.update(auto_follow_back: false, auto_follow_back_aspect: nil)`

app/controllers/users_controller.rb
```diff
@@ -79,7 +79,7 @@ def update
         if password_changed
           redirect_to new_user_session_path
         else
-          render :edit
+          redirect_to edit_user_path
```
jhass: This doesn't break showing validation errors on the fields? Or I guess we don't do it there at all...

svbergerem: You are right, this breaks the validation errors. Will open a new issue for that instead of fixing it in this PR.

app/controllers/aspects_controller.rb
svbergerem: done

app/controllers/aspects_controller.rb
svbergerem: done

svbergerem:
Should be ready to merge.

jhass:
thank you!

PR #3919, opened 2013-01-27
Activity stream keeping retracted participations

desyncr:
Fixes https://github.com/diaspora/diaspora/issues/3393. Please review.

Issue #3393, opened 2012-06-17
"My Activity" Keeps unliked posts.


kas70:
when you like and then unlike a post, it's still in your activity stream. is this how it should be?

PR #5852, opened 2015-04-06
#3393 Activity stream keeping retracted participations

margori:
Solve #3393 Continues PR #3919

I need code review, specially in the migration file. It has a complex query that I didn't know how to write it in Ruby style.

### DIFF
--- a/app/models/like.rb
+++ b/app/models/like.rb
@@ -12,6 +12,7 @@ class Like < Federated::Relayable
       {:target => @target, :positive => true}
     end
   end
+  has_one :participation, :dependent => :destroy, :foreign_key => :target_id, :primary_key => :target_id
 
   after_commit :on => :create do
     self.parent.update_likes_counter
--- a/app/models/like.rb
+++ b/app/models/like.rb
@@ -12,7 +12,6 @@ class Like < Federated::Relayable
       {:target => @target, :positive => true}
     end
   end
-  has_one :participation, :dependent => :destroy, :foreign_key => :target_id, :primary_key => :target_id
 
   after_commit :on => :create do
     self.parent.update_likes_counter
@@ -20,6 +19,8 @@ class Like < Federated::Relayable
 
   after_destroy do
     self.parent.update_likes_counter
+    participation = author.participations.where(target_id: target.id).first
+    participation.unparticipate! if participation.present?
   end
 
   xml_attr :positive
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -377,10 +377,10 @@ class User < ActiveRecord::Base
       retraction = Retraction.for(target)
     end
 
-   if target.is_a?(Post)
-     opts[:additional_subscribers] = target.resharers
-     opts[:services] = self.services
-   end
+    if target.is_a?(Post)
+      opts[:additional_subscribers] = target.resharers
+      opts[:services] = services
+    end
 
     mailman = Postzord::Dispatcher.build(self, retraction, opts)
     mailman.post
--- a/app/models/user/social_actions.rb
+++ b/app/models/user/social_actions.rb
@@ -1,6 +1,6 @@
 module User::SocialActions
   def comment!(target, text, opts={})
-    find_or_create_participation!(target)
+    update_or_create_participation!(target)
     Comment::Generator.new(self, target, text).create!(opts)
   end
 
@@ -9,17 +9,17 @@ module User::SocialActions
   end
 
   def like!(target, opts={})
-    find_or_create_participation!(target)
+    update_or_create_participation!(target)
     Like::Generator.new(self, target).create!(opts)
   end
 
   def participate_in_poll!(target, answer, opts={})
-    find_or_create_participation!(target)
+    update_or_create_participation!(target)
     PollParticipation::Generator.new(self, target, answer).create!(opts)
   end
 
   def reshare!(target, opts={})
-    find_or_create_participation!(target)
+    update_or_create_participation!(target)
     reshare = build_post(:reshare, :root_guid => target.guid)
     reshare.save!
     Postzord::Dispatcher.defer_build_and_post(self, reshare)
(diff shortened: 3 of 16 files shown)


lib/evil_query.rb
```diff
@@ -17,7 +17,10 @@ def initialize(user)
     end
 
     def posts
-      Post.joins(:participations).where(:participations => {:author_id => @user.person.id}).order("posts.interacted_at DESC")
+      Post.joins(:participations)
+        .where(:participations => {:author_id => @user.person.id})
+        .where("participations.count > 0")
```
jhass: The participations table is quite big already, I think instead we should check if the counter reaches 0 while updating it and destroy the participation in that case.

app/models/like.rb
```diff
@@ -19,6 +19,10 @@ def relayable_options
 
   after_destroy do
     self.parent.update_likes_counter
+    participation = Participation.where({ 
+      target_id: self.target_id, author_id: self.author_id
+    }).first()
+    participation.update!(count: participation.count.pred) if participation.present?
```
jhass: That code is redundant, let's extract it somewhere, `Participation::add_reference` or something seems reasonable. That would also be a good place to check whether the counter reached 0 then and it would simplify the unit tests for Comment & Like, you just check whether the method was called, functionality of it only needs to be tested once then in the unit tests for Participation.

app/controllers/status_messages_controller.rb
```diff
@@ -78,7 +78,8 @@ def create
         current_user.dispatch_post(photo)
       end
 
-      current_user.participate!(@status_message)
+      participation = current_user.participate!(@status_message)
+      participation.update!(count: participation.count.next)
```
jhass: I don't quite follow, don't you do this in `find_or_create_participation!` already?

db/migrate/20150404193023_participation_counter.rb
```diff
@@ -0,0 +1,23 @@
...
+               from likes l
+               where
+                  l.target_id = participations.target_id
+                  and l.author_id = participations.author_id) +
+              (select count(c.id)
+               from comments c
+               where
+                  c.commentable_id = participations.target_id
+                  and c.author_id = participations.author_id)
+          )"
```
jhass: What I can come up with is not really worth it:

```ruby
likes_count = Like.select("COUNT(likes.id)")
                  .where("likes.target_id = participations.target_id")
                  .to_sql
comments_count = Comment.select("COUNT(comments.id)")
                        .where("comments.commentable_id = participations.target_id")
                        .where("comments.author_id = participations.author_id")
                        .to_sql
Participation.update_all("count = (#{likes_count}) + (#{comments_count)")
```

So maybe just capitalize the SQL keywords to make it a bit more readable.

app/controllers/status_messages_controller.rb
margori: No, it doesn't.

But, since default value of 'participation.count' field is 1, update is not required.

margori:
Improvement: when participation.count == 0 -> destroy , included in migration

app/models/like.rb
```diff
@@ -19,6 +19,14 @@ def relayable_options
...
   after_destroy do
     self.parent.update_likes_counter
+    participation = Participation.where(target_id: target_id, author_id: author_id).first
+    if participation.present?
+      if participation.count == 1
+        participation.destroy
+      else
+        participation.update!(count: participation.count.pred)
+      end
+    end
```
jhass: You're still duplicating code here, I'd like to see that extracted somewhere, possibly a class method on `Participation`.

margori:
Created Participation.unparticipate! , called when user retract a participation.

Feedback, welcome, specially related to tests.

app/models/user.rb
```diff
@@ -377,10 +377,14 @@ def retract(target, opts={})
...
-     opts[:services] = self.services
-   end
+    if target.is_a?(Post)
+      opts[:additional_subscribers] = target.resharers
+      opts[:services] = self.services
+    elsif target.is_a?(Like)
+      participations.where(target: target.target).first.unparticipate!
+    elsif target.is_a?(Comment)
+      participations.where(target: target.post).first.unparticipate!
+    end
```
jhass: I think I liked calling it in the callbacks better though. I never checked, passing target like that restricts on type & id?

margori:
"Unparticipating" when destroying 'like' or 'comment'

Migration increased in complexity, due poll participation is an accountable participation as well.

db/migrate/20150404193023_participation_counter.rb
```diff
@@ -0,0 +1,36 @@
...
+      set count = count + 1
+      where exists(
+        select 1
+        from poll_participations pp
+          inner join polls po
+            on po.id = pp.poll_id
+        where
+          pp.author_id = participations.author_id
+          and po.status_message_id = participations.target_id
+      )"
```
jhass: Let's not mix both styles, either do it via AR or write a full query. Also I think this could still be one query.

app/models/user/social_actions.rb
```diff
@@ -49,6 +49,11 @@ def build_message(conversation, opts={})
   end
 
   def find_or_create_participation!(target)
```
jhass: I think we need to change the name of this, it no longer just finds it. `update_or_create_participation!`?

jhass:
Thank you!

margori:
Welcome! This PR ley me learn about cucumber. Amazing tool, easy to ensure features functionality and quality.

Flaburgan:
I just upgraded my pod from 0.5 to 0.6-pre and the db migration toke ages. I run postgreSQL with a ~10GB database. Any chance the migration time can be improved?

margori:
> the db migration toke ages.

I think is the cost of a single update query.

> Any chance the migration time can be improved?

Of course there is. Please, I don't have such a big db to test my suggestions properly, may you run them?

Flaburgan:
Sure, you give me the file you want me to test.

Issue #5862, opened 2015-04-15
there is no default-* in public/assets in 0.5.0.0


pravi:
```ruby
Index: diaspora/script/server
===================================================================
--- diaspora.orig/script/server
+++ diaspora/script/server
@@ -120,7 +120,7 @@ then
 fi
 
 # Check if assets are precompiled
-if [ "$RAILS_ENV" = "production" -a -z "$(find public/assets -maxdepth 1 -name 'default-*.css' -print -quit)" ]
+if [ "$RAILS_ENV" = "production" -a -z "$(find public/assets -maxdepth 1 -name 'home-*.css' -print -quit)" ]
 then
   fatal "You're running in production mode without having assets
 precompiled. Now and after each update before you restart the
```

jhass:
Nice catch, want to open a pull request?

pravi:
done!

PR #5863, opened 2015-04-15
fix for #5862, check for compiled assets

pravi:
default-\* is not there anymore in assets folder, check for home-\* Fix for #5862


jhass:
Merged as 2613896cfdd8c9d9feabcdb5a3358354fef42b93

Thank you!

PR #5867, opened 2015-04-20
fix control icon alignment in chromium browser

SuperTux88:
The X icon in chromium was not aligned. In Firefox it was OK, now it's 2px lower, but OK. :)

It is still slightly different, but better than before.
### Chromium:

before:
chromium-before
after:
chromium-after
### Firefox:

before:
firefox-before
after:
firefox-after


jhass:
Thank you!

PR #5873, opened 2015-04-21
Update markdown-it, disable fuzzy links

svbergerem:
Now `http://` or `https://` is required for links. This fixes false positives like `0.5.0.0` or `test.sh`.


jhass:
Merged as a14713b985841aa55bad063e59bdb35338cf0cf1

Thank you!

PR #5877, opened 2015-04-22
fix notification mails for limited posts

SuperTux88:
there were mails like this:

> XYZ liked your post:
> 
> There's a new private message in diaspora\* for you to check out.

### DIFF
--- a/spec/mailers/notifier_spec.rb
+++ b/spec/mailers/notifier_spec.rb
@@ -105,6 +105,32 @@ describe Notifier, :type => :mailer do
     end
   end
 
+  describe ".mentioned limited" do
+    before do
+      @user = alice
+      @post = FactoryGirl.create(:status_message, public: false)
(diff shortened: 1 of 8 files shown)


spec/mailers/notifier_spec.rb
```diff
@@ -131,7 +157,31 @@
...
+  end
+
+  describe ".liked limited" do
+    before do
+      @post = FactoryGirl.create(:status_message, author: alice.person, public: false)
+      @like = @post.likes.create!(author: bob.person)
+      @mail = Notifier.liked(alice.id, @like.author.id, @like.id)
+    end
+
+    it "TO: goes to the right person" do
```
denschub: nit: we usually try to build valid sentences here, so "it delivers the mail to the right person" and "it contains the name of the person liking in the body", for example

denschub:
:cookie:

PR #5879, opened 2015-04-22
encode unencoded urls and normalize IDN urls

SuperTux88:


denschub:
Nice catch, thank you very much!

:cookie:

PR #5882, opened 2015-04-23
Fixes #5416-Adds nprogress gem and modifies header links

### LINKED ISSUE #5416: Clicking on 'Stream' should reload the stream (not the page) when already on the stream page
With the chat feature in now, this becomes more important.

This is should work for both, the sidebar and the header. We could think think about applying it to the header icon too.

Also I think a loading indication while the stream is loading would be nice, similar to what we already have for infinite scroll.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6450282-clicking-on-stream-should-reload-the-stream-not-the-page-when-already-on-the-stream-page?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zachrab:
@jhass initial stab at #5416


jhass:
I checked this out locally.

* It breaks the top links on non-stream pages, such as profiles (loads the stream into the profile), /contacts (doesn't do anything at all) etc.
* I'm not sure about the place and color of the nprogress spinner, I think it should be inside the empty stream. Ideally all uses of ajax-loader.gif would be replaced by it too, but that's maybe something for a different PR.
* It doesn't reload the stream when clicking on the already active link, though adding that in a separate PR is acceptable.

denschub:
Closing due to excessive inactivity. Feel free to reopen if you want to continue working on this.

PR #5883, opened 2015-04-24
use camo for medium and small avatars

SuperTux88:
on the contacts page, external urls were used without camo


denschub:
Thank you!

:cookie:

PR #5886, opened 2015-04-25
Fix unconventional image asset inclusion in css asset

artoartnik:
Instead of using .css.erb, it's more conventional to use the asset_path helper in the css file.


Raven24:
thanks, try to stick to our git workflow: https://wiki.diasporafoundation.org/Git_workflow#Create_an_Issue-Specific_Feature_Branch
keeps your git branches clean and nice ;)

jhass:
Thank you!

Issue #5874, opened 2015-04-21
Bolded markdown URLs broken in two distinct ways.


gordon-morehouse:
### WHAT ACTUALLY HAPPENS

1. Enter a bolded bare URL such as `**http://www.google.com/**`
2. Post
3. URL is not presented in bold and has two asterisks appended when clicked.

Italics not tested.

### WHAT I EXPECT TO HAPPEN

The URL should be presented in bold without markup cruft dangling from either end in the href.

svbergerem:
> URL is not presented in bold

The HTML for those URLs is ok. We just need to change the CSS:
![strong url](https://cloud.githubusercontent.com/assets/3798614/7256677/e7272598-e850-11e4-9ffb-6ee2f3832996.png)

> has two asterisks appended when clicked

I am not able to reproduce that.

svbergerem:
#5887 fixed the first issue. Since I wasn't able to reproduce the second one, I'll close this. Feel free to reopen if you are able to reproduce the second issue again.

PR #5887, opened 2015-04-25
Issue #5874 Fix styled links

szlend:
I don't think applying font-weight to \* is a good idea. This fixes the issue where you're unable to make bold links. I could not reproduce the appended asterisk.


svbergerem:
Thank you!

Issue #5508, opened 2014-12-30
Presenters code is not reloaded in development


collimarco:
This issue affects `app/presenters/post_presenter.rb` and *probably* the other presenters.

This morning I wasted a lot of time figuring out why the modifications I've done to `post_presenter.rb`were not producing the desired effects. Then I realized that whatever I wrote in that file, nothing happened.

**When I restarted the server everything worked as expected.** This means that **the code in that file is not automatically reloaded in development environment**. Instead one expects it is reloaded as any other file (models, controllers, assets, etc.).

PR #5888, opened 2015-04-25
Issue #5508: Remove post_presenter from initializers

szlend:
Libraries included in the initializer are not reloaded automatically. It's also not necessary because rails automatically loads classes in app/ as long as they follow conventions.


jhass:
Thanks!

Issue #5485, opened 2014-12-19
Mobile: Comment button turns uresponsive when trying to submit an empty comment


Makailija:
Commenting on the mobile view, if you have an empty comment textarea and if you click the Comment button, the button turns unresponsive, saying just ”Commenting...”and staying like that infinitely. Curiously enough, the post that you tried to "comment" appears in your "My activity" stream after this.

I guess the expected behaviour would be to focus on the textarea, just like on the desktop view.

polsVoice:
So, this has been fixed?

jhass:
No, #5890 is not merged.

theworldbright:
The desktop view has the following code block that causes the comment box to be focused again if the comment is empty.

```  
createComment: function(evt) {
    if(evt){ evt.preventDefault(); }
    
    var commentText = $.trim(this.$('.comment_box').val());
    this.$(".comment_box").val("");
    this.$(".comment_box").css("height", "");
    if(commentText) {
      this.model.comment(commentText);
      return this;
    } else {
      this.$(".comment_box").focus();
    }
  },
```

Is it sufficient to prevent the comment creation request from reaching the controller with a client side check in the mobile view as follows? 

```
  $(document).on("click", ".new_comment", function(evt) {
    var form = $(this);
    var commentBox = form.find('.comment_box')
    var commentText = $.trim(commentBox.val());
    if (!commentText) {
        commentBox.focus();
        return false;
    }
  });
```

PR #5890, opened 2015-04-25
Issue #5485: Reset comment submit button on mobile if request fails

artoartnik:
This is a temporary hackish fix for the comment button freezing on the mobile interface, proper error handling should be implemented in the controllers, a validation fail shouldn't trigger an exception.


denschub:
Thanks for your work. However, we are not able to accept "temporary hackish fix"es for issues...

You are right, proper error handling should be implemented and I see no reason to add a dirty fix if we can do better. Feel free to implement proper error handling and submit a pull request for that! :) Keep in mind that we usually have tests for our bugfixes, so you should also write a test to ensure the error handling will not break in the future.

denschub:
Eh, reopening. You can add further/new code to this PR. I closed by accident, sorry. ;)

jhass:
Superseded by #6543 for now, thank you.

PR #5892, opened 2015-04-25
automatically use the correct database config depending on ENV["DB"] setting

mrbrdo:


jhass:
Sorry, but I will not accept this.

Once https://github.com/bundler/bundler/pull/3531 is released, the `DB` environment variable will be removed and the decision point about which adapter is active will be in fact `database.yml`.

Nice thought otherwise, thanks anyway!

PR #5896, opened 2015-04-27
fix minor errors in db:seed task

Raven24:
someone at rossconf stumbled over the broken seed db task, so here is a fix for it ;)


jhass:
What do you think about just dropping the seeds file?

Raven24:
I wouldn't be very sad for long  :D

jhass:
Let's do that then

Raven24:
updated PR, not sure how to satisfy hound correctly in this case

jhass:
It's set to table style:
```ruby
{
  a:   b,
  cc:  d,
  eee: f
}

Raven24:
got it, updated.

jhass:
Thanks!

Issue #3076, opened 2012-03-28
bookmarklet gives a blank page when a lot of text is selected


pravi:
Go to http://library.gnome.org/misc/release-notes/3.4/index.html.ml

and select three paragraphs below the first picture or all paragraphs in the beginning just before the first picture. This is an arbitrary limitation.

PR #5904, opened 2015-05-01
prefill bookmarklet with markdown-formatted content

Raven24:
this is based on #5384 and also adds some calculations to center the bookmarklet in the parent browser window.
the inline JS is now a bit longer, but hopefully more readable.

### DIFF
--- a/app/assets/javascripts/app/views/bookmarklet_view.js
+++ b/app/assets/javascripts/app/views/bookmarklet_view.js
@@ -1,7 +1,7 @@
 // @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-v3-or-Later
 
 app.views.Bookmarklet = Backbone.View.extend({
-  separator: ' - ',
+  separator: "\n\n",
 
   initialize: function(opts) {
     // init a standalone publisher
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/app/views/bookmarklet_view.js
```diff
@@ -25,8 +25,12 @@ app.views.Bookmarklet = Backbone.View.extend({
     var p = this.param_contents;
     if( p.content ) return p.content;
 
-    var contents = p.title + this.separator + p.url;
-    if( p.notes ) contents += this.separator + p.notes;
+    var contents = "### " + p.title + this.separator;
```
Raven24: I'd be okay with simply using bold text instead of a heading... opinions?

jhass:
Mh, I wonder if we can externalize this into a file and call uglifier on it.

jhass:
Or yet better hook it into the asset pipeline and read the compiled file, like I hooked generating the static HTML pages into it, would avoid the execjs dependency on runtime that uglifier would add, mh.

Raven24:
ok, gotta have a look at how you did that :)

jhass:
yeah, it might be overkill, but that inline JS...

Raven24:
alright, I played around with asset compilation a bit, and I think running uglifier manually is the easiest way in this case. I added that step to the precompile rake task, while also keeping the JS updated in development.

jhass:
Nice!

svbergerem:
@Raven24 Nice! :) I just found #3076 where you wrote that you didn't want to touch the bookmarklet code. Since you touched the code in this PR it would be great if you could limit the string length and also fix that issue. (Obviously it's up to you if you would like to do that)

Raven24:
right, totally forgot about that one (well, it's 3 years old...)
sure, I can limit the GET length to something reasonable, while I'm at it

jhass:
Thank you!

Issue #4218, opened 2013-06-13
install script does not check for required binaries


chaosmatic:
I'm running Ubuntu 12.04

From the prereqs mentioned in http://wiki.diasporafoundation.org/Installing_on_Debian I had installed only nodejs, rvm, ruby, rubygems and rails. I had not installed any of the dependencies for imagemagick, redis, mysql, mysqlite, openssl or ffi.

The install script did not complain until it crashed during 'bundle install' due to the missing binaries.

jhass:
We decided to drop the install scripts.

Issue #4750, opened 2014-01-29
Improvements for install.sh


Ryuno-Ki:
Good morning :)

I'm able to run a pod in development mode since 7 hours :D

But I have some suggestions to improve the script!

First, make sure (or check) to install ruby 2.0.0-p??? in forehand. <code>rvm get stable</code> pulls a v1.9 on my Sabayon Linux machine ;-)

Second, make sure (or check) for a login shell: <code>/bin/bash --login</code>.

Third, check, wether RUBYOPT is unset (common issue on Gentoo systems).

Fourth, maybe offer a hint to <pre>/usr/bin/mysqlaccess.conf</pre> for database credentials (this was my obstacle). I'm not sure, wether this file exist on non-Sabayon distros, though. But a little guidance for checking of a database user & permission is appreciated.

jhass:
We decided to drop the install scripts.

Issue #4275, opened 2013-07-01
problems with install.sh


nieve:
On Ubuntu 12.04 I run install.sh locally and get the following errors:

```
nieve@ubuntu:~/rails_projects$ bash install.sh 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1666  100  1666    0     0   5387      0 --:--:-- --:--:-- --:--:--  9103
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   376  100   376    0     0   1535      0 --:--:-- --:--:-- --:--:--  1978
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1088  100  1088    0     0   4457      0 --:--:-- --:--:-- --:--:--  6044
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  5206  100  5206    0     0  21679      0 --:--:-- --:--:-- --:--:-- 27989
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4735  100  4735    0     0  18050      0 --:--:-- --:--:-- --:--:-- 25052
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    69  100    69    0     0    275      0 --:--:-- --:--:-- --:--:--   387
#####################################################################

DIASPORA* INSTALL SCRIPT

----

This script will guide you through the basic steps
to get a DEVELOPMENT setup of Diaspora* up and running

For a PRODUCTION installation, please do *not* use this script!
Follow the guide in our wiki, instead:
-- http://wiki.diasporafoundation.org/Installation_guides

#####################################################################

Press [Enter] to continue... 

[ info] -- checking for rvm...   found

Where would you like to put the git clone, or, where is your existing git clone?
-> ~/rails_projects/diaspora

install.sh: line 60: realpath: command not found
[debug] -- pt1 - existing path: /
[debug] -- pt2 - new path: diaspora/
the folder you specified does not exist.
create '/diaspora/'?
Press [Enter] to create it and continue... 
[...]

PR #5907, opened 2015-05-02
add execution permission to scripts

pravi:
these are shell scripts without execution permissions, adding execution permissions.


jhass:
There's no need for them to be executable, https://github.com/diaspora/diaspora/blob/develop/script/install.sh#L28

However do they even still work? What's the state here, @Raven24 ?

pravi:
lintian gives script-not-executable since it has a shebang line at start. If it doesn't hurt, it could be made executable or if that is required can we remove the shebangs?

jhass:
I think we can remove the hashbangs, @Raven24 needs to confirm. Or remove them altogether.

Raven24:
I just tried and it seems they should still run. They don't necessarily need to be executable, since the referenced files are never run directly, but insted `eval`d into the main script context and contain only function definitions and no directly runnable code on their own. So feel free to remove the `#!`.

I've always found those scripts to be quite the ugly hack. also that code being bash, I don't really feel confident to assess any part of their quality (even though I wrote them).
I'd rather point people to [the Ruby version](https://github.com/Raven24/diaspora-install) I wrote about a year ago, instead of the bash mess that hasn't really been touched in two years...

jhass:
Can we get a clear statement please? Do you think we should drop them?

Raven24:
tl;dr
if we should, for some reason keep them, drop the `#!`
but please let's remove them completely ;)

jhass:
Okay, thanks.

@pravi do you want to update the PR to drop them or should I do it?

svbergerem:
I guess removing the install script would fix #4218, #4275 and #4750.

pravi:
I have updated the PR to drop them

jhass:
Thanks!

goobertron:
Thanks, all, for this work.

I'm guessing we should delete the [install script page](https://wiki.diasporafoundation.org/Installation/Script) on the wiki, and the link to it from the development setup section of the [main installation page](https://wiki.diasporafoundation.org/Installation) (and any other links to it on the wiki). I won't do this myself in case I'm mistaken! Apologies if someone was just about to do that; I just saw this PR and wanted to comment before I forget.

PR #5915, opened 2015-05-03
Check request.format :mobile in home controller

### LINKED ISSUE #4518: Missing landing page in mobile view
I've just noticed that by going to a pod's landing page/root page when logged out, and then clicking 'toggle mobile', the page becomes blank (with just the mobile background), and the message

    put something in app/views/home/_show.html

is displayed.

I've checked this on a number of pods, running 0.2.0.0, 0.1.1.0 and development code, and all of those which have the 'toggle mobile' link at the bottom right suffer from this problem, with the exceptions (out of those I tried) of https://diasp.eu/ (which has a 'beta' design for mobile - 'A creative way to remix your world') and https://diapod.net (which has a bespoke page displaying stats).

When visiting the site root from a mobile device (my phone), I'm automatically redirected to the sign-in page (except for the two pods mentioned above which have their own landing pages in mobile view), so it seems that mobile users won't be hitting this problem. Even so, it would be good to tidy up this little loophole.

<bountysource-plugin>
---
There is a **[$25 open bounty](https://www.bountysource.com/issues/896970-missing-landing-page-in-mobile-view?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #4518.
- additional check if the request format is `:mobile` in the home controller
- allow using `app/home/_show.haml` instead of `_show.html.haml` and `_show.mobile.haml`
- merge `show.html.haml` and `show.mobile.haml`

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,7 @@
 #   the COPYRIGHT file.
 
 class ApplicationController < ActionController::Base
+  before_action :force_tablet_html
   has_mobile_fu
   protect_from_forgery :except => :receive
 
@@ -38,9 +39,7 @@ class ApplicationController < ActionController::Base
 
   # Overwriting the sign_out redirect path method
   def after_sign_out_path_for(resource_or_scope)
-    # mobile_fu's is_mobile_device? wasn't working here for some reason...
-    # it may have been just because of the test env.
-    if request.env['HTTP_USER_AGENT'].try(:match, /mobile/i)
+    if is_mobile_device?
       root_path
     else
       new_user_session_path
@@ -124,10 +123,11 @@ class ApplicationController < ActionController::Base
   def mobile_switch
     if session[:mobile_view] == true && request.format.html?
       request.format = :mobile
-    elsif request.format.tablet?
-      # we currently don't have any special tablet views...
-      request.format = :html
     end
+  end
+
+  def force_tablet_html
+    session[:tablet_view] = false
   end
 
   def after_sign_in_path_for(resource)
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -7,15 +7,17 @@ class HomeController < ApplicationController
     partial_dir = Rails.root.join("app", "views", "home")
     if user_signed_in?
       redirect_to stream_path
-    elsif is_mobile_device?
+    elsif request.format == :mobile
       if partial_dir.join("_show.mobile.haml").exist? ||
-         partial_dir.join("_show.mobile.erb").exist?
+         partial_dir.join("_show.mobile.erb").exist? ||
+         partial_dir.join("_show.haml").exist?
         render :show
       else
         redirect_to user_session_path
       end
     elsif partial_dir.join("_show.html.haml").exist? ||
-          partial_dir.join("_show.html.erb").exist?
+          partial_dir.join("_show.html.erb").exist? ||
+          partial_dir.join("_show.haml").exist?
       render :show
     else
       render :default,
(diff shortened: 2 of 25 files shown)


app/controllers/home_controller.rb
```diff
@@ -7,15 +7,17 @@ def show
     partial_dir = Rails.root.join("app", "views", "home")
     if user_signed_in?
       redirect_to stream_path
-    elsif is_mobile_device?
+    elsif is_mobile_device? || request.format == :mobile
```
Flaburgan: I don't know the body of the is_mobile_device method but shouldn't this test be included in it?

svbergerem: No, it's not included in that method: https://github.com/brendanlim/mobile-fu/blob/master/lib/mobile_fu.rb#L89.

@jhass currently it's not possible to manually switch to the desktop version when you are using a mobile device. (because of `is_mobile_device?`) I'd like to remove that method call and just check `request.format == :mobile`. I tested the changes with the FirefoxOS simulator and everything worked like I planned. Do you know if we set the `request.format` somewhere if it is unset so it would be safe to remove the method call?

jhass: I have no idea.

app/controllers/application_controller.rb
```diff
@@ -122,6 +122,12 @@ def grammatical_gender
   # use :mobile view for mobile and :html for everything else
   # (except if explicitly specified, e.g. :json, :xml)
   def mobile_switch
+    # mobile_fu's is_mobile_device? wasn't working here for some reason...
+    # it may have been just because of the test env.
+    if session[:mobile_view].nil? && request.env['HTTP_USER_AGENT'].try(:match, /mobile/i)
```
jhass: shouldn't this at least reuse `ActionController::MobileFu::MOBILE_USER_AGENTS`?

features/support/env.rb
```diff
@@ -21,9 +21,19 @@
 Capybara.register_driver :selenium do |app|
   require 'selenium/webdriver'
   Selenium::WebDriver::Firefox::Binary.path = ENV['FIREFOX_BINARY_PATH'] || Selenium::WebDriver::Firefox::Binary.path
-  Capybara::Selenium::Driver.new(app, :browser => :firefox)
+  Capybara::Selenium::Driver.new(app, browser: :firefox)
 end
 
+Capybara.register_driver :mobile do |app|
+  require 'selenium/webdriver'
```
jhass: Move the require to the ones at the top and remove the duplicate from the earlier driver registration

features/support/env.rb
```diff
@@ -21,9 +21,19 @@
...
+  Selenium::WebDriver::Firefox::Binary.path = ENV['FIREFOX_BINARY_PATH'] || Selenium::WebDriver::Firefox::Binary.path
```
jhass: This seems to be a global, maybe you can move it out of and above the two register_driver calls? And thus deduplicate it.

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,7 @@
 #   the COPYRIGHT file.
 
 class ApplicationController < ActionController::Base
+  before_filter :force_tablet_html
```
jhass: That should actually throw a deprecation warning, it's `before_action` now.

svbergerem:
Should be ready to review and merge. :)

jhass:
Merged as 399fdb2c7d746d934cf85821e005c90a30294b44

Thank you!

Issue #5923, opened 2015-05-04
Facebook crosspost silently fails, access token not removable, connection has no ID


oya3un:
Apologies if this is limited to joindiaspora.com, but apparently its connection to service "Facebook" for posting there via diaspora* is broken. Twitter works, Facebook doesn't (persistently since late last week). Just in case this is because FB blocks everything non-compliant with their API 2.x since last Thursday - has anybody seen this coming?

Unlike before, disconnecting-reconnecting in Settings --> Services doesn't solve the issue. What's different this time is that upon any renewed connection after a disconnect, the corresponding Facebook ID isn't mentioned any longer: the status message says "connected to Facebook as" like it's supposed to when there actually is a connection - but that message used to have an e-mail-address trailing that indicated the FB ID, that's gone now. 

By the way, remove the Facebook service/token altogether and install the whole connection afresh? I'd love to, but the removal feature has ceased to exist...

Flaburgan:
Yeah, the bridge with Facebook is kinda broken... There is several issues already open about it, see https://github.com/diaspora/diaspora/issues/5260 and https://github.com/diaspora/diaspora/issues/5085

denschub:
Closing this. Additional configuration available with 7c0e50c2c270e696b7dbd6540d00b4cc38e25375, additional information regarding this change and issues in #5085. Feel free to reopen if you feel the urge of doing so.

Loelo:
Hi, any hope about this facebook cross-posting working anew ?

denschub:
There is always hope, isn't it?

No, seriously, I see no reason why it should not work...

PR #5889, opened 2015-04-25
upgrade Twitter gem - fixes Faraday deprecation warning

mrbrdo:
tested connect twitter, post on twitter and delete on twitter - working

Issue #5260, opened 2014-09-30
Working on FaceBook API 2.x


Anime-Machina:
Hello there,

Today I noticed that FaceBook is closing down API 1.x on 25th December and all API's will have to be updated prior 25th December. Is D* already running API 2.x or still under the old API or is it already running API 2.X???

Proof:
http://i.imgur.com/jx0Iup7.png

Regards,
Anime Machina

jaywink:
The dates are wrong as discussed on various forums:

> With the launch of v2.0 of the API, we have two versions in service at the same time:
> 
> Version 1.0, which is what we call the API as it existed the day before v2.0 was launched. We'll support v1.0 for one year and it will expire on April 30th, 2015.
> Version 2.0, which is what this upgrade guide covers. Version 2.0 is supported for at least two years. At the earliest, it will expire on April 30th, 2016.

https://developers.facebook.com/docs/apps/upgrading#v1tov2

The problem I have is I can't figure out from anywhere *why* our current integration would even work, since we've *never* required afaict *publish_actions* - **and this is not changing in v2**.. Maybe time for a podmin poll..

But AFAICT our code should be all good with api v2. I don't see anything in the v2 API that would break our current implementation.

Anime-Machina:
I did had error messages showing up about post actions, it definitely needs more info in future. Although I personally dropped FaceBook and Twitter support on my pod cause some one violated their terms.

denschub:
Closing this. Additional configuration available with 7c0e50c2c270e696b7dbd6540d00b4cc38e25375, additional information regarding this change and issues in #5085. Feel free to reopen if you feel the urge of doing so.

Issue #5085, opened 2014-07-24
Facebook rejected app integration for publish_actions


Zelex:
Web
For the login dialog, please use the JavaScript SDK without specifying an explicit display type. If you have special requirements precluding use of the SDK, set height and width on your window.open() call to properly size the dialog. People must have the complete option to change their audience settings. Learn more about Facebook Login.

TwizzyDizzy:
Well... that is odd. I just tried to re-submit my application for publish_actions (detailing the above facts) and therefore tested whether it still does not work. Obviously it did work -  as you can see:

  https://www.facebook.com/permalink.php?story_fbid=1528305794106210&id=100007804127198&pnref=story

The facebook app created for my diaspora pod obviously still has no publish_actions permission.

I am somewhat puzzled...

jaywink:
Well they never before requested "publish_actions" - and that has never been a part of our instructions. At some point people started saying new integrations cannot be made without this permission which never before was required. I'm still guessing it's not requested and podmins just assume it's required ;)

TwizzyDizzy:
I, in fact, assumed that we need that permission, obviously because the post I made back then had not been posted to facebook.

So we're basically back to Post #1, because the error itself is still displayed (I just verified that).

I am kind of embarassed for making such a fuss.

jaywink:
I hope you were wrong and no such permissions are required ;) And no worries, no need to be embarrased about anything. :dancer: These things happen all the time..

@Zelex would you like to try crossposting *without* this permission, ie [strictly following the guide](https://wiki.diasporafoundation.org/Integrating_other_social_networks#Facebook) in our wiki and just 1) creating the app, 2) adding keys to diaspora config, 3) connecting diaspora account to facebook account and 4) crossposting?

Flaburgan:
@jaywink I think we need the permission, the facebook API changed and is now the 2.0, the 1.0 is not working anymore. Btw I'm not even sure that we are compatible with the 2.0 API see https://github.com/diaspora/diaspora/issues/5260

jaywink:
Well I just went through the effort of creating a pod, creating a facebook app and following our wiki to make the integration - and everything worked smoothly - *there was at no point any need to submit anything for review*.

![](http://i.imgur.com/yRMHlqa.png)
![](http://i.imgur.com/QHbYmbP.png)

So closing this unless someone can prove that our current documentation is somehow out of date and publish_actions needs to be requested.

qtheuret:
[...]

PR #5924, opened 2015-05-04
bump oauth related dependencies

jhass:
This is less of a PR as more of an issue related to #3855, #5411, #5889,  #5923 etc.

The task here is to validate that the crossposting functionality is still working, for Facebook that means leaving #5085 and #5260 unsolved would be okay, but publishing to your own profile should still work.

If there are no additional code changes required, just confirming here that it's still working is all that's needed, I'm just too annoyed with dealing with those APIs myself to do the verification work on my own.


jaywink:
I've pulled these gem changes to my pod and can verify cross-posting works to Twitter as before and after fixing Facebook scopes #6019 also that works for myself.

Not sure if related, but username doesn't show like before when authorizing with Facebook. Maybe that is related to the new API instead, or incomplete permissions. Anyhow, posting works in this single user case at least.

Also checked Tumblr, works.

Is there anything else to verify with this before merging these changes in? I'm doubt these gem changes actually fix anything, but it's good to have them updated of course.

jhass:
If you say it's good it's good. Tbh. I pretty much stopped caring about the services stuff personally.

denschub:
Merged as ea63343b6942597199b60f399263dc4517bce836. We'll see if we break anything.

PR #5936, opened 2015-05-07
Pass along current_user in ContactPresenter#full_hash_with_person

svbergerem:


denschub:
Merged as 05ec235ea32cc8e30673ddb3d5521c104895418c, thank you!

Flaburgan:
Just to better understand the code, how is that useful?

svbergerem:
Before the relationship status was always "false" when we called the `PersonPresenter` from the `ContactPresenter` (because `current_user` wasn't present). We never used that information for contacts but nevertheless that information was incomplete.

PR #5938, opened 2015-05-07
Suppress some more exceptions in workers

skyschub:
Based on a production-analysis I have done on Geraspora, looking at the top reasons for jobs to fail. Suppressing those will reduce the amount of retries.
- `Diaspora::PostNotFetchable` is raised by the new fetchers trying to fetch missing posts. This is already a failure prevention method and there is no reason to retry fetching posts here.
- `ActiveRecord::RecordNotUnique` seems to be a frequent exception caused by Friendica nodes doing multiple pushes of posts/comments/likes. I analyzed over 3100 samples collected by Sentry and I was not able to find any actual GUID collisions, all incoming contents matched the ones that already were stored in the database. Since our GUID has a pretty high entropy, I feel like it is safe to mute those.

The filter for `ActiveRecord::RecordNotUnique` contains six cases, three for PostgreSQL and three for MySQL. As far as I can tell, there is unfortunately no nicer way to check the actual reason for collision.

### DIFF
--- a/app/workers/base.rb
+++ b/app/workers/base.rb
@@ -14,7 +14,10 @@ module Workers
     rescue Diaspora::ContactRequiredUnlessRequest,
            Diaspora::RelayableObjectWithoutParent,
            # Friendica seems to provoke these
-           Diaspora::AuthorXMLAuthorMismatch => e
+           Diaspora::AuthorXMLAuthorMismatch,
+           # We received a private object to our public endpoint, again something
+           # Friendica seems to provoke
+           Diaspora::NonPublic => e
       Rails.logger.info("error on receive: #{e.class}")
     rescue ActiveRecord::RecordInvalid => e
       Rails.logger.info("failed to save received object: #{e.record.errors.full_messages}")
--- a/app/workers/base.rb
+++ b/app/workers/base.rb
@@ -13,12 +13,22 @@ module Workers
       yield
     rescue Diaspora::ContactRequiredUnlessRequest,
            Diaspora::RelayableObjectWithoutParent,
-            # Friendica seems to provoke these
+           # Friendica seems to provoke these
            Diaspora::AuthorXMLAuthorMismatch => e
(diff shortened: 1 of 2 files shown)


jhass:
`RecordNotUnique` is fine by me, however `PostNotFetchable` might happen for private posts which parent we didn't receive yet, say we got the parent and the relayable pushed to the queue at the same time or shortly after another, which one is processed first is not deterministic.

Flaburgan:
`Diaspora::PostNotFetchable`: If I understand correctly the way diaspora* deals with that actually, if it receives a comment for a post it didn't receive, it kept the comment in the sidekiq queue and wait to try again later, hoping that the parent will be received in the same time.

Why don't we simply store the comment in the DB even if we don't have the parent? If we receive the parent, then the comment is already there and immediately available without overloading sidekiq, if the parent is never receive (worst case), then we simply have an useless row in the DB. This is not that bad, and maybe we can also add a check like if the comment is one month old and still orphan then the parent will never be received and we delete it.

denschub:
@jhass You are right. I will implement nicer error handling for `Diaspora::PostNotFetchable`, but disabling all retries here is not a solution, indeed. I haven't thought about that very well. :)

I removed muting `Diaspora::PostNotFetchable` now, however, I muted exceptions if a private object gets send to the public enpoint (931 samples from all Friendica hosts in my tests) since there is no way of fixing this by retrying the job.

@Flaburgan Just saving invalid data into the database is not an option and even impossible in this case. Comments do have the root_id in their data row, which we need to look up first (it's not the GUID, it's the databases auto incremented ID) and we're not able to do that if the parent is not there.

jhass:
Merged as 6273c2959ab044bbf00214b8c4727875323f91d7 915dee3c4b9ec0b3f8413cf00b8251116143e3fd

Thank you!

Flaburgan:
@denschub I always wondered why we have a local id, the global one is not enough?

PR #5941, opened 2015-05-08
Mobile: add button to follow/unfollow tags

svbergerem:
and refactor js-routes because of camelCase errors in jshint.

It looks like one needs to run `bin/rake tmp:cache:clear` to be able to use the new js-routes paths.

### DIFF
--- a/app/assets/javascripts/app/views/help_view.js
+++ b/app/assets/javascripts/app/views/help_view.js
@@ -40,7 +40,7 @@ app.views.Help = app.views.StaticContentView.extend({
     this.CHAT_SUBS = {
       add_contact_roster_a: {
         toggle_privilege: this.getChatIcons(),
-        contacts_page: this.linkHtml(Routes.contacts_path(), Diaspora.I18n.t('chat.contacts_page'))
+        contacts_page: this.linkHtml(Routes.contacts(), Diaspora.I18n.t("chat.contacts_page"))
       }
     };
 
--- a/spec/javascripts/app/models/person_spec.js
+++ b/spec/javascripts/app/models/person_spec.js
@@ -1,69 +1,69 @@
 
 describe("app.models.Person", function() {
   beforeEach(function() {
-    this.mutual_contact = factory.person({relationship: 'mutual'});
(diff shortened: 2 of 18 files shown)


app/assets/javascripts/app/views/help_view.js
```diff
@@ -40,7 +40,7 @@ app.views.Help = app.views.StaticContentView.extend({
     this.CHAT_SUBS = {
       add_contact_roster_a: {
         toggle_privilege: this.getChatIcons(),
-        contacts_page: this.linkHtml(Routes.contacts_path(), Diaspora.I18n.t('chat.contacts_page'))
+        contacts_page: this.linkHtml(Routes.contacts(), Diaspora.I18n.t("chat.contacts_page"))
```
svbergerem: This is part of the translations, so I guess I won't fix it in this PR.

spec/javascripts/app/models/person_spec.js
```diff
@@ -1,69 +1,69 @@
...
 
   context("#block", function() {
     it("POSTs a block to the server", function() {
-      this.sharing_contact.block();
+      this.sharingContact.block();
       var request = jasmine.Ajax.requests.mostRecent();
 
       expect(request.method).toEqual("POST");
-      expect($.parseJSON(request.params).block.person_id).toEqual(this.sharing_contact.id);
+      expect($.parseJSON(request.params).block.person_id).toEqual(this.sharingContact.id);
```
svbergerem: This is not in camel case because it is data which is sent to the server. Again: won't fix.

svbergerem:
This PR should be ready to review and merge.

mobile_tag_follow
mobile_tag_stop_following

app/assets/javascripts/app/views/help_view.js
jhass: That it detects that in a string at all seems like a bug...

svbergerem: "contacts_page" occurs two times in that line.

denschub:
Merged, thanks!

PR #5943, opened 2015-05-10
Mute ActiveRecord::RecordInvalid due to ignores

skyschub:
Follow up for #5938 because I missed a `ActiveRecord::RecordInvalid` and @jhass was to faster with the merge. Also contains a commit to change the message to get logged for `ActiveRecord::RecordNotUnique` since we usually do not have `.records` there.


jhass:
Merged as 1a4e3cef4860d5b46adf959bbbffb868dcb9a737 32f95a860c4e47539c8feb909a3c7c696d1ae7ef

Thanks!

PR #5953, opened 2015-05-12
Fix inactive user removal not respecting configuration for daily limits

jaywink:
It seems ActiveRecord ignores .limit() if .find_each() is used to iterate the query (http://stackoverflow.com/a/6680541/1489738). Using .each() instead. Added a test.


jaywink:
Umm is hound stuck?

denschub:
Merged as aec2467e9895a4e70240274d1174f10218113d8d, thank you very much!

PR #5962, opened 2015-05-16
fix empty searchable in hcard, parse empty as false

SuperTux88:


jhass:
Merged as 30cc330747958c7111e7d1d4d4887234a42a91be

Thank you!

PR #5967, opened 2015-05-17
Use case insensitive unconfirmed email

### LINKED ISSUE #2569: Email should be stored with the user-entered case preserved, and then matched case-insensitive
email of some users are not downcased, so they couldn't reset the password.
just checked it, and emails are not downcased.

you could check in your db with this query:

     SELECT distinct email FROM diaspora.users
     WHERE (SUBSTRING(email, 1, 1) COLLATE utf8_bin) = UPPER(SUBSTRING(email, 1, 1));
(igonore thoose with numbers at the beginning)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/114657-email-should-be-stored-with-the-user-entered-case-preserved-and-then-matched-case-insensitive?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #2569.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -56,7 +56,7 @@ class UsersController < ApplicationController
       elsif u[:email]
         @user.unconfirmed_email = u[:email]
         if @user.save
-          @user.mail_confirm_email == @user.email
+          @user.send_confirm_email
           if @user.unconfirmed_email
             flash[:notice] = I18n.t 'users.update.unconfirmed_email_changed'
           end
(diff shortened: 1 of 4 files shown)


denschub:
Merged as ceec5084af803c3c4b17178b581d1248c377dfb8, nice catch!

PR #5968, opened 2015-05-18
improved handling of reshares with deleted root-post

SuperTux88:


app/assets/templates/reshare_tpl.jst.hbs
```diff
@@ -37,10 +35,11 @@
...
   {{else}}
-
-    <p>
-      {{t "stream.original_post_deleted"}}
-    </p>
+    <div class="media">
+      <p>
+        {{t "stream.original_post_deleted"}}
+      </p>
+    </div>
```
jhass: I'd say instead of duplicating the tag, just pull it outside the if/else.

app/models/reshare.rb
```diff
@@ -26,7 +26,7 @@ class Reshare < Post
   end
 
   def root_diaspora_id
-    self.root.author.diaspora_handle
+    root ? root.author.diaspora_handle : nil
```
jhass: `root.try(:author).try(:diaspora_handle)`, I sense we need null objects...

denschub:
Merged as 289753e068bd351f1916d4ca280ed68cba38ad2e cde9b24476b3066ac7084b0c7163940b62feff71 267a1df3c46fd4a7fcb1d749d8bbe2d1c582d91f. Thanks much.

PR #5974, opened 2015-05-19
Update listen configuration

jhass:
- Drop deprecated port setting
- Listen to unix socket by default

### DIFF
--- a/config/defaults.yml
+++ b/config/defaults.yml
@@ -192,7 +192,7 @@ development:
     autofollow_on_join: false
     autofollow_on_join_user: ''
 production:
-  i_am_a_dummy: # Remove if you add an actual override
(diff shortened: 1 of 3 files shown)


denschub:
Did you check for any outdated documentation in the wiki?

jhass:
Nope, I'll just update it as people complain ;P

denschub:
Sounds good. Thanks for your PR!

Flaburgan:
Does that imply any manual change if the default setting was not modified? Looks like not, but I want to be sure.

jhass:
Yes it does, the develop branch in production mode will no longer listen to port 3000 anymore by default, but to `tmp/diaspora.sock`.

PR #5975, opened 2015-05-20
Logger refactoring

SuperTux88:
- replace silence_assets.rb with gem
- use logging gem for better logging
  - timestamp, log-level and class in the log
  - basic logrotate


config/logging.rb
```diff
@@ -0,0 +1,85 @@
...
+  # The rolling file appender uses shared file locks to ensure that only one
+  # process will roll the log file. Each process writing to the file must have
+  # its own open file descriptor for `flock` to function properly. Reopening
+  # the file descriptors after forking ensures that each worker has a unique
+  # file descriptor.
+  if defined? PhusionPassenger
+    PhusionPassenger.on_event(:starting_worker_process) do |forked|
+      Logging.reopen if forked
+    end
+  end
```
jhass: So this needs to happen in unicorns after_fork hook too, no?

SuperTux88: I'm not 100% sure, but it doesn't hurt :)

denschub:
This is quite useful for production setups to debug issues, so I'm merging this into 0.5.1.0.

Merged as be3b2852be9788c0cfafe0576fafff391c1be0e3 ac96232d85e52282803343f89d5968388ca59f6f 49f0fb6dcd8058587700529f2045b94e73402b34

Thanks much! :cookie:

PR #5978, opened 2015-05-22
use id from preloaded gon.post to fix broken liking with guid

SuperTux88:


denschub:
Very nice catch, merged as 68f7208ff5e30e1870d7234efe0316af9db6793d, thank you very much!

PR #5983, opened 2015-05-23
gracefully handle when a like is already deleted again

SuperTux88:
also:
- fail when sender or recipient is not found (should never happen)
- better logging

### DIFF
--- a/app/mailers/notification_mailers/base.rb
+++ b/app/mailers/notification_mailers/base.rb
@@ -8,8 +8,8 @@ module NotificationMailers
 
     def initialize(recipient_id, sender_id=nil, *args)
       @headers = {}
-      @recipient = User.find_by_id(recipient_id)
-      @sender = Person.find_by_id(sender_id) if sender_id.present?
+      @recipient = User.find(recipient_id)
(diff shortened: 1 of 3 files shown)


jhass:
Could you try to add a test please?

SuperTux88:
tests added

app/mailers/notification_mailers/base.rb
```diff
@@ -29,11 +29,12 @@ def name_and_address(name, email)
...
     private
+
     def default_headers
       headers = {
-        :from => AppConfig.mail.sender_address.get,
-        :host => "#{AppConfig.pod_uri.host}",
-        :to => name_and_address(@recipient.name, @recipient.email)
+        from: AppConfig.mail.sender_address.get,
+        host: "#{AppConfig.pod_uri.host}",
+        to:   name_and_address(@recipient.name, @recipient.email)
```
jhass: @salbertson Hound seems to ignore our [rubocop.yml](https://github.com/diaspora/diaspora/blob/develop/.rubocop.yml#L34) here.

denschub:
Merged as 8531b160a6c8407749411d44ac6bf72ca40408c1, :cookie:

PR #5985, opened 2015-05-23
Introduce 'authorized' configuration option for services

jaywink:
Since the Facebook API has changed and additional permissions are required for all users on a pod to cross-post, an additional 'authorized' flag is needed to be set for the Facebook service.
This flag allows either all users, one user or no users to use the cross-posting service.

Clarifies the situation for #5923, #5260 and #5085.

### DIFF
--- a/lib/configuration_methods.rb
+++ b/lib/configuration_methods.rb
@@ -34,6 +34,13 @@ module Configuration
     end
     attr_writer :configured_services
 
+    def show_service?(service, user)
+      return false unless self["services.#{service}.enable"]
+      # Return true only if 'authorized' is true or equal to user username
+      (user && self["services.#{service}.authorized"] == user.username) ||
+        self["services.#{service}.authorized"] == true
+    end
+
     def secret_token
       if heroku?
         return ENV['SECRET_TOKEN'] if ENV['SECRET_TOKEN']
--- a/app/presenters/statistics_presenter.rb
+++ b/app/presenters/statistics_presenter.rb
@@ -98,7 +98,7 @@ class StatisticsPresenter
   def all_services_helper
     result = {}
     Configuration::KNOWN_SERVICES.each {|service, options|
-      result[service.to_s] = AppConfig["services.#{service}.enable"]
+      result[service.to_s] = AppConfig.show_service?(service, nil)
     }
     result
   end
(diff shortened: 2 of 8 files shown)


lib/configuration_methods.rb
```diff
@@ -34,6 +34,22 @@ def configured_services
     end
     attr_writer :configured_services
 
+    def show_service?(service, user)
+      unless self["services.#{service}.enable"]
+        return false
+      end
+      # If service has an 'authorized' attr, return true only if it is true or equal to user username
+      if self["services.#{service}"].has_key?("authorized")
```
jhass: Won't work for the environment provider or properly fall through the stack. I really should refuse to return hashes... Anyway, `self["services.#{service}.authorized"].present?` should work I think.

lib/configuration_methods.rb
```diff
@@ -34,6 +34,22 @@ def configured_services
...
+        if self["services.#{service}.authorized"].is_a?(TrueClass)
+          return true
+        end
+        unless user && self["services.#{service}.authorized"] == user.username
+          return false
+        end
+      end
+      true
+    end
```
jhass: I think modifer-if makes this a bit more readable:

```ruby
def show_service?(service, user)
  return false unless self["services.#{service}.enable"]
  # If service has an 'authorized' attr, return true only if it is true or equal to user username
  if self["services.#{service}.authorized"].present?
    return (user && self["services.#{service}.authorized"] == user.username) ||
      self["services.#{service}.authorized"] == true
  end

  true
end
```

lib/configuration_methods.rb
jaywink: I did try all the cases I could see and it seemed to work all throughout, but will change :)

lib/configuration_methods.rb
jaywink: Personally I prefer verbose, it seems clearer to follow the logic, but this is ok, will change.

jaywink:
@denschub While this doesn't strictly fill in the requirements for inclusion into 0.5.1.0, I'd still like to propose this as included. It affects such a large amount of users who are able to connect with Facebook but cannot in reality post anything. This change will force podmins to either set the authorized flag, or disable the service. I can include a changelog entry depending on which release this goes in with. What do you think?

This doesn't (afaict) solve the issue where existing connections have been made to Facebook but posting is broken except for the app admin. This is a bit tricky since we can't just have a migration to disconnect users as some pods might be properly authorized. But at least any new connections will not be able to be made unless the app is properly authorized.

denschub:
Since this was accepted on Loomio, thanks for implementing that.

I have no objections merging it into 0.5.1.0 since it does not break anything if the podmin does not change their configuration (Facebook is broken anyway), so merged as 7c0e50c2c270e696b7dbd6540d00b4cc38e25375. Thanks!

PR #5988, opened 2015-05-24
wrap the sidekiq logger to add the context info again

SuperTux88:
add the sidekiq context info to the logging-gem logger


denschub:
No objections from anyone so far and I have no idea for a solution without using the `SimpleDelegator`, so merged as b491ecac31b45a1da9f24b10d300c76f0f77355f. Thanks!

PR #5989, opened 2015-05-25
split eye output in 2 logfiles and forward stdout to terminal in dev mode

SuperTux88:


denschub:
Merged as d4312fbaba31843e423942b7d9f0c8bdeb7fe981, thanks!

PR #5991, opened 2015-05-26
handle when the diaspora xml parser returns nil

SuperTux88:


jhass:
So you want retries for this? Else please catch the exception somewhere

SuperTux88:
I catch the exception now in Workers::Base

denschub:
Merged as 21ae93e6583850cee2ac9581ddcf28cbac9298eb, :cookie:, thanks

PR #5992, opened 2015-05-26
Do not mute ActiveRecord::RecordNotUnique for now

### LINKED ISSUE #5971: Some private objects are not received
This is a regression from #5938, more precise from muting `ActiveRecord::RecordNotUnique`. Some private objects gets received but don't get share visibilities for all recipients. Some visibilities seems to get added when retrying a `Postzord::Receiver::Private` previously failed because of a non-unique GUID.

This is mainly a tracking bug and to block the release, but feel free to add any thoughts if you have any.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15834814-some-private-objects-are-not-received?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skyschub:
fixes #5971, regression from #5938

### DIFF
--- a/app/workers/base.rb
+++ b/app/workers/base.rb
@@ -25,16 +25,6 @@ module Workers
         "already been taken"
         "is ignored by the post author"
       ).any? {|reason| e.message.include? reason }
-    rescue ActiveRecord::RecordNotUnique => e
-      Rails.logger.info("failed to save received object: #{e.message}")
-      raise e unless %w(
-        index_comments_on_guid
-        index_likes_on_guid
-        index_posts_on_guid
-        "duplicate key in table 'comments'"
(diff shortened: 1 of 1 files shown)


denschub:
Merged without third-party review as 62e8f87f2c17c21d641ffe577adfbade8336e4b9 as a hotfix. It is not possible to mute the `ActiveRecord::RecordNotUnique` without refactoring large portions of the federation source, which will not happen in a minor release.

PR #5999, opened 2015-05-27
Updated capitalisation in conversation.rb

ghost:
Continuing the correct capitalisation of diaspora\* I have changed 'no subject' to become 'No subject'

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -71,7 +71,7 @@ class Conversation < ActiveRecord::Base
   end
 
   def subject
-    self[:subject].blank? ? "no subject" : self[:subject]
+    self[:subject].blank? ? "No subject" : self[:subject]
   end
 
   def subscribers(user)


jhass:
Oh wow, good catch. Could you extract this to a translation key?

RichKavanagh:
It needs it for sure, but I'm afraid that is way beyond my non-existent RoR skillset :(

jhass:
Hey, it's how  I started! Just take a peek at config/locals/en.yml and call `I18n.t('my.super.awesome.key')`.

denschub:
@RichKavanagh Ping!

RichKavanagh:
@denschub Pong!

denschub:
You are not done with this PR and I get very sad when pull requests are orphaned... Do you mind extracting your translation into actual langauge files as Jonne suggested?

RichKavanagh:
Don't be sad mate, it's Thursday!

This hasn't changed I'm afraid:

> It needs it for sure, but I'm afraid that is way beyond my non-existent RoR skillset :(

So if you're worried about this being 'orphaned', just kill it mate :+1:

denschub:
Nah. That's an excuse I won't accept here. ;)

As Jonne said, just add the key to `config/locales/diaspora/en.yml` (the syntax should be familiar to you) and replace the string with `I18n.t('my.super.awesome.key')`, where `my.super.awesome.key` is the chain of labels in your `yml` file. You can do it!

RichKavanagh:
For some reason I was unable to edit this #5999 PR. So instead I created two new ones as detailed above and I'll close this one :+1:

PR #6001, opened 2015-05-27
Replaces the zipzip gem with the rubyzip gem

strivedi183:
The PR replaces the zipzip gem with the rubyzip gem. There was a time when the selenium-webdriver gem needed the zipzip gem to interface with rubyzip < 1 but the current version of selenium-webdriver uses rubyzip ~> 1.

The only place other than from selenium-webdriver that rubyzip is used is in the user model (User.new.perform_export_photos! - https://github.com/diaspora/diaspora/blob/develop/app/models/user.rb#L328)

Also added an initializer for rubyzip configurations in case anyone wanted or needed the ability to configure it.

Hopefully this helps and let me know if there are any issues with it, thanks!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -173,9 +173,9 @@ gem "rails-timeago",           "2.11.0"
 
 gem "logging-rails", "0.5.0", require: "logging/rails"
 
-# Workarounds
-# https://github.com/rubyzip/rubyzip#important-note
-gem "zip-zip"
+# # Workarounds
+# # https://github.com/rubyzip/rubyzip#important-note
+gem "rubyzip"
 
 # Prevent occasions where minitest is not bundled in
 # packaged versions of ruby. See following issues/prs:
--- a/Gemfile
+++ b/Gemfile
@@ -173,9 +173,9 @@ gem "rails-timeago",           "2.11.0"
 
 gem "logging-rails", "0.5.0", require: "logging/rails"
 
-# # Workarounds
-# # https://github.com/rubyzip/rubyzip#important-note
-gem "rubyzip"
+# Reading and writing zip files
+
+gem "rubyzip", "1.1.7"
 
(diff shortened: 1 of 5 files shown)


config/initializers/rubyzip.rb
```diff
@@ -0,0 +1,7 @@
+# https://github.com/rubyzip/rubyzip#configuration
+# Zip.setup do |c|
+#   c.on_exists_proc = true
+#   c.continue_on_exists_proc = true
+#   c.unicode_names = true
+#   c.default_compression = Zlib::BEST_COMPRESSION
+# end
```
jhass: I don't think we need this, we can always add it if we do want to configure it ;)

Gemfile
```diff
@@ -173,9 +173,9 @@ gem "rails-timeago",           "2.11.0"
 
 gem "logging-rails", "0.5.0", require: "logging/rails"
 
-# Workarounds
-# https://github.com/rubyzip/rubyzip#important-note
-gem "zip-zip"
+# # Workarounds
+# # https://github.com/rubyzip/rubyzip#important-note
```
jhass: Remove the comments since the no longer apply

Gemfile
```diff
@@ -173,9 +173,9 @@ gem "rails-timeago",           "2.11.0"
...
+gem "rubyzip"
```
jhass: We lock down the version in your Gemfile to protect our users from too much impact when running `bundle update` on their own, please add `, "1.1.7"`

strivedi183:
Thanks @jhass , made the changes, please let me know if there is anything else, thanks again

jhass:
Merged as 36d6c5d2ba2fa71ddad396e19c8c42051cab14ee 2ba2ef1b9c0b3e00c9096710e1c67dabc730b3e2
 
Thank you!

PR #6003, opened 2015-05-28
Replace DB environment variable with optional bundler groups

jhass:
See the changes to the changelog for more details


denschub:
Wheeee! \o/

PR #6005, opened 2015-05-29
Replaces facebox vendored asset with rails gem #5194

strivedi183:
Replaces facebox vendored asset with a rails gem, references issue #5194

Here is a link the rails gem in case anyone is interested - https://github.com/bai/facebox-rails


jhass:
Merged as b026a2afe8d6215c79054be612d91f7aafcb0089

Thank you!

PR #6007, opened 2015-05-29
Only return the current_users participation for post interactions

jhass:
Nobody needs to know who else participated.

### DIFF
--- a/app/presenters/post_interaction_presenter.rb
+++ b/app/presenters/post_interaction_presenter.rb
@@ -1,0 +1,30 @@
+class PostInteractionPresenter
+  def initialize(post, current_user)
+    @post = post
+    @current_user = current_user
+  end
+
+  def as_json(_options={})
+    {
+      likes:          as_api(@post.likes),
+      reshares:       PostPresenter.collection_json(@post.reshares, @current_user),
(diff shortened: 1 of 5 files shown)


app/presenters/post_interaction_presenter.rb
```diff
@@ -0,0 +1,30 @@
...
+    }
+  end
+
+  def participations
+    return @post.participations.none unless @current_user
+    @post.participations.where(author: @current_user.person)
+  end
+
+  def as_api(collection)
+    collection.includes(author: :profile).map {|element|
```
jhass: This check is disabled in our config.

app/presenters/post_interaction_presenter.rb
```diff
@@ -0,0 +1,30 @@
+class PostInteractionPresenter
+  def initialize(post, current_user)
+    @post = post
+    @current_user = current_user
+  end
+
+  def as_json(_options={})
+    {
+      likes:          as_api(@post.likes),
```
jhass: This check is different in our config

Flaburgan:
Thank you!

denschub:
Merged as ac52cef546929ed10640e148dbbb009fbe247b76, muchas gracias.

PR #6014, opened 2015-05-31
Dont save default avatars in the database

SuperTux88:
the avatar-urls from other pod were saved with the digest from the other pod but the domain from the local pod ... -> 404

also: older user have the url without digest in the database, which is no longer generated -> 404

with NULL in the database the default avatar with digest from the local pod is used :)

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -68,7 +68,7 @@ class Profile < ActiveRecord::Base
     (self.person) ? self.person.diaspora_handle : self[:diaspora_handle]
   end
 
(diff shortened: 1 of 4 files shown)


denschub:
Merged as b8a6af2742fa57ea4063dadcfdcd130a78af4f02 03dbb75dd58a410a5579aaecd9611b5cc5a2de7c.

This PR includes a migration, but this one is very, very fast. After some discussion on IRC, we decided to pull it into 0.5.1.0.

Issue #5987, opened 2015-05-24
<hr> color in Markdown should be styled


abimelex:
The `<hr>` element in  class `markdown-content` has currently `border-top: 1px solid #eee;` inherited from bootstrap styling.
So the horizontal line is almost invisible, I would suggest to give it a more strong color, either like border color or text color, since horizontal line is actually text element in markdown content.

EdzeronK:
a track : 

in `app/assets/stylesheets/bootstrap-complete.scss`,  line 2: 

 @import 'bootstrap';
 
the problem is in a gem 'bootstrap' ?

jhass:
There's no problem, the issue simply asks for a prettier override for those used in posts.

PR #6016, opened 2015-05-31
Change markdown-content hr style #5987

KentShikama:
The hr tag in the markdown-content was #eee,
and thus almost invisible. It has been changed to
$border-gray (currently #ddd) to be consistent
with the other hr tags.

### DIFF
--- a/app/assets/stylesheets/stream_element.scss
+++ b/app/assets/stylesheets/stream_element.scss
@@ -111,9 +111,8 @@
     position: relative;
     .markdown-content hr {
       width: 85%;
-      margin: 0.8em auto 0.8em auto;
+      margin: 0.8em auto;
       border-top: 1px solid $border-grey;
-      border-bottom: 0px;
     }
     .expander {
       position: absolute;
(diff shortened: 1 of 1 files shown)


jhass:
I think I'd like it a bit more distinct from our usual borders surrounding the post, how about something like a width of 85% and centering it?

theworldbright:
What reason is there for having no margin after the last paragraph element in a post?

#main_stream .stream_element > .media .post-content p:last-of-type {
  margin-bottom: 0;
}

jhass:
Because it looks better?

theworldbright:
Yeah I guessed as much. I originally had a problem with the way the vertical margins after last paragraph elements were working because of that specific style but I found a way around it.

The following markup

```
# Heading of post
<hr>
The title and this line have been separated with a hr tag
<hr>
This line and the line above have been separated with a hr tag
<hr>
A longer paragraph text text text text text text text text text text text text text text text text text text text text text text text text text text text text end of paragraph
```
now produces

screen shot 2015-06-01 at 1 06 11 am

jhass:
Merged as 45a5e3d1c5b76b8d83caf3bfec730cdbab7bdae2

Thank you!

PR #6019, opened 2015-05-31
Fix Facebook scopes for cross-posting

jaywink:
Facebook deprecated their v1.0 API and thus the old ones no longer work (for new authorizations).


denschub:
Merged as d2e0b3ceae02987a154d499514c57167a87ae9b3

PR #6020, opened 2015-05-31
Atom feed: set post title length to 50 and use configured pod name in the feed title

svbergerem:
Fixes #5517.

PR #6022, opened 2015-05-31
Truncate post title to 20 characters for notifications

svbergerem:
Before the title could have been 200 characters long if there was a heading in the post. Now the length is consistent.

### DIFF
--- a/spec/models/services/wordpress_spec.rb
+++ b/spec/models/services/wordpress_spec.rb
@@ -1,33 +1,37 @@
-require 'spec_helper'
+require "spec_helper"
 
-describe Services::Wordpress, :type => :model do
-  
+describe Services::Wordpress, type: :model do
   before do
     @user = alice
-    @post = @user.post(:status_message, 
-                       :text => "Hello there. This is a **Wordpress** post that we hope to turn into something else.", 
-                       :to => @user.aspects.first.id)
-                       
-    @service = Services::Wordpress.new(:nickname => "andrew", 
-                                       :access_token => "abc123", 
-                                       :uid => "123")
+    @post = @user.post(:status_message,
+                       text: "Hello there. This is a **Wordpress** post that we hope to turn into something else.",
+                       to:   @user.aspects.first.id)
+
+    @service = Services::Wordpress.new(nickname:     "andrew",
+                                       access_token: "abc123",
+                                       uid:          "123")
     @user.services << @service
   end
-  
+
   describe "#post" do
-    it 'posts a status message to wordpress' do
-      stub_request(:post, "https://public-api.wordpress.com/rest/v1/sites/123/posts/new").to_return(:status => 200, :body => {:ID => 68}.to_json, :headers => {})
+    it "posts a status message to wordpress" do
+      stub_request(:post, "https://public-api.wordpress.com/rest/v1/sites/123/posts/new").to_return(
+        status:  200,
+        body:    {ID: 68}.to_json,
+        headers: {}
+      )
       @service.post(@post)
     end
   end
-  
+
   describe "#post_body" do
     it "truncates content for use in title" do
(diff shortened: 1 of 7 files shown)


Flaburgan:
IMO the 20 limit is too short, what about changing it to at least 30? (for both the title defined by the # and when we're taking the beginning of the post)

svbergerem:
Currently we have the following limits:
* [default](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/message_renderer.rb#L235): 20, for headings 200
* [wordpress](https://github.com/diaspora/diaspora/blob/develop/app/models/services/wordpress.rb#L25): 40
* [subject for emails](https://github.com/diaspora/diaspora/blob/develop/app/models/status_message.rb#L131): 70
* [open graph](https://github.com/diaspora/diaspora/blob/develop/app/helpers/open_graph_helper.rb#L59): 140
* atom feeds: 50 (after merging #6020)

Maybe we should decide on some sane defaults and use those whenever possible.

jaywink:
I tend to agree, 20 is very short. What about 70 like email subjects?

svbergerem:
20:
post_title_20
40:
post_title_40
50:
post_title_50
70:
post_title_70

denschub:
+1 for 70.

svbergerem:
These are the lines I found where we use the `title` method from the `MessageRenderer`
```bash
$ grep -Rin "message.title" app lib
app/helpers/posts_helper.rb:13:        post.message.title opts
app/models/services/wordpress.rb:25:     title: post.message.title(length: 40),
app/models/status_message.rb:131:    message.title length: 70
app/presenters/post_presenter.rb:57:    @post.message.present? ? @post.message.title : I18n.t('posts.presenter.title', name: @post.author_name)
app/views/users/public.atom.builder:33:      entry.title post.message.title(length: 50)

$ grep -Rin "post_page_title" app lib
app/helpers/posts_helper.rb:6:  def post_page_title(post, opts={})
app/helpers/report_helper.rb:8:      raw t('report.post_label', title: link_to(post_page_title(post), post_path(id)))
app/helpers/open_graph_helper.rb:59:        og_title(post_page_title(post, :length => 140)),
app/helpers/notifications_helper.rb:14:                    post_link: link_to(post_page_title(post, length: 20), post_path(post)).html_safe)
app/helpers/notifications_helper.rb:24:                    post_link:   link_to(post_page_title(post, length: 20),
app/presenters/o_embed_presenter.rb:33:    post_page_title(@post)
app/views/posts/show.html.haml:6:  = post_page_title @post
```

I'd propose the following:
* For the `MessageRenderer` set the default length to 70 and always truncate the title. (Even if there is some heading defined in the post)
* Keep 140 for OpenGraph, use the default everywhere else (wordpress, email, post presenter, atom feed, post reports, notifications, oEmbed, ...)

Does that sound ok to you? @jhass @denschub @jaywink

Flaburgan:
Yeah sounds fine.

jhass:
Yup, sounds fine.

spec/models/services/wordpress_spec.rb
```diff
@@ -1,33 +1,35 @@
...
-    @post = @user.post(:status_message, 
-                       :text => "Hello there. This is a **Wordpress** post that we hope to turn into something else.", 
+    @post = @user.post(:status_message,
+                       :text => "Hello there. This is a **Wordpress** post that we hope to turn into something else.",
                        :to => @user.aspects.first.id)
-                       
-    @service = Services::Wordpress.new(:nickname => "andrew", 
-                                       :access_token => "abc123", 
+
+    @service = Services::Wordpress.new(:nickname => "andrew",
```
jhass: Ruby 1.9 hash syntax.

svbergerem:
This PR should be ready to review.

jhass:
Merged as 1b77258d0e1486c8d4a8f5e4ee9fb327963e5822

Thanks!

PR #6015, opened 2015-05-31
Port to Bootstrap 3

jhass:
This is the continuation for #6002 

I'm not going to work on this personally, instead it will be used to track progress towards merging this to develop. Please see #6002 for the details. To contribute to this please send pull requests against the bootstrap3 branch.

Remaining issues:
- [x] <del>some hounds remarks ([1](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31260555), [2](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31343411), [3](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31343412), [4](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31343413), [5](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31350131), [6](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31350132), [7](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31379537), [8](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31379538), [9](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31379540), [10](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31379541), [11](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31379542), [12](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31381082), [13](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31381083))</del> #6036
- [x] <del>code refactor ([1](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31264375), <del>[2](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31264501)</del>, <del>[3](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31385096)</del>, <del>[4](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31385137)</del>)</del> #6024 #6043 #6046
- [x] <del>hash syntax (<del>[1](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31385150)</del>, <del>[2](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31385153)</del>, [3](https://github.com/diaspora/diaspora/pull/6002#discussion-diff-31385172))</del> #6037
[...]

PR #6024, opened 2015-06-01
#6015 - Fix inline css 1

KentShikama:


jhass:
Thanks!

app/views/conversations/index.mobile.haml
```diff
@@ -20,11 +20,7 @@
       = render partial: 'conversations/conversation', collection: @conversations,
           locals: { authors: @authors,  unread_counts: @unread_counts }
     - else
-      %br
-      %br
-      %br
-      %br
-      %div{ style: 'text-align:center;' }
+      %div{ class: 'no-messages' }
```
Flaburgan: This can be replaced by `.no-messages`: the default element is a `<div>`.

theworldbright: Ahh yes :+1:

PR #6025, opened 2015-06-01
Vertically align mail icon with text

KentShikama:
It now looks as follows:

screen shot 2015-06-01 at 11 48 13 pm

### DIFF
--- a/app/assets/stylesheets/new_styles/_forms.scss
+++ b/app/assets/stylesheets/new_styles/_forms.scss
@@ -86,11 +86,10 @@ form.block-form {
       text-align: center;
       color: $text-grey;
       font-size: 20px;
-    }
-
-    .entypo.mail {
-      top: 7px;
-      font-size: 19px;
+      &.mail {
+        font-size: 19px;
+        top: 7px;
+      }
     }
 
     .entypo:nth-of-type(2) {
--- a/app/assets/stylesheets/new_styles/_forms.scss
+++ b/app/assets/stylesheets/new_styles/_forms.scss
@@ -80,24 +80,29 @@ form.block-form {
 
     .entypo {
       position: absolute;
-      top: 7px;
-      left: 14px;
+      top: 10px;
+      left: 10px;
       width: 20px;
       text-align: center;
       color: $text-grey;
+      font-size: 20px;
+    }
+
+    .entypo.mail {
+      top: 7px;
       font-size: 19px;
     }
 
     .entypo:nth-of-type(2) {
-      top: 50px;
+      top: 47px;
     }
 
     .entypo:nth-of-type(3) {
-      top: 90px;
+      top: 87px;
     }
 
     .entypo:nth-of-type(4) {
-      top: 130px;
+      top: 127px;
     }
 
     ::placeholder { text-transform: uppercase; }
--- a/app/assets/stylesheets/new_styles/_forms.scss
+++ b/app/assets/stylesheets/new_styles/_forms.scss
@@ -80,12 +80,12 @@ form.block-form {
 
     .entypo {
       position: absolute;
-      top: 10px;
-      left: 10px;
+      top: 7px;
+      left: 14px;
       width: 20px;
       text-align: center;
       color: $text-grey;
-      font-size: 20px;
+      font-size: 19px;
     }
 
     .entypo:nth-of-type(2) {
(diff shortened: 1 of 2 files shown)


jhass:
Uh, that affects none of the other icons?

AugierLe42e:
Meh... I think it does. Completely unverified, but I think it will mess with the trash icon at least.

theworldbright:
Let me check. I just realized that that scss styling isn't as isolated as I thought.

AugierLe42e:
Haha, nope, it isn't. Actually, here, you have modified the styling of the entypo icons for the whole application :p
If I'm not wrong, SCSS in `new_styles` directory define rules for the whole application like button styling.

theworldbright:
Here is what the forgot your password page looks like:

screen shot 2015-06-02 at 1 28 23 am

Here is what the login page looks like:

screen shot 2015-06-02 at 1 30 12 am

Here is what the signup page looks like:

screen shot 2015-06-02 at 1 32 01 am

theworldbright:
I searched all occurrences of '.entypo.mail' in the project. There were three results, and the signup page and the forgot your password page were the only two places where the entypo mail icon was used in a block-form > fieldset.

jhass:
I guess we can live with that, but if you want it perfect I think either the vertical or the bottom baseline should match.... :P

Current: 
Your screenshot:

app/assets/stylesheets/new_styles/_forms.scss
```diff
@@ -86,18 +86,22 @@ form.block-form {
...
-      top: 50px;
+      top: 47px;
     }
 
     .entypo:nth-of-type(3) {
-      top: 90px;
+      top: 87px;
     }
 
     .entypo:nth-of-type(4) {
```
AugierLe42e: Oh crap, that's so an ugly selector :/
What does it corresponds to ?

Flaburgan: The fourth item. Icons are placed from the top so it's the cleanest solution I found.

AugierLe42e: K, I trust you :p
As I said, I suck at JS :D

theworldbright:
I find it weird that the username icon is out of place. I didn't make any changes in my code that would affect that icon. It seems like in the develop branch both icons are aligned;  however, in the latest bootstrap3 branch (without my changes in these PR), the icons are slightly lower than they should be. Furthermore, I did a diff across the two branches and there has been no changes to the _form.scss. I will  continue to look but if anyone can give me a tip as to why this is happening, I would be interested to know.

AugierLe42e:
@theworldbright : It is likely possible that BS3 just broke everything. E.g: By default, all div have a null height, on the contrary to BS2, which introduced weird sides effects I had hard time to deal with when working on this (headache, bitch ! :p)

theworldbright:
@AugierLe42e Thanks for the tip. Yup it seems like there was a change in line height from BS2 to BS3. BS2 uses a line height of 20px while BS3 uses a line height of 1.42857143, causing a shift is all icon heights. For now, I'll just change the line height back to 20px.

AugierLe42e:
> For now, I'll just change the line height back to 20px.

Don't. [A PR has just been merged](https://github.com/diaspora/diaspora/commit/781ca2b76e7c6083978c9cf4a5f47274cb0c9e48) making font size a global Bootstrap variable. I wonder it it could solve your problem. Try to update your branch and see if it changes something.

theworldbright:
Yeah just saw that PR myself.

theworldbright:
I am now tempted to change the following line in bootstrap-variables.scss

```
// $line-height-base:        1.428571429 // 20/14
```

to 20px. Do you know if that would have any major side-effects off the top of your head?

jhass:
No expert either here, but yeah, I'd expect it to affect about everything.

AugierLe42e:
That could be a good idea. 1.428571429 seems to be 20/14, that is `baseline height / font size`. As our font size is 13px and not 14, that would be compilant

theworldbright:
Did some reading and it seems like using the "unitless" line height is recommended as it computes line heights based on the font size. It seems like overriding the value locally to 20px does the job, so I'll leave it at that for now.

@AugierLe42e Are you suggesting to change the value to 20/13 ~= 1.5...?

AugierLe42e:
> Are you suggesting to change the value to 20/13 ~= 1.5...?

Yep, I think the unit of this line height variable is in `em` or `rem` that is, a unit which is proportionnal to the body height of the font. So, it is a good practice to choose a size that is related to our font size (13px) than Bootstrap's default (14px).

PR #6027, opened 2015-06-01
Better settings display

ghost:
I think the settings page would fit better with a larger display as some buttons mess up with long traductions. TBH, these settings views are a mess. They'll have to be greatly refactored.

Before:

screen shot 2015-06-01 at 21 08 00
capture d ecran 2015-06-01 a 20 25 32

After:

capture d ecran 2015-06-01 a 21 06 18
capture d ecran 2015-06-01 a 21 06 27
capture d ecran 2015-06-01 a 21 06 51
capture d ecran 2015-06-01 a 21 06 58
capture d ecran 2015-06-01 a 21 07 15
capture d ecran 2015-06-01 a 21 07 27


AugierLe42e:
Crap... One cuke to solve :'(
BTW, it solves inline CSS n°2.

jhass:
Maybe put location and gender next to each other?

jhass:
And I don't think we need to stretch the birthday inputs to the full width.

AugierLe42e:
> Maybe put location and gender next to each other?

Good idea.

> And I don't think we need to stretch the birthday inputs to the full width.

This id Bootstrap default. I don't think we should change such tiny details. It will be harder to maintain if we customize too much.

jhass:
Uh, can't you just put it into a span3 or whatever half the width is?

jhass:
Or what about http://getbootstrap.com/css/#forms-inline ?

AugierLe42e:
Yes, I can. What I'm worried about is the reaction when resizing to very small screens.

jhass:
So what about the inline forms?
> This only applies to forms within viewports that are at least 768px wide.

AugierLe42e:
I don't know. I'm not sure it fits well either. As you see it, the interface is in englush but some translation in french take much more space. That can be a problem and that is the reason I've enlarged the with here.

AugierLe42e:
You can see an exemple on that picture, the button at the bottom:



The translation fits well in english but not in french.

AugierLe42e:
@jhass: Looks better like this, right ? Cleaner !

capture d ecran 2015-06-02 a 20 37 38
capture d ecran 2015-06-02 a 20 37 51
capture d ecran 2015-06-02 a 20 38 11
capture d ecran 2015-06-02 a 20 38 23
capture d ecran 2015-06-02 a 20 38 34

jhass:
Yeah, only nitpick I'd have is that the Cancel button seems a bit lonely, why not stick it right to the left of the other one?

jhass:
Also language select feels inconsistent.

AugierLe42e:
> Yeah, only nitpick I'd have is that the Cancel button seems a bit lonely, why not stick it right to the left of the other one?

Seems reasonnable. Let's give a try.

> Also language select feels inconsistent.

What do you mean ?

jhass:
Well, all the inputs drift to the right but that language select is stuck to the left.

AugierLe42e:
Yeah... I don't know... It'd feel weird to have all the elements (the input and the button) on the right. The ohter elements are on the right because they have a label on the left. Here what I can propose : the input and the button on the same line, like the *Cancel* and the *Change password* button just above?

jhass:
Mh yeah, I guess same line would help

AugierLe42e:
That's it. I don't think I'll be able to refacor more than this :p

capture d ecran 2015-06-03 a 00 04 16
capture d ecran 2015-06-03 a 00 04 38
capture d ecran 2015-06-03 a 00 04 49
capture d ecran 2015-06-03 a 00 04 58

Question is: why is Travic so silent ?

app/helpers/people_helper.rb
```diff
@@ -32,7 +32,8 @@ def person_link(person, opts={})
   end
 
   def person_image_tag(person, size = :thumb_small)
-    image_tag(person.profile.image_url(size), :alt => person.name, :class => 'avatar', :title => person.name, 'data-person_id' => person.id)
+    image_tag(person.profile.image_url(size), :alt => person.name, :class => 'avatar img-responsive center-block',
```
jhass: And as always, while touching please upgrade to Ruby 1.9 hash syntax ;)

AugierLe42e: Of course.

jhass:
Alright, if you could just squash together to get rid of those meaningless commit messages (you know `git commit --amend` btw? ;) ) we're good I think.

AugierLe42e:
Nope, I don't know `git comit --amend`. Is she sexy ? :p

denschub:
Dude... please be a bit more serious.

AugierLe42e:
Why so ? You never laugh ?

jhass:
Thanks!

svbergerem:
@AugierLe42e I think this PR broke one cuke: https://travis-ci.org/diaspora/diaspora/jobs/65163504#L2572

AugierLe42e:
@svbergerem Indeed. Nothing crtical, hopefully. I submit a patch tonight.

PR #6030, opened 2015-06-02
Overwrite Bootstrap variables and fix publisher

svbergerem:
- Start using and overwriting Bootstrap variables, set the base font-size to 13px
- Fix mentions in the publisher + some entypo fixes in the publisher
  - before publisher_before
  - after publisher_after
- Fix modal title style
  - before modal_before
  - after modal_after

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,9 +1,12 @@
+/* variables */
+@import 'bootstrap-variables'; // our overwrites
+@import 'bootstrap/variables'; // the original bootstrap ones
+
 @import 'bootstrap-fix';
 
 @import 'perfect-scrollbar';
 
 @import "colors";
-@import 'sizes';
 @import 'mixins';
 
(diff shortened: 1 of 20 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -1,9 +1,12 @@
+/* variables */
+@import 'bootstrap-variables'; // our overwrites
+@import 'bootstrap/variables'; // the original bootstrap ones
```
jhass: Wrong order and thus the duplicate below?

svbergerem: Bootstrap variables [are defined here](https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss) with `!default` so the values will be used from `bootstrap/variables` unless they are defined before.

jhass: Cool, thanks!

Issue #4297, opened 2013-07-12
Own template


max258:
Maybe for someone is it only brainstorm idea, but for me not.
Very usefull can be option use own template.
Not only changing some values in scss or something else, but complex solution.
Now is changing the template very hard and update can destroy new structure.
New folder with templates can be good idea and in confing specifing the template.

efstrian:
We could create fixed color themes for the users to choose, i.e. dark red, dark green, magenta. This way, the ui would be a bit more customizable, but still elegant and true to diaspora's initial design. What do you think about it?

efstrian:
Hi! We are working on implementing a way for the user to opt for one of a few fixed color themes, but we've run into a problem. The header is one of the elements that should change color, and this could be easily done with CSS3 shadows in order to recreate the gradient effect of the current implementation. The problem is that the current logo is not transparent. Any idea how to get past this?

(This feature is still desired, isn't it?)

max258:
@efstrian Good work! I think you can change the header with contributing to git. But not sure if will be this accepted. I think yes, but inform in creators of diaspora*

efstrian:
@max258 No problem if this does not get accepted, we're just trying to help after all. Going from the current png to a CSS background should be both visually impressive and highly configurable, yet one should carefully consider performance issues (if there are any).

We do need a transparent logo to use in the header to try it, though.

Flaburgan:
We are talking about https://github.com/diaspora/diaspora/blob/develop/app/assets/images/branding/header-logo.png

No problem to remove the background of that file (and the @2x version). Just do it and open a pull request.

efstrian:
@Flaburgan, I'm not sure how to remove the background, because of the relatively low resolution. There isn't any transparent version available, is it?

jhass:
We decided to implement color themes and that's done by #6033.

PR #6033, opened 2015-06-02
Issue #4297: Color themes

ghost:
This enables the user to choose a color theme from a given list of themes. The way color themes are specified and made available to the application is similar to the way locales work.

What's missing right now is the transparent header logo (desktop view) needed for the header to display right.

This is how the original theme looks now:

diaspora_original

And below is an example of a dark green color theme.
Desktop:

diaspora_dark_green

Mobile:

diaspora_dark_green_mobile


app/helpers/users_helper.rb
```diff
@@ -6,4 +6,48 @@ def owner_image_tag(size=nil)
...
+  #   current_color_theme # => "color_themes/original"
+  #   # if user Alice has not selected a color theme
+  #   current_color_theme # => "color_themes/original"
+  #   # if user Alice has selected a "magenta" theme
+  #   current_color_theme # => "color_themes/magenta"
+  def current_color_theme
+    if user_signed_in?
+      color_theme = current_user.color_theme
+    end
+    color_theme ||= DEFAULT_COLOR_THEME
```
jhass: How about making this configurable via `AppConfig`? Just move the default to `config/defaults.yml` and add an example in `config/diaspora.yml.example`.

app/helpers/users_helper.rb
```diff
@@ -6,4 +6,48 @@ def owner_image_tag(size=nil)
...
+    current_color_theme = "color_themes/" + color_theme
```
jhass: Please use `"string #{interpolation}"`

app/helpers/users_helper.rb
```diff
@@ -6,4 +6,48 @@ def owner_image_tag(size=nil)
...
+  #   available:
+  #     original: "Original dark"
+  #     dark_green: "Dark green"
+  #   # and AVAILABLE_COLOR_THEMES is accordingly initialized,
+  #   # then:
+  #   available_color_themes
+  #   # => [["Original dark", "original"], ["Dark green", "dark_green"]]
+  def available_color_themes
+    opts = []
+    AVAILABLE_COLOR_THEMES.each do |theme_code, theme_name|
```
jhass: This is a `.map`

app/models/user.rb
```diff
@@ -23,12 +23,14 @@ class User < ActiveRecord::Base
 
   before_validation :strip_and_downcase_username
   before_validation :set_current_language, :on => :create
+  before_validation :set_default_color_theme, :on => :create
```
jhass: Use Ruby 1.9 hash syntax here.

app/models/user.rb
```diff
@@ -198,6 +200,10 @@ def set_current_language
     self.language = I18n.locale.to_s if self.language.blank?
   end
 
+  def set_default_color_theme
+    self.color_theme ||= DEFAULT_COLOR_THEME
+  end
+
```
jhass: This seems redundant to what you do in `setup`?

app/views/layouts/application.html.haml
```diff
@@ -27,6 +27,8 @@
     - if rtl?
       = stylesheet_link_tag :rtl, :media => 'all'
 
+    = stylesheet_link_tag current_color_theme, :media => "all"
```
jhass: Use Ruby 1.9 hash syntax here.

config/initializers/color_themes.rb
```diff
@@ -0,0 +1,18 @@
+require 'pathname'
+
+# Generate the path to the .yml file with the available color themes.
+color_themes_file = Pathname.new(__FILE__).dirname.join('..').expand_path.join("color_themes.yml")
```
jhass: You should have `Rails.root` available here.

app/helpers/users_helper.rb
Flaburgan: @jhass if I get it correctly, this PR would allow each user to set the color theme they want to. So it's not a podmin setting.

jhass: But we could make the default theme if none is set one.

app/models/user.rb
efstrian: Any chance ```create``` is called instead of ```build```?

jhass: Oh well, I guess since we do that for the locale too we can leave figuring that out to another time.

app/helpers/users_helper.rb
efstrian: Is is better to keep ```DEFAULT_COLOR_THEME``` inside the model and the controller and set it to ```AppConfig.settings.color_theme``` in the initializer or directly replace it with the setting?

jhass: I think I see no point in the indirection, just replace it directly.

efstrian:
Improved the code a bit, and made the default theme configurable. Anything else you have in mind?

jhass:
I'm sorry but #6015 got in first, so you'll unfortunately run into a lot of conflicts here while rebasing. On the plus side Bootstrap 3 should make the color themes a bit easier, perhaps as easy as setting a couple of its variables and replacing a couple of hardcoded colors in our default theme with them/new variables.

efstrian:
@jhass So what do you propose we do next (instead of solving the conflicts)?

jhass:
I didn't mean you shouldn't even attempt to resolve the conflicts, the backend side is still good and can be pulled in, I only meant that you might consider an approach to actually defining the themes that utilizes Bootstrap 3's powers.

efstrian:
Oh, I'm sorry, I wanted to say 'except for' not 'instead of'. Sorry... We DO solve the conflicts. We'll have a look at the changes made and see what we can do.

efstrian:
There weren't many conflicts to resolve, actually. We've avoided to integrate color theme logic in the main SCSS so that creating a color theme is very flexible. The other option was to set variables in place of the hardcoded colors in the original SCSS files (for example, header.scss), but this would mean the whole thing would be less flexible and customizable, wouldn't it?

@jhass If you have something in mind, tell us, we'll do it!

P.S. Does porting to Bootstrap 3 mean you're going flat?

jhass:
Are you sure you rebased against the right branch? Because the PR is still not in a merge conflict free state :)

Regarding your other points, I guess somebody from the design team needs to say something about that, /cc @pablocubico @svbergerem and whoever I forgot

svbergerem:
While porting everything to Bootstrap 3 we started to overwrite the [bootstrap variables](https://github.com/diaspora/diaspora/blob/develop/app/assets/stylesheets/bootstrap-variables.scss).

One possible way to handle colors in the future would be the following:
1. Set the base colors in `bootstrap-variables.scss`
2. Define some additional colors in `colors.scss`. Most of those could be based on the existing base colors. (with `darken`, `lighten` etc.)
3. Use those colors for the UI. Don't use any other hardcoded colors.

For a color theme you would only have to overwrite the variables in those two files. (To change the background-color of the navbar you could change [this variable](https://github.com/diaspora/diaspora/blob/develop/app/assets/stylesheets/bootstrap-variables.scss#L369) and we would have to use it in all other SCSS files to define the background of the navbar.)

This of course implies some refactoring of the existing SCSS code.

svbergerem:
> Does porting to Bootstrap 3 mean you're going flat?

Currently on the develop branch: yes. But I don't know if it will stay that way. I'd like to try a diaspora* UI with a flat design but I'm not the only one who decides that. ;)

pablocubico:
@efstrian I think the migration's goal is not to adopt BS3 styles, but to have a solid (and updated) framework to work on top of it. Once it's migrated we'll start experimenting with the UI, and polishing section by section. I'll try to catch up over the weekend to help.

efstrian:
@jhass I have only rebased the branch locally.

@svbergerem Overriding the bootstrap variables means that the theme files cannot be loaded directly in ```application.html.haml```. They should be imported in application.scss instead. To get a theme file there, I have added a line like ```@import "color-theme-variables"``` to the ```application.scss``` and implemented a custom Sass importer to choose which file to import based on which theme a user has selected. I'm not finished yet, but I think it will work. That way, color themes are basically .scss files similar to ```bootstrap-variables.scss```.

What do you think of this implementation? Is there any other/better idea?

jhass:
Will that work in production, where there's no on demand asset compilation? I think I see no issue in generating a combined css file for each theme and switch including that, so like you do it now but that the file just has the entire application CSS.

efstrian:
@jhass I'm not sure if I got you right, do you mean that for each theme there should be a file with both the customized variables and the application CSS?

jhass:
Yes.

efstrian:
This is much better than importing the theme file into the ```application.scss```, and easier to maintain. I tried it the first time and ran into a problem with ```bootstrap-complete.scss```. I'll try it again and check back tomorrow, then.

efstrian:
Are both ```bootstrap-complete.scss``` and ```application.scss``` needed?
Couldn't ```@import "bootstrap-sprockets"``` and ```@import "bootstrap"``` be included in ```application.scss```?

jhass:
Yeah, moving towards a single unified file wouldn't be bad I guess.

efstrian:
What do you think of the changes?

app/views/layouts/application.html.haml
```diff
@@ -20,8 +20,7 @@
     = chartbeat_head_block
     = include_mixpanel
 
-    = include_base_css_framework
```
jhass: Could you drop the helper definition too?

config/diaspora.yml.example
```diff
@@ -533,6 +533,12 @@ configuration: ## Section
     ## If not set your pod will provide a downloadable archive.
     #source_url: 'https://example.org/username/diaspora'
 
+    ## Default color theme
+    ## You can change which color theme is displayed when a user is not signed in
+    ## or has not selected any color theme from the available ones. You simply have
+    ## to enter the name of the theme's .scss file in
+    ## app/assets/stylesheets/color_themes/ without the file extension.
```
jhass: It's a folder now.

config/application.rb
```diff
@@ -87,6 +87,7 @@ class Application < Rails::Application
       mobile/mobile.css
       rtl.css
       home.css
+      color_themes/*.css
```
jhass: Same, nested folders now.

jhass:
Looking good, though I do wonder, the overrides in each theme look the same except for the variable values, might it be worth to extract a standard `_color_theme_override.scss` and import it into the theme after the variable definitions?

jhass:
Oh and looks like the Jasmine suite's setup needs to be fixed too https://travis-ci.org/diaspora/diaspora/jobs/65736007#L611

app/views/layouts/application.html.haml
efstrian: Isn't it dropped?

jhass: Yeah, I failed to read the diff even though I checked twice m( Sorry, ignore that.

efstrian:
@jhass You're right, perhaps we could do it only for standard overrides, for example, the background of the header. But what if the author of a theme wants to skip some (not all) of the standard overrides, while also adding his own?
(Sure, I'll check the problem with Jasmine.)

jhass:
Then they shouldn't use the common override since it's not the same anymore ;)

config/application.rb
efstrian: Should I change it ```color_themes/*/*.css```?

jhass: I think so, but I didn't test. Maybe even just `color_themes/*/{desktop,mobile}.css`?

jhass: Oh and I guess `bootstrap-complete.css` above can be removed.

efstrian:
I changed ```application.scss``` to ```_application.scss``` so that it is not precompiled.
Jasmine now complains about two failures in ```spec/javascripts/app/views/publisher_view_spec.js``` and ```spec/javascripts/app/pages/contacts_spec.js```.
It did not display these failures when I had ```application.scss```.
Any idea what that might be?

jhass:
Did you try running `bin/rake tests:generate_fixtures` (again)?

efstrian:
I did, but the failures persisted...

jhass:
Well, push it and we'll see if it's a setup issue :)

efstrian:
It fails here too. https://travis-ci.org/diaspora/diaspora/jobs/65766819#L612

config/application.rb
```diff
@@ -81,12 +81,12 @@ class Application < Rails::Application
       templates.js
       validation.js
 
-      bootstrap-complete.css
       error_pages.css
       admin.css
       mobile/mobile.css
       rtl.css
       home.css
+      color_themes/*/{desktop, mobile}.css
```
jhass: Did you test if `bin/rake assets:precompile` generates the right files (you can get rid of them again with `bin/rake assets:clobber`)? Pretty sure that space here is wrong :)

svbergerem:
@efstrian in spec/javascripts/support/jasmine.yml is a list of stylesheets which are included for the jasmine tests. `default.css` doesn't exist anymore so you can remove it. There are also `bootstrap.css` and `application.css` which you can remove and add `assets/color_themes/original/desktop.css` instead. This fixes the tests on my laptop.

efstrian:
@svbergerem Thanks, this was indeed why the tests failed.
But there is also a problem in ```color_themes/*/{desktop, mobile}.css```, as @jhass noted, but it's not the space. I could not find out a proper globbing pattern, so I did an ugly:

```
color_themes/*/desktop.css
color_themes/*/mobile.css
```
Is there any other idea?

jhass:
It doesn't work without the space either? `{desktop,mobile}` ?

efstrian:
Apparently, it doesn't.

jhass:
Well, then let's stick to your way.

efstrian:
Done!

efstrian:
I've seen [this](https://travis-ci.org/diaspora/diaspora/jobs/65787847#L1564) failure in other builds as well. What is it about? Does it have to do with our commits?

jhass:
No it's another random failure. The issue with these is that none of us can reproduce them locally, at least not reliably enough, so they're very hard to debug.

efstrian:
Ok, so, is there anything else to do here in this PR?

jhass:
No, thank you! We can always improve the themes in smaller follow up PRs ;)

efstrian:
The current implementation supports changes in all of the style rules, so different changes could be also considered. Successfully and easily overriding the bootstrap variables requires considerable refactoring of the current SCSS, so that rules based on the bootstrap variables are not overriden where unnecessary (This way, themes would include mostly variables and not much raw CSS). Themes could be so much more than what they are now.

Thanks, @jhass for all your help and support!
Thank @Flaburgan @svbergerem @pablocubico (and @houndci) as well!

PR #6045, opened 2015-06-04
Use darker version of creation-blue

svbergerem:
**before:**
creation_blue_before
**after:**
creation_blue_after
**before:**
creation_blue2_before
**after:**
creation_blue2_after


Flaburgan:
I like the hover blue in the left side bar:
capture du 2015-06-04 11 35 06

I think it can be used in the .nav-tabs in both notifications page (your first screenshot) and contacts page.

Flaburgan:
Wow, that was a quick merge.

PR #6046, opened 2015-06-04
Use media objects for mobile posts/comments/notifications

svbergerem:
This uses [Bootstrap's media objects](http://getbootstrap.com/components/#media) instead of an ugly grid system hack.

**left: before, right: after**
mobile_notifications
mobile_posts


jhass:
:heart:

Flaburgan:
Nice! Thank you!

PR #6050, opened 2015-06-04
Bootstrap3 various fixes

svbergerem:
- Remove inline css, use font-size-small
- Remove bootstrap-fix.scss
- Use [nav-tabs](http://getbootstrap.com/javascript/#tabs) for ToS:
  - **before:** tos_before
  - **after:** tos_after


jhass:
Thank you!

PR #6053, opened 2015-06-04
Fix pagination issue for mobile view

SansPseudoFix:
Before:

screenshot_2015-06-04-13-35-10

After:

screenshot_2015-06-04-17-30-23


jhass:
Merged as f3938a173e81c23bffa9192f60db6481a172e763

Thank you!

PR #6057, opened 2015-06-04
Fixed variable drawer width

ghost:
That's the correct way to handle the drawer ;)

### DIFF
--- a/app/assets/stylesheets/mobile/header.scss
+++ b/app/assets/stylesheets/mobile/header.scss
@@ -1,5 +1,8 @@
 /* This file contains the CSS code corresponding to the header and the drawer (including the menu) of the mobile version */
 
+$drawer-width: 400px;
+$drawer-width-offset: $drawer-width + 50px;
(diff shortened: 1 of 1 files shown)


svbergerem:
That doesn't work for browser windows with a width smaller than 400px. I don't know how common that is but I think the mobile version should also work for those resolutions.
320px

AugierLe42e:
Let's say the following, then. I don't think there is any device thiner than 250px. And event then, I don't think we should support them.

svbergerem:
Another idea: Set the drawer width via media query.
max-width: 400px => `$drawer-width: 100%`
else: `$drawer-width: 400px`

Everything else like you did that before. (`transform: translateX(-$drawer-width);`)

AugierLe42e:
Done. `$drawer-width-offset` is to prevent toggle button to be hidden when looking on a 400px screen.

svbergerem:
Thank you!

PR #6059, opened 2015-06-04
Bootstrap3: fix likes avatars and conversation participants

svbergerem:
For conversations the participant avatars (which are visible on hover) were cut off. The avatars for likes were displayed one below another.


jhass:
Thanks!

PR #6068, opened 2015-06-06
gracefully handle duplicate entry while receiving share-visibility parallel

SuperTux88:
refactoring:
- remove unused return-values (were used for caching, which was removed again)
- remove transaction (doesn't help here, added in 2615126)


denschub:
Merged as aa2297a8c0bd557a3b5aa7e562c4452f84c8e4f1, thanks!

PR #6072, opened 2015-06-07
Fix Bootstrap3 regression: reshares in SPV

svbergerem:
### root post

**before:** 
spv_reshare_root_before

**after:**
spv_reshare_root_after
### reshare

**before:**
spv_reshare_reshare_before

**after:**
spv_reshare_reshare_after

### DIFF
--- a/app/assets/stylesheets/single-post-view.scss
+++ b/app/assets/stylesheets/single-post-view.scss
@@ -4,7 +4,6 @@
 
 #single-post-content {
   border-right: solid 1px #cccccc;
-  padding-right: 12px;
   #head {
     padding-bottom: 10px;
     border-bottom: 1px solid $border-grey;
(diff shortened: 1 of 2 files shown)


Flaburgan:
Oh, was working on this one :) The border is still not fine imo but let's merge this and I'll polish it.

svbergerem:
@Flaburgan what is wrong with which border?

Flaburgan:
@svbergerem the horizontal borders: they are not aligned with the avatars in the left, and are touching the vertical border at the right. Currently master is looking like that:

capture du 2015-06-07 13 22 20

svbergerem:
Well, I'm ok with the current borders. If you have a different opinion about that then just open a new PR afterwards and we can discuss over there.

Flaburgan:
Fair enough :)

jhass:
Thanks!

PR #6084, opened 2015-06-08
Update gems

jhass:
#3855


denschub:
Merges as appropriate. Thanks!

Issue #5312, opened 2014-10-08
Error 500 when displaying the services page on the mobile version


Flaburgan:
Steps to reproduce:

Go to the /services page.
Click on the "toggle mobile" link in the footer

```
Started GET "/services" for 83.157.17.132 at 2014-10-08 22:19:51 +0200
Processing by ServicesController#index as HTML
Completed 500 Internal Server Error in 5.8ms

ActionView::MissingTemplate (Missing template services/index, application/index with {:locale=>[:fr, :en], :formats=>[:mobile], :handlers=>[:erb, :builder, :markerb, :haml]}. Searched in:
  * "/home/diaspora/diaspora/app/views"
  * "/home/diaspora/.rvm/gems/ruby-2.0.0-p481@diaspora/gems/rails_admin-0.4.9/app/views"
  * "/home/diaspora/.rvm/gems/ruby-2.0.0-p481@diaspora/gems/kaminari-0.15.1/app/views"
  * "/home/diaspora/.rvm/gems/ruby-2.0.0-p481@diaspora/gems/devise-3.2.4/app/views"
):
  lib/rack/chrome_frame.rb:39:in `call'
  lib/unicorn_killer.rb:35:in `call'
```

Flaburgan:
I'm working on this issue for both privacy and services routes.

svbergerem:
Fixed by #6086.

PR #6086, opened 2015-06-08
Add Services and Privacy settings page to mobile

Flaburgan:


app/views/users/_edit.haml
```diff
@@ -5,242 +5,241 @@
...
+= render 'shared/settings_nav'
+
 .container-fluid
-  = render 'shared/settings_nav'
+  .col-lg-8.col-lg-offset-2
+    .row
+      .col-md-6
+        %h3= t('.your_handle')
 
-  .container-fluid
```
Flaburgan: Git diff is not really powerful here, the only thing I did was to remove a `.container-fluid` and so moving everything two spaces to the left.

Flaburgan:
Wow, no hound remark at all? Wasn't expected.

app/views/profiles/edit.mobile.haml
```diff
@@ -2,13 +2,14 @@
 -# licensed under the Affero General Public License version 3 or later. See
 -# the COPYRIGHT file.
 
-= render "shared/settings_nav"
+.stream_element
```
svbergerem: That class doesn't look correct in this context. Maybe think of a new name and apply the same styles? (Or pick those that are necessary in this case)

Flaburgan: Yeah I agree. I think that we should have two layouts in the mobile version, the pages which contain elements, like the stream, the profile, the notifications or the contacts pages, and the pages are more about text, like the settings pages, the admin panel, the getting started...

The pages of the first group should have the grey background and then a white bg around each elements, but the text pages should directly have a white background, to avoid losing width space and look overloaded.

What do you think about that? It's a bigger work than what I'm doing here, that's why I simply used `.stream_element` as a temporary solution, but I agree, temp can become long, so I can make this one cleaner.

Flaburgan:
This closes https://github.com/diaspora/diaspora/issues/5312 btw

app/views/users/privacy_settings.mobile.haml
```diff
@@ -0,0 +1,41 @@
...
+        = t('.ignored_users')
+
+      - if @blocks.length.zero?
+        = t('.no_user_ignored_message')
+
+      - @blocks.each do |block|
+        = block.person_name
+        \-
+        = link_to t('.stop_ignoring'), block_path(block),
+          method: :delete
```
AugierLe42e: No way to render the desktop view instead ?

app/views/profiles/edit.mobile.haml
Flaburgan: Added a new class.

app/views/users/privacy_settings.mobile.haml
Flaburgan: I could use a partial to do that yeah. What's your opinion everybody? Is this view good enough like this for mobile or should it be improved?

capture d ecran 2015-06-08 a 23 07 59

Flaburgan: I extracted it in a partial.

Flaburgan:
Rebased, this is ready to review.

svbergerem:
I'll review this tonight. (Unless someone else is faster)

app/views/users/_edit.haml
```diff
@@ -5,242 +5,241 @@
...
-            %div= t('.email_awaiting_confirmation', email: @user.email, :unconfirmed_email => @user.unconfirmed_email)
+              .clearfix= f.label :close_account_password, t("users.edit.current_password"), for: :close_account_password,
+                      class: "pull-right", style: "padding-top: 9px;"
+              .small-horizontal-spacer
+              .clearfix= f.password_field :current_password, id: :close_account_password, class: "form-control"
+              %p
+              .clearfix= f.submit t('users.edit.close_account_text'), class: "btn btn-danger pull-right",
+                      id: "close_account_confirm", data: { confirm: t("are_you_sure_delete_account") }
+
+        - if not mobile
```
svbergerem: Shouldn't this be just `- else`?

app/views/users/_edit.haml
svbergerem: This adds some inconsistency for the desktop views. /user/edit has no surrounding `.container-fluid` anymore while all other settings pages still have one.

app/views/services/index.mobile.haml
```diff
@@ -0,0 +1,15 @@
...
+-#   licensed under the Affero General Public License version 3 or later.  See
+-#   the COPYRIGHT file.
+
+.settings_container.services_page
+  - content_for :page_title do
+    = t('.edit_services')
+
+  = render 'shared/settings_nav'
+
+  .col-md-12
```
svbergerem: This is not the correct usage of Bootstrap's grid system. See http://getbootstrap.com/css/#grid
There is some `.container-fluid` missing. You have the same problem on most other pages.

app/views/users/_edit.haml
Flaburgan: I didn't touch this code at all, I only reindented it that's why it is shown as modified.

Flaburgan:
Hm, you removed the users/_edit.haml file in
https://github.com/diaspora/diaspora/commit/4534ba5de882e42ee9eb95ce2d5812145a3aca63 yesterday and now use only one view for both mobile and desktop. Why did you do that?

app/views/users/_edit.haml
svbergerem: I removed that in #6102 so don't worry about it anymore.

svbergerem:
@Flaburgan Because there were no differences between the rendered files anymore.

Flaburgan:
Well, in develop, but not in this PR...

svbergerem:
If you need different views for desktop and mobile you can move `app/views/users/edit.haml` to `app/views/users/edit.html.haml` and save a copy to `app/views/users/edit.mobile.haml`.

Flaburgan:
The partial is fine, I'm going to redo it. I just wanted to check if you did that for a good reason or just as a refactor (it seemed weird because you reviewed this code just before). Anyway, back to work :)

Flaburgan:
I refactored the settings pages to correctly use the bootstrap grid. I'll have a final look at it soon and then it'll be ready.

Flaburgan:
This should be ready.

app/views/users/edit.html.haml
```diff
@@ -0,0 +1,2 @@
+.container-fluid
+  = render "edit", mobile: false
```
svbergerem: the variable `mobile` is not used anywhere

app/views/users/edit.mobile.haml
```diff
@@ -0,0 +1,2 @@
+.settings_container
+  = render "edit", mobile: true
```
svbergerem: the variable `mobile` is not used anywhere

svbergerem:
**services before**:

services_before

**services after**:

services_after

app/views/users/edit.mobile.haml
AugierLe42e: Hmm ? When has this change ? I used it to render a small block on mobile view. As I recall, it was the the password fied above the close account button that was present in mobile view and not on desktop. That section did not have the same class names and IDs than desktop and I kept it as is to avoid having to modify the JS.

svbergerem: #6102 Now both views use the bootstrap modal.

Flaburgan:
@svbergerem problems fixed.

Btw @goobertron is `There are no services on this pod available.` a nice sentence? I would have placed `available` just after `services` but eh, you're the native speaker so I ask ;)

goobertron:
You are a native English speaker! Yes, 'There are no services available on this pod.' is far better - grammatically correct, and clearer in meaning.

I'm just going to have a think about whether there's a more 'positive'/friendly way we can say this. Probably not, and it's fine as is - just a good opportunity to see whether there's any improvement that can be made.

theworldbright:
Maybe 'There are no third party sharing services available on this pod.' is clearer? Just plain 'services' is a bit ambiguous in terms of what 'services' are if you didn't know what they could be in the first place. Without context, it might even make it sound like this pod can't do anything.

goobertron:
There is context given by the notice to the right in the Services tab - see screenshots above. However it's not currently that informative.

I like your 'third-party sharing services', which I think makes it a bit clearer. I would add this to the notice on the right-hand side, so:

> Connecting to third-party sharing services gives you the ability to publish your posts to them as you write them in diaspora*. 

on the right, and stay with

> There are no services available on this pod.

for the notice when there are none connected.

Fla, would you be happy to incorporate that extra change to the en locale in this PR?

Flaburgan:
@goobertron done.

svbergerem:
Thank you!

Flaburgan:
Nice :)

PR #6087, opened 2015-06-08
Improve mobile conversation design

Flaburgan:
## Before

capture d ecran 2015-06-08 a 22 32 46
## After

capture d ecran 2015-06-08 a 22 31 18

### DIFF
--- a/app/views/conversations/_show.mobile.haml
+++ b/app/views/conversations/_show.mobile.haml
@@ -2,16 +2,21 @@
 -# licensed under the Affero General Public License version 3 or later. See
 -# the COPYRIGHT file.
 
-.conversation_participants
-  .right
-    = link_to(raw("<i class='entypo trash'></i>"), conversation_visibility_path(conversation),
-            method: 'delete', data: { confirm: "#{t('.delete')}?" }, class: "remove")
+.conversation
+  .conversation_participants.header-full-width
+    .delete_conversation.pull-right
+      = link_to(raw("<i class='entypo trash'></i>"), conversation_visibility_path(conversation),
+              method: 'delete', data: { confirm: "#{t('.delete')}?" }, class: "remove")
 
-  %h3{ class: direction_for(conversation.subject) }
-    = conversation.subject
+    %h3
+      = conversation.subject
+    .last_message_timeago
+      != t('.last_message', timeago: timeago(conversation.updated_at))
 
-  - for participant in conversation.participants
-    = person_image_link(participant, size: :thumb_small)
+    .avatars
+      - for participant in conversation.participants
+        .avatar
+          = person_image_link(participant, size: :thumb_small)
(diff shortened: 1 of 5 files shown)


app/views/conversations/_show.mobile.haml
```diff
@@ -2,16 +2,21 @@
 -# licensed under the Affero General Public License version 3 or later. See
 -# the COPYRIGHT file.
 
-.conversation_participants
-  .right
-    = link_to(raw("<i class='entypo trash'></i>"), conversation_visibility_path(conversation),
-            method: 'delete', data: { confirm: "#{t('.delete')}?" }, class: "remove")
+.conversation
+  .conversation_participants.header-full-width
+    .delete_conversation.right
```
svbergerem: You can use `.pull-right` instead of `.right`.

app/views/conversations/_show.mobile.haml
```diff
@@ -2,16 +2,21 @@
...
-    = person_image_link(participant, size: :thumb_small)
+    %div
+      - for participant in conversation.participants
+        .avatar
+          = person_image_link(participant, size: :thumb_small)
+    .clear
 
-.span-15.last
-  = render partial: 'messages', locals: { conversation: conversation }
+  .span-15.last
```
svbergerem: That has no effect anymore. Remove it and use a simple `%div` instead or use Bootstrap's grid.

svbergerem:
Here is a screenshot of a conversation with many participants.

conversation with many participants

The avatars could need some bottom margin and because of the missing left margin for the first avatar the avatars in the first and second line are not properly aligned. I'd suggest using `margin: 0.15em` for all avatars.

I'd also like to see some padding for the messages in a conversation. Changing [this line](https://github.com/Flaburgan/diaspora/blob/improve-mobile-conversation-design/app/assets/stylesheets/mobile/conversations.scss#L48) from
```scss
.new_message {
```
to
```scss
.stream .stream_element {
```
would do that.

svbergerem:
You added a timeago under the conversation title. Currently it is not clear what this timeago should tell us. One could argue that the timeago is not really useful but I'm ok with adding it under the title. I'd just add some text like "last message received 10 days ago" or "updated 10 days ago".

Flaburgan:
>You added a timeago under the conversation title.

I took this one from the desktop view. Do you want me to add a legend on both views?

svbergerem:
In the desktop view we have no timeago under the conversation title. We have timeagos next to each message and there is also a timeago in the list of conversations where we also show the last message so I think in all those contexts it is pretty clear what the timeago refers to.

Flaburgan:
@svbergerem changes done.

svbergerem:
Thank you! :)

Flaburgan:
:)

PR #6090, opened 2015-06-09
make sql and federation debug-logging configurable

SuperTux88:
by default both are disabled now, you can enable them in the diaspora.yml if you need them.


jhass:
Merged as ce380d78c9343aa60181b6bc595403d9e5376eeb

Thanks!

PR #6094, opened 2015-06-12
fix-overflow-in-publisher-bookmarklet

arlogn:
Fixing overflow in publisher bookmarklet when the selected text exceeds the size of the textarea.

_publisher-overflow
It overlaps the text about use of markdown and prevents to click the link and the entypo icons.


svbergerem:
This would just hide the extra text. I think the correct solution would be to expand the publisher.

jhass:
Superseded by #6104 

Thank you!

Issue #5339, opened 2014-10-13
Drop old services keys from statistics.json


jaywink:
See https://github.com/diaspora/diaspora/pull/5296

Old keys should be dropped earliest in 0.6 release.

PR #6097, opened 2015-06-13
Remove old service keys in statistics.json

Flaburgan:
Close https://github.com/diaspora/diaspora/issues/5339

### DIFF
--- a/app/presenters/statistics_presenter.rb
+++ b/app/presenters/statistics_presenter.rb
@@ -8,8 +8,6 @@ class StatisticsPresenter
     base_data.merge(user_counts)
              .merge(post_counts)
              .merge(comment_counts)
-             .merge(all_services)
-             .merge(legacy_services) # Remove in 0.6
   end
 
   def base_data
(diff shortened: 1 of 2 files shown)


app/presenters/statistics_presenter.rb
```diff
@@ -8,8 +8,6 @@ def as_json options={}
     base_data.merge(user_counts)
              .merge(post_counts)
              .merge(comment_counts)
-             .merge(all_services)
```
Flaburgan: Data are already included in base_data, this was useless.

app/presenters/statistics_presenter.rb
```diff
@@ -95,28 +93,10 @@ def local_comments
...
-    result
-  end
-
-  def all_services
-    @all_services ||= all_services_helper
-  end
-
   def available_services
     Configuration::KNOWN_SERVICES.select {|service|
       AppConfig.show_service?(service, nil)
```
Flaburgan: I don't get the test at https://github.com/diaspora/diaspora/blob/develop/lib/configuration_methods.rb#L41 with "authorized" etc. We just need to know if the service is enabled or not.

Flaburgan:
@jaywink is http://the-federation.info updated to check the new syntax? Maybe you want to still support both syntaxes for pod which are still running 0.4.

@dmorley same remark for http://podupti.me

app/presenters/statistics_presenter.rb
jhass: #5985

Flaburgan: Oh right! Thank you :)

jaywink:
@Flaburgan not up to date yet, will need to update asap once this is merged in :)

Note, however merges this is - as agreed earlier, this is 0.6 stuff, not a minor release.

app/presenters/statistics_presenter.rb
jaywink: :+1:

jaywink:
Hmm, might as well merge now - thanks!

@dmorley - pods running on develop will not show with correct services after they pull this in so time to update the code if you haven't already. I'd suggest supporting the old services keys until forever for old pods, friendica and redmatrix.

Issue #5815, opened 2015-03-25
Armenian punctuation marks in tags.


ghost:
Hello,

In Armenian some of [these](https://en.wikipedia.org/wiki/Armenian_alphabet#Punctuation) symbols are used inside words, between letters.

Thus I would like to ask to make possible to use following symbols in tags
- [՜](https://en.wiktionary.org/wiki/%D5%9C)
- [՞](https://en.wiktionary.org/wiki/%D5%9E)
- [֊](https://en.wiktionary.org/wiki/%D6%8A)
- [՛](https://en.wiktionary.org/wiki/%D5%9B)

Thus we can write ինչո՞ւ, այո՜ո, սեւ֊սպիտակ, գժանո՛ց, etc. Thank you.
##

jhass:
This should be uncontroversial. Solving this would involve patching our markdown it plugin for tag parsing and the regexp used in the backend code.

svbergerem:
Just to clarify (because “patching our markdown it plugin” sounds like a bigger task): You only need to change the regexp [here](https://github.com/diaspora/diaspora/blob/3298373a7fd57b6f7f23cb5a36bce0806476b07d/app/assets/javascripts/app/helpers/text_formatter.js#L57). This is related to #5765.

Issue #5765, opened 2015-03-12
Incorrect hashtag split for Malayalam


balasankarc:
While giving hashtags in Malayalam language, the hashtag gets split at the character ്   (Unicode codepoint - +0D4D ) . This character is used to to join two consonants so as to form a conjunct and is part of a whole word. Hence, hashtag must not split at this character and must continue till a space is occured.
Example : ണ്മ is formed by ണ + ് + മ 
In the post https://poddery.com/posts/1672504 , the hashtag #മലയാണ്മ gets split at ്  and hence becomes #മലയാണ (് മ gets cut off).
##

jhass:
To clarify, the frontend code seems to parse it correctly, but 04D4 is not in `[[:alnum:]]`, so the backend associates the wrong tag.

svbergerem:
After changing the markdown parser from pagedown to markdown-it this also won't be parsed in the frontend code because we now use [[[:alnum:]]](https://github.com/diaspora/diaspora/blob/develop/lib/assets/javascripts/posix-bracket-expressions.js#L7).

jhass:
> /[[:word:]]/ - A character in one of the following Unicode general categories Letter, Mark, Number, Connector_Punctuation

Am I missing something or would this be good to use?

balasankarc:
Hi, any update on this?? This is actually breaking the hashtag functionality in Malayalam as there are numerous words which use +0D4D and are commonly used for hashtags.

Issue #5542, opened 2015-01-10
Full Unicode support for client side tag parsing


jhass:
We had to remove some ranges in #5541, we should find a way to make them work

Flashwalker:
We had removed it and if you wana back, pay to us, mthrfckr.

SuperTux88:
Excuse me, can I help you?

PR #6099, opened 2015-06-13
hashtags accept Armenian punctuation marks and Malayalam virama

polsvoice:
Fixes issues #5815 and #5765.


jhass:
That looks very specific, I wonder if there's a unicode category or something we want to include?

Also did you verify the backend regexp matches them already?

polsVoice:
@jhass For your first question: I wasn't sure how to include this. This was the best way I could think of to do it. I haven't been able to find a category that covers this, but I can keep looking.

For your second question: I'm not sure I understand. Are you asking if I tested the code to see if it works? All I do was verify that it wasn't working (by making a post with hashtags containing the characters) and then verify that the tags were working after adding the character codes. If I need to do any more tests, please let me know, and I will do my best.

Also, should I fix the hound violations? Should the + sign be at the end of the lines instead of the beginning?

jhass:
> Also, should I fix the hound violations? Should the + sign be at the end of the lines instead of the beginning?

Yes.

I'm talking about the expression used in the backend to associate posts to a tag, that is extract tags from a post. this one: https://github.com/diaspora/diaspora/blob/develop/app/models/acts_as_taggable_on-tag.rb#L7

svbergerem:
@jhass suggested in #5765 that one could use `[[:word:]]`.

polsVoice:
@svbergerem I'm not sure it's that simple. On line 39, `[[:alnum]]` is PosixBracketExpressions.alnum. It's not actually referenced as a character class, but is defined in an object called PosixBracketExpressions. Adding `+[[:word]]` to the end of the regex doesn't work. Also, according to [this,](http://www.regular-expressions.info/posixbrackets.html) `[:word]` is only `[A-Za-z0-9_]`...no Armenian or Malayalam punctuation.

jhass:
Notice that the Unicode version includes `\p{Pc}`, http://www.regular-expressions.info/unicode.html and Oniguruma should follow that version IME.

PosixBracketExpressions is something we did on our own since JS lacks them, it was adapted from Onigurumas mapping to get consistent results: https://github.com/ruby/ruby/blob/trunk/enc/unicode/name2ctype.src#L10 You can find the mapping for `[[:word:]]` further down: https://github.com/ruby/ruby/blob/trunk/enc/unicode/name2ctype.src#L3125

polsVoice:
`[[:word]]` contains `0x0d4d` but not `0x055b`, `0x055c`,`0x055e`, or `0x058a`, so that should fix #5765, but not #5815. (Those are included in `[[:punct]]`.) Note that including the Armenian question mark and exclamation mark is different from including their English equivalents, because Armenian apparently has those marks *inside* of the words, rather than at the end as in English. `0x055b` is inserted after a vowel to indicate stress. `0x058a`, however, is a hyphen (!!), and since English hyphens aren't allowed in hashtags, probably Armenian ones shouldn't either.

Personally, I think that hashtags should allow a much wider array of characters, but that's obviously [a much bigger discussion.](http://www.dailydot.com/lol/hashtags-fails-punctuation/)

jhass:
Well, including `[[:punct::]]` is a no go since it includes `.`, `)`, `?`, `!` etc etc.

jhass:
Maybe switch to `[[:word:]]` and leave #5815 unsolved for now?

svbergerem:
Or use `"[[:word:]]\u055b\u055c\u055e_-"` since that seems to be uncontroversial. `\u058a` could be discussed in #5815 after merging this PR.

lib/assets/javascripts/posix-bracket-expressions.js
```diff
@@ -4,445 +4,569 @@
...
-        + '\\ufe70-\\ufe74'
-        + '\\ufe76-\\ufefc'
-        + '\\uff10-\\uff19'
-        + '\\uff21-\\uff3a'
-        + '\\uff41-\\uff5a'
-        + '\\uff66-\\uffbe'
-        + '\\uffc2-\\uffc7'
-        + '\\uffca-\\uffcf'
-        + '\\uffd2-\\uffd7'
-        + '\\uffda-\\uffdc'
```
jhass: I don't think it harms to keep those

jhass:
Alright, so the open points are

* [ ] Please also update the backend regexp
* [ ] Please add some Jasmine specs for the change
* [ ] Please add some RSpec specs for the backend regexp change

lib/assets/javascripts/posix-bracket-expressions.js
polsVoice: I'm not sure I understand. You think we should have both `[[:alnum]]` and `[[:word]]`?

jhass: I think it doesn't harm to build up all the expressions here as needed, might come handy elsewhere at some point

polsVoice: That's a yes?

jhass: We should not use both in the hashtag regex, but there's no maintenance cost to keeping both available in the library.

polsVoice:
As for the last 3 items that you mentioned, I don't know where to start. Are there any tutorials for that? I have some experience with Jasmine and writing unit tests... Where can I find the backend regexp?

jhass:
https://github.com/diaspora/diaspora/blob/develop/spec/javascripts/app/helpers/text_formatter_spec.js#L10

and

https://github.com/diaspora/diaspora/blob/develop/spec/shared_behaviors/taggable.rb

polsVoice:
I was actually wrong. The English hyphen *does* work (e.g. #foo-bar is a valid hashtag).

spec/javascripts/app/helpers/text_formatter_spec.js
```diff
@@ -24,6 +33,8 @@ describe("app.helpers.textFormatter", function(){
         expect(formattedText).toContain(link);
       });
     });
+    
+
```
jhass: Accidental whitespace

jhass:
Alright, add the examples to the backend specs and we should be good.

spec/shared_behaviors/taggable.rb
```diff
@@ -29,6 +34,11 @@ def tag_link(s)
       expect(formatted_string).to include(tag_link('what'))
       expect(formatted_string).to include(tag_link('hey'))
       expect(formatted_string).to include(tag_link('vöglein'))
+      expect(formatted_string).to include(tag_link('ինչո՞ւ'))
+      expect(formatted_string).to include(tag_link('այո՜ո'))
+      expect(formatted_string).to include(tag_link('սեւ֊սպիտակ'))
+      expect(formatted_string).to include(tag_link('գժանո՛ց'))
+      expect(formatted_string).to include(tag_link('മലയാണ്മ'))
     end
```
jhass: Also make sure the extraction works correctly by adding cases to the `it 'extracts tags despite surrounding text'` spec.

spec/shared_behaviors/taggable.rb
```diff
@@ -139,7 +149,12 @@ def tag_link(s)
         '#-initialhyphen'        => '-initialhyphen',
         '#-initialhyphen tag'    => '-initialhyphen',
         '#-initial-hyphen'       => '-initial-hyphen',
-        "\u202a#\u200eUSA\u202c" => 'USA'
+        "\u202a#\u200eUSA\u202c" => "USA",
+        "test #ինչո՞ւ test"      => "ինչո՞ւ",
+        "test #այո՜ո test"       => "այո՜ո",
+        "test #սեւ֊սպիտակ test"  => "սեւ֊սպիտակ",
+        "test #գժանո՛ց test"     => "գժանո՛ց",
+        "test #മലയാണ്മ test"      => "മലയാണ്മ"
```
jhass: Alignment.

spec/javascripts/app/helpers/text_formatter_spec.js
```diff
@@ -216,7 +225,7 @@ describe("app.helpers.textFormatter", function(){
       });
 
       it("correctly encodes image src to punycode", function() {
-        _.each(this.evilUrls, function(url, num) {
+        _.each(this.evilUrls, function(url, num) {
```
jhass: Trailing whitespace

spec/javascripts/app/helpers/text_formatter_spec.js
```diff
@@ -261,7 +270,7 @@ describe("app.helpers.textFormatter", function(){
       it("doesn't crash Firefox", function() {
         var content = "antifaschistisch-feministische ://";
         var parsed = this.formatter(content);
-        expect(parsed).toContain(content);
+        expect(parsed).toContain(content);
```
jhass: Trailing whitespace

lib/assets/javascripts/posix-bracket-expressions.js
```diff
@@ -4,445 +4,1010 @@
...
+			"\\uff10-\\uff19" +
+			"\\uff21-\\uff3a" +
+			"\\uff3f" +
+			"\\uff41-\\uff5a" +
+			"\\uff66-\\uffbe" +
+			"\\uffc2-\\uffc7" +
+			"\\uffca-\\uffcf" +
+			"\\uffd2-\\uffd7" +
+			"\\uffda-\\uffdc" +
+			"\\u10000-\\u1000b" +
```
svbergerem: I guess you will have to remove those. See #5541 and [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#character-classes) 

> `\uhhhh` 	Matches the character with the Unicode value `hhhh` (four hexadecimal digits).

polsVoice: So, that's all Unicode values with 5 digits?

svbergerem: Yes, unfortunately. ECMAScript 6 supports also other ways to define unicode characters but that would imply updating markdown-it-hashtag and using a js compiler like [babeljs](https://babeljs.io/) to add support for older browsers. For this PR I'd say we only use 4 digit unicode values and fix #5542 later.

spec/shared_behaviors/taggable.rb
polsVoice: What do you mean?

polsVoice: It's weird. I can't get it to line up. If I remove the space, it moves over too far to the left. I tried deleting characters until I could get it to line up. The problem seems to be with the virama, which is taking up *half a space* (even in a monospace font?) It's the 4th character after the hash.

jhass: Hmm, indeed, anyway, the `=` is in column 34 and this one is in column 35, I'd say make them the same column regardless.

polsVoice:
Well, I'm stymied on this one. Either I'm writing the tests wrong, or there's some weird bug. I don't know much Ruby, so that could have something to do with it. Right now, I feel like I'm taking shots in the dark.

Is there any way I can run the tests locally? Or do I have to push to Github every time?

svbergerem:
https://wiki.diasporafoundation.org/Testing_workflow

polsVoice:
Thanks, @svbergerem. I'm getting a 502 gateway error from that page right now, though...

svbergerem:
@polsVoice At least for me it is working again.

polsVoice:
@svbergerem Okay, it's working for me now, too.

balasankarc:
Hoping to see this fixed (pulled) soon, as the number of people using Malayalam is growing each day. . :)

polsVoice:
The changes to ```spec/shared_behaviors/taggable.rb``` produce failures in ```spec/models/profile_spec.rb```. The string ```@str``` (on line 17 of ```spec/shared_behaviors/taggable.rb```) has 8 words, but it's truncated to 5 when it's saved (because there can only be 5 tags in a profile?). As a result, the last 3 words can't be tested, and that produces errors.
```
 Failure/Error: expect(@object.tags(true).map(&:name)).to include('սեւ֊սպիտակ')
       expected ["what", "hey", "vöglein", "ինչո՞ւ", "այո՜ո"] to include "սեւ֊սպիտակ"
```
I'm working on a solution, but I guess it'll have to be something that doesn't butt heads with the profile spec, as there's a limitation in the number of tags it can process.

jhass:
No, that function shouldn't limit the number of items, the regexp simply doesn't match then.

polsVoice:
When I do ```puts @object.tags``` after line 20 of ```spec/shared_behaviors/taggable.rb```, the last 3 tags don't show up in the console, either. I run ```bin/rspec spec/models/profile_spec.rb``` again, and here's what I get:
```
what64 |=========== 51 ============>                           |  ETA: 00:00:01 
hey
vöglein
ինչո՞ւ
այո՜ո
what64 |============ 53 ============>                          |  ETA: 00:00:01 
hey
vöglein
ինչո՞ւ
այո՜ո
```

jhass:
I wouldn't expect otherwise if my assumption holds.

polsVoice:
Hashtags with Roman characters get rejected, too. I changed the string to
```
@str = "#what #hey #vöglein #my #dear #aunt #sally"
```
and the error turned into
```
 Failure/Error: expect(@object.tags(true).map(&:name)).to include("ինչո՞ւ")
       expected ["what", "hey", "vöglein", "my", "dear"] to include "ինչո՞ւ"
```
It still returns only 5 strings. I also tried
```
@str = "#ինչո՞ւ #այո՜ո #սեւ֊սպիտակ #գժանո՛ց #മലയാണ്മ"
```
and got this:
```
  Failure/Error: expect(@object.tags(true).map(&:name)).to include("vöglein")
       expected ["ինչո՞ւ", "այո՜ո", "սեւ֊սպիտակ", "գժանո՛ց", "മലയാണ്മ"] to include "vöglein"
```

jhass:
Welp, https://github.com/diaspora/diaspora/blob/develop/spec/models/profile_spec.rb#L285 I didn't think profile would just drop them.

Anyway, do multiple expectations a 5 then.

polsVoice:
I don't know much about RSpec (or Ruby), so this might take a while. I'm pretty lost in the code at the moment.

SuperTux88:
Superseded by #7350

PR #6104, opened 2015-06-14
Replace jquery.autoresize with autosize

svbergerem:
Replaces the old jquery.autoresize plugin with [autosize](https://github.com/jackmoore/autosize) which seems to be well-maintained.

Supersedes #6094 and fixes a Bootstrap 3 style regression in the bookmarklet code.


jhass:
Thanks!

PR #6105, opened 2015-06-15
precompile images from facebox gem

SuperTux88:
regression from #6005

maybe we should replace this with a bootstrap-modal anyway ...


jhass:
Merged as beb06bc42d830708c27b13f93a2fdf858ebf39f6

Thanks!

PR #6106, opened 2015-06-15
Remove all faceboxes

svbergerem:
Converts the last facebox (public_explain) into a Bootstrap modal and removes the facebox code from our codebase. I reverted the changes of #6105 but this PR uses Bootstrap 3 so we still need #6105 for v0.5.2.0.

The modal is as bad as the facebox was before and real improvements are out of scope of this PR. ;)

public explain modal


jhass:
Thank you!

Issue #1960, opened 2011-09-12
User menu inaccessible for people with low resolutions


cparker15:
Due to the header element's fixed positioning, it does not scroll horizontally with the rest of the page. This makes the user menu on the right side of the header completely inaccessible for people stuck on lower resolutions.

This behavior can be seen by resizing the browser window so the user menu disappears, and then scrolling the window horizontally.

svbergerem:
This has been fixed by #6109.

Issue #5780, opened 2015-03-14
Opening the dropdown doesn't update the notification count


AugierLe42e:
I though I solved the bug but seemd not. As @jhass  noticed [here](https://github.com/diaspora/diaspora/pull/5707#discussion-diff-25560611) I think the notifications should not be directly transmitted in HTML but in a JSON structure and rendered client side. This way, the total count of unread notifications could be transmitted too.
##

goobertron:
Posted in #6181:
 
I've noticed that when I open a page by clicking on a notification and then click the browser's 'back' button, while the new notifications dropdown is updated (showing the notification you just clicked as read), the notification badge in the header still shows the previous number, as though that notification just clicked was in fact unread.

(Reposting here in case it provides extra information for debugging.)

PR #6109, opened 2015-06-16
Bootstrap header rewrite

### LINKED ISSUE #5283: Nav badge icons icons should use Entypo
I noticed on high resolution screens that the nav badges for messages and notifications look fuzzy. This is due to the fact that we are using images for these assets that are not high resolution, obvi. Since we are using Entypo Pictograms in other places on the site, we should use Entypo for other places where we have icons that are image-based where possible. This includes the nav badges and probably the `.rightBar` sections as well. Entypo has a good range of icons we could easily replace many of these with. This will also allow us more control over the style, color, size of these elements without relying on images.

On my own branch I have replaced them in the Nav area. I am happy to do this for the right sidebar as well.

Before:
screen shot 2014-10-04 at 6 51 42 pm

screen shot 2014-10-04 at 6 57 57 pm

After:

screen shot 2014-10-04 at 6 52 23 pm

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/5095166-nav-badge-icons-icons-should-use-entypo?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
This fixes #5283 and #1960. This is a rewrite so I am not trying to get the old design back. I am pretty much done with the design so you can already comment on that.

I still need to check if there are any failing tests, missing tests for new functions I wrote or some code style issues. The color themes are also currently not supported.
## header for big screens (>= 992px)

20150616_normal
20150616_normal_logout
## header for small screens (desktop view)

20150616_small


jhass:
The notification dropdown feels kind of detached for me, Maybe we can keep changing the background color of the button if active?



I never liked the full height of the search in bootstrap headers, I think a slightly larger margin on the top and bottom looks better. Also the spinner while loading should be a lot smaller and not touch the borders of the input IMO. Finally I think we should take the chance and remove the transparency of the dropdown.



The user dropdown again feels kind of detached to me, I think keep the header background color looks nicer.



I guess I don't need to say much about the notifications dropdown in the collapsed header version ;)

svbergerem:
There shouldn't be any notification dropdown for the collapsed header version. Looks like that broke while I did some refactorings. All other points also seem legit. I'll see what I can do.

Flaburgan:
Hey, nice work @svbergerem I was also thinking to make the header flat, you arrived first :p
I'll wait to be able to try it locally to tell my remarks.

AugierLe42e:
Flat ? I though you hated flat, fl

Flaburgan:
@AugierLe42e I like to know where I can click easily. So, I don't like when elements with which I interact are flat, like buttons for example. For the header it's fine. That's what's done here on github for example: flat header but real buttons :)

AugierLe42e:
Myeah, you're right. Flat buttons are weird on BS3. I really like the design @pablocubico did months ago.

jaywink:
I miss @pablocubico :(

But great work @svbergerem - welcome refresh.

AugierLe42e:
I'd also like to remind that [this design](https://www.loomio.org/d/QdrGgw7u/new-header-design) has been voted on Loomio 2 months ago.

pablocubico:
Hey guys, I'm under a shitload of work and didn't have time to contribute,
also not having BS3 was a blocker, I didn't want to push or suggest changes
before migrating. Now that we are close, I hope we can start polishing the
UI section by section. I'm on release week right now, but I'll be in touch
next week to catch up and start contributing again.

This is a huge step forward! Thanks Augier for all the hard work and not
giving up.

jaywink:
@pablocubico ooh nice to hear you're still there - was afraid you had left us for good :)

svbergerem:
20150617_overview

features/step_definitions/conversations_steps.rb
```diff
@@ -5,7 +5,7 @@
 end
 
 Then /^I should have (\d+) unread private messages?$/ do |n_unread|
-  find("header #conversations_badge .badge_count").should have_content(n_unread)
+  find("header #conversations-link .badge").should have_content(n_unread)
```
jhass: Please update to `expect(...).to` syntax while touching.

app/assets/templates/header_tpl.jst.hbs
```diff
@@ -1,109 +1,121 @@
...
+        <li class="dropdown" id="user_menu">
+          <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">
+            <span class="user-avatar pull-left">
+              {{{personImage current_user "small"}}}
+            </span>
+            <span class="user-name">{{current_user.name}}</span>
+            <span class="caret"></span>
+          </a>
 
+          <ul class="dropdown-menu" role="menu">
```
AugierLe42e: I think you should add `width: 100%;` to that element. It feel weird that the user menu is not the same width than the button.

Flaburgan:
I tested it locally and I like it :)

- I would have kept a `box-shadow: 1px 0px 2px black;` on `.navbar-inverse`
- It would be nice to use the same bootstrap class than the stream page to align "diaspora*" in the header with the left side bar, and the user menu with the right of the right side bar. This looks really better imo.
- I agree with Augier on the `width: 100%` on `.dropdown-menu`, I would add a `padding-left` to make the li text aligned with the username and not the avatar though.
- Maybe stream and my activity links can have a font-weight: bold or something like this, they looks tiny
- I would have kept the `background-color: #444;` of the search field when it's not focused.

What nice work you did :)

AugierLe42e:
> It would be nice to use the same bootstrap class than the stream page to align "diaspora*" in the header with the left side bar, and the user menu with the right of the right side bar. This looks really better imo.

Oh! Yes! As I recall, it's `col-md-offset-1 col-md-10`

> I agree with Augier on the width: 100% on .dropdown-menu, I would add a padding-left to make the li text aligned with the username and not the avatar though.

And `min-width: 160px;` on the `.dropdown-toggle` to prevent the button to be smaller that the dropdown (I forgot to mention it)

> And maybe also the transition?

+1, I liked it :D

And it would be nice to also have to extras butons next to *My Activity*: *Likes* and *Reshares*.

jhass:
> And it would be nice to also have to extras butons next to My Activity: Likes and Reshares.

Let's not creep "features" into this and discuss it separately.

pablocubico:
This looks awesome, I just saw the screenshots, I thought I was answering another thread before.

Thanks @svbergerem, I'm feeling that by the end of this year the Diaspora UI will kick ass. :fist:

pablocubico:
Hey @svbergerem two suggestions:
- Make the search border more subtle when black, maybe 50% gray or something.
- Maybe the dropdown menu on the user side could fit the size of the name + avatar.
- Also, I would make the dropdown options bigger, same font size but a little more padding, so the clickable region is taller. If we remember to have big clickable sections through the whole app, it will be "tablet-ready" (finger-friendly?) with no extra effort.

svbergerem:
Thank you all for your feedback. First some new screenshots, then I'll go through your suggestions.

20150617_colorthemes

(color theme support, the user menu in those screenshots is opened but not part of the screenshot)

20150617_new

> I would have kept a box-shadow: 1px 0px 2px black; on .navbar-inverse

done

> align "diaspora*" in the header with the left side bar, and the user menu with the right of the right side bar

done, but only for screens >= 1200px. For screens between 992px and 1200px we need the full width.

> width: 100% on .dropdown-menu, add padding-left to text

done

> stream and my activity links can have a font-weight: bold

done

> I would have kept the background-color: #444; of the search field when it's not focused

I like the current subtle search field

> And maybe also the transition?

done

> And min-width: 160px; on the .dropdown-toggle to prevent the button to be smaller that the dropdown

done

> would be nice to also have to extras butons next to My Activity: Likes and Reshares

like @jhass said: won't happen in this PR

> Make the search border more subtle when black, maybe 50% gray or something.

done

> make the dropdown options bigger, same font size but a little more padding (-> tablet-ready)

I'd like to keep this PR as simple as possible and would keep the Bootstrap defaults currently. I guess the Bootstrap guys chose the defaults deliberately and this PR already improves the situation for tablet users. We can talk about some further improvements in separate issues or follow-up PRs.

Flaburgan:
Wow, awesome work steffen! I just noticed that the header is not higher (50px, was 40px before). Any reason for this change?

svbergerem:
> I just noticed that the header is not higher (50px, was 40px before). Any reason for this change?

It's the default height for Bootstrap headers which makes a lot of things easier and I like the 50px header.

Flaburgan:
Fair enough ;)

pablocubico:
> I'd like to keep this PR as simple as possible and would keep the Bootstrap defaults currently. I guess the Bootstrap guys chose the defaults deliberately and this PR already improves the situation for tablet users. We can talk about some further improvements in separate issues or follow-up PRs.

It's just some padding on the options! BS3 dropdown don't make a great experience on tablets unless you upsize them a bit, I work with Bootstrap all the time. Anything >= 40px tall/wide is good for a finger on pretty much any device.

Anyway, looks great!

AugierLe42e:
> Thank you all for your feedback. First some new screenshots, then I'll go through your suggestions.

I don't know about the others, but it feels weird that the open user menu does not have the same color than the open notifications dropdown.

Besides this, it's wonderful ! Really nice work.

svbergerem:
@AugierLe42e 

> it feels weird that the open user menu does not have the same color than the open notifications dropdown

The current header does exactly the same thing. Let's keep it like that in this PR and discuss afterwards if we want to change it.

>  I think padding: 6px 8px 6px 40px; would be perfect for #user-menu > li > a

I tried this (with left-padding 55px instead of 40px) and I'm not really convinced. As I said before, let's stick with Bootstrap's defaults in this PR and discuss afterwards how we could further improve the UX for tablet users.

svbergerem:
This PR should be ready to review and merge.

AugierLe42e:
> The current header does exactly the same thing. Let's keep it like that in this PR and discuss afterwards if we want to change it.

Not really. Actually, the current user menu have gray borders when opened:

user-menu

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -70,7 +70,8 @@ app.views.Notifications = Backbone.View.extend({
 
     [allNotes, typeNotes, headerBadge].forEach(function(element){
       element.text(function(i, text){
-        return parseInt(text) + change });
+        return Math.max(parseInt(text) + change, 0);
```
Flaburgan: How is it possible to have a negative result here?

AugierLe42e: Theorically, yes. As the notifications count is not deduced from an actual list, but computed taking account user actions. It's not likely to happen, but it's possible. The best solution would be to have a model for notifications user-side.

svbergerem: I got some negative results after playing with the notification dropdown for some time. As @AugierLe42e wrote there are still some improvements needed but this one was an easy fix to improve the UX. Otherwise you could mark some notifications as unread and the counter would still be negative or zero so you won't see it.

Flaburgan: Ok. It just feels like hiding a bug. Maybe add a Todo or something?

svbergerem: Alright, I reverted this change. It is related to #5780 and I agree that one could interpret this as hiding the bug.

features/step_definitions/custom_web_steps.rb
```diff
@@ -201,7 +201,7 @@ def take_screenshots_with_login
 
 
 Then /^the notification dropdown should be visible$/ do
-  find(:css, "#notification_dropdown").should be_visible
+  find(:css, "#notification-dropdown").should be_visible
```
jhass: Please update to `expect(...).to` while touching.

features/step_definitions/modal_steps.rb
```diff
@@ -1,3 +1,12 @@
...
+Then /^I should see a modal$/ do
+  step %{I should see a ".modal.in"}
+end
+
+Then /^I should see the mention modal$/ do
   step %{I should see a "#mentionModal.in"}
 end
+
+When /^I put in my password in the close account modal$/ do
+  execute_script("$(\"#closeAccountModal input#close_account_password\").val(\"#{@me.password}\")")
```
jhass: Please add a comment explaining why the `execute_script` solution is used instead of the Capybara helper.

features/step_definitions/custom_web_steps.rb
svbergerem: done

features/step_definitions/modal_steps.rb
svbergerem: done

Flaburgan:
Looks ready from my side :)

AugierLe42e:
The notifications dropdown is overlapping the header: 

header

svbergerem:
@AugierLe42e done.

AugierLe42e:
The [border top](https://github.com/diaspora/diaspora/pull/6109/files#diff-1d3b7aaf9a68499c7fa63be5194906caR66) on the *view all* button feels weird, isn't it ?

header

svbergerem:
@AugierLe42e I didn't use your diff and the user menu is still working on my dev pod.

Flaburgan:
I merged it on diaspora-fr, let's see if we have more feedback. But this feels like a great improvement to me :)

AugierLe42e:
Ok, cool.

AugierLe42e:
`.navbar-fixed-top` still have 1px border, so the dropdown still overlaps from 1px.

AugierLe42e:
Wouldn't it be better if `#user_menu > .dropdown-menu` had no padding ?

capture d ecran 2015-06-19 a 23 36 30

capture d ecran 2015-06-19 a 23 38 11

SansPseudoFix:
Really cool, thanks!

I notice one weird thing: when you click the bell (one or two times), notification stream goes out the box.

svbergerem:
> .navbar-fixed-top still have 1px border, so the dropdown still overlaps from 1px.

I don't know which browser you used to test this but with Firefox 38.0.5 it has no border:

notification dropdown border 1

notification dropdown border 2

AugierLe42e:
Meh... FF 37.0.2. Is it possible that I'm not up to date ?

svbergerem:
> Wouldn't it be better if #user_menu > .dropdown-menu had no padding ?

The padding is the Bootstrap default that we also use for aspect membership dropdowns and I like it.

jhass:
I would consider dropping the box shadow from the notification dropdown, it feels irregular to the button it extends from, like the dropdown is still separate. I also wonder if it's possible to move it down a pixel or two without disconnecting it from the button, it's still over the headers border (or is that the box shadow already? Anyway, the dark part) here but I think it would look better if it starts below so the contrast on the right edge isn't so high.

AugierLe42e:
> it's still over the headers border (or is that the box shadow already?

I think it is. I also have thi impression.

AugierLe42e:
Hmm... No, it's because the `li` height is still 49px:

header2

It looks really better at 50px:

header3

pablocubico:
Last screenshot looks awesome to me.

svbergerem:
> it's still over the headers border (or is that the box shadow already? Anyway, the dark part)

I posted a screenshot. When you zoom in you can see that it is the box shadow. I don't see an easy solution to move down the dropdown without disconnecting it.

> I would consider dropping the box shadow from the notification dropdown

That is the default Bootstrap box shadow for dropdowns that we have for each dropdown including the notification dropdown and the user menu.

AugierLe42e:
@svbergerem : Look at the two pics I posted. It's just because the button is 49px instead of 50px ;)

svbergerem:
@AugierLe42e Are you sure that you are using the code with the latest changes? Here is a screenshot from my dev pod:

dropdown li

width: 49px, height: 50px

AugierLe42e:
Hmmf... Nope. No way. It does change. fetch, pull, remote remove... :/

svbergerem:
I'm still not able to reproduce this with Firefox but I just reproduced it with Chromium so I now set the height of the button manually to 50px which fixes the problem.

AugierLe42e:
Nice. Thx.

AugierLe42e:
*Sing out* option is not displayed in mobile size:

screen shot 2015-06-20 at 09 32 11

svbergerem:
@AugierLe42e It is, you just have to scroll down.

AugierLe42e:
Oh yeah. Hmm... It's weird. Can't make it higher ? I think it will be hardly accessible on mobile view.
And I just spotted another glitch with the conversations button the is overlapped by the user menu:

screen shot 2015-06-20 at 10 51 07

and completely desappears when user menu is open:

screen shot 2015-06-20 at 10 51 37

It's on the old version currently deployed on d-fr, though.

AugierLe42e:
Also, I think it would be better if you used the sandwish icon of the current mobile view and change the current mobile view sandwish icon to the Bootstrap's default for compilence.

svbergerem:
> Can't make it higher ?

The current version of the PR already improves the situation for mobile devices. The height is a Bootstrap default so the same applies that I already said about the paddings in the user menu: let's keep the defaults in this PR and discuss later what we want to do about them. There might be good reasons for those defaults and while I did some testing I saw that they even added breakpoints depending on the height of the screen for the header.

> I just spotted another glitch with the conversations button the is overlapped by the user menu

I am not able to reproduce this. Neither on Firefox nor on Chromium.

> I think it would be better if you used the sandwish icon of the current mobile view

I don't think so. With [Bootstrap 3](http://getbootstrap.com/components/#navbar) you use 3 simple spans to display that element. I don't see why loading an image and adding it to the header (with additional css because just adding it won't work) should be better than the current solution.

> and change the current mobile view sandwish icon to the Bootstrap's default

No, I won't touch the mobile view in this PR. This PR is only about the header in the desktop view. One could think about removing the image in the mobile view and using the Bootstrap classes instead but that is out of scope of this PR.

jhass:
Alright, let's do further polishing in follow up PRs, thank you!

goobertron:
I've just noticed that - I think - this PR uses the pod name in the header. This will hit problems with pods with long names such as `Diaspora* @ Oude RKZ Groningen`.

I think that, for length purposes, it will be better to use simply `diaspora*`, which was, in any case, the outcome of the vote [here](https://www.loomio.org/d/x0Th4Xwl/identity-use-a-white-diaspora-wordmark-instead-of-the-black-star-logo-in-the-navbar?proposal=1sjQFEBC).

If I've understood this correctly - and unfortunately I'm having problems getting my rails server running at the moment so I can't check it directly - I can open an issue for it, unless @svbergerem wants to pick it up from this comment.

PR #6110, opened 2015-06-16
Remove top margin for first heading in a post

jhass:
That is, if it's the first element in the post

I think that's more consistent with posts that don't start with a heading.
That gap always felt wrong to me

I think that's more consistent with posts that don't start with a heading.
That gap always felt wrong to me..


svbergerem:
Thank you!

PR #6111, opened 2015-06-17
A minor typo on the tag list that was causing an empty element to be created

pablocubico:
captura de pantalla 2015-06-17 a la s 06 28 34


Flaburgan:
Nice catch!

jhass:
Merged as c050c606c23ebdcb09326261d2e9e79dedb6cfa2

Good catch indeed. Since it's present in stable too I had to cherry-pick your commit over, which means you'll have to throw it out with an interactive rebase in #6112 

Thanks!

PR #6112, opened 2015-06-17
Left sidebar polishing preserving current styles.

pablocubico:
Hi all, I just did some minor poilshing of broken styles regarding the left sidebar. See the results on the pics.

Tell me what you guys think.

ping: 
@AugierLe42e @svbergerem @Flaburgan @goobertron 

captura de pantalla 2015-06-17 a la s 06 39 37

captura de pantalla 2015-06-17 a la s 06 39 51

PR #6113, opened 2015-06-17
Polishing of the right sidebar, less weight, minor styling

pablocubico:
Guys, take a look. Same approach as the left, but this time, going all the way to the right. I'll do the same with the left one.

I removed the icons and the bold text. Why? To reduce weight and increase legibility.

Tell me what you think.

captura de pantalla 2015-06-17 a la s 09 11 11

### DIFF
--- a/app/assets/stylesheets/colors.scss
+++ b/app/assets/stylesheets/colors.scss
@@ -26,6 +26,7 @@ $blue: #3F8FBA;
 $dark-blue: darken(#0984C8,10%);
 
 $right-sidebar-background: #f0f0f0;
+$right-sidebar-sub-background: darken($right-sidebar-background, 12%);
 
 /* colors : http://www.colourlovers.com/palette/2134203/Awezome_in_argyle */
 $cyan : rgb(8,204,249);
--- a/app/assets/stylesheets/colors.scss
+++ b/app/assets/stylesheets/colors.scss
@@ -25,6 +25,8 @@ $red: #A80000;
 $blue: #3F8FBA;
 $dark-blue: darken(#0984C8,10%);
 
+$right-sidebar-background: #f0f0f0;
+
 /* colors : http://www.colourlovers.com/palette/2134203/Awezome_in_argyle */
 $cyan : rgb(8,204,249);
 $yellow : rgb(242,244,9);
--- a/app/assets/stylesheets/sidebar.scss
+++ b/app/assets/stylesheets/sidebar.scss
@@ -22,7 +22,6 @@
         font-weight: normal;
         font-size: $font-size-base;
         margin: 0;
-        text-transform: uppercase;
       }
     }
 
--- a/app/assets/stylesheets/sidebar.scss
+++ b/app/assets/stylesheets/sidebar.scss
@@ -1,21 +1,28 @@
+#right-sidebar-fixed-background {
+  height: 100%;
+  background: $right-sidebar-background;
+  position: fixed;
+  right: 0;
+}
 .rightBar {
   padding-top: 20px;
+  background: $right-sidebar-background;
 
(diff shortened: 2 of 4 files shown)


pablocubico:
The uppercase is just a matter of trends/fashion, it's being used quite often for sidebar thingys, categories, etc.

jhass:
Mmh, but it doesn't work well for localization afaik

AugierLe42e:
@jhass : What do you have in mind as problems ? No opinion on this.
I like the color, same thing than #6112, cleanier, easier to read.

pablocubico:
Man, I thought that wasn't a problem anymore. Still is for some chars. I guess we'll have to ditch it.

http://www.w3.org/International/tests/repo/results/text-transform

I don't know too much about those tests, anyway, I can't find the proper dating (except an "updated on" mark) or the browser versions... meh.

jhass:
Even if the transform would work, consider scripts that don't have a case distinction, I could imagine it looks weird if used as a design element which then simply is missing in those scripts.

pablocubico:
Not sure if I follow, by scripts you mean the locale texts that will take those places?

Anyway, I didn't take localization into account, it's always a problem for capitalization and such, so we better get that out.

jhass:
Script is used as term for writing systems, e.g. Latin, Cyrillic, Greek etc, see https://en.wikipedia.org/wiki/Writing_system

pablocubico:
Uppercasing removed:

captura de pantalla 2015-06-17 a la s 22 08 22

pablocubico:
Better capture:

captura de pantalla 2015-06-17 a la s 22 10 16

jaywink:
Looks cleaner without the icons imho, :+1:

goobertron:
Thanks for this work, Pablo. I like what you've done a lot.

Re capitalisation, we had a vote in which we decided to use sentence case (i.e. normal capitalisation for prose text) throughout the UI for English locale. See [this Loomio proposal](https://www.loomio.org/d/kY0wClao/capitalisation-of-the-entire-ui-in-the-english-language?proposal=9DylKaxP). So your revised PR fits this decision.

My only comment would be to have a max-width for this div. Probably that white input field for the invitation doesn't want to get any wider than it is in your very first screen-shot. I'm thinking of really wide screens - it would be good to have a side-bar of a certain width only, with the grey background extending to the edge of the screen - just as with the left side-bar. You might already have set a max-width - I can't tell without poking around in the code, which I'm afraid I don't have time to do, so I hope you don't mind this (possibly unnecessary) comment.

pablocubico:
I added an inner element with a 300px max-width on it, I have a 1920px screen so I'll send some screenshots.

I think this approach is the most reasonable.

captura de pantalla 2015-06-22 a la s 19 32 11
captura de pantalla 2015-06-22 a la s 19 32 00
captura de pantalla 2015-06-22 a la s 19 31 49
captura de pantalla 2015-06-22 a la s 19 31 37
captura de pantalla 2015-06-22 a la s 19 31 29
captura de pantalla 2015-06-22 a la s 19 31 19

pablocubico:
Right sidebar following same approach as the left one (see: #6112)

captura de pantalla 2015-06-22 a la s 20 32 55
captura de pantalla 2015-06-22 a la s 20 32 31
captura de pantalla 2015-06-22 a la s 20 32 22
captura de pantalla 2015-06-22 a la s 20 32 14
captura de pantalla 2015-06-22 a la s 20 32 03

AugierLe42e:
Just like #6112, love it. Just one thing: can you add `.hidden-sm.hidden-xs` to `.rightBar` ?

svbergerem:
Superseded by #6309. Thank you.

PR #6049, opened 2015-06-04
Fix for too long username in notification dropdown

SansPseudoFix:
Fix for [Long username breaks notifications dropdown](https://github.com/diaspora/diaspora/pull/6015#issuecomment-107565374).
Before:

bs3_diaspora_too_long_pseudo_notification

After:

bs3_diaspora_too_long_pseudo_notification_ok2

PR #6116, opened 2015-06-18
Fix issue with too long username and post link in notifications dropdown

SansPseudoFix:
To avoid this:

overflow_too_long_username_dropdown_wrong

I pushed [this PR](https://github.com/diaspora/diaspora/pull/6049), but it breaks display too:

overflow_too_long_username_dropdown_mistake

With this PR, I select `a` to be sure:

overflow_too_long_username_dropdown_ok

Sorry for the previous PR.

### DIFF
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -177,7 +177,10 @@ body > header {
           }
 
           .media-body {
-            word-break: break-all;
+            a {
+              display: inline-block;
+              word-break: break-all;
+            }
           }
 
           .pull-right > .aspect_membership_dropdown { display: none; }


goobertron:
Thanks for this work! It would be really nice if the long post title in the first notification in your example could start on the same line as the preceding text, i.e.

> a really long nick name for test commented on
> your post ThisisareallylongtitlebecauseIwantto
> testsomethingareyouokwithit?.

Would this be possible? If we're forcing a line break in the title, hopefully it wouldn't be difficult.

Also, the full stop that follows the post title is on a new line. Again, it would be good to have this on the same line as the end of the title - or even to omit the full stop if the last character of the title is a punctuation stop (`.`, `!` or `?`) - but I guess that's a different PR...!

Thanks again.

SansPseudoFix:
I have to confess I don't know how do better. This PR is here to find someone to help me with this strange behavior, but indeed, I prefer display the begining of the sentence in the same line =/

And, I really don't know why this ````.```` is here! I have only a interrogation mark.

I'll keep trying.

denschub:
Obsoleted by our bootstrap "relaunch", closing.

PR #6117, opened 2015-06-18
Add link to pod statistics in right navigation

SansPseudoFix:
Result:

link_for_stat

### DIFF
--- a/app/views/shared/_links.haml
+++ b/app/views/shared/_links.haml
@@ -2,6 +2,7 @@
 %li= link_to 'wiki', "https://wiki.diasporafoundation.org"
 %li= link_to t('layouts.application.whats_new'), changelog_url
 %li= link_to t('layouts.header.code') + " " + pod_version, "#{source_url}", {:title => t('layouts.application.source_package')}
+%li= link_to t("layouts.application.statistics_link"), statistics_path
 %li= link_to(t('layouts.application.toggle'), toggle_mobile_path)
 - if AppConfig.settings.terms.enable?
     %li= link_to(t('_terms'), terms_path)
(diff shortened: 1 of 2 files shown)


app/views/shared/_links.haml
```diff
@@ -2,6 +2,7 @@
 %li= link_to 'wiki', "https://wiki.diasporafoundation.org"
 %li= link_to t('layouts.application.whats_new'), changelog_url
 %li= link_to t('layouts.header.code') + " " + pod_version, "#{source_url}", {:title => t('layouts.application.source_package')}
+%li= link_to t('layouts.application.statistics_link'), statistics_path
```
svbergerem: please use doublequotes

goobertron:
Good idea. What would you think about using 'Network statistics' as the label, to make it more clear what these statistics are?

app/views/shared/_links.haml
SansPseudoFix: Okay. Do I do the same for other lines, or just mine?

SansPseudoFix:
Yes, I didn't know which word to use, so I just used "Statistics". To be more specific, I can use "Pod statistics" maybe ("Network statistics" can be confused -> all diaspora* network).

goobertron:
Oh, I'm sorry. Without checking the code, I assumed this  was a link to the network statistics page (http://the-federation.info/). If it is to the pod's own statistics page, then yes, 'Pod statistics' would be appropriate. I suppose this proves that 'Statistics' alone would be confusing!

goobertron:
(And now I noticed that your PR has 'pod statistics' in its title - time for another eye test...)

SansPseudoFix:
:D 
But you're right, "Statistics" only is confusing. I will change that for "Pod statistics".

app/views/shared/_links.haml
svbergerem: Just change the lines you already changed.

SansPseudoFix:
Done.

svbergerem:
Thank you.

PR #6118, opened 2015-06-18
Fixed more-button wording when there are <15 posts

### LINKED ISSUE #6093: Incorrect wording / functionality of the More button in Mobile view
The introduction of v0.5.0.0 seems to have brought a new issue in with regards to the wording and/or functionality of the More button in Mobile view.

Steps to reproduce:

1. Switch to mobile view
2. Head over to something like */tags/exmachina* and you’ll see a small number of posts
3. At the bottom, where the 'More' button usually appears, you'll see a button saying, “There are no posts yet.”

Well, there *are* posts there...

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/21655766-incorrect-wording-functionality-of-the-more-button-in-mobile-view?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Faldrian:
Added the scenario where there are <15 posts and the user should see "You have reached the end of the stream.".

"More" is only shown when there are 15 (maximum posts on a page) or more (we won't know, since the amount of items is limited by 15) posts. This can lead to one page showing 15 posts and the "More"-Button and when the button is clicked the next page presents a "You have reached the end of the stream."-Message.
Technically correct and without additional data from the database / an extra query I don't think this could get more convenient.

This fixes #6093

### DIFF
--- a/features/mobile/more-button.feature
+++ b/features/mobile/more-button.feature
@@ -1,5 +1,5 @@
 @javascript @mobile
-Feature: posting from the mobile main page
+Feature: using the more button on mobile stream
     As a mobile user
     I want to navigate the stream
     And I want to test the text of the more-button in different environments
@@ -23,21 +23,8 @@ Feature: posting from the mobile main page
 
     Scenario: There are 15 posts
       Given I am on the home page
+      Given there are 15 public posts from "bob@bob.bob"
       And "bob@bob.bob" has a public post with text "post 1"
-      And "bob@bob.bob" has a public post with text "post 2"
-      And "bob@bob.bob" has a public post with text "post 3"
-      And "bob@bob.bob" has a public post with text "post 4"
-      And "bob@bob.bob" has a public post with text "post 5"
-      And "bob@bob.bob" has a public post with text "post 6"
-      And "bob@bob.bob" has a public post with text "post 7"
-      And "bob@bob.bob" has a public post with text "post 8"
-      And "bob@bob.bob" has a public post with text "post 9"
-      And "bob@bob.bob" has a public post with text "post 10"
-      And "bob@bob.bob" has a public post with text "post 11"
-      And "bob@bob.bob" has a public post with text "post 12"
-      And "bob@bob.bob" has a public post with text "post 13"
-      And "bob@bob.bob" has a public post with text "post 14"
-      And "bob@bob.bob" has a public post with text "post 15"
 
       When I go to the stream page
       Then I should see "More"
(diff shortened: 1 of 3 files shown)


features/mobile/more-button.feature
```diff
@@ -0,0 +1,71 @@
+@javascript @mobile
+Feature: posting from the mobile main page
```
svbergerem: I think you forgot to change this line. ;-)

features/mobile/more-button.feature
```diff
@@ -0,0 +1,71 @@
...
+      And "bob@bob.bob" has a public post with text "post 6"
+      And "bob@bob.bob" has a public post with text "post 7"
+      And "bob@bob.bob" has a public post with text "post 8"
+      And "bob@bob.bob" has a public post with text "post 9"
+      And "bob@bob.bob" has a public post with text "post 10"
+      And "bob@bob.bob" has a public post with text "post 11"
+      And "bob@bob.bob" has a public post with text "post 12"
+      And "bob@bob.bob" has a public post with text "post 13"
+      And "bob@bob.bob" has a public post with text "post 14"
+      And "bob@bob.bob" has a public post with text "post 15"
```
svbergerem: I think this should be a new step definition. You can find a similar one here https://github.com/Faldrian/diaspora/blob/6093-fix-mobile-pagination-wording/features/step_definitions/user_steps.rb#L102.

This one could be
```ruby
Given /^(\d) public posts from "([^\"]*)"$/ do |n_posts, username|
```

Faldrian: Okay, will do. In the past I had added a step and got the comment that there is already a step doing what I wanted, so this time I wanted to do as much as possible with existing steps.

features/step_definitions/posts_steps.rb
```diff
@@ -27,6 +27,13 @@
   user.post(:status_message, :text => text, :public => true, :to => user.aspect_ids)
 end
 
+Given /^there are (\d+) public posts from "([^"]*)"$/ do |n_posts, email|
+  user = User.find_by_email(email)
+  (1..n_posts.to_i).each do |n|
+    user.post(:status_message, :text => "post nr. #{n}", :public => true, :to => user.aspect_ids)
```
svbergerem: I don't know why hound isn't complaining here but I think that you should use the new ruby hash syntax instead.

Faldrian: what syntax? I looked for how to evaluate a variable inside a string and found this. How to do it "right"?

jhass: Steffen is referring to the shorter syntax for hashes with Symbol keys introduced with Ruby 1.9: `user.post(:status_message, text: "post nr. #{n}", public: true, to: user.aspect_ids)`

Faldrian:
travis is stuck somewhere again

svbergerem:
Thank you. :)

RichKavanagh:
Great work @Faldrian - thanks man! :)

PR #6123, opened 2015-06-19
Fix broken avatar getting start page

SansPseudoFix:
Result:

6119-avatar-ok

### DIFF
--- a/app/assets/stylesheets/getting-started.scss
+++ b/app/assets/stylesheets/getting-started.scss
@@ -1,5 +1,5 @@
 #hello-there {
-  .avatar {
+  #profile_photo_upload .avatar {
     width: 200px;
     height: 200px;
   }
--- a/app/views/people/_add_contact.haml
+++ b/app/views/people/_add_contact.haml
@@ -3,6 +3,7 @@
   .media
     .pull-right
       = aspect_membership_dropdown(contact, inviter, false)
-    = person_image_link(inviter, :size => :thumb_small, :class => 'img')
-    .bd
+    .media-left
+      = person_image_link(inviter, size: :thumb_small, class: "media-object")
+    .media-body
       = person_link(inviter)


app/assets/stylesheets/getting-started.scss
```diff
@@ -1,7 +1,11 @@
 #hello-there {
-  .avatar {
-    width: 200px;
-    height: 200px;
```
svbergerem: This should set the width and height of the new avatar which the user might upload on the getting started page so removing these lines is not an option. The avatar in the well has a default width and height of 50px so we don't have to set that explicitly. (see below)

To have an avatar with the correct size you only need to make sure that this style will only be applied to the uploaded avatar and not to any other avatar on the page.

app/assets/stylesheets/getting-started.scss
```diff
@@ -1,7 +1,11 @@
...
+  .well {
+    padding-bottom: 30px;
+  }
+
+  .well .avatar {
+    width: 50px;
+    height: 50px;
```
svbergerem: This is not needed. (see above)

app/assets/stylesheets/getting-started.scss
```diff
@@ -1,7 +1,11 @@
 #hello-there {
-  .avatar {
-    width: 200px;
-    height: 200px;
+  .well {
+    padding-bottom: 30px;
+  }
```
svbergerem: You could also fix the missing padding by updating `app/views/people/_add_contact.haml` to use correct Bootstrap 3 media objects. (See http://getbootstrap.com/components/#media and https://github.com/diaspora/diaspora/blob/develop/app/views/conversations/_message.haml)

For me that would be the preferred solution.

app/assets/stylesheets/getting-started.scss
SansPseudoFix: I removed it because it didn't affect the new avatar.

> To have an avatar with the correct size you only need to make sure that this style will only be applied to the uploaded avatar and not to any other avatar on the page.

D'oh! You're right, I've even not thought about it, sorry. I will do.

app/assets/stylesheets/getting-started.scss
SansPseudoFix: Thanks! Not a good PR at all.

svbergerem: I neither said nor thought that. Your PR would be a correct solution for the problem. I'd just prefer changing the mentioned file to make Bootstrap apply the correct styles for us.

app/assets/stylesheets/getting-started.scss
svbergerem: > I removed it because it didn't affect the new avatar.

It does affect the new avatar:

new user avatar

SansPseudoFix:
6119-avatar-ok2

(I will squash my commit if this PR is good)

app/views/people/_add_contact.haml
```diff
@@ -3,6 +3,6 @@
   .media
     .pull-right
       = aspect_membership_dropdown(contact, inviter, false)
-    = person_image_link(inviter, :size => :thumb_small, :class => 'img')
+    = person_image_link(inviter, :size => :thumb_small, :class => 'media-left')
```
svbergerem: Given the examples on http://getbootstrap.com/components/#media I'd say this should have class `media-object`, wrapped by a `.media-left` div. Also, while touching this line, please use double quotes and the new ruby hash syntax (`size: :thumb_small, class: "media-object"`).

app/views/people/_add_contact.haml
```diff
@@ -3,6 +3,6 @@
...
     .bd
```
svbergerem: When you change this to `.media-body` the username will be to the right of the avatar.

SansPseudoFix:
6119-well

app/views/people/_add_contact.haml
```diff
@@ -3,6 +3,7 @@
   .media
     .pull-right
       = aspect_membership_dropdown(contact, inviter, false)
-    = person_image_link(inviter, :size => :thumb_small, :class => 'img')
-    .bd
+    .media-left
+      = person_image_link(inviter, :size => :thumb_small, :class => "media-object")
```
svbergerem: Just fix the hash syntax
```ruby
= person_image_link(inviter, size: :thumb_small, class: "media-object")
```
and squash the commits. Then this should be ready to merge.

SansPseudoFix:
Done!

svbergerem:
Thank you. :)

Issue #6021, opened 2015-05-31
Welcome messages to users


TwizzyDizzy:
Hi folks,

a feature request: It would be great to, as a podmin, have a sort of default welcome message you could post to a newly registered user's inbox to show them how things are handled on your pod.

There could be config options in diaspora.yaml:

* welcome_sendmsg: BOOLEAN, Determines if welcome messages shall be sent
* welcome_msg: VARCHAR The text of the message, maybe containing a token (e.g. [USERNAME]) to be replaced with the user's name. Diaspora shall provide a sound default. (I would write one, if this feature request gets accepted)

Cheers
Thomas

fap-:
I would like to work on this issue, if this is alright?

If we drop the message in the user's inbox we need a sender account.
Maybe use the admin account specified in 'diaspora.yml' and 
ignore the 'welcome_sendmsg' flag if no admin account is specified?

jhass:
Admin account sounds good, I think falling back to the autofollow_on_join user wouldn't be too bad too, so by default `diasporahq@joindiaspora.com`.

TwizzyDizzy:
Hi folks,

using the specified admin account seems like the way to, ignoring the welcome_sendmsg flag if no account is specified, too.

I'll compose an intriguing ( ;) ) welcome message up until saturday.

Maybe we could shortly agree on the contents. I'd like to mention at least the following things:

* "hi, welcome to this pod"
* some links to sites containing information on how to use diaspora (any ideas?)
* link to legal terms
* some information on whom to contact in case the pod doesn't behave as expected
* some encouragement to support your podmin (bitcoin, paypal)

If something else comes to your mind, just mention it :)

Cheers so far,
Thomas

jaywink:
Aww shucks, sorry, accidental mouse click :dancer: 

Was just going to say "great idea", then decided not to spam with a comment and accidently clicked Close. So, great idea ;)

Flaburgan:
I'm definitely in favour of introducing the software to the new users (and @SansPseudoFix probably has suggestion about that, he is welcoming manually new users on framasphere for a while now) but I would prefer to improve the already existing "getting_started" panels. You can reactivate it in the settings if you forgot what is this about, but to summarize, there is a welcome message, a presentation of the publisher, a presentation of the audience control component and a presentation of the stream. We can help the new user discovers the other part of diaspora* with this mechanism.

jaywink:
[...]

PR #6128, opened 2015-06-20
Adds support for automated welcome message

fapdash:
This works on #6021

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -24,7 +24,8 @@ class Conversation < ActiveRecord::Base
   def local_recipients
     recipients.each do |recipient|
       if recipient.local?
-        if recipient.owner.contacts.where(:person_id => self.author.id).count == 0
+        unless recipient.owner.contacts.where(person_id: author.id).any? ||
+            (author.owner && author.owner.podmin_account?)
           errors.add(:all_recipients, "recipient not allowed")
         end
       end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -468,12 +468,29 @@ class User < ActiveRecord::Base
     aq
   end
 
+  def send_welcome_message
+    return unless AppConfig.settings.welcome_message.enabled? && AppConfig.admins.account?
+    sender_username = AppConfig.admins.account.get
+    sender = User.find_by(username: sender_username)
+    conversation = sender.build_conversation(
+      participant_ids: [sender.person.id, person.id],
+      subject:         AppConfig.settings.welcome_message.subject.get,
+      message:         {text: AppConfig.settings.welcome_message.text.get % {username: username}})
+    if conversation.save
+      Postzord::Dispatcher.build(sender, conversation).post
+    end
+  end
+
   def encryption_key
     OpenSSL::PKey::RSA.new(serialized_private_key)
   end
 
   def admin?
     Role.is_admin?(self.person)
+  end
+
+  def podmin_account?
+    username == AppConfig.admins.account
   end
 
   def mine?(target)
--- a/config/diaspora.yml.example
+++ b/config/diaspora.yml.example
@@ -410,6 +410,23 @@ configuration: ## Section
     ## please consider resharing diaspora* HQ's posts for your pod's users!
     #autofollow_on_join_user: 'diasporahq@joindiaspora.com'
 
+    ## Welcome Message on registration (default=false)
+    ## Send a message to new users after registration
+    ## to tell them about your pod and how things
+    ## are handled on it.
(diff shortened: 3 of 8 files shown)


app/models/user.rb
```diff
@@ -468,6 +468,20 @@ def seed_aspects
     aq
   end
 
+  def send_welcome_message
+    unless AppConfig.settings.welcome_message? && AppConfig.admins.account? then return end
```
jaywink: The syntax here is a bit confusing imho, what about more verbosity?

    unless AppConfig.settings.welcome_message? && AppConfig.admins.account?
      return
    end

jhass: I'd vote for modifier unless:
```rb
return unless ...
```

app/models/user.rb
```diff
@@ -468,6 +468,20 @@ def seed_aspects
...
+    sender_username = AppConfig.admins.account.get
+    sender = User.find_by_username(sender_username)
```
jhass: Aren't those finders deprecated in favour of `User.find_by(username: sender_username)`?

app/models/user.rb
```diff
@@ -476,6 +490,10 @@ def admin?
     Role.is_admin?(self.person)
   end
 
+  def is_podmin?
```
jhass: I'd go for `podmin_account?` even.

app/models/user.rb
```diff
@@ -468,6 +468,20 @@ def seed_aspects
...
   end
 
+  def send_welcome_message
+    unless AppConfig.settings.welcome_message? && AppConfig.admins.account? then return end
+    sender_username = AppConfig.admins.account.get
+    sender = User.find_by_username(sender_username)
+    conversation = sender.build_conversation(
+      participant_ids: [sender.person.id, self.person.id],
+      subject:         AppConfig.settings.welcome_message_subject.get,
+      message:         {text: AppConfig.settings.welcome_message_text.get.gsub("[USERNAME]",
```
jhass: How about using `String#%`, `"Hello, %{username}" % {username: "alice"}`

config/diaspora.yml.example
```diff
@@ -410,6 +410,23 @@ configuration: ## Section
...
+    ## The content of your welcome message.
+    ## The placeholder '[USERNAME]' will be replaced by the username
+    ## of the new user.
+    #welcome_message_text='Hello [USERNAME], welcome to diaspora.'
+
+    ## Welcome Message subject (default='Welcome Message')
+    ## The subject of the conversation that is started
+    ## by your welcome message.
+    #welcome_message_subject='Welcome Message'
+
```
jhass: I think this would make a nice subsection.

```yaml
welcome_message:
  enabled: false
  subject: "..."
  text: "..."
```

Also note that `foo=bar` is not valid YAML afaik.

spec/models/user_spec.rb
```diff
@@ -934,6 +934,58 @@
     end
   end
 
+  describe "#send_welcome_message" do
+    let(:user) {
```
jhass: I don't think you need a multiline block at all here, single line should be fine.

spec/models/user_spec.rb
```diff
@@ -934,6 +934,58 @@
...
+      FactoryGirl.create(:user)
+    }
+    let(:podmin) {
+      FactoryGirl.create(:user)
+    }
+
+    context "with welcome message enabled" do
+      it "should send welcome message from podmin account" do
+        AppConfig.settings.welcome_message = true
```
jhass: At least this one you can extract into a `before` hook for the context.

app/models/user.rb
fap-: Cool, didn't know about that function.

app/models/conversation.rb
```diff
@@ -24,7 +24,7 @@ def max_participants
   def local_recipients
     recipients.each do |recipient|
       if recipient.local?
-        if recipient.owner.contacts.where(:person_id => self.author.id).count == 0
+        if (recipient.owner.contacts.where(person_id: author.id).count == 0) && (!author.owner.podmin_account?)
```
jhass: Actually the parens seem unnecessary here, we could also get rid of the negation here: `unless recipient.owner.contacts.where(person_id: author.id).any? || author.owner.podmin_account?`

Thinking about this, are we sure the author is local? I think it's not guaranteed that we have an owner.

jhass:
Merged as 1061e101fde21f7acb7659ff701a4cf6f9c3d0f3 073f028f88d1989f4149706c7c852bd71966e12e

Thank you!

PR #6130, opened 2015-06-21
Fix autocomplete

svbergerem:
- position: fixed for search results: otherwise the search results move when you scroll down / up
- use the background color of the navbar for search results. Saw that on framaspere-redesign and thought that it would be also nice for the default diaspora\* design

search results


jhass:
Sure, let's try it.

Issue #6143, opened 2015-06-28
Add ability to access JSXC chat from all pages not just the main stream


KentShikama:
https://github.com/diaspora/diaspora/issues/5398#issuecomment-70303170
##

kas70:
+1

SuperTux88:
The chat was removed with #8069, this is obsolete now.

PR #6144, opened 2015-06-28
Add ability to access chat from all pages #6143

KentShikama:
I tested this on a number of pages using localhost:3000 across two accounts on Firefox and Chrome.

### DIFF
--- a/app/assets/javascripts/jsxc.js
+++ b/app/assets/javascripts/jsxc.js
@@ -3,7 +3,7 @@
 // initialize jsxc xmpp client
 $(document).ready(function() {
   if (app.currentUser.authenticated()) {
-    $.post("/api/v1/tokens", null, function(data) {
+    $.post(Routes.apiV1Tokens(), null, function(data) {
       if (jsxc && data['token']) {
         var jid = app.currentUser.get('diaspora_id');
         jsxc.init({
--- a/app/assets/javascripts/jsxc.js
+++ b/app/assets/javascripts/jsxc.js
@@ -3,7 +3,7 @@
 // initialize jsxc xmpp client
 $(document).ready(function() {
   if (app.currentUser.authenticated()) {
-    $.post('/api/v1/tokens', null, function(data) {
+    $.post("/api/v1/tokens", null, function(data) {
       if (jsxc && data['token']) {
         var jid = app.currentUser.get('diaspora_id');
         jsxc.init({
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/jsxc.js
```diff
@@ -3,7 +3,7 @@
 // initialize jsxc xmpp client
 $(document).ready(function() {
   if (app.currentUser.authenticated()) {
-    $.post('api/v1/tokens', null, function(data) {
+    $.post("/api/v1/tokens", null, function(data) {
```
jhass: Actually please switch this to the `Routes` helper while touching it.

jhass:
The downside is that because diaspora is not (yet) a single page app, this will increase the sign out/sign in stuff quite a bit, I already get online notifications from JSXC every time I reload my stream.

Did you also test this on the signed out routes? /users/sign_in, /users/sign_up, the default home page etc.

theworldbright:
@jhass Yeah it does not work as expected on those 3 pages or more like it silently fails. Hmm…is there a way to only load the chat once? Do we already use turbolinks?

app/views/layouts/application.html.haml
```diff
@@ -40,6 +40,10 @@
     = current_user_atom_tag
 
     = yield(:head)
+    - if AppConfig.chat.enabled?
```
jhass: I guess you could just add `&& user_signed_in?`

jhass:
No we don't use turbolinks and I'd expect it to break enough of our JS that making everything a backbone view and use backbones router instead would be easier.

theworldbright:
@jhass, @Zauberstuhl Actually I am a bit confused why users are being re-logged in on page refresh. JSXC already supports pre-binding, so how come chat sessions are not being persisted across page loads?

Zauberstuhl:
@theworldbright a few things changed in the last months. I have no trouble reloading pages in my dev setup and on sechat.org 

besides that in the main-stream we have a container for jsxc(to avoid overlays) https://github.com/Zauberstuhl/jsxc/blob/caa6ceb93583bd4818fe1cb9fa067641c434fc35/src/jsxc.lib.gui.js#L1483

we should do the same for the rest of the pages, no?!

Flaburgan:
+1 for this, as a user, I don't want to have to ask myself "will the chat be present on the page I'm going or will I be disconnected" each time I click on a link.

Flaburgan:
The chat doesn't log out / log in at each refresh now. Is that solution still a problem?

Issue #6136, opened 2015-06-23
"Email/Username is already taken" modal appears on the sign in page after failed sign up


theworldbright:
Steps to reproduce:

1. Sign up with an email and username that has already been taken
2. Click the "Sign up" button
3. The "Email is already taken. -Username is already taken.-That username has already been taken" modal should appear
4. Click sign in in the header
5. The "Email is already taken. -Username is already taken.-That username has already been taken" modal appears again on the sign in page

PR #6146, opened 2015-06-29
Fix bug where flash error reappeared in sign in page #6136

KentShikama:


denschub:
Merged as d410d2ee761e0d9c1d940142c9ff59425acc5b25, thank you very much!

Issue #5963, opened 2015-05-16
Implement RFC 7033 Webfinger


Flaburgan:
As voted on https://www.loomio.org/d/gfBaSiAE/implement-rfc7033-webfinger we should improve our webfinger implementation. @SuperTux88 I saw you were touching things like that recently, would you want to have a look?
##

denschub:
We should not **update** our existing implementation, we have to add new methods. Our implementation needs to be backwards compatible.

Flaburgan:
Thanks Dennis I edited the issue :) Feel free to edit my post when I make mistakes ;)

Flaburgan:
For the record, https://github.com/diaspora/diaspora/pull/6310 is now merged.

SuperTux88:
this is still the old webfinger. I have not implemented the RFC-webfinger yet. But it is prepared to add it :)

yookoala:
Let's make a checklist for [RFC7033 Webfinger](https://tools.ietf.org/html/rfc7033) compatible implementation in case someone wants to work on it. From what I can see:

- [Section 4. Webfinger Protocol](https://tools.ietf.org/html/rfc7033#section-4)
  - [ ] The path component of a WebFinger URI MUST be the well-known path "`/.well-known/webfinger`".

  - [ ] A WebFinger URI MUST contain a query component that encodes the query target and optional link relation types as specified in [Section 4.1](https://tools.ietf.org/html/rfc7033#section-4.1).

- [Section 4.1. Constructing the Query Component of the Request URI
](https://tools.ietf.org/html/rfc7033#section-4.1)
  - [ ] A WebFinger client issues a query using the GET method to the well-
   known [[3](https://tools.ietf.org/html/rfc7033#ref-3)] "`resource`" identified by the URI whose path component is "`/.well-known/webfinger`" and whose query component MUST include the "`resource`" parameter exactly once and set to the value of the URI for which information is being sought.

  - [ ] If the "`resource`" parameter is absent or malformed, the WebFinger resource MUST indicate that the request is bad as per [Section 10.4.1 of RFC 2616](https://tools.ietf.org/html/rfc2616#section-10.4.1) [[2](https://tools.ietf.org/html/rfc7033#ref-2)].

cmrd-senya:
So now this feature has been merged in `diaspora_federation`, so after merging #7436 this will be fixed.

SuperTux88:
[...]

PR #6151, opened 2015-07-01
Federation Gem Step 1: Webfinger/HCard generation

### LINKED ISSUE #5143: Drop caches_page
We should not use features that got removed from Rails core and extracted into separate gems. Especially in this case since Rails 4 provides new improved caching facilities.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4026390-drop-caches_page?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SuperTux88:
I'm working [on a new federation gem](https://github.com/SuperTux88/diaspora_federation). This is built as a rails engine, so it can simply be mounted to a rails-app and provide the URLs that are needed for the federation. Some parts are based on @Raven24's old federation gem. After the diaspora_federation gem is complete, we can move this to the diaspora organization.

I also integrate the new gem into diaspora and I want to create several small pull requests for this, so that it is better to review with small changes.

In the first step I replaced the host-meta, webfinger and hcard generation with the new diaspora_federation gem. Everything else is still working as before.

I wrote a new spec, which generates the fixtures with the new controllers. So the existing parser is tested with the new generator. This can be removed again when the parsing is also done in the gem.

Also done in this PR:
- refactored the atom_url to contain the .atom suffix
- dropped caches_page, this fixes #5143
- removed unused publics_helper, this was unused since 2010
- don't get `person.url` from the hcard, this is already set from the webfinger

@denschub reviews my work at the gem on a regular basis, I think he can also review this PR :)


app/models/person.rb
```diff
@@ -241,15 +238,52 @@ def exported_key= new_key
...
+      url:              url,
+      photo_large_url:  image_url,
+      photo_medium_url: profile.image_url_medium,
+      photo_small_url:  profile.image_url_small,
+      pubkey:           serialized_public_key,
+      searchable:       searchable,
+      first_name:       first_name,
+      last_name:        last_name
+    }
+  end
```
denschub: I am not sure if I like those hashes. It would be nicer to just have a list of attributes that are included.

spec/workers/publish_to_hub_spec.rb
```diff
@@ -2,15 +2,15 @@
...
+require "spec_helper"
 
 describe Workers::PublishToHub do
-  describe '.perform' do
-    it 'calls pubsubhubbub' do
-      url = "http://publiczone.com/"
-      m = double()
+  describe ".perform" do
+    it "calls pubsubhubbub" do
+      url = "http://publiczone.com/public/username.atom"
```
denschub: Please use example.com everywhere.

Flaburgan:
Wow, awesome!

Gemfile
```diff
@@ -15,6 +10,10 @@ gem "responders", "2.1.0"
 
 gem "unicorn", "4.9.0", require: false
 
+# Federation
+
+gem "diaspora_federation", "0.0.1"
```
jaywink: :cookie: :+1:

app/models/person.rb
```diff
@@ -241,15 +238,52 @@ def exported_key= new_key
...
+
+  def webfinger_hash
+    {
+      acct_uri:    "acct:#{diaspora_handle}",
+      alias_url:   "#{url}people/#{guid}",
+      hcard_url:   url[0...-1] + DiasporaFederation::Engine.routes.url_helpers.hcard_path(guid),
+      seed_url:    url,
+      profile_url: profile_url,
+      atom_url:    atom_url,
+      salmon_url:  receive_url,
```
jhass: How about cleaning up all the URL generation a bit and making it more robust:

```ruby
def uri
  @uri ||= URI.parse(url)
  @uri.dup
end

def url_to(path)
  uri.tap {uri|  uri.path = path }.to_s
end
```

And then while at it also cleaning up `clean_url` (still used, can usages be replaced?) and `update_url` below.

app/models/person.rb
```diff
@@ -241,15 +238,52 @@ def exported_key= new_key
...
+      guid:        guid,
+      pubkey:      serialized_public_key
+    }
+  end
+
+  def hcard_profile_hash
+    {
+      guid:             guid,
+      nickname:         username,
+      full_name:        "#{first_name} #{last_name}",
```
jhass: Both fields are optional, so this may end up to be the empty space. If that's not relevant because the other side never uses that field, then I think we already have a `full_name` method somewhere that can be used. Else this should probably have a `.strip`

app/models/person.rb
```diff
@@ -241,15 +238,52 @@ def exported_key= new_key
...
+      url:              url,
+      photo_large_url:  image_url,
+      photo_medium_url: profile.image_url_medium,
+      photo_small_url:  profile.image_url_small,
```
jhass: That feels inconsistent, either further add to the delegation and use only the delegated helpers, or use all fields from the profile explicitly

app/models/person.rb
```diff
@@ -241,15 +238,52 @@ def exported_key= new_key
...
+      first_name:       first_name,
+      last_name:        last_name
+    }
+  end
+
+  # database calls
   def self.by_account_identifier(identifier)
-    identifier = identifier.strip.downcase.gsub('acct:', '')
-    self.where(:diaspora_handle => identifier).first
+    identifier = identifier.strip.downcase.gsub("acct:", "")
```
jhass: A simple `sub` should be enough and communicate that we only expect one occurrence.

config/initializers/diaspora_federation.rb
```diff
@@ -0,0 +1,8 @@
+# configure the federation engine
+DiasporaFederation.configure do |config|
+  # the pod url
+  config.server_uri = AppConfig.pod_uri
+
+  # the class to be used for a person
+  config.person_class = Person.to_s
```
jhass: I think the accessor should duck type here and call `to_s` on its own. Assigning the actual classes is just too common, especially in the Rails world.

SuperTux88: The Rails documentation says it should be done with a String: http://guides.rubyonrails.org/engines.html#configuring-an-engine (the red block)

> It's very important here to use the String version of the class, rather than the class itself. If you were to use the class, Rails would attempt to load that class and then reference the related table. This could lead to problems if the table wasn't already existing. Therefore, a String should be used and then converted to a class using constantize in the engine later on.

So, because it's red, I thought it's important to use a String here ;) But I can move the .to_s to the gem and still hold it as a String in the gem.

jhass: Yes, I'm just saying it's safer to do in the accessor, since you then 1) don't need that knowledge 2) aren't confused by it since it's common not to in the Rails API (e.g. `has_many :foo, class: Bar`)

app/models/person.rb
SuperTux88: @jhass: Do you know why we need the rescue in the ```url```-method? do we have invalid URLs anywhere? Or can I expect parseable URLs here?

jhass: No I don't know, I'd assume defensive coding.

SuperTux88: I'll have a look at it. But I think that should be a valid URL anyway, or the other generated URLs wouldn't work ...

spec/models/person_spec.rb
```diff
@@ -376,6 +411,62 @@
...
+
+    describe "#webfinger_hash" do
+      let(:hash) { person.webfinger_hash }
+
+      it "should contain all needed keys" do
+        complete = %i(
+          acct_uri alias_url hcard_url seed_url
+          guid profile_url atom_url salmon_url pubkey
+        ).all? {|k| hash.key? k }
+        expect(complete).to be_truthy
```
denschub: Suggestion for more helpful error messages:

```ruby
%i(acct_uri alias_url ...).each do |key|
  expect(hash).to have_key key
end
```

AugierLe42e:
@SuperTux88 : When do you think this first step will be done? There's a discussion going on [here](https://github.com/AugierLe42e/diaspora/pull/5#discussion-diff-33924414) between @theworldbright  and me regarding Webfinger. The thing is that OpenID Connect also uses Webfinger for discovering. So, there's a Q&D implementation of Webfinger in this work. But I'd prefer not to overlap your work, so we agreed to let the Webfinger implementation out of the scope of our work until the rest is done. Then, it would be cool if we can use your work.

SuperTux88:
I'm working as fast as I can :) I have done some refactorings and cleanups to the gem, based on the feedback of this PR. I think I can update this PR in one or two days.

The first step is only the old webfinger as it is now in diaspora. I have also planned to add the RFC-7033 webfinger (see #5963), but I want to add the current functionality first. Also this first step is only the generation and will not yet replace the Webfinger class. The Webfinger class will be replaced in a later step, when I also replace the parsing.

AugierLe42e:
> I'm working as fast as I can :)

Wasn't a criticism at all ;)
Ok, then I think what we'll do, if we are finished before you, is a canned webfinger discovery and see later how it can be integrated into your own work.

config/initializers/diaspora_federation.rb
```diff
@@ -0,0 +1,47 @@
...
+  # the pod url
+  config.server_uri = AppConfig.pod_uri
+
+  config.define_callbacks do
+    on :person_webfinger_fetch do |handle|
+      person = Person.find_local_by_diaspora_handle(handle)
+      if person
+        DiasporaFederation::WebFinger::WebFinger.new(
+          acct_uri:    "acct:#{person.diaspora_handle}",
+          alias_url:   url_to("/people/#{person.guid}"),
```
jhass: I would consider moving this into a model method, like we already have `profile_url` and such.

SuperTux88:
I have moved the url_to to the AppConfig and wrote some tests for the callbacks.

denschub:
Okay, I merged this.

Thank you **very much** for your ongoing work, looking forward to the upcoming pull request. :heart:

Flaburgan:
Yay! This is **awesome**, definitely the most important work on diaspora* for months. Thank you so much for what you're doing.

Fensterbank:
That's definitely great! Keep up the good work. :+1:

denschub:
As discussed with @SuperTux88, I backported this to 0.5.2.0. We merged the other federation refactorings into minors as well and I see no harm in having additional production time here.

PR #6157, opened 2015-07-03
updated hash syntax to ruby 1.9

askl56:
I have fixed the hash syntax in the entire project to the standard ruby 1.9


jhass:
Thank you for your effort but I'm afraid we won't accept this.

Please see #5561 and [the last entry in Hound's FAQ](https://houndci.com/faq) for the reasoning. This was also mentioned in the [Loomio discussion](https://www.loomio.org/d/bLH78pEh/styleguides).

PR #6160, opened 2015-07-03
Remove aspects-dropdown.js

svbergerem:
Looks like we don't use it anymore.


jhass:
Looks like it's still being loaded though ;)

jhass:
Thanks!

Issue #5684, opened 2015-02-18
Allow additional profile fields (previously private profile) to be set public


svbergerem:
@jaywink tried to implement this in #4596 but closed the PR. I wasn't able to find a similar open issue.

cmrd-senya:
In original PR there was mentioned a toggle slider as a way to switch beetwen private and public modes. Is there any examples of the toggle slider used in Diaspora? Is [this](http://www.bootstrap-switch.org/) is what we need?

svbergerem:
Bootstrap-switch seems legit. There is even a [gem](https://rubygems.org/gems/bootstrap-switch-rails) which you can add to our Gemfile.

jhass:
We decided to make this a local only feature for now, since it's a lot easier and the common usecase will be referring somebody to your `/u/username` anyway.

PR #6162, opened 2015-07-03
Allow additional profile fields (previously private profile) to be set public (new updated version)

cmrd-senya:
Pull request for #5684.  

Currently I came up to point, where @jaywink was in #4596.  

There are some issues that require discussion before I can proceed further.

The first is whether I should write some tests for that? And the second I'll append on the diff.


app/assets/javascripts/app/views/profile_sidebar_view.js
```diff
@@ -10,7 +10,7 @@ app.views.ProfileSidebar = app.views.Base.extend({
   },
 
   _shouldShowProfileInfo: function() {
-    return (this.model.isSharing() || this.model.get('is_own_profile'));
+    return true;
```
cmrd-senya: Ok, this is a disputable place. I removed this checks, because they duplicate the same checks, that already have been done in app/presenters/person_presenter.rb (see below) on the server side. @jhass wasn't agree with me on IRC about it. So lets discuss, what is the best way to do it - leave it how I did it, or add check this.model.get('is_public') here just as in PersonPresenter.

db/migrate/20150630221004_add_public_to_profiles.rb
```diff
@@ -0,0 +1,5 @@
+class AddPublicToProfiles < ActiveRecord::Migration
+  def change
+    add_column :profiles, :is_public, :boolean, :default => false
```
jhass: Can't we drop the `is_` prefix here and use the predicate accessor so it becomes idiomatic ruby, `.public?`?

app/presenters/person_presenter.rb
```diff
@@ -24,7 +24,7 @@ def full_hash_with_avatar
   def full_hash_with_profile
     attrs = full_hash
 
-    if own_profile? || person_is_following_current_user
+    if is_public || own_profile? || person_is_following_current_user
```
jhass: My argument was to extract this into a method and also provide its result as attribute in the returned hash, and then use that attribute in the backbone view. That would de-duplicate and centralize the logic without having the client side guess depending on the returned fields.

db/migrate/20150630221004_add_public_to_profiles.rb
cmrd-senya: If it is supported as database field name, sure.

jhass: The database field name would be `public`, AR automatically adds a `foo?` attribute method.

jhass:
Oh and we'll most definitely need some tests for this.

db/migrate/20150630221004_add_public_to_profiles.rb
cmrd-senya: Done in 65f30dc7c6f46177dec025bc421ccebc62494ee6.

app/presenters/person_presenter.rb
cmrd-senya: Probably c16a3b86ffa242bea5c1b31e461caf76597ef120 would make you more happy?

app/assets/javascripts/app/views/profile_sidebar_view.js
```diff
@@ -4,13 +4,7 @@ app.views.ProfileSidebar = app.views.Base.extend({
...
   presenter: function() {
-    return _.extend({}, this.defaultPresenter(), {
-      show_profile_info: this._shouldShowProfileInfo(),
-    });
-  },
-
-  _shouldShowProfileInfo: function() {
-    return (this.model.isSharing() || this.model.get('is_own_profile'));
+    return _.extend({}, this.defaultPresenter(), {});
   }
```
jhass: Can't you just delete the presenter override if no attributes are added?

config/locales/diaspora/ru.yml
```diff
@@ -1403,4 +1407,4 @@ ru:
   welcome: "Добро пожаловать!"
   will_paginate:
     next_label: "вперед &raquo;"
-    previous_label: "&laquo; назад"
\ No newline at end of file
+    previous_label: "&laquo; назад"
```
jhass: Please only edit the English locale, translations are handled via WebTranslateIt.

app/assets/javascripts/app/views/profile_sidebar_view.js
```diff
@@ -1,17 +1,7 @@
...
-
-  presenter: function() {
-    return _.extend({}, this.defaultPresenter(), {
-      show_profile_info: this._shouldShowProfileInfo(),
-    });
-  },
-
-  _shouldShowProfileInfo: function() {
-    return (this.model.isSharing() || this.model.get('is_own_profile'));
-  }
```
svbergerem: I agree with @jhass: the logic should stay in the view.

app/presenters/person_presenter.rb
```diff
@@ -34,14 +31,12 @@ def full_hash_with_profile
...
-        location: @presentable.location,
-        birthday: @presentable.formatted_birthday,
-        bio:      @presentable.bio
-      )
+    attrs = full_hash
+
+    if attrs[:show_profile_info]
+      attrs.merge!(ProfilePresenter.new(profile).private_hash) { |key, v1, v2| v1 }
+    else
+      attrs.merge!(ProfilePresenter.new(profile).public_hash) { |key, v1, v2| v1 }
```
jhass: Which keys conflict here?

app/presenters/person_presenter.rb
```diff
@@ -34,14 +31,12 @@ def full_hash_with_profile
...
-        birthday: @presentable.formatted_birthday,
-        bio:      @presentable.bio
-      )
+    attrs = full_hash
+
+    if attrs[:show_profile_info]
+      attrs.merge!(ProfilePresenter.new(profile).private_hash) {|_, v1, _| v1 }
+    else
+      attrs.merge!(ProfilePresenter.new(profile).public_hash) {|_, v1, _| v1 }
     end
```
jhass: Call `full_hash_with_profile` if you have to include it?!

cmrd-senya: :id conflicts here, we need the one from person, and the one in profile is different.

cmrd-senya: And there is a difference in the format, notice:
```ruby
attrs.merge!(profile: ProfilePresenter.new(profile).private_hash)
```
and
```ruby
attrs.merge!(ProfilePresenter.new(profile).private_hash)
```

jhass: That seems wrong anyway, why can't we return the profile as a subkey? It's a different model after all.

cmrd-senya: We can, but that would change the way it was before and make tests fail.

jhass: Well, this is restructuring it a bit anyway, I would have no issue with the tests being adapted accordingly if everything that depends on the old format is as well.

cmrd-senya: Then how to find users of PersonPresenter.as_json?

jhass: If the cucumber run still passes after the failing specs are adapted,  we should be good.

cmrd-senya: The change to the as_json function make a big number of cucumber tests fail.

jhass: Well, I expect some JS would need to get adjusted too.

cmrd-senya: No, I was wrong. I receive messages "Unexpected modal dialog (text: A script on this page may be busy, or it may have stopped responding. You can stop the script now, or you can continue to see if the script will complete." with both versions of the as_json function. Dunno how to deal with that. Probably my test machine is too weak.

However, I found out, that the as_json function doesn't get called during run of the cucumber testsuite. So its changes can't probably break anything at all.

jhass: Maybe [this call](https://github.com/diaspora/diaspora/blob/develop/app/controllers/people_controller.rb#L78) being the common one should be replaced by as_json, the method made private and as_json just delegate to it?

cmrd-senya: Done in 0a99e6c199e6bff69b1d23eb0b357545e90aa789

cmrd-senya:
Are this tests enough?

app/controllers/profiles_controller.rb
```diff
@@ -79,6 +80,8 @@ def munge_tag_string
   end
 
   def profile_params
-    params.require(:profile).permit(:first_name, :last_name, :gender, :bio, :location, :searchable, :tag_string, :nsfw, :date => [:year, :month, :day]) || {}
+    params.require(:profile).permit(:first_name, :last_name, :gender, :bio,
+                                    :location, :searchable, :tag_string, :nsfw,
+                                    :public_details, :date => %i(year month day)) || {}
```
jhass: Please update this to the Ruby 1.9 hash syntax while touching, `date: %i(year month day)`

app/views/profiles/_edit_private.haml
```diff
@@ -3,7 +3,13 @@
 -#   the COPYRIGHT file.
 
 %hr
-%h3= t('profiles.edit.your_private_profile')
+%h3{:class=>"inline"}
```
jhass: `%h3.inline`

app/views/profiles/_edit_private.haml
```diff
@@ -3,7 +3,13 @@
...
+  = t('profiles.edit.additional')
```
jhass: Double quotes.

app/views/profiles/_edit_private.haml
```diff
@@ -3,7 +3,13 @@
...
+= t('profiles.edit.additional_visibility_text')
+= check_box_tag 'profile[public_details]', true, profile.public_details, {'data-size' => 'mini', 'data-on-text' => t('profiles.edit.public'), 'data-off-text' => t('profiles.edit.private')}
+%span{ :rel => 'facebox', :title => t('profiles.edit.additional_hint') }
```
jhass: `rel: "facebox", title: t("profiles.edit.additional_hint")`

config/locales/diaspora/en.yml
```diff
@@ -1008,8 +1008,13 @@ en:
 
   profiles:
     edit:
-      your_public_profile: "Your public profile"
-      your_private_profile: "Your private profile"
+      basic: "Basic"
+      additional: "Additional"
```
jhass: I'm actually not sure about the wording here, "Additional" on its own feels a bit cut off, but also "Additional profile data" feels a bit technical. @goobertron any ideas?

config/locales/diaspora/ru.yml
jhass: This is still true.

db/migrate/20150630221004_add_public_to_profiles.rb
```diff
@@ -0,0 +1,5 @@
+class AddPublicToProfiles < ActiveRecord::Migration
+  def change
+    add_column :profiles, :public_details, :boolean, :default => false
```
jhass: `default: false`

spec/presenters/person_presenter_spec.rb
```diff
@@ -5,27 +5,38 @@
...
   describe "#as_json" do
-    context "with no current_user" do
-      it "returns the user's public information if a user is not logged in" do
+    context "with no current_user (a user is not logged in)" do
+      it "returns the user's basic profile" do
         expect(PersonPresenter.new(person, nil).as_json).to include(person.as_api_response(:backbone).except(:avatar))
       end
+
+      it "returns the user's additional profile if the user has set additional profile public" do
+        person.profile[:public_details] = true
```
jhass: `person.profile.public_details = true` should work here.

spec/presenters/person_presenter_spec.rb
```diff
@@ -5,27 +5,38 @@
   let(:person) { profile_user.person }
 
   describe "#as_json" do
-    context "with no current_user" do
-      it "returns the user's public information if a user is not logged in" do
+    context "with no current_user (a user is not logged in)" do
```
jhass: I I think the current description is clear, to the presenter current_user being present or not is what matters, it has no concept of that meaning "somebody is signed in"

spec/presenters/person_presenter_spec.rb
```diff
@@ -5,27 +5,38 @@
...
 
     context "with a current_user" do
       let(:current_user) { FactoryGirl.create(:user)}
       let(:presenter){ PersonPresenter.new(person, current_user) }
+      # here private information == addtional user profile, because additional profile by default is private
 
       it "doesn't share private information when the users aren't connected" do
-        expect(presenter.full_hash_with_profile[:profile]).not_to have_key(:location)
+        expect(presenter.as_json[:profile]).not_to have_key(:location)
       end
```
jhass: This case is no longer complete, it needs to differentiate whether public_details is true, like above.

app/views/profiles/_edit_private.haml
svbergerem: #6106 removed all faceboxes so I don't think that this will work.

cmrd-senya: What was wrong with faceboxes? And what is the "real improvement" as stated in #6106?

svbergerem: In the PR I wrote
> real improvements are out of scope of this PR

Bootstrap already has [modals](http://getbootstrap.com/javascript/#modals) so we didn't need the faceboxes anymore.

cmrd-senya: Ok, then I should replace the facebox with a modal?

svbergerem: In most cases: yes. In this case I am not sure. What should the facebox do? AFAIR faceboxes had some URL before to load the content which is missing here. Depending on the content you could use a modal, a [tooltip](http://getbootstrap.com/javascript/#tooltips) or a [popover](http://getbootstrap.com/javascript/#popovers).

cmrd-senya: I'm not quite sure, I derived this code from @jaywink's PR #4596. Maybe @jaywink could tell us?

config/locales/diaspora/en.yml
goobertron: How about 'Your basic profile'/'Your core profile' and 'Your detailed profile'/'Your extended profile'?

Would 'My' read better than 'Your' in these strings? It can make the page feel more like something created by the user themselves, rather than someone else (the server) talking to them. There are pros and cons on both sides, I think. What do others think?

Will the basic profile fields (ID, name, avatar, tags) continue to be public, or will these fields also be optionally public/private? If they will remain public, which seems to me the sensible thing, it might be worth adding a note to this effect under the Profile tab at the top of the page. Perhaps something like:

> Every item in your profile is optional. Your basic profile will always be publicly visible; each item in your extended [or detailed] profile can be set as public or private by you.

One other thing: is 'private' the right word here? It's not being kept private; it's being given a limited scope, rather than public. Perhaps 'limited' or 'contacts only' would be more accurate.

jhass: One clarification: this PR introduces a flag to make either all of the details public or all non-public, it's not a per field setting.

cmrd-senya: I'd vote for 'My basic profile'/'My extended profile'.

I think it is a good idea to change the word "private" to the word "limited" there.

Currently we have a note near additional profile (see 'additional_hint' below). I could add similiar hint saying "Your basic profile will always be publicly visible" to basic profile section.

spec/presenters/person_presenter_spec.rb
```diff
@@ -6,26 +6,43 @@
...
-        expect(presenter.full_hash_with_profile[:profile]).not_to have_key(:location)
+        expect(person.profile.public_details).to be_falsey
+        expect(presenter.as_json[:show_profile_info]).to be_falsey
+        expect(presenter.as_json[:profile]).not_to have_key(:location)
       end
 
       it "has private information when the person is sharing with the current user" do
         expect(person).to receive(:shares_with).with(current_user).and_return(true)
-        expect(presenter.full_hash_with_profile[:profile]).to have_key(:location)
+        expect(person.profile.public_details).to be_falsey
```
cmrd-senya: I wanted to add
```ruby
expect(presenter.as_json[:show_profile_info]).to be_truthy
```
here, but for some reason after I add the line it starts failing at spec/presenters/person_presenter_spec.rb:36 (two lines above) with [the message](https://gist.github.com/cmrd-senya/7fbc54ca94ff44b0335a). Weird.

config/locales/diaspora/en.yml
theworldbright: +1 for 'My basic profile'/'My extended profile'

app/views/profiles/_edit_private.haml
jaywink: This particular hint was shown on mouse-over, I think :P

cmrd-senya: I wiped the facebox mention and it still works the same way on mouse-over. So it probably just not required.

spec/presenters/person_presenter_spec.rb
```diff
@@ -6,26 +6,43 @@
...
       let(:current_user) { FactoryGirl.create(:user)}
       let(:presenter){ PersonPresenter.new(person, current_user) }
+      # here private information == addtional user profile, because additional profile by default is private
 
       it "doesn't share private information when the users aren't connected" do
-        expect(presenter.full_hash_with_profile[:profile]).not_to have_key(:location)
+        expect(person.profile.public_details).to be_falsey
+        expect(presenter.as_json[:show_profile_info]).to be_falsey
+        expect(presenter.as_json[:profile]).not_to have_key(:location)
       end
```
jhass: I'm still missing the variation of this case with `public_details` set to true.

config/locales/diaspora/en.yml
```diff
@@ -1008,8 +1008,14 @@ en:
...
-      your_public_profile: "Your public profile"
-      your_private_profile: "Your private profile"
+      basic: "My basic profile"
+      additional: "My extended profile"
+      settings: "Profile settings"
+      additional_visibility_text: "Visibility of your extended profile:"
+      public: "Public"
+      limited: "Limited"
+      basic_hint: "Every item in your profile is optional. Your basic profile will always be publicly visible."
+      additional_hint: "Click the switch to set your extended profile data visibility. Public means it is visible to the internet, limited means only people who you share with will see this information."
```
jhass: Maybe rename the keys from `additional` to `extended` too :)

jhass:
Thank you!

jaywink:
Awesome stuff, thanks @cmrd-senya , @jhass and everyone else involved!

cmrd-senya:
@jhass, @jaywink, the pleasure is mine :)

Issue #6145, opened 2015-06-29
404 error page Tom is missing


Zauberstuhl:
![](https://pod.tchncs.de/uploads/images/scaled_full_60a007293d8c2faa5308.png)
since [e7d0a978bad4ae403dbbfc5b142e51ee6d80a930](https://github.com/diaspora/diaspora/commit/e7d0a978bad4ae403dbbfc5b142e51ee6d80a930) Tom died ([error_pages.scss#L37](https://github.com/diaspora/diaspora/blob/develop/app/assets/stylesheets/error_pages.scss#L37))
There is something wrong with the CSS, if you remove `postion:fixed` from `error_404` Tom will rise again!

[original post](https://sechat.org/posts/4ee33940ff9801321b0b00163cfa796c#575adc90ffba0132e98014dae9b624ce)

Haidy777:
Should be doable as my first contribution to diaspora.

Zauberstuhl:
I have no idea about CSS and how my solution would work on multiple browser ... but still I want to throw some code into the pot:

replace `position:fixed;` with `background-attachment:fixed;` and at least in firefox it'll be the old 404 page.

lgl

Flaburgan:
The CSS is dirty, I can clean it when I'll be back from vacations.

Haidy777:
@Zauberstuhl That would also fix the problem of Tom, but the "html"-Tag would be still a very small height (sometimes even zero Pixel).
Which isn't a problem in most cases, but I think mobile Safari does sometimes strange things.
(Well, mobile safari does always strange things. ;) )

PR #6169, opened 2015-07-05
Issue #6145: Fixed position of Tom on the 404-Page

Haidy777:
Just one small addition to set the height of html and body to 100%
Testet it with several browsers, didn't notice that it would break anything.

### DIFF
--- a/app/assets/stylesheets/new_styles/_base.scss
+++ b/app/assets/stylesheets/new_styles/_base.scss
@@ -2,6 +2,7 @@ html,
 body {
   /* hack to ensure fixed elements at height: 100%; are in relation to the window */
   max-height : 100%;
+  height: 100%;
 }
 
 body {


goobertron:
Thanks a lot for contributing to Diaspora, it's good to have you on board!

It looks to me as though you can replace `max-height:100%` with `height:100%` rather than adding the second, as `height:100%` should imply `max-height:100%`. But do test to make sure this is OK.

goobertron:
Quickly also, does `height:100%` need to be set for both `html` and `body`? If not, that element could be deleted and `height:100%`  added to the `body` element below. Again, won't have time to test this myself, so it's just a suggestion by looking at that file.

Haidy777:
Body might not be absolutely necessary, but html is.
The main Problem is, that the 'html'-Tag is not big enough.

Body is 100% of the height, but html is not. Because of that Tom is rendered in a space that can't be displayed because it lacks of height.

(Apologies my English, I didn't speak it for a while now and so I got a bit rusty. ;) )

goobertron:
You English is fine! I understand you perfectly. Thank you for explaining that, it makes complete sense.

denschub:
Merged as 990c678cd3a9cc529230cc417f49d2ac874b6324, thak you very much!

Also, goob: 

> You English is fine! 

*Your. (Sorry.)

denschub:
Yikes. Holdon, I had to back this out. I missed that the change was applied globally. No idea how that happened, I am very sorry.

@Haidy777 Please only change the height for `body#error_404` and not globally... :)

Haidy777:
@denschub If you look here: https://github.com/diaspora/diaspora/blob/develop/app/assets/stylesheets/error_pages.scss#L29 you will see that the height of body#error_404 is already set to 100%.
As I explained above, changing the height of body might not be necessary. (Because it is already set to 100%) The problem is that the html-Tag isn't high enough.

To explain what I mean, here are some images:
bildschirmfoto 2015-07-07 um 10 20 20
This is the Version with the current CSS. As you can see, body is the full height of my screen.
bildschirmfoto 2015-07-07 um 10 22 25
In the second you can see the width and height of html in the current CSS.
bildschirmfoto 2015-07-07 um 10 24 50
In picture three you see Tom again with only the value of html's height changed.

Hopefully that helps to understand the problem behind that bug.
Adding the height to html only (above this lines: https://github.com/diaspora/diaspora/blob/develop/app/assets/stylesheets/new_styles/_base.scss#L1-L5) would also work, then the change would only affect all html-Tags. (Which again shouldn't be bad, because max-height is already set)

jhass:
Superseded by #6226 

Thanks!

PR #6173, opened 2015-07-08
Very minor changes in vendor/assets/javascripts/

ghost:


jhass:
Thank you, but the stuff in `vendor/` is upstream code that's not distributed in any other way so we have to copy it over, or at least we didn't replace the copies with another distribution format yet.

Thus please send your changes upstream.

RyDroid:
Ok.
There is no indication of upstream in  "vendor/assets/javascripts/keycodes.js". Where does it come from? There is also the question about the license.

jhass:
Looks like it was added in https://github.com/diaspora/diaspora/commit/0ec2d918bbde3461f624eb137ebe3d8ec6c1c238 and moved to vendor/assets incorrectly, it should go to lib/assets if it's a in source tree library.

PR #6174, opened 2015-07-08
Very minor changes to graphics

ghost:


jhass:
Uhm, why?

RyDroid:
It is a bit lighter and easier to understand without a GUI + HTTPS > HTTP for privacy.

jhass:
I doubt anybody wants to edit any of this by hand, the next edit to them will just readd tags like that, so it's unnecessary code churn.

If you want the URLs fixed you should make that effort in the editors generating them.

RyDroid:
Some URLs are generated by an Adobe proprietay program, so it is impossible to fix except by working in Adobe.

jhass:
"that effort" = submitting an issue to them.

Flaburgan:
There are tools to automatically strip comments and compress svg files to make them lighter. It would be nice to see that applied on our image files.

jhass:
We're not using graphics/ as material on the website but rather as source material to generate things from that are possibly used on the website.

PR #6183, opened 2015-07-09
Refactoring tests on photo saved with and without EXIF data

joebew42:


spec/models/photo_spec.rb
```diff
@@ -45,8 +45,7 @@ def with_carrierwave_processing(&block)
   describe '#diaspora_initialize' do
     before do
       @image = File.open(@fixture_name)
-      @photo = Photo.diaspora_initialize(
-                :author => @user.person, :user_file => @image)
+      @photo = Photo.diaspora_initialize(:author => @user.person, :user_file => @image)
```
jhass: Please update to Ruby 1.9 hash syntax while touching: `author: @user.person, user_file: @image`

spec/models/photo_spec.rb
```diff
@@ -141,36 +141,43 @@ def with_carrierwave_processing(&block)
...
 
   context 'with a saved photo containing EXIF data' do
-    before do
-      @exif_filename  = 'exif.jpg'
-      @exif_name      = File.join(File.dirname(__FILE__), '..', 'fixtures', @exif_filename)
-    end
 
-    it 'should contain EXIF data if user prefer' do
-      @alice_photo  = alice.build_post(:photo, :user_file => File.open(@exif_name), :to => alice.aspects.first.id)
+    let (:base_path)       { File.dirname(__FILE__) }
```
jhass: `let` is a method call, the space between `let` and `(` seems wrong.

joebew42:
https://travis-ci.org/diaspora/diaspora/jobs/70276060

The build failed due a 404 returned from `rubygems.global.ssl.fastly.net`. Is it normal for build to fail for these reasons? Is there some automatic retry?

jhass:
Looks like rubygems has mirror issues atm, we should retry at a later point.

Meanwhile it would be great if you could squash the commits using `git rebase -i develop`. You can also edit commits with `git commit --amend`. Both operations need a `git push -f` afterwards, which is okay in feature branches.

jhass:
There's also still one open style note about the hash syntax and the notes about the quotes are valid too.

jhass:
Merged as e21dad780da22326a5071234abec6d0ad9627326

Thanks!

PR #6191, opened 2015-07-12
Rephrase 'Enter an email'

ghost:
One does not 'enter an email' (should I enter a random email from my Inbox?), instead one enters ones own email address.


jaywink:
Thanks!

PR #6195, opened 2015-07-14
fix federation-gem for ruby 2.0

SuperTux88:
see SuperTux88/diaspora_federation#2


denschub:
Merged as 0401de0e8a25c3ba44a0a1840c71fa9f01d72ec9, thanks.

PR #6122, opened 2015-06-19
Add a ToS design change mention in the changelog, close #6121

Flaburgan:

PR #6196, opened 2015-07-14
Update Changelog to improve terms changes indication.

Flaburgan:
Following https://github.com/diaspora/diaspora/pull/6122

I didn't exactly put the message of @goobertron is this one better?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -28,7 +28,7 @@ bind to an UNIX socket at `unix:tmp/diaspora.sock`. Please change your local
 
 ## Terms of Use design changes
 
-With the port to Bootstrap 3, the app/views/terms/default.haml file changed. If you edited the terms by adding a terms.erb file, you probably want to backup these changes in it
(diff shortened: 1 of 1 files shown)


goobertron:
It's a simpler sentence, but I think it leaves open more possibility of someone misunderstanding and just making a few edits to their existing terms.haml or terms.erb rather than starting with the new default.haml and adding their existing customisations into that new file. I thought the latter approach is the better one, as it reduces the possibility of a podmin missing out some important elements of the new structure when copying over.

However, as far as the English goes, it is perfectly good and perfectly clear about what should be done *if* a podmin reads it carefully enough. If our technical wizards such as @jhass, @denschub, @svbergerem and @jaywink think it is clear enough and won't lead to podmins making the abovce mistake, I'm happy with it. It is definitely a more elegant sentence at the end!

denschub:
I'm fine with it, ans @svbergerem seems to be fine with it, too. I'll merge this now.

PR #6199, opened 2015-07-15
6192 refactor rspec syntax

zaziemo:
first model specs refactored using let instead of before blocks and other minor things


spec/models/account_deletion_spec.rb
```diff
@@ -5,88 +5,73 @@
 require 'spec_helper'
 
 describe AccountDeletion, :type => :model do
+  let(:ad_new)    { AccountDeletion.new(person: alice.person) }
```
jhass: I think it would be worth to write `account_deletion` out here, just use search & replace ;)

spec/models/acts_as_taggable_on_tag_spec.rb
```diff
@@ -1,45 +1,45 @@
 require 'spec_helper'
 
 describe ActsAsTaggableOn::Tag, :type => :model do
+  subject(:tag) { ActsAsTaggableOn::Tag }
```
jhass: One alternative here would be to use `described_class`

spec/models/aspect_membership_spec.rb
```diff
@@ -7,25 +7,25 @@
 describe AspectMembership, :type => :model do
 
   describe '#before_destroy' do
-    before do
-      @aspect = alice.aspects.create(:name => "two")
-      @contact = alice.contact_for(bob.person)
+    let(:aspect)  { alice.aspects.create(name: "two") }
+    let(:contact) { alice.contact_for(bob.person) }
+    let(:am)      { alice.aspects.where(name: "generic").first.aspect_memberships.first }
```
jhass: Again, let's also get rid of abbreviations, `aspect_membership` sounds better here IMO.

spec/models/contact_spec.rb
```diff
@@ -147,73 +143,70 @@
...
-      @person = FactoryGirl.build(:person)
+    let(:contact) { Contact.new }
+    let(:user)    { build(:user) }
+    let(:person)  { build(:person) }
 
-      @contact.user = @user
-      @contact.person = @person
+    before do
+      contact.user = user
+      contact.person = person
```
jhass: The real benefit of let, lazy evaluation, feels a bit lost here. Doesn't doing `Contact.new user: user, person: person` work?

spec/models/conversation_spec.rb
```diff
@@ -111,63 +107,58 @@
...
-          participant_ids:     [peter.person.id, @user1.person.id],
-          subject:             "cool stuff",
-          messages_attributes: [{author: peter.person, text: "hey"}]
+      let(:invalid_hash) do
+        {
+          author: peter.person,
+          participant_ids: [peter.person.id, user1.person.id],
+          subject: "cool stuff", messages_attributes: [{author: peter.person, text: "hey"}]
         }
       end
```
jhass: Please update to Weirichs rule while touching. In `let` we care about the blocks return value, not its side effects, thus `{`/`}` should be used.

jhass:
You can run the style checks locally with [Pronto](https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review)

spec/models/aspect_membership_spec.rb
zaziemo: @jhass yeah, that's what we thought as well but we didn't want to change so much. But then, we'll do it :)

spec/models/contact_spec.rb
zaziemo: We'll try it out.

spec/models/conversation_spec.rb
zaziemo: Yep, we switched in between those two possibilities but decided for this one because of better readability. Thanks for the comment, we'll change that.

spec/models/acts_as_taggable_on_tag_spec.rb
```diff
@@ -1,45 +1,45 @@
...
-        "other\twhitespace\n"            => 'otherwhitespace',
-        'hash#inside'                    => 'hashinside',
-        'f!u@n#k$y%-<c>^h&a*r(a)c{t}e[r]s' => 'funky-characters'
+        "node.js"                          => "nodejs",
+        ".dotatstart"                      => "dotatstart",
+        "you,inside"                       => "youinside",
+        "iam(parenthetical)"               => "iamparenthetical",
+        "imeanit?maybe"                    => "imeanitmaybe",
+        "imeanit!"                         => "imeanit",
+        "how about spaces"                 => "howaboutspaces",
```
jhass: Ignore this one

spec/models/contact_spec.rb
```diff
@@ -7,14 +7,14 @@
 describe Contact, :type => :model do
   describe 'aspect_memberships' do
     it 'deletes dependent aspect memberships' do
-      expect{
+      expect {
```
jhass: Ignore this one

denschub:
I've merged this into develop and stable. Thank you for your contributions!

For future reference: 20 commits for changes on 9 files is a bit too much. Merging the commits would be hard in this PR since you both worked on the same files, but please squash the commits in your further PRs down to a smaller number. If you need help with that, just ping us on IRC.

zaziemo:
@denschub thank you! We will find out how to squash the commits :)

PR #6200, opened 2015-07-15
Better display for stream faces

ghost:
Regression introduced by port to BS3.

Before:
before

After:
after


denschub:
flexbox is strong in this one... but IE will kill us anyway in our new version, so who cares. thanks!

AugierLe42e:
IE still doesn't support lfexboxes !? O__O
Hmmf... How, therefore, is it possible to make something proper ?

denschub:
IE 11 does, so we are fine.

PR #6204, opened 2015-07-17
Fix profile visibility icon

svbergerem:
- set the correct entypo class
- add some margin
- use Bootstrap tooltip
- change cursor from `pointer` to `default`: nothing happens when you click the icon so it shouldn't be `pointer`


jhass:
Thanks!

PR #6205, opened 2015-07-17
Added new translation key

ghost:
Added conversations.new.subject_default: "No subject"

Ref #5999


RichKavanagh:
Also ref #6206

PR #6206, opened 2015-07-17
Updated capitalisation in conversation.rb (v3)

ghost:
Value extracted to a translation key.  This pull request replaces PR #5999

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -83,7 +83,7 @@ class Conversation < ActiveRecord::Base
   end
 
   def subject
-    self[:subject].blank? ? I18n.t('conversations.new.subject_default') : self[:subject]
+    self[:subject].blank? ? I18n.t("conversations.new.subject_default") : self[:subject]
   end
 
   def subscribers(user)
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -83,7 +83,7 @@ class Conversation < ActiveRecord::Base
   end
 
   def subject
-    self[:subject].blank? ? "no subject" : self[:subject]
+    self[:subject].blank? ? I18n.t('conversations.new.subject_default') : self[:subject]
   end
 
   def subscribers(user)


Flaburgan:
For the future, you can update a pull request by pushing to the same branch ;)

app/models/conversation.rb
```diff
@@ -83,7 +83,7 @@ def ordered_participants
   end
 
   def subject
-    self[:subject].blank? ? "no subject" : self[:subject]
+    self[:subject].blank? ? I18n.t('conversations.new.subject_default') : self[:subject]
```
RichKavanagh: I used a single quote as that seemed the most common. Do you mean like this?

> I18n.t("conversations.new.subject_default")

?

RichKavanagh:
REF #6205

SuperTux88:
Please do not split one change to two different PRs. You can do both changes in the same branch (even in the same commit).

app/models/conversation.rb
jhass: Yes.

RichKavanagh:
@SuperTux88 You're welcome.

app/models/conversation.rb
RichKavanagh: Done.

goobertron:
Hi Rich, thanks for spotting this.

To complete the PR, you need to add your text into the English locale file. I've found the location for you:

Under https://github.com/diaspora/diaspora/blob/develop/config/locales/diaspora/en.yml#L352 , add a line (with the same indentation as `subject: "Subject"`), which should read:

    	      subject_default: "No subject"

That's how to 'extract it to a translation key', as @jhass asked in your earlier PR. That way it can be translated into other languages in the other locale files. My skills are way below yours, so if I can learn it, you definitely can!

goobertron:
Oh, hang on a sec, you've done that in a separate PR. It would be good to do these two changes in one PR, otherwise Travis won't be able to confirm that the change is correct.

jhass:
With 48f5aed63ef867952ee9ce4178180fa503b4fd94 I combined all your commits and gave them a proper commit message. Learning the basics of git and Github would be a well spent effort nonetheless, give https://try.github.com a shot.

RichKavanagh:
Thanks for the tidy-up @jhass mate!

Point(s) well taken everyone, I'll leave the code changes alone now and stop messing with things I don't actually understand as I only seem to create more work for you all.  I made a right mess of this lol, I'll get my coat :)

PR #6209, opened 2015-07-18
Update rails assets gems

svbergerem:
Updates a few rails-assets gems and removes jquery-idletimer which was introduced in https://github.com/diaspora/diaspora/commit/2c0cfa53dceaeb276b0f15e2e27ec502792963c5 but there were no matches for `idle` in `app/` anymore. (except for the `require` line in `app/assets/javascripts/main.js`) If we would still use jquery-idletimer there would be some `idleTimer()` call somewhere. (See https://github.com/thorst/jquery-idletimer)


denschub:
merged into 0.5.2.0, stable and dev. thanks for your work!

Issue #6137, opened 2015-06-23
Dropdown button stays active after closing


AugierLe42e:
On the new header, the dropdown buttons stay active after clicking them the close the dropdown:

![header](https://cloud.githubusercontent.com/assets/3429231/8308333/77fff0f0-19c4-11e5-9746-e3f5fcd468cb.png)

Way to reproduce: open one of the dropdown and close it by clicking the button.

svbergerem:
All other links in the header also have that active link color and it might be useful for accessibility reasons.

morashid92:
that's strange, It works on and off, but only when it wants to.

morashid92:
I agree with @AugierLe42e. I have added some javascript to make this #fff disappear when closing the notifications instead of adding some funny css checkbox tricks which didn't work and are terrible. And before I do anything. I would like to know  where is the suitable place to add this javascript code. I presumed it was in: https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/header_view.js. 

Please tell me if that is ok? or is there another place we put our styling javascript code?

svbergerem:
I think `notification_dropdown_view.js` would be the correct place but in this case setting the font color of `.nav-badge:focus:not(:hover)` in the header should be sufficient.

morashid92:
I did    
<tt>.nav-badge:focus:not(:hover){
        color:#9d9d9d;
      }
</tt>
that doesn't take away the hover color on closing the icon,  but only after hovering away. I think that the hover should go away after closing the dropdown. What do you think? Iv'e accomplished that with a bit of javascript?

svbergerem:
As I wrote [before](https://github.com/diaspora/diaspora/issues/6137#issuecomment-114671120) I think we need some style change for focused links in the header. All other links in the header have that hover color plus some outline that we removed for the nav badge [here](https://github.com/diaspora/diaspora/blob/develop/app/assets/stylesheets/header.scss#L28). Maybe we should readd that outline by removing the linked line to make the focused link styling consistent. Then we might also remove the color change for all links in the header on focus. What do others think about that?

> I think that the hover should go away after closing the dropdown.

I think we should keep the font color on hover no matter whether we just moved the mouse to the link or we still hover over the link because we closed the dropdown.

morashid92:
Sounds like a good idea to me. I get the part where you want to re add the outline but how will you remove the linked line?

svbergerem:
> but how will you remove the linked line?

[...]

PR #6210, opened 2015-07-18
Fix Hover problem with Entypo Bell

morashid92:
In reference to https://github.com/diaspora/diaspora/issues/6137


jhass:
Looks like a couple of unrelated commits got in. That's because you didn't create feature branches for your contributions. Please read and follow our [git workflow](https://wiki.diasporafoundation.org/Git_workflow).

You should also read up on properly associating your commits with your Github account, Githubs help pages explain that in various places.

AugierLe42e:
Besides, the solution won't solve the problem for the user dropdown. I'm suggesting you to better use `.dropdown > a` as a selector.

PR #6212, opened 2015-07-19
Fix default home page

korun:
Hi!

In this PR I just fix indentation level for `p` block in default home page, and remove GitHub reference from wiki link title, because:

> [Please refer to that one instead of the information here](https://github.com/diaspora/diaspora/wiki/Home/0d45a6ac30090074f1744a0bcbe07c142eb3bd8d)

:)


svbergerem:
Thank you!

PR #6216, opened 2015-07-21
Add dashboard to admin page

### LINKED ISSUE #3588: admins should be informed about updates through the web-interface
<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/476-admins-should-be-informed-about-updates-through-the-web-interface?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
dashboard
error
outdated
uptodate

Fixes #3588.

I changed the admin navigation for all admin pages because there were now too many items to display them in one line on some screens.

### DIFF
--- a/app/assets/javascripts/app/pages/admin_dashboard.js
+++ b/app/assets/javascripts/app/pages/admin_dashboard.js
@@ -1,0 +1,75 @@
+// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-v3-or-Later
+
+app.pages.AdminDashboard = Backbone.View.extend({
+  initialize: function() {
+    this.updatePodStatus();
+  },
+
+  updatePodStatus: function() {
+    var self = this,
+        tagName = "";
+    $.get("https://api.github.com/repos/diaspora/diaspora/releases/latest")
+      .done(function(data) {
+        // the response might be malformed
+        try {
+          /* jshint camelcase: false */
+          tagName = data.tag_name;
+          /* jshint camelcase: true */
+          if(tagName.charAt(0) !== "v") {
+            self.updatePodStatusFail();
+            return;
+          }
+        } catch(e) {
+          self.updatePodStatusFail();
+          return;
+        }
+
+        // split version into components
+        self.latestVersion = tagName.slice(1).split(".").map(Number);
+        if(self.podUpToDate() === null) {
+          self.updatePodStatusFail();
+        } else {
+          self.updatePodStatusSuccess();
+        }
+      })
+      .fail(function() {
+        self.updatePodStatusFail();
+      });
+  },
+
+  updatePodStatusSuccess: function() {
+    $("#pod-status .alert").removeClass("alert-info");
+    var podStatusMessage = Diaspora.I18n.t("admins.dashboard.up_to_date");
+    if(this.podUpToDate()) {
+      $("#pod-status .alert").addClass("alert-success");
+    } else {
+      podStatusMessage = Diaspora.I18n.t("admins.dashboard.outdated");
+      $("#pod-status .alert").addClass("alert-danger");
+    }
+    $("#pod-status .alert")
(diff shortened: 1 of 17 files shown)


Flaburgan:
Awesome @svbergerem! But I'm not sure admins go often on the admin panel. Maybe this info should be displayed directly in the right column of the main page, or a small /!\ in the header which would link to this page?

svbergerem:
> Maybe this info should be displayed directly in the right column of the main page

Because we all know that users pay a lot of attention to the right column ;-) 

In both of your proposed solutions we would either have to fetch the current d* version every time the admin reloads the page or set up a background job that fetches the current version every n minutes. The first solution sounds bad, the second one will be more complicated.

The dashboard is an easy solution, we might want to add some more information about the pod in the future and we can still add some info in the header, the right sidebar or somewhere else in another PR if we think that the dashboard is not enough.

Flaburgan:
>Because we all know that users pay a lot of attention to the right column ;-)

With a red background like you designed it, it will definitely be seen ;)
This would be displayed only to users with administrator rights though, but I think that's what you meant.

>In both of your proposed solutions we would either have to fetch the current d* version every time the admin reloads the page or set up a background job that fetches the current version every n minutes.

Hm, maybe this could be done with something like `if I didn't check the version in the last hour, perform the check, else just skip it`?

>The dashboard is an easy solution, we might want to add some more information about the pod in the future and we can still add some info in the header, the right sidebar or somewhere else in another PR if we think that the dashboard is not enough.

Yeah no problem, I just wanted to point that it's probably not very efficient in that state, but it can be a good first step

AugierLe42e:
> With a red background like you designed it, it will definitely be seen ;)

This is typically the reason why it *shouldn't* be here. On stream page, the right bar must not draw any attention unless the user searches something.

@svbergerem : I see a warning box saying *Error fetching the latest diaspora version*. I don't find the wording clear : do you talk about the version number, or about the source code ?

svbergerem:
> do you talk about the version number, or about the source code ?

My code fetches https://api.github.com/repos/diaspora/diaspora/releases/latest and uses the `tag_name` (the version number). Should I change the wording to

> Fetching latest diaspora* version number...

and

> Error fetching the latest diaspora* version number.

?

Flaburgan:
>This is typically the reason why it shouldn't be here. On stream page, the right bar must not draw any attention unless the user searches something.

Read more carefully, this is only about the podmins, and we **want** the user to pay attention about this critical message: an update is required.

jhass:
"version number" sounds a bit odd to me, I think it's fine the way it is.

jhass:
If anybody feels strong about changing something we can always do that in a follow up PR.

Thanks!

Flaburgan:
Thanks Steffen and Jonne.

goobertron:
Hey, fantastic, Steffen!

> Should I change the wording to
> 
> > Fetching latest diaspora* version number...
> 
> and
> 
> > Error fetching the latest diaspora* version number.
> 
> ?

How about:

> Determining which is the latest version of diaspora*

and 

> Error determining which is the latest version of diaspora*

'Fetching the latest version' suggests to me fetching the files rather than simply the meta-information about version number. Perhaps 'fetch' has a very specific meaning to sysadmins, and so is perhaps not ambiguous in this circumstance. But if there is any possibility of ambiguity, I'd suggest the 'plain English' versions I've given above.

jhass:
Yes, determining seems the better word. How about "Determined latest diaspora* version"?

goobertron:
> How about "Determining latest diaspora* version"?

I'd be happy with that. However, 'Error determining latest diaspora* version' sounds a bit cumbersome to me, so how about, for the error string: 'Unable to determine latest diaspora* version'?

jhass:
Sounds good to me.

PR #6222, opened 2015-07-21
Issue #6137: Add focus:not(:hover) color

morashid92:
Thanks for pointing me towards the workflow page @jhass. And i've also linked my account with my commits. Now back to CSS. This commit changes the hover away color of the .dropdown instead of .nav-badge which I think is convenient.

### DIFF
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -26,6 +26,7 @@ header > .dark-header > nav {
       margin-left: 20px;
       & > li { height: 50px; }
       .dropdown > a:focus { outline: 0px none; }
+      .dropdown > a:focus:not(:hover){color:#9d9d9d;}
       .dropdown-open {
         background-color: $dropdown-bg;
         & > a { color: $dropdown-link-color; }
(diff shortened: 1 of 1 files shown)


svbergerem:
Now you have the consistency and accessibility issues that I mentioned in #6137. There is no way to see if the bell icon is focused and all other links in the header will be white on focus except for the bell icon.

Also please use a color variable. I guess in this case [`$navbar-inverse-link-color`](https://github.com/diaspora/diaspora/blob/develop/app/assets/stylesheets/bootstrap-variables.scss#L399) would be the correct one.

morashid92:
Ok, i've changed hex to the variable now. Also, so your suggesting that .nav-badge should be used instead then?

svbergerem:
For consistency you would have to add the color to all links in the header, so maybe something like
```scss
.navbar-nav > li > a:focus:not(:hover) { ... }
```

svbergerem:
@morashid92 why did you close this PR? :-(

morashid92:
@svbergerem Whoops, I was creating another fix which I thought had to do with this, but came to an understanding that its browser related. Not any more it isn't!

denschub:
Ping?

svbergerem:
#6137 has been closed. Thanks for your work.

Issue #5772, opened 2015-03-14
accessibility: new conversation labels are not correct


pravi:
'to' field is not read by orca. It just says entry, it should say 'to entry'.  Same for subject. Body is incorrectly read as subject.

pravi:
Reply field also missing label. It just says entry.

svbergerem:
@pravi #6227 should have improved this. Could you check again if this is fixed now?

pravi:
@svbergerem thanks! I will try to test it soon, but it may take a few days.

svbergerem:
@pravi Ping. If you don't have access to a pod running the develop code you can try https://sechat.org/, https://iliketoast.net/ or any other pod that is listed on https://the-federation.info running 0.5.99.0 with open signups.

Flaburgan:
@pravi this issue is currently a blocker for 0.6, please tell us if the problem is solved.

abhijithb21:
@Flaburgan I tested it on sechat.org with orca version 3.20 running on Debian Stretch XFCE. It reads "to entry". I noticed it is reading to and entry together. My friend who uses orca at a speed higher than normal says it is fine.

abhijithb21:
I compared this with 5.9 on shrekislove.us. Orca read "entry" soon after it read what the page contents are. In sechat.org I had to press tab, shift+tab etc to bring focus manually to to entry. It reads properly but only after I bring focus.

cmrd-senya:
@abhijithb21 
I tested both sechat.org and shrekislove. I also tested it on my development install with the latest develop branch. I use Ubuntu 14.04 and Firefox version from the distribution for the test. Everywhere it read "to entry" as expected. Probably in your installation it didn't focus on the entry after load for some reason. Your browser version might be important in order to reproduce that.

abhijithb21:
It works fine in Firefox 45 now. Autofocus still does not work in midori 0.5.11 but no problem with reading "to entry".

cmrd-senya:
So it's resolved then, I guess

abhijithb21:
Yes.

svbergerem:
Thank you for testing this!

PR #6227, opened 2015-07-22
Improve accessibility of signin/signup/password and conversation pages

svbergerem:
- Remove css hack for labels and use `.sr-only` instead
- Add labels for textareas on conversation pages
- Use `aria-labelledby` on the mentioned pages

This should improve #5772. I don't know if this PR fixes the issue.

`.sr-only` has been introduced by Bootstrap 3 so this only works for 0.6.0.0.


jhass:
Thanks!

PR #6229, opened 2015-07-22
Add js for aria-labelledby in conversations

svbergerem:
I forgot to add this line in #6227. `#conctact_ids` is added by our js so we also need to add the attribute with js.


jhass:
Thanks!

PR #6230, opened 2015-07-22
Add viewport to desktop view

Flaburgan:
Since some part of the desktop view are now responsive, we should specify a viewport to make them usable. Notice the menu, the size of the font of users posts and the size of the publisher.


denschub:
Thank you!

PR #6231, opened 2015-07-22
Add my aspects link in the header

Flaburgan:
Add a link to the /aspects page in the header.
Users usually have too many contact to be interested by the raw stream, so the /aspects page becomes the "root" page. This link allows to access it directly from anywhere. It is a customization @taratatach did on diaspora-fr years ago and which was well welcomed, so I propose to merge it upstream, = here :)

### DIFF
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -23,7 +23,6 @@ header > .dark-header > nav {
     }
     .navbar-nav:not(.nav-badges) > li > a { font-weight: bold; }
     .nav-badges {
-      margin-left: 20px;
       & > li { height: 50px; }
       .dropdown > a:focus { outline: 0px none; }
       .dropdown-open {
--- a/app/assets/templates/header_tpl.jst.hbs
+++ b/app/assets/templates/header_tpl.jst.hbs
@@ -17,6 +17,7 @@
         <div class="collapse navbar-collapse" id="navbar-collapse">
           <ul class="nav navbar-nav navbar-left">
             <li><a href="/stream">{{t "my_stream"}}</a></li>
+            <li><a href="/aspects">{{t "my_aspects"}}</a></li>
             <li><a href="/activity">{{t "my_activity"}}</a></li>
           </ul>
 


svbergerem:
Doesn't work for some screen widths: (>1024px)
navbar1
navbar2

Flaburgan:
Nice catch. Solved by removing the margin-left on the nav-badge. Is that fine for you?

svbergerem:
That doesn't fix the issues. (You can see in the screenshots that I tried this in German where the titles in the header are a bit longer)

Flaburgan:
Yeah I set my language to German to test it, and I had the issue only with the search bar focused. Without the 20px margin, no problem anymore for me.

svbergerem:
Try 1024px and 1200px with the focused search bar.

Flaburgan:
@svbergerem maybe we could remove the "Stream" link, as it's already the link on the name of the pod?

jaywink:
> maybe we could remove the "Stream" link, as it's already the link on the name of the pod?

Hmm personally I'd like to keep the stream even if the name of the pod is a link. What if the multi stream won't always be the "home" stream? I think "home" and "stream" should be kept separate, even if they go to the same place at the moment.

Flaburgan:
Fair enough, I'll try to fix it by changing the size of the elements.

svbergerem:
Hmm, tbh. I don't think that we need the aspects link in the header.

Flaburgan:
@svbergerem it's up to you guys, I find it way more useful than the "My Activity" link but as said, this is already on d-fr for at least two years and I only propose it here because it's easier for me to maintain :)

svbergerem:
For me the link isn't really useful and it adds another element to the header. I'd prefer a header with as few links as possible. This is just my personal opinion and if we merge this I can still remove it for my own pod. It would be great if some others could also write what they think about this so we can decide who of us will have to do more work to maintain his pod. ;-)

Flaburgan:
Any other opinions everyone?

jhass:
I'd like a minimal header too, I don't jump to my activity either from the other pages, always back to the stream page. From there I hit other kinds of streams via the left sidebar. So actually I wouldn't even be opposed to dropping "My activity", and given the podname links to the stream, even dropping that explicit link.

jaywink:
I tend to agree about not cluttering the header. I use My Activity sometimes, but My Aspects almost never. You could always do a poll with #diaspora tag if you want some larger audience to vote :)

SuperTux88:
I don't use these links too, so we can remove the existing "My Stream" and "My Activity" and not add even more links.

AugierLe42e:
I personnally use My Activity very often. Since notifications tend to bug sometimes, I usually use it to assure that nobody reacted on a post I also commented and I regularly discover think thanks to this link, so I'd vote for keeping it. On the other hand, I don't use Stream at all so no need to keep it for me.

theworldbright:
I personally just click the logo. I do keep a link to the Stream on my right hand side just for people who might not find it intuitive to navigate with the logo (picture below). I would vote to either have ALL the stream kinds up in the header (and perhaps remove the sidebar links) or only one.

screen shot 2015-07-29 at 10 38 39 pm

Globulle:
I agree with @AugierLe42e . Due to the very frequent bugs in notifications, I use *My activity* very often. This is also useful to retrieve some posts I have liked/commented. I have several questions:

1. Is there a reason to keep *My activity* link both in the left-hand side bar and in the header? (if we drop one, I suggest to keep the one in the header: I think this is good to invite people visiting it since they could miss lots of activity if they don't, as long as the bug remains)
2. Why are there 2 buttons (Logo and *stream*)side by side which do the same thing? 
3. I know we can access our likes and comments typing */liked* or */commented*... I don't like the fact this is kind of secret trick... Don't you think we could add these link in the left panel?

Finally, I like the suggestion of @Flaburgan , but maybe we could make it more compact and more clear by having just a star button, which would be our favourites aspects/tags, which we could defined elsewhere. Another option should be to allow *My Aspects* to remain expanded...

Flaburgan:
>Is there a reason to keep My activity link both in the left-hand side bar and in the header? (if we drop one, I suggest to keep the one in the header: I think this is good to invite people visiting it since they could miss lots of activity if they don't, as long as the bug remains)

The header link is only a shortcut, added because it's handy. The leftnav bar has to contain every stream links.

>Why are there 2 buttons (Logo and stream)side by side which do the same thing?

Currently the "home page" is the stream. This can change in the future.

>I know we can access our likes and comments typing /liked or /commented... I don't like the fact this is kind of secret trick... Don't you think we could add these link in the left panel?

https://github.com/diaspora/diaspora/pull/5502

AugierLe42e:
> Currently the "home page" is the stream. This can change in the future.

I don't see why it would change. It's pretty a convention now.

> > I know we can access our likes and comments typing /liked or /commented... I don't like the fact this is kind of secret trick... Don't you think we could add these link in the left panel?
>
> #5502

It also let acces to one's own posts which can't be reached from `/commented` if I'm not wrong.

taratatach:
I added this link because this is the only stream I use. The Stream is all cluttered and what I care about are the posts from the people I've added to my aspects.
Going through the main stream to go the aspects page is super inefficient, adding a superfluous click and a big load time. I would agree that a link isn't needed if it was a subfunction of the main stream but, like My Activity, it is not and therefore has its place in the main header.

Like @Globulle suggested, we might want to replace those textual links with icons to reclaim space in the header and even add access to more streams (like the mentions and followed tags or new ones).

AugierLe42e:
There is actutally a good reason to add *My aspects* to the header: The link would be accessible from everywhere when navigating on desktop version from a mobile without having to scroll all the way up.

Flaburgan:
Not a majority, I'll keep that for d-fr only.

PR #6233, opened 2015-07-22
Rework mobile drawer transition

Flaburgan:
I cleaned the mobile drawer code a bit. Now only the header and the drawer translate, that means we are **finally** fluid. This also fixes the bug with the avatar hidden in the profile `li`.

This mostly supersedes https://github.com/diaspora/diaspora/pull/5556 although I didn't include Hammer.js so no swipe opening at the moment.

I tested this with Firefox OS and we have no performance issue anymore. @SansPseudoFix tested it on Android 4.4 with Firefox and Chrome and confirm the drawer is now fluid.

Please test this with other phones, I don't have iPhone, iPad, Windows Phone, Ubuntu nor Jolla to test it. This commit is currently on diaspora-fr.org to test it.

### DIFF
--- a/app/assets/javascripts/mobile/mobile.js
+++ b/app/assets/javascripts/mobile/mobile.js
@@ -38,21 +38,21 @@ $(document).ready(function(){
   };
 
   /* Drawer menu */
-  $('#menu_badge').bind("tap click", function(evt){
+  $("#menu_badge").bind("tap click", function(evt){
     evt.preventDefault();
-    $("#app").toggleClass('draw');
+    $("#app").toggleClass("draw");
   });
 
   /* Show / hide aspects in the drawer */
-  $('#all_aspects').bind("tap click", function(evt){
+  $("#all_aspects").bind("tap click", function(evt){
     evt.preventDefault();
-    $("#all_aspects + li").toggleClass('hide');
+    $("#all_aspects + li").toggleClass("hide");
   });
 
   /* Show / hide followed tags in the drawer */
-  $('#followed_tags').bind("tap click", function(evt){
+  $("#followed_tags > a").bind("tap click", function(evt){
     evt.preventDefault();
-    $("#followed_tags + li").toggleClass('hide');
+    $("#followed_tags + li").toggleClass("hide");
   });
 
   /* Heart toggle */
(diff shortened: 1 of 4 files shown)


AugierLe42e:
> This also fixes the bug with the avatar hidden in the profile li.

That one was already solved by #6057 :p

Flaburgan:
I don't call `display: none` a fix.

SansPseudoFix:
Not tested with chrome, only Firefox 39. I haven't chrome :P

AugierLe42e:
> I don't call `display: none` a fix.

This was only on small screens because it was ugly. On large screens it was displayed.

Flaburgan:
I know, but too bad, mobiles have small screens, and that's why we're building a mobile version :p

Flaburgan:
I have a bug with the default Android Browser, have to fix that.

AugierLe42e:
> I know, but too bad, mobiles have small screens, and that's why we're building a mobile version :p

Me not a web-dev, sadly. I just fixed what was a technically incorrect way to handle the sliding.

Flaburgan:
OK looks like the bug on webkit / blink is fixed. Can anyone else try this patch with any mobile device he finds? Especially iOS and Windows Phone, I have no device to test.

Flaburgan:
Tested with iOS 7 iPhone 4, Windows Phone 8.1 Lumia, Firefox OS 2.1, 2.2 and 3.0.

IMO it can be merged, and if there are regression on some devices I'll fix them.

app/assets/javascripts/mobile/mobile.js
```diff
@@ -51,7 +51,7 @@ $(document).ready(function(){
   });
 
   /* Show / hide followed tags in the drawer */
-  $('#followed_tags').bind("tap click", function(evt){
+  $('#followed_tags > a').bind("tap click", function(evt){
```
svbergerem: Please fix this.

Flaburgan:
@svbergerem : fixed double quote for the related par of mobile.js

svbergerem:
Thanks!

Flaburgan:
Awesome <3

PR #6234, opened 2015-07-22
Enable the jasmine testsuite to work across LAN

cmrd-senya:
Enable the jasmine testsuite to work across LAN by setting its Host variable to "0.0.0.0" (instead of default 127.0.0.1) so you could test it with a foreign web browser.


jhass:
Merged as e30293b9c1e0cce53b300a5727606caf580a31de

Thanks!

PR #6238, opened 2015-07-23
Add location and polls to mobile website

svbergerem:
This just adds them to the stream and the SPV. You are still unable to create a post with locations and polls from the mobile website. Voting is also not possible. This PR just implements the basics. Everything else should happen in other PRs.


jhass:
Looking good, maybe a cuke that asserts that they're displayed?

jhass:
Thank you!

jaywink:
@svbergerem https://iliketoast.net/posts/681449 any ideas why?

svbergerem:
@jaywink Thank you for the quick bug report. That was kind of a silly mistake. #6249 should fix this.

PR #6239, opened 2015-07-23
Implement NodeInfo

jhass:
With this, my current plan would be to drop statistics.json
after 0.6 was released.


app/controllers/node_info_controller.rb
```diff
@@ -0,0 +1,24 @@
+class NodeInfoController < ApplicationController
+  respond_to :json
+  respond_to :html, only: :statistics
+
+  def jrd
+    render json: NodeInfo.jrd(CGI.unescape(node_info_url("123.123").sub("123.123", "%{version}")))
```
Flaburgan: Why can't you give directly `%{version}` to `node_info_url`?

jhass: Because of the constraint on the route. But true, I'm rather unhappy about  this "solution".

app/views/node_info/_statistics.haml
```diff
@@ -0,0 +1,21 @@
+-#   Copyright (c) 2010-2011, Diaspora Inc.  This file is
```
jaywink: I feel bad about including this header always.

What about just;

> # This file is licensed under the Affero General Public License version 3 or later.  See
> # the COPYRIGHT file.

Diaspora Inc doesn't exist any more and this file wasn't amde in 2010-2011 and the current files are not even copyright Diaspora, Inc according to the agreement with the FSSN.

I know this is not related to this PR, but still for this file the first sentence could be removed ;) Also, do files even need the license? Isn't a repo license enough?

jhass: Right, I just moved the file, I don't include it to new files for the time being.

jaywink: Ah ok, sorry :dancer:

jaywink:
@annando, @redmatrix - diaspora plans to drop the current `statistics.json` in development releases after 0.6 and (I'm assuming here) fully in 0.7. Our major releases (these are major increments according to our version scheme) don't have release schedules, but it is likely 0.6 happens this year and 0.7 probably sometime early next year (just guessing).

Anyway, it would be great if, as communicated before, Friendica and RedMatrix would also support the new more flexible and useful NodeInfo, fully detailed [here](http://nodeinfo.diaspora.software/schema.html).

annando:
@jaywink Just tell me how it will be called (path) and provide a sample Installation, so I can compare my implementation with this.

jhass:
@annando if you follow @jaywink's link you can a full description of the format and if you click on the "{}" symbol you can view the raw JSON schema document. If you follow the "Protocol" link at the top you can learn about how to announce the actual location.

app/views/node_info/_statistics.haml
denschub: Technically, the header has to include a AGPL notice. But let's not make this part of this PR :)

denschub:
Merged. Thank you very much! :)

redmatrix:
Probably should also add 'hubzilla' to the list of server software; which will eventually take the place of redmatrix. You're likely to see both in the wild for quite some time.

denschub:
Please file issues regarding the actual nodeinfo spec at [jhass/nodeinfo](https://github.com/jhass/nodeinfo).

app/views/node_info/_statistics.haml
Flaburgan: @jaywink just for information, I changed the meta author in https://github.com/diaspora/diaspora/pull/6230/files#diff-b7186ba040870c69d2ae394fa317ea64R14 using "diaspora* contributors".

annando:
@jaywink it is implemented on my local server and is awaiting a pull request.

See https://pirati.ca/.well-known/nodeinfo and https://pirati.ca/nodeinfo/1.0

I have two questions:

1. Is it implemented correctly on my system?
2. Can we use the same registering process at http://the-federation.info/register/ like the one with statistics.json?

jhass:
@annando there shouldn't be any `\` in the URIs inside the JRD.

The following validation errors occurred when validating against the schema:

```
The property '#/' did not contain a required property of 'metadata' in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
The property '#/services/4' value " friendica" did not match one of the following values: appnet, blogger, buddycloud, diaspora, dreamwidth, drupal, facebook, friendica, gnusocial, google, insanejournal, libertree, linkedin, livejournal, mediagoblin, myspace, pinterest, posterous, pumpio, redmatrix, smtp, tent, tumblr, twitter, wordpress, xmpp in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
The property '#/usage/users/total' of type String did not match the following type: integer in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
The property '#/usage/users/activeHalfyear' of type String did not match the following type: integer in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
The property '#/usage/users/activeMonth' of type String did not match the following type: integer in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
The property '#/usage/localPosts' of type String did not match the following type: integer in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
The property '#/usage/localComments' of type String did not match the following type: integer in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
```

annando:
@jhass Okay, changed.

BTW: Friendica does not directly support LinkedIn - but can post to buffer which offers a connector to LinkedIn. Is it okay that I include it in the list?

And: I think it would be great to extend the "services" value with an "inbound" and "outbound" value. Friendica can import from app.net, pump.io, Twitter and GNU Social. It would be great to show this.

jhass:
@annando please open an issue at https://github.com/jhass/nodeinfo to discuss such things.

JRD looks good now. The `metadata` data key is placed wrong:

```
The property '#/' did not contain a required property of 'metadata' in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
The property '#/usage' contains additional properties ["metadata"] outside of the schema when none are allowed in schema http://nodeinfo.diaspora.software/ns/schema/1.0#
```

annando:
@jhass oops ... copy'n paste problem. "metadata" is moved.

jhass:
@annando almost there, it should be an object, not an array :)

annando:
@jhass done. Thanks for your patience!

jhass:
Nice work, you're welcome.

annando:
Just a final question: Should "services" only contain connections via third party APIs or the internal protocols as well?

jhass:
It's intended to publish information indicating roughly "you can post (and with your inbound idea read) content to (from) twitter / facebook / whatever" here too. Note that the values aren't free form http://nodeinfo.diaspora.software/schema.html

annando:
Friendica is able to speak the OStatus protocol directly - and it is able to use GNU social accounts via their API. These are two different things and maybe it would be a good idea to show this by having the protocol stuff under "protocols" and the API stuff under "services"? So systems with disabled OStatus protocol but enabled GNU Social connector would have an entry under "services" but not "protocols" and vice versa? Should I file an issue for the clarification?

jhass:
I feel like the distinction should be user oriented. The main distinction I see between `protocols` and `services` is that for the latter the user has to authenticate a specific account and thus have an account on the target service, while the former comes for free because the software is or pretends to be a regular node of that network with no need for another account by the user.

So essentially the answer is yes. And also yes, please open another issue if you want more clarification/discussion on that topic, that keeps it more organized and easier to find later :)

annando:
Issue is opened: https://github.com/jhass/nodeinfo/issues/8

jaywink:
@annando 

> Can we use the same registering process at http://the-federation.info/register/ like the one with statistics.json?

Once I (or someone else) implement NodeInfo into the-federation.info, then yes. Currently it still depends on statistics.json, so don't ramp that down ;) I'll hopefully get it done within a few weeks, been a little occupied with the relay stuff and one not related Django site.

jhass:
@annando We just noticed (and I wasn't paying enough attention earlier, sorry!) that your JRD (https://pirati.ca/.well-known/nodeinfo) is invalid, `links` should be an array of objects, not pointing to an object directly.

PR #6240, opened 2015-07-24
Update number of contacts on contact page when adding/removing contacts

### LINKED ISSUE #2900: Number of contacts in the aspects not updated after updating
To reproduce:
1. Create a new aspect
2. Click on it
3. Edit it
4. Select a few contacts
5. Click "Done"

There are two problems here actually. The main one is that the number of contacts is not updated in the aspects list on the left. The other problem is that there is no confirmation message appearing that says that the changes were applied. It thus gives the impression that it's not working.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/114627-number-of-contacts-in-the-aspects-not-updated-after-updating?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #2900.

### DIFF
--- a/app/assets/javascripts/app/models/contact.js
+++ b/app/assets/javascripts/app/models/contact.js
@@ -2,12 +2,12 @@
 
 app.models.Contact = Backbone.Model.extend({
   initialize : function() {
-    this.aspect_memberships = new app.collections.AspectMemberships(this.get('aspect_memberships'));
-    if( this.get('person') ) this.person = new app.models.Person(this.get('person'));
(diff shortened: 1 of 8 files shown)


jhass:
Thank you!

PR #6243, opened 2015-07-24
Fix mobile photos index page

svbergerem:
There is no link to the photos index page on the mobile website but the page exists so it should work. It should be possible to backport this. Small design improvements will come in a follow-up PR for Bootstrap 3.


denschub:
Merged as 6fea450777f412f2805840c25ee644ea81d0b647, thanks much!

PR #6246, opened 2015-07-24
Better alignement of interaction icons

ghost:
Before:

capture d ecran 2015-07-24 a 20 08 53

After:

capture d ecran 2015-07-24 a 20 07 49


svbergerem:
Thank you!

PR #6074, opened 2015-06-07
Fix some Bootstrap 3 regressions

svbergerem:
- mobile notification padding ([before](https://cloud.githubusercontent.com/assets/3798614/8023628/26a8f736-0d14-11e5-9872-4553a8df1dc1.png)) ([after](https://cloud.githubusercontent.com/assets/3798614/8023631/321cc0fc-0d14-11e5-9b50-fec89b89c1f5.png))
- width of mobile notifications and stream page

PR #6247, opened 2015-07-24
Fix page width on people view (BS3 regression)

ghost:
People page regains its normal width.
Before:

capture d ecran 2015-07-24 a 20 35 59

After:

capture d ecran 2015-07-24 a 20 33 42

### DIFF
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -312,8 +312,7 @@ h3 {  margin-top: 0; }
 }
 
 .header-full-width {
-  margin: -10px;
-  margin-bottom: 1.5rem;
+  margin: -10px -10px 0;
   padding-top: 5px;
   background-color: #fff;
   border-bottom: 1px solid #aaa;
--- a/app/views/people/show.mobile.haml
+++ b/app/views/people/show.mobile.haml
@@ -15,19 +15,19 @@
   .bottom_bar
     - if !@person.tag_string.blank? && user_signed_in?
       = Diaspora::Taggable.format_tags(@person.tag_string)
-
-.col-md-12.profile_stream
-  - if @stream.stream_posts.length > 0
-    - if @post_type == :photos
-      = render 'photos/index', photos: @stream.stream_posts
+.row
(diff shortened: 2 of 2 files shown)


svbergerem:
http://stackoverflow.com/a/18882813

We use the Bootstrap grid system also on other mobile pages (e.g. on those pages where we have one combined template for the mobile and desktop view) so I'd suggest that we start using the Bootstrap grid in a proper way instead of removing it from some pages and keeping it on others.

AugierLe42e:
https://github.com/diaspora/diaspora/pull/6074/files#diff-9eac80d828cb5c27b2a4ae46132e070fL5

svbergerem:
On the people show page [that stream is already wrapped](https://github.com/diaspora/diaspora/blob/develop/app/views/people/show.mobile.haml#L25) in a `.col-md-12` and having a `.col-**-**` inside a `.col-**-**` without a `.row` in between is wrong,  but there are some other places where this is missing which means that it is inconsistent, that's right.

AugierLe42e:
So basically, a `.row` is missing? Ok, let me correct that.

svbergerem:
One way to make that consistent throughout our codebase would be a rule like
> 1. Whenever you use `col-**-**`, wrap it in a `.row` in the same file. 
> 2. Whenever you render some file in a `.row` make sure there is a `.col-**-**` in between.

svbergerem:
> So basically, a .row is missing?

It's not just the `.row`. The structure of our code should be

```haml
.container # (or .container-fluid) should be somewhere in the layout files
  .row
    .col-**-**
      .row
        .col-**-**
  .row
    .col-**-**
```

AugierLe42e:
I don't think there is any `.container` anywhere on the mobile view. At least, I don't find any on stream page.

Flaburgan:
I noticed this problem today, thank you for solving it :)

svbergerem:
Thanks! As I wrote in [this comment](https://github.com/diaspora/diaspora/pull/6247#issuecomment-124620276) I think we should properly use the Bootstrap grid on the mobile website but that is out of scope of this PR.

PR #6255, opened 2015-07-27
Update gems

jhass:
See #3855


denschub:
Merged! Thanks.

denschub:
Okay, apparently I broke a cuke while backporting. https://travis-ci.org/diaspora/diaspora/builds/73310194 Will investigate this evening.

PR #6259, opened 2015-07-28
Update home_controller.rb

mikicaivosevic:
Refactor method toggle_mobile

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -26,13 +26,7 @@ class HomeController < ApplicationController
   end
 
   def toggle_mobile
-    if session[:mobile_view].nil?
-      # we're most probably not on mobile, but user wants it anyway
-      session[:mobile_view] = true
-    else
-      # switch from mobile to normal html
-      session[:mobile_view] = !session[:mobile_view]
-    end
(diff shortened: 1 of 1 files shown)


app/controllers/home_controller.rb
```diff
@@ -26,13 +26,7 @@ def show
...
 
   def toggle_mobile
-    if session[:mobile_view].nil?
-      # we're most probably not on mobile, but user wants it anyway
-      session[:mobile_view] = true
-    else
-      # switch from mobile to normal html
-      session[:mobile_view] = !session[:mobile_view]
-    end
+    session[:mobile_view] = session[:mobile_view].nil? ? true : false
```
jhass: I'm not sure this is equivalent.
```
new value:

old value | old variant | new variant
----------+-------------+------------
nil       | true        | true
true      | false       | false
false     | true        | false
```

If the tests pass, a test exercising the last case would be great though. Or maybe I'm wrong and it can never get `false`?

mikicaivosevic: Hm, my bad, I will fix.

jhass:
Closing in favour of #6260 

Note that if you edit a commit and force push to the branch the PR will update ;)

PR #6266, opened 2015-07-29
Fix conversations view with no contacts

### LINKED ISSUE #6265: Bug in message view when having no contacts?
Hi folks,

I just encountered a strange issue when testing the new welcome message feature (#6021).

A newly created account was not able to click on the welcome message (to show it on the right side) while having no contacts.

After adding the new profile as a contact on my own profile (and vice versa) I was able to click the message with the new user account.

Maybe you have an idea about what that issue could possibly be, otherwise I'd of course be willing to help narrow it down.

Browser: 43.0.2357.130 Built on Ubuntu 14.04, running on elementary OS 0.3 (64-bit)

Cheers
Thomas

svbergerem:
Fixes #6265.


denschub:
Merged as d8505b2ae0b674bff70efbe02d73ded495cda84d, thank you!

Issue #6235, opened 2015-07-22
Sidebar links are not fully clickable


maliktunga:
Here is the clickable zone of Stream:
![stream](https://cloud.githubusercontent.com/assets/6393838/8840115/f6d6c906-30aa-11e5-9396-3f77cc8f1ea0.png)

Here is the clickable zone of My Activity:
![myactivity](https://cloud.githubusercontent.com/assets/6393838/8840113/f032447c-30aa-11e5-8058-87049dd9d41e.png)

Here is the clickable zone of @Mentions:
![mentions](https://cloud.githubusercontent.com/assets/6393838/8840136/3762c9e8-30ab-11e5-8ab3-ea0c6c8fd7cc.png)

Here is the clickable zone of My Aspects:
![myaspects](https://cloud.githubusercontent.com/assets/6393838/8840166/7498c51a-30ab-11e5-9d04-b689c6d21b3d.png)

Here is the clickable zone of #Followed Tags
![followedtags](https://cloud.githubusercontent.com/assets/6393838/8840182/b6c04ff8-30ab-11e5-92db-e4b191d6845c.png)

I would expect the clickable zone to cover the blue zone just like with My Aspects and #Followed Tags.

I use Diaspora 0.5.99.0-p0713edaf on https://social.mrzyx.de/
Firefox 39.0 on Fedora 22

Flaburgan:
https://github.com/diaspora/diaspora/pull/6112 should have solved this.

PR #6267, opened 2015-07-29
Issue #6235: Add hoverable class inside the a tag

morashid92:
I initially altered leftnavbar.scss by adding in <tt>#stream_selection > li {cursor: pointer}</tt>, but realised thats insanity because all it does is changes pointer on hovering and is still un-clickable!. However after inspections, adding the class <tt>hoverable</tt> inside the a tag instead, gives a much better result!


svbergerem:
Thank you!

PR #6271, opened 2015-07-30
Raise rubocop abc size metric from 15 to 20

KentShikama:
```
      def save_request_parameters
        session[:client_id] = @o_auth_application.client_id
        session[:response_type] = @response_type
        session[:redirect_uri] = @redirect_uri
        session[:scopes] = scopes_as_space_seperated_values
        session[:request_object] = @request_object
        session[:nonce] = params[:nonce]
      end
```

The above method has an ABC metric of 16 and is not currently passing the check.

```
      def restore_request_parameters
        req = Rack::Request.new(request.env)
        req.update_param("client_id", session[:client_id])
        req.update_param("redirect_uri", session[:redirect_uri])
        req.update_param("response_type", response_type_as_space_seperated_values)
        req.update_param("scope", session[:scopes])
        req.update_param("request_object", session[:request_object])
        req.update_param("nonce", session[:nonce])
      end
```

The above method has an ABC metric of 20.02 and is not currently passing the check.

I just feel like 15 is a bit too strict.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,6 +13,10 @@ Metrics/MethodLength:
 # The guiding principle of classes is SRP, SRP can't be accurately measured by LoC
 Metrics/ClassLength:
   Max: 1500
+  
+# Raise AbcSize from 15 to 20
+Metrics/AbcSize:
+  Max: 20
 
 # No space makes the method definition shorter and differentiates
 # from a regular assignment.


jaywink:
Totally agree :+1:

jhass:
Yeah, I wanted to try to run with it for a while, but it is.

Merged as 23bea8e3d37c2e2fbb24a489dd08b78cacbe91b9

Thanks

PR #6277, opened 2015-07-31
Use css spinner instead of gif

svbergerem:
Uses a css spinner for
- infinite streams
- notification dropdown
- publisher:
  - location
  - loader while publishing a post
  - ~~photo uploader~~
- modals
- changing the name in the getting started view

To use the spinner add `.spinner` to your code. The spinner will use width and height of the parent element so putting it inside of a `.loader` div sets the width and height to 32px. You can also use it without the `.loader` div and manually set width and height for the spinner.

I replaced the ajax-loader gif only for the desktop view. There are also four other places where we still use gif (because I didn't want to touch the code yet):

``` bash
$ grep -Rin "loader2" app lib spec features 
app/assets/javascripts/mobile/mobile_file_uploader.js:33:            "<img alt='Ajax-loader2' src='"+ImagePaths.get('ajax-loader2.gif')+"' />" +
app/assets/javascripts/widgets/lightbox.js:181:      self.image.attr("src", ImagePaths.get("ajax-loader2.gif"));
app/assets/stylesheets/vendor/autoSuggest.css:14:  background: asset_path("ajax-loader2.gif") right center no-repeat;
app/assets/stylesheets/oembed.scss:2:  background: image-url("ajax-loader2.gif") no-repeat center center;
app/assets/stylesheets/autocomplete.scss:46:  background: white image-url('ajax-loader2.gif') right 5px center no-repeat;
```

Here are two videos where you can see the new spinner:
http://videobin.org/+9gs/cmt.html
http://videobin.org/+9gs/cmu.html


svbergerem:
I reverted the changes for the photo uploader and added the new spinner to the search input in the header. I removed the spinner for autoSuggest inputs since neither on develop nor on stable it is displayed.

The two photo uploaders and the lightbox display the old spinner while they are loading an image, replace the img src of the spinner with the one of the image and the browser displays the image. If we would use the new spinner we would have to load the image in the background and replace the spinner with the image once it is completely loaded. jQuery [has a deprecated `load` function](https://api.jquery.com/load-event/) which doesn't seem to be reliable when used with images so I'd keep the gif loader for the uploaders and the lightbox in this PR.

There is a similar problem with the oEmbed stuff. The current implementation has a bug because it doesn't remove the spinner once the oEmbed content has been loaded. (AFAIR we have an issue somewhere) So also in this case I'd keep the gif loader in this PR and we can switch to the new spinner while fixing the other bug.

AugierLe42e:
Nice. Like it. Dunno why but CSS spinner never worked on my computer before port to BS3. I suppressed it during this work because of that.

svbergerem:
This PR should be ready to merge.

denschub:
No objections. :)

PR #6280, opened 2015-08-02
Refactor status messages controller

KentShikama:
For another API route

### DIFF
--- a/app/controllers/status_messages_controller.rb
+++ b/app/controllers/status_messages_controller.rb
@@ -1,36 +1,33 @@
 #   Copyright (c) 2010-2011, Diaspora Inc.  This file is
 #   licensed under the Affero General Public License version 3 or later.  See
 #   the COPYRIGHT file.
-
 class StatusMessagesController < ApplicationController
   before_action :authenticate_user!
 
-  before_action :remove_getting_started, :only => [:create]
+  before_action :remove_getting_started, only: :create
 
-  respond_to :html,
-             :mobile,
-             :json
+  respond_to :html, :mobile, :json
 
-  layout 'application', only: :bookmarklet
+  layout "application", only: :bookmarklet
 
   # Called when a user clicks "Mention" on a profile page
   # @param person_id [Integer] The id of the person to be mentioned
   def new
-    if params[:person_id] && @person = Person.where(:id => params[:person_id]).first
+    if params[:person_id] && fetch_person(params[:person_id])
       @aspect = :profile
       @contact = current_user.contact_for(@person)
-      @aspects_with_person = []
       if @contact
-        @aspects_with_person = @contact.aspects
-        @aspect_ids = @aspects_with_person.map{|x| x.id}
+        @aspects_with_person = @contact.aspects.load
+        @aspect_ids = @aspects_with_person.map(&:id)
         gon.aspect_ids = @aspect_ids
-        @contacts_of_contact = @contact.contacts
-        render :layout => nil
+        render layout: nil
+      else
+        @aspects_with_person = []
       end
-    elsif(request.format == :mobile)
+    elsif request.format == :mobile
       @aspect = :all
-      @aspects = current_user.aspects
-      @aspect_ids = @aspects.map{ |a| a.id }
+      @aspects = current_user.aspects.load
+      @aspect_ids = @aspects.map(&:id)
       gon.aspect_ids = @aspect_ids
     else
       redirect_to stream_path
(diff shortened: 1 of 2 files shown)


app/controllers/status_messages_controller.rb
```diff
@@ -1,128 +1,97 @@
...
   def new
-    if params[:person_id] && @person = Person.where(:id => params[:person_id]).first
+    if params[:person_id] && get_person(params[:person_id])
       @aspect = :profile
       @contact = current_user.contact_for(@person)
-      @aspects_with_person = []
       if @contact
         @aspects_with_person = @contact.aspects
-        @aspect_ids = @aspects_with_person.map{|x| x.id}
+        @aspect_ids = @aspects_with_person.map(&:id)
```
jhass: We know that `@aspects_with_person` is a relation here (`@contact.aspects`), if the aspect names are used in the view, we probably should signal that and eagerly fetch them with `load`, else this line should use `pluck(:id)` to avoid instantiating them at all.

app/controllers/status_messages_controller.rb
```diff
@@ -1,128 +1,97 @@
...
+  respond_to :html, :mobile, :json
 
-  layout 'application', only: :bookmarklet
+  layout "application", only: :bookmarklet
 
   # Called when a user clicks "Mention" on a profile page
   # @param person_id [Integer] The id of the person to be mentioned
   def new
-    if params[:person_id] && @person = Person.where(:id => params[:person_id]).first
+    if params[:person_id] && get_person(params[:person_id])
```
jhass: `get_` prefixes are always smelly in Ruby, I'd opt for `fetch_person` or `find_person`.

app/controllers/status_messages_controller.rb
```diff
@@ -1,128 +1,97 @@
...
+        render layout: nil
+      else
+        @aspects_with_person = []
       end
-    elsif(request.format == :mobile)
+    elsif request.format == :mobile
       @aspect = :all
       @aspects = current_user.aspects
-      @aspect_ids = @aspects.map{ |a| a.id }
+      @aspect_ids = @aspects.map(&:id)
```
jhass: Similar like above, either `load` `current_user.aspects` or `pluck` the IDs.

app/controllers/status_messages_controller.rb
```diff
@@ -1,128 +1,97 @@
...
   end
 
+  def get_person(person_id)
+    @person = Person.where(id: person_id).first
+  end
+
   def bookmarklet
     @aspects = current_user.aspects
-    @aspect_ids = @aspects.map{|x| x.id}
+    @aspect_ids = @aspects.map(&:id)
```
jhass: And thinking a bit more about it, Rails even has a convenience method for the plucking on the association, `current_user.aspect_ids`.

app/services/status_message_service.rb
```diff
@@ -0,0 +1,86 @@
...
+    build_poll(params)
+    build_photos(params)
+    build_services(params)
+  end
+
+  def save
+    @status_message.save
+  end
+
+  def process_status
```
jhass: I think I'd call it `process_message`, since it's a `StatusMessage`, not a `MessageStatus`. I'm also not sure about returning it and relying on the return value here, I think I'd just require calling the accessor in any case.

app/services/status_message_service.rb
```diff
@@ -0,0 +1,86 @@
...
+  attr_reader :user, :services, :destination_aspect_ids
+
+  def normalize_aspect_ids(params)
+    params[:status_message][:aspect_ids] = [*params[:aspect_ids]]
+  end
+
+  def normalize_public_flag!(params)
+    sm = params[:status_message]
+    public_flag = (sm[:aspect_ids] && sm[:aspect_ids].first == "public") || sm[:public]
+    public_flag.to_s.match(/(true)|(on)/) ? public_flag = true : public_flag = false
```
jhass: put the assignment in front of the ternary, else it would be control flow.

app/services/status_message_service.rb
```diff
@@ -0,0 +1,86 @@
...
+  def build_services(params)
+    @services = [*params[:services]].compact
+  end
+
+  def add_status_to_streams
+    aspects = user.aspects_from_ids(@destination_aspect_ids)
+    user.add_to_streams(@status_message, aspects)
+  end
+
+  def dispatch_status
```
jhass: Same naming issue as above.

app/services/status_message_service.rb
theworldbright: Hmm I suppose in terms of "Adjective + Noun" it would be `process_message` but I hear a lot of people refer to status messages as just "status" (since just "message" would make it seem like something different) and thus `process_status` sounds better to me.

And yeah I guess it isn't clear by the method name that it will return a status message.

jhass: `process_status_message` isn't all that long actually and might be the best compromise after all. Though given the class is already `StatusMessageService`, a plain `process` would actually be fine with me too.

theworldbright: Yeah `process_status_message` sounds good to me.

app/controllers/status_messages_controller.rb
```diff
@@ -1,36 +1,33 @@
...
   def new
-    if params[:person_id] && @person = Person.where(:id => params[:person_id]).first
+    if params[:person_id] && fetch_person(params[:person_id])
       @aspect = :profile
       @contact = current_user.contact_for(@person)
-      @aspects_with_person = []
       if @contact
         @aspects_with_person = @contact.aspects
-        @aspect_ids = @aspects_with_person.map{|x| x.id}
+        @aspect_ids = @aspects_with_person.load.pluck(&:id)
```
jhass: Oh, I meant the load to be on the previous line of course.

jhass:
Merged as cae5f94af091cca9bd833d455860f0377e266b08

Thank you!

PR #6283, opened 2015-08-02
Suppress useless images

ghost:
With the port of header to BS3, some images are not used anymore. Plus, I wonder about the utility of [these glyphicons](glyphicons-halflings-blue.png).


svbergerem:
I think we can remove the glyphicons in this dir: https://github.com/diaspora/diaspora/tree/develop/app/assets/images/img

AugierLe42e:
Seems like it's ok to merge.

svbergerem:
Thank you.

Issue #6164, opened 2015-07-04
It is impossible to type cyrillic Б in profile/edit tags field


cmrd-senya:
If you try to type Б in russian keyboard layout to tags entry at the profile/edit page nothing happens, symbol doesn't appear. That possibly because it is on the same key that "," character, and somehow get blocked along with comma. I thought my key has been broken and started to disassemble my keyboard. And then just tried different input field.

AntonTkachov:
I am not sure that using keypress will not cause any side effects, but I haven't find any problems during my research and manual testing. Jasmine also passed ok

cmrd-senya:
I've just tested it on my development setup, and it works fine. However
on 0.5.1.1 at socializer.cc it doesn't. Probably it just has been solved
recently.

jaywink:
OK so closing as fixed already, thanks.

svbergerem:
@AntonTkachov Since you were trying to fix the issue with your PR: have you been able to reproduce the issue on the develop branch?

AntonTkachov:
@svbergerem Yes. In fact I still can reproduce it on both develop branch and UA pod -> https://diaspora.scorry.net. I'm confused that @cmrd-senya haven't this issue on his development setup.

AntonTkachov:
Since the issue had beed reopened I re-created pull request

jaywink:
Sorry, misunderstood a comment :(

cmrd-senya:
Ok, I tested last development version again on firefox and chromium.
Firefox has no issue, while chromium has. That's why I saw it on
production and didn't see on development - because I use firefox for
development and chromium for surfing. So we should probably rename topic
accordingly.

PR #6292, opened 2015-08-03
Issue #6164: fix cyrrilic 'б' in tags input. Move comma validation fr…

AntonTkachov:
…om keydown to keypress event

### DIFF
--- a/vendor/assets/javascripts/jquery.autoSuggest.custom.js
+++ b/vendor/assets/javascripts/jquery.autoSuggest.custom.js
@@ -197,16 +197,8 @@
               if(input.val() == ''){
                 break;
               }
-						case 13: case 188: // return, comma
-							tab_press = false;
-							var active = $("li.active:first", results_holder);
-							if(active.length > 0){
-								active.click();
-								results_holder.hide();
-							}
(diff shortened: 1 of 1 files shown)


AntonTkachov:
Fixes #6164

svbergerem:
Our autoSuggest code is based on an outdated version of autoSuggest and [it looks like the project is no longer maintained](https://github.com/wuyuntao/jquery-autosuggest). I think instead of adding yet another patch to some piece of code that is unmaintained and not covered by any tests, we should look for alternatives and use those. I started to do something similar for autocomplete in #6293 so I think a typeahead.js based alternative would be great.

AntonTkachov:
@svbergerem So, do you think we should just suspend this task and close pull request for now or should I try to make this pull request properly path Travis validation?

svbergerem:
> do you think we should just suspend this task and close pull request for now

Yes, I think so. Since we don't have any jasmine tests here we can neither make sure that everything else works like it did before (without doing manual tests) nor make sure that the fixed behaviour won't break with further changes. Since we are planning to replace that unmaintained code I don't think it is worth the effort if we would write our own tests.

Nevertheless, thank you for your work.

PR #6307, opened 2015-08-09
Extract services from comments controller

KentShikama:
For the comments API routes


app/services/comment_creation_service.rb
```diff
@@ -0,0 +1,12 @@
+class CommentCreationService
+  def initialize(params)
+    @user = params[:user]
+    @post_id = params[:post_id]
+    @text = params[:text]
+  end
+
+  def create_comment!
```
jhass: I tend to prefer Ruby core's naming convention of not having a `bang!` method if there's no `nonbang` variant.

app/controllers/comments_controller.rb
```diff
@@ -3,73 +3,73 @@
...
-    if current_user.owns?(@comment) || current_user.owns?(@comment.parent)
-      current_user.retract(@comment)
-      respond_to do |format|
-        format.js { render :nothing => true, :status => 204 }
-        format.json { render :nothing => true, :status => 204 }
-        format.mobile{ redirect_to :back }
-      end
+    deletion_service = CommentDeletionService.new(comment_id: params[:id], user: current_user)
+    if deletion_service.allow_destroy?
+      deletion_service.destroy_comment
```
jhass: Since we don't differentiate between not allowed and other failures, I'd say move the allowance check into `destroy_comment` and return `true`/`false` from it.

jhass:
For `Post` you did one that service that does both, creation and destruction. Having separate ones for `Comment` now feels a bit inconsistent to me.

theworldbright:
For posts I put the `show` and `destroy` together in one service and the `create` in another service. I suppose I should just separate the `show` and `destroy` service for the posts?

jhass:
I think I wouldn't mind a service per model. Currently we have about what, a dozen resources? With two to four of the CRUD actions that's 30-40 services.

theworldbright:
Yeah sounds good; I'll merge them together then (y)

jhass:
Merged as d7c92431aea25c56356b9678690798350ff7a38b

Thanks!

PR #6319, opened 2015-08-14
[Quickfix] Body padding-top use BS' SCSS variable

ghost:
Use of BS variable for `body`'s `padding-top`. This is to prevent unexpected behaviours if podmins want to change the header height. E.g, diaspote uses 45px instead of default's 50.


AugierLe42e:
Wut? WTF `Undefined variable: "$navbar-height"`!?

AugierLe42e:
Ready to merge.

svbergerem:
Thank you.

PR #6321, opened 2015-08-16
[Quickfix] Admin panel use 'form-control' class for selects

ghost:


AugierLe42e:
Ready to merge.

app/views/admins/stats.html.haml
```diff
@@ -7,7 +7,7 @@
 
       .pull-right
         = form_tag('/admins/stats', :method => 'get', class: 'form-inline') do
-          %select{:name => 'range'}
+          %select.form-control{:name => 'range'}
```
jhass: `name: "range"`

AugierLe42e:
Done.

app/views/admins/stats.html.haml
svbergerem: ping

AugierLe42e: I think you pinged an outdated diff :p

svbergerem: You only changed to single quotes to double quotes but didn't change the hash syntax so the comment is still valid.

AugierLe42e: That's right.

svbergerem:
Thanks!

PR #6333, opened 2015-08-21
Bump jsxc alpha

ghost:
this is tested for stable (alpha) branch


Flaburgan:
Do we want to mention in the changelog that improvements to the chat were integrated? Saying that this is still alpha of course.

PR #6338, opened 2015-08-22
bump diaspora-vines to v0.2.0-develop

ghost:
I think using pre-releases for vines isn't a bad idea as well as for jsxc


jhass:
Merged as f44bb613b95ea02a7506fa8a2e9e865f344fc8ae

Thanks!

PR #6339, opened 2015-08-23
Disable the handlebars auto-indent feature

### LINKED ISSUE #6253: Tabulation seems to be inserted after each line break in code blocks of a reshare in stream.
Here is an exemple:

* original post:

original

* reshare:

reshare1

It doesn't do this on SPV:

* [original post](https://diaspora-fr.org/posts/1412151),
* [reshare](https://diaspora-fr.org/posts/1412225).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25014797-tabulation-seems-to-be-inserted-after-each-line-break-in-code-blocks-of-a-reshare-in-stream?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #6253.

> `preventIndent`: By default an indented partial-call causes the output of the whole partial being indented by the same amount. This leads to unexpected behavior when the partial writes `pre`-tags. Setting this option to `true` will disable the auto-indent feature.

(http://handlebarsjs.com/reference.html)

### DIFF
--- a/config/initializers/handlebars_assets.rb
+++ b/config/initializers/handlebars_assets.rb
@@ -1,0 +1,1 @@
+HandlebarsAssets::Config.options = {preventIndent: true}


jhass:
Nice catch!

Merged as a7287f4f937cbacee4bfc647b93e586cc016a4c6

Thanks!

PR #6309, opened 2015-08-09
Left and right navbar polishing

ghost:
Supersedes #6112 and #6113.

PR #6342, opened 2015-08-23
Fix style regressions and code style for sidebar polishing PR

svbergerem:
Some of the things #6112 implemented broke in a rebase in #6309.

**before**
before

**after**
after


jhass:
Thank you!

PR #6351, opened 2015-08-26
5324 create new role moderator

zaziemo:
The people who are assigned the moderator role should be able to see, mark and delete reported comments/posts. They also should get an email when a new post/comment is reported. Mostly everything works fine, only the email's language doesn't change according to the person's language settings (it takes the language settings of the first one).


app/controllers/application_controller.rb
```diff
@@ -150,7 +157,7 @@ def gon_set_current_user
     return unless user_signed_in?
     a_ids = session[:a_ids] || []
     user = UserPresenter.new(current_user, a_ids)
-    gon.push({:user => user})
+    gon.push({user: user})
```
diaspora-code-review: Redundant curly braces around a hash parameter.

app/controllers/application_controller.rb
```diff
@@ -88,7 +88,14 @@ def set_locale
...
   def redirect_unless_admin
     unless current_user.admin?
-      redirect_to stream_url, :notice => 'you need to be an admin to do that'
+      redirect_to stream_url, notice: "you need to be an admin to do that"
+      return
+    end
+  end
+
+  def redirect_unless_admin_or_moderator
+    unless current_user.moderator? || current_user.admin?
```
diaspora-code-review: Use a guard clause instead of wrapping the code inside a conditional expression.

app/controllers/report_controller.rb
```diff
@@ -4,7 +4,7 @@
 
 class ReportController < ApplicationController
   before_filter :authenticate_user!
-  before_filter :redirect_unless_admin, :except => [:create]
+  before_filter :redirect_unless_admin_or_moderator, except: [:create]
```
diaspora-code-review: Prefer `before_action` over `before_filter`.

app/mailers/report_mailer.rb
```diff
@@ -1,30 +1,32 @@
...
-    Role.admins.each do |role|
-      person = Person.find(role.person_id)
-      if person.local?
-        user = User.find_by_id(person.owner_id)
-        unless user.user_preferences.exists?(:email_type => :someone_reported)
-          resource[:email] = user.email
-          format(resource)
-        end
+    person = Person.find(role.person_id)
+    if person.local?
```
diaspora-code-review: Use a guard clause instead of wrapping the code inside a conditional expression.

app/models/user.rb
```diff
@@ -489,6 +489,10 @@ def admin?
     Role.is_admin?(self.person)
   end
 
+  def moderator?
+    Role.moderator?(self.person)
```
diaspora-code-review: Redundant `self` detected.

Flaburgan:
Awesome <3

app/assets/templates/header_tpl.jst.hbs
```diff
@@ -94,6 +94,9 @@
                 {{#if current_user.admin}}
                   <li><a href="/admins/dashboard">{{t "header.admin"}}</a></li>
                 {{/if}}
+                {{#if current_user.moderator}}
```
troessner: Let's rather use obvious predicate methods here:

```Ruby
current_user.moderator?
```

app/assets/templates/header_tpl.jst.hbs
```diff
@@ -94,6 +94,9 @@
...
+                  <li><a href="/report">Moderator</a></li>
```
troessner: Shouldn't this be internationalized to support multiple languages?

app/controllers/application_controller.rb
```diff
@@ -35,7 +35,7 @@ def default_serializer_options
   end
 
   def ensure_http_referer_is_set
-    request.env['HTTP_REFERER'] ||= '/'
+    request.env["HTTP_REFERER"] ||= "/"
```
troessner: Was it intentional to clean this up in this PR as well? Personally I prefer "one PR == one feature" and then have cleanup PR's afterwards. Otherwise it's hard to see what change is coming from where and why.

app/controllers/application_controller.rb
```diff
@@ -68,11 +68,11 @@ def ensure_page
...
 
   def set_diaspora_header
-    headers['X-Diaspora-Version'] = AppConfig.version_string
+    headers["X-Diaspora-Version"] = AppConfig.version_string
 
     if AppConfig.git_available?
-      headers['X-Git-Update'] = AppConfig.git_update if AppConfig.git_update.present?
-      headers['X-Git-Revision'] = AppConfig.git_revision if AppConfig.git_revision.present?
+      headers["X-Git-Update"] = AppConfig.git_update if AppConfig.git_update.present?
+      headers["X-Git-Revision"] = AppConfig.git_revision if AppConfig.git_revision.present?
```
troessner: That's what I meant above - by doing those cleanup tasks as well it's harder for me, the reviewer, to actually .....review this PR.:)

app/mailers/report_mailer.rb
```diff
@@ -1,30 +1,32 @@
 class ReportMailer < ActionMailer::Base
-  default :from => AppConfig.mail.sender_address
+  default from: AppConfig.mail.sender_address
 
-  def new_report(type, id)
+  def self.new_report(type, id)
```
troessner: Personally I'd prefer the use of "module_function" instead of the "self" syntax but maybe that's just me.:)

app/models/role.rb
```diff
@@ -17,6 +18,14 @@ def self.add_admin(person)
     find_or_create_by(person_id: person.id, name: "admin")
   end
 
+  def self.moderator?(person)
+    exists?(person_id: person.id, name: "moderator")
+  end
```
jhass: Maybe we should just consider all admins moderators? So the scope would become `where(name: %w(moderator admin))`, this would become `moderators.exists?(person_id: person.id)`, the filter would become `redirect_unless_moderator` and reports would be mailed to `Role.moderators.map`

app/mailers/report_mailer.rb
```diff
@@ -1,30 +1,32 @@
...
+    [Role.admins + Role.moderators].flatten.map {|role| super(type, id, role) }
```
troessner: You can use Enumerable#flat_map here which is more concise. Bonus points: it's significantly faster as well -> https://github.com/JuanitoFatas/fast-ruby#enumerablemaparrayflatten-vs-enumerableflat_map-code

app/mailers/report_mailer.rb
```diff
@@ -1,30 +1,32 @@
...
+  end
+
+  def new_report(type, id, role)
```
troessner: How about using keyword arguments here to improve the readability of the method signature?

app/mailers/report_mailer.rb
jhass: Not applicable here, `#new_report` and `::new_report` are intentionally different methods.

app/mailers/report_mailer.rb
```diff
@@ -1,30 +1,32 @@
...
-      if person.local?
-        user = User.find_by_id(person.owner_id)
-        unless user.user_preferences.exists?(:email_type => :someone_reported)
-          resource[:email] = user.email
-          format(resource)
-        end
+    person = Person.find(role.person_id)
+    if person.local?
+      user = User.find_by_id(person.owner_id)
+      unless user.user_preferences.exists?(email_type: :someone_reported)
```
troessner: Slight LoD violation: http://devblog.avdi.org/2011/07/05/demeter-its-not-just-a-good-idea-its-the-law/

app/mailers/report_mailer.rb
jhass: Note it's called through ActionMailer via the `super` call above.

app/models/user.rb
troessner: :+1:  ;)

app/presenters/user_presenter.rb
```diff
@@ -6,16 +6,16 @@ def initialize(user, aspects_ids)
...
-    self.user.person.as_api_response(:backbone).update(
-      { :notifications_count => notifications_count,
-        :unread_messages_count => unread_messages_count,
-        :admin => admin,
-        :aspects => aspects,
-        :services => services,
-        :following_count => self.user.contacts.receiving.count,
-        :configured_services => self.configured_services,
-      }
+  def to_json(options={})
```
troessner: Instead of using hashes I'd use keyword arguments here as well.

app/mailers/report_mailer.rb
jhass: But it wasn't introduced by this PR, so I would welcome fixing it in a separate PR.

app/presenters/user_presenter.rb
jhass: The API is mandated by the json stdlib.

app/presenters/user_presenter.rb
```diff
@@ -24,14 +24,14 @@ def services
...
+    user.services.map(&:provider)
   end
 
   def aspects
     @aspects ||= begin
                    aspects = AspectPresenter.as_collection(user.aspects)
-                   no_aspects = self.aspects_ids.empty?
-                   aspects.each{ |a| a[:selected] = no_aspects || self.aspects_ids.include?(a[:id].to_s) }
+                   no_aspects = aspects_ids.empty?
+                   aspects.each {|a| a[:selected] = no_aspects || aspects_ids.include?(a[:id].to_s) }
```
troessner: This "each" is pretty odd. Or maybe it's because I don't know the domain language. What is an "aspect" when it comes to an user?

app/presenters/user_presenter.rb
```diff
@@ -46,4 +46,8 @@ def unread_messages_count
   def admin
     user.admin?
   end
+
+  def moderator
```
troessner: I'd make that a properly syntaxed predicate method (with a "?"). Additionally, you could use the Forwardable module in ruby to use implicit delegation instead of explicit one even though I realize this might be a little overkill here.

app/presenters/user_presenter.rb
jhass: Finding a way to refactor this is completely out of scope of this PR

app/views/report/index.html.haml
```diff
@@ -1,7 +1,8 @@
 .container
   .row
     .col-md-3
-      = render partial: "admins/admin_bar"
+      - if @current_user.admin?
+        = render partial: "admins/admin_bar"
```
troessner: My rails days are long gone but i think you can just do

```Ruby
= render "admins/admin_bar"
````

File: spec/controllers/report_controller_spec.rb [hunk 19]
```diff
@@ -2,128 +2,180 @@
...
+      it "succeeds" do
+        put :update, id: @message.id, type: "post"
+        expect(response.status).to eq(302)
+        expect(Report.where(reviewed: true, item_id: @message.id, item_type: "post")).to be_truthy
+      end
+    end
+    context "mark comment report as moderator" do
+      before do
+        Role.add_moderator(alice.person)
+      end
```
troessner: Newline missing.

spec/controllers/report_controller_spec.rb
```diff
@@ -2,128 +2,180 @@
...
-        expect(Report.where(:reviewed => true, :item_id => @comment.id, :item_type => 'comment')).to be_truthy
+        expect(Report.where(reviewed: true, item_id: @comment.id, item_type: "comment")).to be_truthy
+      end
+    end
+
+    context "mark post report as moderator" do
+      before do
+        Role.add_moderator(alice.person)
+      end
+      it "succeeds" do
```
troessner: Newline missing.

app/presenters/user_presenter.rb
jhass: We're inside Rails, no need for forwardable, http://api.rubyonrails.org/classes/Module.html#method-i-delegate

spec/controllers/report_controller_spec.rb
```diff
@@ -2,128 +2,180 @@
...
-        expect(Report.where(:reviewed => false, :item_id => @message.id, :item_type => 'post')).to be_truthy
+        expect(Report.where(reviewed: false, item_id: @message.id, item_type: "post")).to be_truthy
       end
     end
-    context 'destroy comment as user' do
-      it 'is behind redirect_unless_admin' do
-        delete :destroy, :id => @comment.id, :type => 'comment'
+    context "destroy comment as user" do
+      it "is behind redirect_unless_admin_or_moderator" do
+        delete :destroy, id: @comment.id, type: "comment"
```
troessner: I'd rather use lazy loaded "lets" instead of boring old instance variables here and everywhere else: http://betterspecs.org/#let

app/mailers/report_mailer.rb
```diff
@@ -1,30 +1,32 @@
...
+        resource[:email] = user.email
+        format(resource)
```
jhass: Does wrapping this into `I18n.with_locale(user.language) do` help regarding the localization issue?

spec/models/role_spec.rb
```diff
@@ -45,6 +45,13 @@
     end
   end
 
+  describe ".add_moderator" do
```
troessner: I'd rather have a generic "add" method that takes :moderator as an argument. Otherwise you'll pretty quick bloat your models with repetitive methods like that.

app/presenters/user_presenter.rb
troessner: Ah, yeah. Like I said, my rails is rusty.:)

app/mailers/report_mailer.rb
troessner: :+1:

app/assets/templates/header_tpl.jst.hbs
zaziemo: yeah, absolutely. We kept it for later and then forgot :)

app/controllers/application_controller.rb
zaziemo: Yes, this was intentional because we thought to clean up the obvious things rubocop suggested. We didn't think about it being confusing.

app/controllers/application_controller.rb
zaziemo: @troessner or @jhass could someone of you explain how that guard clause should look like? We didn't manage to change the code accordingly.

jhass:
```rb
def redirect_unless_admin_or_moderator
  return if current_user.moderator? || current_user.admin?
```

app/models/role.rb
zaziemo: @jhass we also reflected some alternatives to the current design... we'll think it over again. I find it a little bit confusing to only mention moderators but include admins. But @Realtin also was pro including both roles somehow.

app/mailers/report_mailer.rb
zaziemo: Yeah! That works. Thx, @jhass!

jhass: Great! Would love a spec that ensures it stays that way ;)

app/controllers/application_controller.rb
troessner: But that would drop the message below @jhass 

@zaziemo Think about what happens when you would like to add another role, say "super_admin"? Then the method name would need to change to "redirect_unless_admin_or_moderator_or_super_admin". And then you'd like to another role and all hell would break loose :)

I'd rather go with something flexible like this:

```Ruby
def authorize
  return if user_allowed?
  redirect_to stream_url, notice: "you don't have the necessary rights to do that"
end

def user_allowed?
  current_user.moderator? || current_user.admin?
end
```

Overall this is still not the best design since i'd rather go with something like "user has_many roles" or a gem that abstracts away role management but I guess thats out of the scope of this PR..;)

app/mailers/report_mailer.rb
```diff
@@ -1,18 +1,21 @@
...
+      id:   id
     }
-
-    Role.admins.each do |role|
-      person = Person.find(role.person_id)
-      if person.local?
-        user = User.find_by_id(person.owner_id)
-        unless user.user_preferences.exists?(:email_type => :someone_reported)
+    person = Person.find(role.person_id)
+    if person.local?
```
diaspora-code-review: Use a guard clause instead of wrapping the code inside a conditional expression.

app/assets/templates/header_tpl.jst.hbs
jhass: That's a Handlebars template rendered on the client side, not Ruby ;)

Realtin:
Ok this should be ready to merge. 
I squashed the refactor commits. do you want more squashed? (I'm getting a hang of it now..)

app/models/role.rb
```diff
@@ -17,6 +18,14 @@ def self.add_admin(person)
     find_or_create_by(person_id: person.id, name: "admin")
   end
 
+  def self.moderator?(person)
+    exists?(person_id: person.id, name: "moderator")
```
jhass: This should be `moderators.exists?(person_id: person.id)` and then the `|| current_user.admin?` above shouldn't be needed anymore.

jhass: And then three specs would be awesome, `expect(Role.moderator?(admin_person)).to be_true`, `expect(Role.moderator?(moderator_person)).to be_true`, `expect(Role.moderator?(normal_person)).to be_false`

Realtin:
We added the specs and refactored the method. what do you think?

jhass:
I think we're almost done here. I'd like the three specs for `Role.moderator?(some_person)` specifically, but the ones for the scope are good too, so don't drop them.

zaziemo:
@jhass do you mean in the same document but outside the scope?

spec/models/role_spec.rb
```diff
@@ -22,6 +23,20 @@
...
+
+      it "should include moderators" do
+        expect(Role.moderators).to include(moderator_role)
+      end
+
+      it "should not include normal users" do
+        expect(Role.moderators).to_not include(person)
+      end
+    end
   end
```
jhass: No. I'm saying that these tests are for the scope `moderators` and it's good to have them, but I'd also like to see them for `Role.moderator?(person)`, like we have specs for `Role.is_admin?(person)` below.

zaziemo:
@jhass should be ready now. We rebased to the current develop state. 
Travis fails are caused by the cucumber tests.

jhass:
Merged as d38741d5ebc8ac8a4e0696f172713626ed1f3a6f c2c6ed5dea50a8ac50141c404afda9e17f7a3126 15b186518c47665f3765bf175aa9f94068f05e8c 098c30c2b7dbfd6138de591b3ca379cb4ef9d87e 0b420d0308046bfac3f63f341f52a9e2806afa90 b2dc77e1e6cbef4702f9ba23605d775126eea306 bc75371b7a95f8ead5f463caf4ef7fe7db32c021 3a3c8817358c3487886ce272c58ac359fa51cb37

Thank you!

Flaburgan:
Nice! Thank you!

Flaburgan:
@zaziemo can you also edit [the wiki](https://wiki.diasporafoundation.org/FAQ_for_pod_maintainers#What_are_roles_and_how_do_I_use_them.3F_.2F_Make_yourself_an_admin) to indicate to podmin how to turn a user into a moderator?

zaziemo:
@Flaburgan sure. How do I indicate that the feature will only be available with the next release?

jhass:
Usually we do wiki updates only with the releases.

Issue #6353, opened 2015-08-27
add an option to force mobile version (for screen readers like orca)


pravi:
/m should redirect to mobile login page and activate mobile/toggle (only if not already in mobile version)

jhass:
I think it should just ensure the mobile force flag is set and redirect to the `stream_path`

PR #6354, opened 2015-08-27
Mobile path

ghost:
`/m` path to solve #6353.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -229,6 +229,7 @@ Diaspora::Application.routes.draw do
   # Mobile site
 
   get 'mobile/toggle', :to => 'home#toggle_mobile', :as => 'toggle_mobile'
+  get "/m", to: "home#force_mobile", as: "force_mobile"
 
   # Help
   get 'help' => 'help#faq', :as => 'help'
--- a/features/mobile/home.feature
+++ b/features/mobile/home.feature
@@ -10,3 +10,10 @@ Feature: Visit the landing page of the pod
     When I toggle the mobile view
     And I go to the root page
     Then I should see "Welcome, friend"
+
+    When I am on the root page
+    Then I should see "Welcome, friend"
+    When I go to the mobile path
+    Then I should see "LOG IN"
+    When I go to the mobile path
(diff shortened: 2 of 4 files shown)


jhass:
I think I prefer `/m` idea, it doesn't require any additional setup changes.

jhass:
Also this doesn't work if the main host of your pod is already second level.

jhass:
Oh and as I said I think it should redirect to `stream_path`.

AugierLe42e:
Looks better ?

config/routes.rb
```diff
@@ -18,6 +18,8 @@
     mount Sidekiq::Web => '/sidekiq', :as => 'sidekiq'
   end
 
+  get "/m", to: "home#force_mobile"
```
jhass: Let's move that down to `mobile/toggle`

features/mobile/home.feature
```diff
@@ -10,3 +10,10 @@ Feature: Visit the landing page of the pod
     When I toggle the mobile view
     And I go to the root page
     Then I should see "Welcome, friend"
+
+    When I am on the root page
+    Then I should see "Welcome, friend"
+    When I go to the mobile path
+    Then I should see "LOG IN"
+    And I go to the mobile path
```
jhass: `When`

features/mobile/home.feature
```diff
@@ -10,3 +10,10 @@ Feature: Visit the landing page of the pod
...
+    And I should see "LOG IN"
```
jhass: `Then`

config/routes.rb
jhass: Also a prettier name would be good, how about `as: `force_mobile`?

AugierLe42e:
I think this is ready for merge.

jhass:
Merged as 2ce7a1e185eeda9cb0b87064febb0e228e65264b

Thanks.

Flaburgan:
Thanks! Will be useful for https://github.com/Flaburgan/diaspora-webapp

PR #6357, opened 2015-08-28
[Quickfix] Correcting BS3 regression on badge color + spelling mistake

ghost:


app/views/conversations/_conversation.haml
```diff
@@ -16,10 +16,10 @@
               = other_participants.count - 1
 
       .bg
-        .badge.badge-dafault.message_count
+        .badge.badge-default.message_count
           = conversation.messages.size
         - if visibility.unread > 0
-          .badge.badge-important.unread_message_count
+          .badge.progress-bar-danger.unread_message_count
```
svbergerem: I don't think we should misuse Bootstrap classes. Instead you could set a red background for `.badge-important` via CSS.

svbergerem:
We have the same spelling mistake in `_conversation_subject.haml`. Would you mind fixing that too?

AugierLe42e:
Sure!

AugierLe42e:
Wow... Lots of random fails. Travis must be busy.

svbergerem:
Thank you.

PR #6360, opened 2015-08-28
Fix back-to-top backbone view

svbergerem:


denschub:
Merged as 1874c377065e854dff751c358729c6811e0a2fe5, thanks. ;)

PR #6371, opened 2015-08-31
pod check - make the error handling more robust

Raven24:
catch StandardError - higher up in the hierarchy
so the scheduled worker doesn't stop checking all pods if there's just one malfolmed pod

avoid echoing the complete body on json error - just a bit more sane :)


jhass:
Thanks

PR #6372, opened 2015-08-31
Fix syntax error in edit password view

KentShikama:


app/views/devise/passwords/edit.mobile.haml
```diff
@@ -23,7 +23,7 @@
               = f.password_field :password_confirmation
 
           .controls
-            = f.submit t('devise.passwords.edit.change_password'), :class: 'btn btn-primary'
+            = f.submit t('devise.passwords.edit.change_password'), class: 'btn btn-primary'
```
diaspora-code-review: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

jhass:
Ugh, good catch!

This really should have a test failing though, since it's only on develop, may I ask you to add one?

theworldbright:
Ahh ok; good to know it is only on develop. I will add the tests when I next get a larger patch of time.

theworldbright:
This has been fixed by #6504; I still plan to add the tests when I have time.

jhass:
Superseded by #6504 and #6525.

Issue #6341, opened 2015-08-23
Something is broken between JSXC and new design


AugierLe42e:
![screen shot 2015-08-23 at 12 45 21](https://cloud.githubusercontent.com/assets/3429231/9427904/e6c379ee-4994-11e5-8ec4-ed25fdc7bb59.png)

AugierLe42e:
I've tried to find a solution for this but it seems like using `position: fixed` on sidebars is not a good solution. The only solution I found was to use `position: sticky` which [is not well-supported by most browsers](http://caniuse.com/#feat=css-sticky). In the end, in short-term, we'll need to solve this with a bit of JS. :/

svbergerem:
Fixed by #6535.

PR #6324, opened 2015-08-16
Styling main stream

ghost:
This is a continuation of #6309. and is intended to be merged _after it_. This is a styling proposition inspired from @svbergerem's post [here](https://github.com/diaspora/diaspora/pull/6112#issuecomment-114666165). @pablocubico : I'd like your opinon on this.

screen shot 2015-08-16 at 14 40 46
screen shot 2015-08-16 at 14 40 58
screen shot 2015-08-16 at 14 41 02

PR #6381, opened 2015-09-04
[WIP] Fixing broken design between jsxc and main stream

ghost:
I removed the fixed position stuff and replaced it with some layers. Has the same effect at least for me ;)
Since I am not a pro at CSS at all I could use some help here and more importantly your opinion.

Does that make sense what I am doing? Can we replace fixed state? Or is that a no-go in respect of supporting multiple browser?

Firefox:

Firefox scrolling down:

Chrome:

Is there a better solution then `height: 100vh;`? Cause it obviously doesn't work in chrome.

related to #6341

@svbergerem @AugierLe42e @pablocubico

### DIFF
--- a/app/assets/stylesheets/stream.scss
+++ b/app/assets/stylesheets/stream.scss
@@ -1,10 +1,60 @@
+body {
+  background: $sidebars-sub-background;
+}
+
 .stream_container {
-	padding-left: 10px;
-	padding-top: 20px;
-	margin-left: -10px;
-	margin-top: -20px;
+  padding-left: 10px;
+  padding-top: 20px;
+  margin-left: -10px;
+  margin-top: -20px;
   #publisher {
     margin-bottom: 15px;
   }
   .well#ignore-info { text-align: center; }
+}
+
+.main_layer_one {
+  float: left;
+  width: 100%;
+  background: $sidebars-sub-background;
+  overflow: hidden;
+  position: relative;
+
+  .main_layer_two {
+    float: left;
+    width: 100%;
+    background: $white;
+    position: relative;
+    right: 25%;
+
+    .main_layer_three {
+      float: left;
+      width: 100%;
+      background: $sidebars-sub-background;
+      position: relative;
+      right: 50%;
+
+      .column_left {
+        float: left;
+        position: relative;
+        left: 75%;
+        overflow: hidden;
+      }
+
+      .column_middle {
+        float: left;
+        width: 45%;
+        position: relative;
+        left: 77%;
+        overflow: hidden;
+      }
+
+      .column_right {
+        float: left;
+        position: relative;
+        left: 80%;
+        overflow: hidden;
+      }
+    }
+  }
 }
--- a/app/views/streams/main_stream.html.haml
+++ b/app/views/streams/main_stream.html.haml
@@ -21,41 +21,39 @@
             = link_to '&times;'.html_safe, getting_started_completed_path, id: "gs-skip-x", class: "close"
 
 %div.container-fluid
-  .row
-    #left-navbar-fixed-sub-background.col-lg-1.visible-lg
-    #left-navbar-fixed-background.col-lg-offset-1.col-lg-2.col-md-3.hidden-sm.hidden-xs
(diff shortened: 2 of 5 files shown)


app/views/streams/main_stream.html.haml
```diff
@@ -21,41 +21,39 @@
...
-        .content
-
-      = render 'shared/right_sections'
-
-      %a{id:"back-to-top", title:"#{t('layouts.application.back_to_top')}", href:"#"}
-        &#8679;
+  .row.main_layer_one
+    .main_layer_two
+      .main_layer_three
+        #left-navbar.column_left.col-lg-offset-1.col-lg-2.col-md-3
```
diaspora-code-review: Classes should be listed before IDs (.column_left should precede #left-navbar)

app/views/streams/main_stream.html.haml
```diff
@@ -21,41 +21,39 @@
...
+              = link_to t("streams.activity.title"), activity_stream_path, rel: "backbone", class: "hoverable"
+            %li{data: {stream: "mentions"}}
+              = link_to t("streams.mentions.title"), mentioned_stream_path, rel: "backbone", class: "hoverable"
+            %li.all_aspects
+              = render "aspects/aspect_listings", stream: @stream
+            %li.followed-tags-sidebar
+              = render "tags/followed_tags_listings"
+
+        .column_middle.col-md-6
+          #aspect_stream_container.stream_container
```
diaspora-code-review: Classes should be listed before IDs (.stream_container should precede #aspect_stream_container)

app/views/streams/main_stream.html.haml
```diff
@@ -21,41 +21,39 @@
...
+            = render 'aspects/aspect_stream', stream: @stream
```
diaspora-code-review: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/views/streams/main_stream.html.haml
```diff
@@ -21,41 +21,39 @@
...
+
+        .column_right.col-md-3.col-lg-2.hidden-xs.hidden-sm.rightbar
+          #selected_aspect_contacts.section
```
diaspora-code-review: Classes should be listed before IDs (.section should precede #selected_aspect_contacts)

app/views/streams/main_stream.html.haml
```diff
@@ -21,41 +21,39 @@
...
+            .title
+              %h5.stream_title
+                = @stream.title
+            .content
+
+          = render 'shared/right_sections'
```
diaspora-code-review: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/views/streams/main_stream.html.haml
```diff
@@ -21,41 +21,39 @@
...
+
+          %a{id:"back-to-top", title:"#{t('layouts.application.back_to_top')}", href:"#"}
```
diaspora-code-review: Style/SpaceAfterColon: Space missing after colon.

app/assets/stylesheets/stream.scss
```diff
@@ -1,10 +1,60 @@
+body {
+  background: $sidebars-sub-background;
```
Zauberstuhl: that affects multiple views e.g. tags

AugierLe42e:
According to @pablocubico (who is the one who basically wrote the design), the fixed position is a common practice to get backgrounds all the way down. As you can see on pic n.2, the background color doesn't cover the whole page which breaks the initial design.

Zauberstuhl:
@AugierLe42e sure but that won't solve the problem, right? ;)

app/assets/stylesheets/stream.scss
AugierLe42e: You can remove this. It's aready a part of #6324 ;)

Zauberstuhl: perfect cause on my end it breaks stuff cheers

svbergerem:
Superseded by #6535.

Issue #3938, opened 2013-01-29
Tumblr service always posts to all blogs


Darael:
It would be nice if the Tumblr service could post to a subset of a user's blogs; many users run side blogs that general posts shouldn't go to.  Failing that, using their primary blog only is probably a saner default than the current approach, which is to post to all of their blogs.

A very nice feature indeed, but one that would require rather more extensive changes, would be the ability to effectively add a Tumblr blog (with the prerequisite that one had authorised its associated account in D*'s services, of course) to an aspect, so that posts to that aspect would also be sent to that blog.

Darael:
...we've had a patch that provided at least a partial fix (posting to the primary blog only - suboptimal but a damn sight better than the current state) for two years now!  Can't we at least have that integrated?

jaywink:
@Darael feel free to push the patch for review to the core if no one has done so yet! Patches don't magically get pulled in - someone has to submit them as PR's.

Darael:
@jaywink Funny thou shouldst mention it - I'm in the process of doing just that.  Should I say in the PR that it closes this bug, and immediately open another one for actually choosing where to post, or should I just mention that it's partial progress on this one?  IIRC github's smartness means it does matter.

gordon-morehouse:
Thanks for working on this.

Darael:
@gordon-morehouse I just wish I'd got to it sooner - like two and a half years ago when I opened the ticket, or two years ago when @meitar first posted a patch.  Hopefully the PR (which has got to a functional state) will be accepted soon and we can actually use the feature.

meitar:
Oh shit are you using my patch? Sorry, I've been meaning to come back to this but in the interim I stopped using Tumblr all together so I dropped the ball on this.

Darael:
@meitar I based it on thy patch but it's been through a fair number of changes since then thanks to helpful feedback from @jhass (eg it uses `.find` with a fallback rather than `.each`) and I adjusted the spec to match.

PR #6386, opened 2015-09-05
Stop Tumblr posting to all blogs

Darael:
Based on meitar/diaspora@b5c01598.

Is a small improvement to issue #3938, but not the optimal solution - the optimal solution would probably involve treating Tumblr blogs as contacts, able to be put in aspects and crossposting anything posted to those aspects.  At the very least, an ideal solution should allow the user to choose which blog to post to, rather than forcing the primary only, as this version does.

### DIFF
--- a/app/models/services/tumblr.rb
+++ b/app/models/services/tumblr.rb
@@ -16,16 +16,18 @@ class Services::Tumblr < Service
   def post(post, url='')
     body = build_tumblr_post(post, url)
     user_info = JSON.parse(client.get("/v2/user/info").body)
-    blogs = user_info["response"]["user"]["blogs"].map { |blog| URI.parse(blog['url']) }
+    blogs = user_info["response"]["user"]["blogs"]
+    primaryblog = blogs.find(blogs[0]) {|blog| blog["primary"] }
     tumblr_ids = {}
-    blogs.each do |blog|
-      resp = client.post("/v2/blog/#{blog.host}/post", body)
-      if resp.code == "201"
-        tumblr_ids[blog.host.to_s] = JSON.parse(resp.body)["response"]["id"]
-      end
+
+    blogurl = URI.parse(primaryblog["url"])
+    resp = client.post("/v2/blog/#{blogurl.host}/post", body)
+    if resp.code == "201"
+      tumblr_ids[blogurl.host.to_s] = JSON.parse(resp.body)["response"]["id"]
+    end
+
     post.tumblr_ids = tumblr_ids.to_json
     post.save
-    end
   end
 
   def build_tumblr_post(post, url)
--- a/spec/models/services/tumblr_spec.rb
+++ b/spec/models/services/tumblr_spec.rb
@@ -11,17 +11,19 @@ describe Services::Tumblr, :type => :model do
 
   describe '#post' do
     it 'posts a status message to tumblr and saves the returned ids' do
-      response = double(body: '{"response": {"user": {"blogs": [{"url": "http://foo.tumblr.com"}]}}}')
+      response = double(body: '{"response": {"user": {"blogs":
+       [{"primary": false, "url": "http://foo.tumblr.com"},
+        {"primary": true, "url": "http://bar.tumblr.com"}]}}}')
       expect_any_instance_of(OAuth::AccessToken).to receive(:get)
       .with("/v2/user/info")
       .and_return(response)
 
       response = double(code: "201", body: '{"response": {"id": "bla"}}')
       expect_any_instance_of(OAuth::AccessToken).to receive(:post)
-      .with("/v2/blog/foo.tumblr.com/post", @service.build_tumblr_post(@post, ''))
+      .with("/v2/blog/bar.tumblr.com/post", @service.build_tumblr_post(@post, ""))
       .and_return(response)
 
-      expect(@post).to receive(:tumblr_ids=).with({"foo.tumblr.com" => "bla"}.to_json)
+      expect(@post).to receive(:tumblr_ids=).with({"bar.tumblr.com" => "bla"}.to_json)
 
       @service.post(@post)
     end


app/models/services/tumblr.rb
```diff
@@ -19,9 +19,11 @@ def post(post, url='')
     blogs = user_info["response"]["user"]["blogs"].map { |blog| URI.parse(blog['url']) }
     tumblr_ids = {}
     blogs.each do |blog|
-      resp = client.post("/v2/blog/#{blog.host}/post", body)
-      if resp.code == "201"
-        tumblr_ids[blog.host.to_s] = JSON.parse(resp.body)["response"]["id"]
+      if blog['primary']
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

jhass:
Can you try to write a spec for this? Probably using webmock to stub out the request for the blogs and expecting the post request to the right one.

Darael:
@jhass I'll see what I can do.

Darael:
No spec yet, but I fixed the Travis failure.  Still working on the spec.

Darael:
I am *hoping* that's enough changes to the spec to test this adequately.  If not, poke me and I'll fix it as need be.  Naturally.

spec/models/services/tumblr_spec.rb
```diff
@@ -11,7 +11,9 @@
 
   describe '#post' do
     it 'posts a status message to tumblr and saves the returned ids' do
-      response = double(body: '{"response": {"user": {"blogs": [{"url": "http://foo.tumblr.com"}]}}}')
+      response = double(body: '{"response": {"user": {"blogs":
+       [{"primary": true, "url": "http://foo.tumblr.com"},
+        {"primary": false, "url": "http://bar.tumblr.com"}]}}}')
```
jhass: Let's make sure it's not just picking the first one and order the primary behind the non-primary one.

app/models/services/tumblr.rb
```diff
@@ -16,12 +16,15 @@ def consumer_secret
...
-      resp = client.post("/v2/blog/#{blog.host}/post", body)
-      if resp.code == "201"
-        tumblr_ids[blog.host.to_s] = JSON.parse(resp.body)["response"]["id"]
+      if blog["primary"]
+        blogurl = URI.parse(blog["url"])
+        resp = client.post("/v2/blog/#{blogurl.host}/post", body)
+        if resp.code == "201"
+          tumblr_ids[blog.host.to_s] = JSON.parse(resp.body)["response"]["id"]
+        end
       end
```
jhass: I think we can make this a bit clearer by using `blog = user_info["response"]["user"]["blogs"].find {|blog| blog["primary"] }` and dropping the `.each` loop. Btw. Are we guaranteed to have a primary blog? Or is it for example possible that only one result is returned and then it's not marked as primary?

spec/models/services/tumblr_spec.rb
```diff
@@ -11,14 +11,16 @@
...
+       [{"primary": false, "url": "http://foo.tumblr.com"},
+        {"primary": true, "url": "http://bar.tumblr.com"}]}}}')
       expect_any_instance_of(OAuth::AccessToken).to receive(:get)
       .with("/v2/user/info")
       .and_return(response)
 
       response = double(code: "201", body: '{"response": {"id": "bla"}}')
       expect_any_instance_of(OAuth::AccessToken).to receive(:post)
-      .with("/v2/blog/foo.tumblr.com/post", @service.build_tumblr_post(@post, ''))
+      .with("/v2/blog/bar.tumblr.com/post", @service.build_tumblr_post(@post, ''))
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/services/tumblr.rb
Darael: Point.  The `.each` loop would need to make a return if this were ever to support choosing a subset of blogs to post to, but it can go for now.

Tumblr's API docs imply that fields may be omitted if irrelevant, but don't say outright whether this includes the primary field - and the only account to which I have access for testing already has multiple blogs, so no help there.

jhass: The defensive approach would be to fallback to the first blog if the find returns no result, though that'd also require another spec ensuring that behaviour ... ;)

Darael: The fallback is easy enough using the optional ifnone argument of Enumerable#find :
```
blogs = user_info["response"]["user"]["blogs"]
blog = blogs.find(blogs[0]) {|blog| blog["primary"]}
```
But I'm insufficiently familiar with rspec to be able to write the spec correctly, at least with any reasonable speed

app/models/services/tumblr.rb
```diff
@@ -16,16 +16,21 @@ def consumer_secret
   def post(post, url='')
     body = build_tumblr_post(post, url)
     user_info = JSON.parse(client.get("/v2/user/info").body)
-    blogs = user_info["response"]["user"]["blogs"].map { |blog| URI.parse(blog['url']) }
+    blogs = user_info["response"]["user"]["blogs"]
+    primaryblog = blogs.find(blogs[0]) { |blog| blog["primary"] }
+    if !primaryblog
```
diaspora-code-review: Favor modifier `if` usage when having a single-line body. Another good alternative is the usage of control flow `&&`/`||`.

diaspora-code-review: Favor `unless` over `if` for negative conditions.

app/models/services/tumblr.rb
```diff
@@ -16,16 +16,21 @@ def consumer_secret
   def post(post, url='')
     body = build_tumblr_post(post, url)
     user_info = JSON.parse(client.get("/v2/user/info").body)
-    blogs = user_info["response"]["user"]["blogs"].map { |blog| URI.parse(blog['url']) }
+    blogs = user_info["response"]["user"]["blogs"]
+    primaryblog = blogs.find(blogs[0]) { |blog| blog["primary"] }
```
diaspora-code-review: Space between { and | detected.

app/models/services/tumblr.rb
```diff
@@ -16,16 +16,18 @@ def consumer_secret
   def post(post, url='')
     body = build_tumblr_post(post, url)
     user_info = JSON.parse(client.get("/v2/user/info").body)
-    blogs = user_info["response"]["user"]["blogs"].map { |blog| URI.parse(blog['url']) }
+    blogs = user_info["response"]["user"]["blogs"]
+    primaryblog = blogs.find(blogs[0]) {|blog| blog["primary"]}
```
diaspora-code-review: Space missing inside }.

jhass:
Ah well, let's let it slip. Please fix the remaining style issues and squash your commits together:
```
$ git remote add upstream https://github.com/diaspora/diaspora.git
$ git fetch upstream
$ git checkout feature/3938-stop-tumblr-posting-to-all-blogs
$ git rebase -i upstream/develop
# Leave the first commit at pick, choose squash or fix for the others
$ git push -f origin feature/3938-stop-tumblr-posting-to-all-blogs
```

You can edit your last commit with `git commit --amend` btw.

Darael:
Yeah, I know about `--amend`, but my understanding was that if a commit had reached any repository other than the local and then one `--amend`ed it, things got hairy.

In any case.

Squashing now.

jhass:
https://travis-ci.org/diaspora/diaspora/jobs/78951946 spec's still off.

Re amending: It's fine for your feature branches, you don't want to force push ever to things others might have pulled.

Darael:
Right.  Just patched the problem with the spec, and the [build on that](https://travis-ci.org/diaspora/diaspora/builds/79023613) is passing so far, but I forgot to rebase again for the three new commits in develop, so the second build (linked as usual) is in the queue.

Darael:
OK, that's *bizarre* - the failing test has nothing to do with *any* external services, including Tumblr.  It's a Cucumber scenario - "post a text-only message to just one aspect", and it only occurred with ruby 2.2 and mysql; all three other cucumber scenarios went fine.  In any case, [the version of this PR from before the latest rebase](https://travis-ci.org/diaspora/diaspora/builds/79023613) passed everything, so I'm inclined to blame what's been merged in the meantime.

jhass:
Yes it's one of the randomly failing ones, don't worry.

jhass:
Merged as 2ace9b47b1de118c5766d780fb4b7f28719668fd

Thanks!

jhass:
Note I wrote a spec for the fallback case and noticed in such that the argument to `Enumerable#find` must be a callable ;)

PR #6395, opened 2015-09-11
Use backbone for flash messages

ghost:


app/assets/javascripts/app/views/feedback_view.js
```diff
@@ -11,14 +11,14 @@ app.views.Feedback = app.views.Base.extend({
...
+    "click .hide_post" : "hidePost"
   },
 
   tooltipSelector : ".label",
 
   initialize : function() {
-    this.model.interactions.on('change', this.render, this);
-    this.initViews && this.initViews(); // I don't know why this was failing with $.noop... :(
+    this.model.interactions.on("change", this.render, this);
+    this.initViews && this.initViews(); // I don"t know why this was failing with $.noop... :(
```
diaspora-code-review: Expected an assignment or function call and instead saw an expression.

jhass: And here's why blindly replacing ' with " is bad.

AugierLe42e:
That should be ready to review.

app/assets/javascripts/app/views/flash_messages_view.js
```diff
@@ -0,0 +1,18 @@
...
+    };
+
+    this.renderTemplate();
+    var flashElement = this.$el;
+
+    flashElement.addClass("expose");
+    setTimeout(function(){
+      flashElement.removeClass("expose");
+    }, 10100);
+  }
```
svbergerem: I'd like to have two additional functions: `success` and `error` (which simply call `flash` with the correct values). Outside of this view only `success` and `error` should be used. You could also rename `flash` to `_flash`. That improves the readability of your code and makes this view easier to use.

AugierLe42e: Ok, it looks like a good idea.

app/assets/stylesheets/colors.scss
```diff
@@ -29,6 +29,8 @@ $sidebars-background: #f0f0f0;
 $sidebars-sub-background: darken($sidebars-background, 12%);
 $left-navbar-drawer-background: darken($sidebars-background, 6%);
 
+$card-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.16), 0 2px 10px 0 rgba(0, 0, 0, 0.12);
```
svbergerem: I don't think that this is the right file. Maybe `variables.scss`?

spec/javascripts/app/views/flash_messages_view-spec.js
```diff
@@ -0,0 +1,14 @@
...
+  var flashMessages = new app.views.FlashMessages();
+
+  describe("render", function() {
+    it("creates a new div for the message and calls flashes.animateMessages", function() {
+      spyOn(flashMessages, "success");
+      flashMessages.success("success!");
+      expect(flashMessages.success).toHaveBeenCalled();
+      spyOn(flashMessages, "error");
+      flashMessages.error("error!");
+      expect(flashMessages.error).toHaveBeenCalled();
```
svbergerem: I'd like to see some tests for the three functions you wrote. These tests don't do anything valuable.

spec/javascripts/helpers/SpecHelper.js
```diff
@@ -190,7 +187,7 @@ spec.retrieveFixture = function(fixtureName) {
   } catch(e) {
     throw new Error("couldn't fetch " + path + ": " + e);
   }
-  var regExp = new RegExp(/Couldn\\\'t load \/fixture/);
+  var regExp = new RegExp(/Couldn\\\"t load \/fixture/);
```
svbergerem: no

AugierLe42e: Love :p

app/assets/stylesheets/colors.scss
AugierLe42e: You mean `boostrap-variables.scss` or a new `variables.scss` file?

svbergerem: A new `variables.scss` file. Although I'm not sure if we really need this variable here. Instead you could just add the value itself in `_flash_messages.scss`.

AugierLe42e: Nope, I will also use it in #6324 [here](https://github.com/diaspora/diaspora/pull/6324/files#diff-b48a3c07783ee2256d72153189c9743bR32). I thought it was more or less a color to me.

spec/javascripts/app/views/flash_messages_view-spec.js
AugierLe42e: Which tests? The functions I wrote do more or less nothing but rendering.

svbergerem: 1. Check that your `error` and `success` functions call `_flash` with the correct values
2. Check that `_flash` renders the correct flash message depending on the input. (right message, `error: true` vs. `error: false`)

svbergerem:
The first flash message after page load is not visible.

AugierLe42e:
> The first flash message after page load is not visible.

Which one ? `You are already signed in.` ? I do see it.

svbergerem:
> Which one ?

Start sharing / stop sharing with someone.

spec/javascripts/app/views/flash_messages_view-spec.js
```diff
@@ -0,0 +1,33 @@
...
+      expect(flashMessages._flash).toHaveBeenCalledWith("success!", false);
+      this.view.error("error!");
+      expect(flashMessages._flash).toHaveBeenCalledWith("error!", true);
+    });
+  });
+
+  describe("render", function(){
+    beforeEach(function(){
+      spec.content().html("<div id='flash-container'/>");
+      flashMessages = new app.views.FlashMessages({ el: $("#flash-container") })
```
diaspora-code-review: Missing semicolon.

AugierLe42e:
> Start sharing / stop sharing with someone.

I do see it too :/

features/support/application_cuke_helpers.rb
```diff
@@ -17,8 +17,8 @@ def flash_message_containing?(text)
 
   def flash_message(opts={})
     selector = opts.delete(:selector)
-    selector &&= "#flash_#{selector}"
-    find(selector || '.message', {match: :first}.merge(opts))
+    selector &&= ".alert-#{selector}"
+    find(selector || '#flash-message', {match: :first}.merge(opts))
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

features/support/application_cuke_helpers.rb
```diff
@@ -27,8 +27,8 @@ def confirm_form_validation_error(element)
   end
 
   def check_fields_validation_error(field_list)
-    field_list.split(',').each do |f|
-      confirm_form_validation_error('input#'+f.strip)
+    field_list.split(",").each do |f|
+      confirm_form_validation_error("input#"+f.strip)
```
diaspora-code-review: Surrounding space missing for operator `+`.

AugierLe42e: Rhaaaaa !

AugierLe42e:
Nice, this should be ok now.

svbergerem:
Mobile view:

alert mobile view

This happens because of [this line](https://github.com/AugierLe42e/diaspora/blob/126125226b9d38bebe12c72a18f882c96a88bef0/app/assets/stylesheets/mobile/mobile.scss#L41).

svbergerem:
I'm not sure which version I'd prefer.

compare both versions

jhass:
The yellow for errors definitely doesn't feel right. I also it being attached to header and the shadow could use a better/sharper contrast. Don't miss the rounded edges though.

AugierLe42e:
>The yellow for errors definitely doesn't feel right.

Hmm... It shouldn't be yellow. Yellow is for warnings.

app/assets/stylesheets/_flash_messages.scss
```diff
@@ -1,53 +1,24 @@
-
 @import 'new_styles/animations';
 
-#flash_notice,
-#flash_alert,
-#flash_error {
-  position : fixed;
+#flash-body {
```
diaspora-code-review: Avoid using id selectors

app/assets/stylesheets/_flash_messages.scss
```diff
@@ -1,53 +1,24 @@
...
-    max-width: 800px;
-
-    color : #fff;
-    background-color : rgba(0,0,0,0.8);
-    border : 1px solid rgba(255,255,255,0.7);
-    border-radius: 6px;
+  width: 100%;
+  &.expose {  @include animation(expose, 10s) }
 
+  #flash-message {
```
diaspora-code-review: Avoid using id selectors

AugierLe42e:
Are we ok with this?

svbergerem:
@AugierLe42e https://github.com/diaspora/diaspora/pull/6395#issuecomment-143014217 is still unresolved.

app/assets/javascripts/app/models/post/interactions.js
```diff
@@ -102,16 +98,11 @@ app.models.Post.Interactions = Backbone.Model.extend({
...
+    var interactions = this;
 
-    reshare.save()
+    this.post.reshare().save()
       .done(function(reshare) {
-        flash.render({
-          success: true,
-          notice: Diaspora.I18n.t("reshares.successful")
-        });
+          app.flashMessages.success(Diaspora.I18n.t("reshares.successful"));
```
svbergerem: You might also want to fix the indentation here. ;)

AugierLe42e:
> @AugierLe42e #6395 (comment) is still unresolved.

I answered:

> So two solution here: consider warning and failure are the same level (like seemingly it was before) or change the level of this flash message.

Up to you to decide. But IMHO, feels weird to display warning messages as error messages.

AugierLe42e:
So, what option do you decide?

AugierLe42e:
screen shot 2015-10-02 at 22 51 42

svbergerem:
Codewise and designwise this looks good to me. This PR needs a rebase but that shouldn't be a problem.

About the issue with the "warning" flash messages on failed logins: [Rails seems to use `:notice` and `:alert` for flash messages](http://api.rubyonrails.org/classes/ActionDispatch/Flash.html), we also have `:error` but right now `:error` and `:alert` look exactly the same. You could change [this line](https://github.com/diaspora/diaspora/pull/6395/files#diff-c70c347141b21b6f03e13ad752ecca26R20) to 
```ruby
{notice: "success", alert: "danger", error: "danger"}[name.to_sym]
```

Then `:alert` and `:error` would also look the same in your PR.

AugierLe42e:
Done.

AugierLe42e:
Should be ready to merge, now.

svbergerem:
Thanks!

Issue #6063, opened 2015-06-04
Bootstrap 3 regressions


svbergerem:
Let's collect the regressions introduced by the Bootstrap 3 port in this issue.

* [ ] Scroll bar for "to" input in conversations view (Chromium 43.0.2357.65) screenshots: [[1]](https://cloud.githubusercontent.com/assets/3798614/7995910/f632ffde-0b1a-11e5-8b6d-85bfb93ab990.png)
  [[2]](https://cloud.githubusercontent.com/assets/3798614/7995909/f630124c-0b1a-11e5-80c0-f8a41644544c.png)
* [ ] Avatars in conversation view hidden (Chromium 43.0.2357.65) [screenshot](https://cloud.githubusercontent.com/assets/3798614/7995927/3bf331b0-0b1b-11e5-9ba7-ca94f665dae6.png)
* [x] The dropdown aspect and the "mark as read" icon are too close to the right of the screen on the notification page. ([comment](https://github.com/diaspora/diaspora/issues/6063#issuecomment-109179380)) ([commit](https://github.com/diaspora/diaspora/commit/7b7700b31722371118b0bfa2eb8c2314b56796a4))
* [x] The dropdown to add a tag in the left side bar of the stream is misplaced (gap between it and the input) ([comment](https://github.com/diaspora/diaspora/issues/6063#issuecomment-109179380)) ([commit](https://github.com/diaspora/diaspora/commit/7b7700b31722371118b0bfa2eb8c2314b56796a4))
* [x] Avatars of people who likes are collapsed in the stream view ([comment](https://github.com/diaspora/diaspora/issues/6063#issuecomment-109180187)) ([commit](https://github.com/diaspora/diaspora/commit/7b7700b31722371118b0bfa2eb8c2314b56796a4))
* [x] Images post by link are cropped ([comment](https://github.com/diaspora/diaspora/issues/6063#issuecomment-109210968)) ([commit](https://github.com/diaspora/diaspora/commit/39fe6bfff231458aca2f4b075e870675bc369c3f))
* [x] No more animation when uploading an image ([comment](https://github.com/diaspora/diaspora/issues/6063#issuecomment-109210968))
* [x] Stream view and notifications view are not as wide as they were on mobile view. ([comment](https://github.com/diaspora/diaspora/issues/6063#issuecomment-109212095))
* [x] notifications content lacks padding (mobile) ([comment](https://github.com/diaspora/diaspora/issues/6063#issuecomment-109212095))
[...]

PR #6397, opened 2015-09-12
Fix adding the first comment in the mobile view

svbergerem:
Fixes https://github.com/diaspora/diaspora/issues/6063#issuecomment-139368965.


jhass:
Thanks!

Issue #4875, opened 2014-03-24
comments should not reload textarea when expanded


Raven24:
follow up tracking issue from #4858 

When comments are expanded to show more than three entries, the whole section is re-drawn, emptying the comment textarea (in case something was already entered).
This was fixed by #4858, but the proper way to handle this would be to only redraw the comment section, leaving the textarea alone.

svbergerem:
Fixed by #6406.

PR #6406, opened 2015-09-16
Comments expansion refactoring

cmrd-senya:
If you look at comment expand process with javascript debugger, you
will notice that at first comments get added to existing stream ("add"
event handler of model.comments gets launched because of model.comments.fetch).
Then the comment stream gets empty and then filled by the postRenderTemplate
handler.

This patch removes comments rerendering on the expansion. The fetching process is
changed so that older comments are added to the correct place, so the order
looks nice without rerendering. Thus, unnecessary job of rerendering is avoided
saving user's browser's resourses.

### DIFF
--- a/app/assets/javascripts/app/views/comment_stream_view.js
+++ b/app/assets/javascripts/app/views/comment_stream_view.js
@@ -21,16 +21,10 @@ app.views.CommentStream = app.views.Base.extend({
 
   setupBindings: function() {
     this.model.comments.bind('add', this.appendComment, this);
-    this.model.bind("commentsExpanded", this.storeTextareaValue, this);
-    this.model.bind("commentsExpanded", this.render, this);
   },
 
   postRenderTemplate : function() {
     this.model.comments.each(this.appendComment, this);
-
-    // add autoexpanders to new comment textarea
-    this.$("textarea").val(this.textareaValue);
-    autosize.update(this.$("textarea"));
   },
 
   presenter: function(){
@@ -62,22 +56,46 @@ app.views.CommentStream = app.views.Base.extend({
     }
   },
 
+  _insertPoint: 0, // An index of the comment added in the last call of this.appendComment
+
+  // This adjusts this._insertPoint according to timestamp value
+  _moveInsertPoint: function(timestamp, commentBlocks) {
+    if (commentBlocks.length === 0) {
+      this._insertPoint = 0;
+      return;
+    }
+
+    if (this._insertPoint > commentBlocks.length) {
+      this._insertPoint = commentBlocks.length;
+    }
+
+    while (this._insertPoint > 0 && timestamp < commentBlocks.eq(this._insertPoint - 1).find("time").attr("datetime")) {
+      this._insertPoint--;
+    }
+    while (this._insertPoint < commentBlocks.length &&
+        timestamp > commentBlocks.eq(this._insertPoint).find("time").attr("datetime")) {
+      this._insertPoint++;
+    }
+  },
+
   appendComment: function(comment) {
     // Set the post as the comment's parent, so we can check
     // on post ownership in the Comment view.
     comment.set({parent : this.model.toJSON()});
 
-    this.$(".comments").append(new app.views.Comment({
-      model: comment
-    }).render().el);
+    var commentHtml = new app.views.Comment({model: comment}).render().el;
+    var commentBlocks = this.$(".comments div.comment.media");
+    this._moveInsertPoint(comment.get("created_at"), commentBlocks);
+    if (this._insertPoint === commentBlocks.length) {
+      this.$(".comments").append(commentHtml);
+    } else {
+      commentBlocks.eq(this._insertPoint).before(commentHtml);
+    }
+    this._insertPoint++;
   },
 
   commentTextareaFocused: function(){
     this.$("form").removeClass('hidden').addClass("open");
-  },
-
-  storeTextareaValue: function(){
-    this.textareaValue = this.$('textarea').val();
   },
 
   expandComments: function(evt){
@@ -86,10 +104,7 @@ app.views.CommentStream = app.views.Base.extend({
 
     this.model.comments.fetch({
       success : function(resp){
-        self.model.set({
-          comments : resp.models,
-          all_comments_loaded : true
-        });
+        self.$("div.comment.show_comments").addClass("hidden");
 
         self.model.trigger("commentsExpanded", self);
       }
--- a/spec/javascripts/app/views/comment_stream_view_spec.js
+++ b/spec/javascripts/app/views/comment_stream_view_spec.js
@@ -5,20 +5,11 @@ describe("app.views.CommentStream", function(){
   });
 
   describe("binds", function() {
-    it("re-renders on a commentsExpanded trigger", function(){
-      spyOn(this.view, "render");
+    it("calls appendComment on insertion to the comments collection", function() {
+      spyOn(this.view, "appendComment");
       this.view.setupBindings();
-      this.view.model.trigger("commentsExpanded");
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/app/views/comment_stream_view.js
```diff
@@ -68,9 +67,15 @@ app.views.CommentStream = app.views.Base.extend({
     // on post ownership in the Comment view.
     comment.set({parent : this.model.toJSON()});
 
-    this.$(".comments").append(new app.views.Comment({
-      model: comment
-    }).render().el);
+    var commentHtml = new app.views.Comment({model: comment}).render().el;
+    var time = this.$(".comments time:last");
+    if(time.length == 0 || comment.attributes.created_at > time.attr("datetime")) {
```
diaspora-code-review: Expected '===' and instead saw '=='.

diaspora-code-review: Identifier 'created_at' is not in camel case.

cmrd-senya:
This is an alternative fix to a problem described before in #6295.

This one fails one Jasmine test: "app.views.CommentStream re-renders on a commentsExpanded trigger", though that is exactly what the patch does - removes re-rendering on a commentsExpanded trgger, because it's considered redundant. So maybe just remove the test?

svbergerem:
Sorry for the delay. With this PR we don't rerender the comment stream anymore so I'd like to see some additional tests for `appendComment` that it is doing the right thing. There might be also some problems with comments that get federated after the user loads the stream but before they click on "show more comments". Those comments might have to be displayed somewhere between the already displayed three comments. I guess you'll find that out while writing the tests.

app/assets/javascripts/app/views/comment_stream_view.js
```diff
@@ -63,14 +62,40 @@ app.views.CommentStream = app.views.Base.extend({
...
+    }
+
+    if(this.insertPoint > commentBlocks.length) {
+      this.insertPoint = commentBlocks.length;
+    }
+
+    while(this.insertPoint > 0 && timestamp < commentBlocks.eq(this.insertPoint-1).find("time").attr("datetime")) {
+      this.insertPoint--;
+    }
+    while(this.insertPoint < commentBlocks.length && timestamp > commentBlocks.eq(this.insertPoint).find("time").attr("datetime")) {
```
diaspora-code-review: Line is too long.

app/assets/javascripts/app/views/comment_stream_view.js
```diff
@@ -63,14 +62,40 @@ app.views.CommentStream = app.views.Base.extend({
...
     // Set the post as the comment's parent, so we can check
     // on post ownership in the Comment view.
     comment.set({parent : this.model.toJSON()});
 
-    this.$(".comments").append(new app.views.Comment({
-      model: comment
-    }).render().el);
+    var commentHtml = new app.views.Comment({model: comment}).render().el;
+    var commentBlocks = this.$(".comments div.comment.media");
+    this.moveInsertPoint(comment.attributes.created_at, commentBlocks);
```
diaspora-code-review: Identifier 'created_at' is not in camel case.

spec/javascripts/app/views/comment_stream_view_spec.js
```diff
@@ -83,6 +74,27 @@ describe("app.views.CommentStream", function(){
...
+          factory.comment({"created_at" : new Date(2000).toJSON(), "text": "2"}),
+          factory.comment({"created_at" : new Date(4000).toJSON(), "text": "4"}),
+          factory.comment({"created_at" : new Date(5000).toJSON(), "text": "5"}),
+        ]
+      }};
+      this.view = new app.views.CommentStream({model : factory.post(postAttrs)});
+      this.view.render();
+      this.view.model.comments.push(factory.comment({"created_at" : new Date(3000).toJSON(), "text": "to_delete"}));
+      this.view.model.comments.push(factory.comment({"created_at" : new Date(6000).toJSON(), "text": "6"}));
+      (new app.views.Comment({el: this.view.$(".comments div.comment.media")[1], model: this.view.model.comments.models[1]})).remove();
```
diaspora-code-review: Line is too long.

app/assets/javascripts/app/views/comment_stream_view.js
```diff
@@ -63,14 +62,41 @@ app.views.CommentStream = app.views.Base.extend({
...
+
+    if(this.insertPoint > commentBlocks.length) {
+      this.insertPoint = commentBlocks.length;
+    }
+
+    while(this.insertPoint > 0 && timestamp < commentBlocks.eq(this.insertPoint-1).find("time").attr("datetime")) {
+      this.insertPoint--;
+    }
+    while(this.insertPoint < commentBlocks.length
+        && timestamp > commentBlocks.eq(this.insertPoint).find("time").attr("datetime")) {
```
diaspora-code-review: Bad line breaking before '&&'.

cmrd-senya:
I found out, that my previous assumptions were wrong, so I had to rewrite this.

Actually, we are in the situation, when we can have any comment count before the expansion. I had to write a general code, so that any comment that we merge gets into the right place. Though I didn't want to do any usual kind of sorting, because it is too consuming. So I did the following: I save the index of the comment, which were added the last in ```insertPoint```. In most situations (page load, comments expansion, new comments by user) every new comment should be stored in the same position, where we inserted the previous one. So we use this information. On every call ```appendComment``` calls ```moveInsertPoint```, which checks by the timestamp if we are in the right place, and if not, it moves insertPoint. Thus we minimize seek operations to the situations where we merge comments non-consequently.

I wrote a test for this, which includes usual comments addition, deleting from the middle and addition of comments in a wrong order.

denschub:
Ping?

cmrd-senya:
Why orphan? That's not orphan.
That was finished and waiting for review. Probably I had to write a
comment that this is for review.

cmrd-senya:
@svbergerem, this PR has a comprehensive test case and running the test with #6295 shows that it doesn't work well in some cases.

So, if you aren't satifsfied with the changes maybe leave all this as is?

app/assets/javascripts/app/views/comment_stream_view.js
```diff
@@ -62,14 +61,41 @@ app.views.CommentStream = app.views.Base.extend({
     }
   },
 
+  insertPoint: 0, // An index of the comment added in the last call of this.appendComment
+
+  //This adjusts this.insertPoint according to timestamp value
+  moveInsertPoint: function(timestamp, commentBlocks) {
+    if(commentBlocks.length === 0) {
+      return;
```
svbergerem: You will have to set `this.insertPoint = 0` first. Otherwise when you delete the only comment and try to comment afterwards the comment won't be added to the comment stream. Please also add a test for that case.

app/assets/javascripts/app/views/comment_stream_view.js
```diff
@@ -22,7 +22,6 @@ app.views.CommentStream = app.views.Base.extend({
   setupBindings: function() {
     this.model.comments.bind('add', this.appendComment, this);
     this.model.bind("commentsExpanded", this.storeTextareaValue, this);
```
svbergerem: Since we don't re-render the whole stream anymore it looks like you can remove this, remove the `storeTextareaValue` method and also all occurrences of the stored textarea value. There is a spec that says "refills the comment textbox on success". I guess the spec should still work but the description would have to be reworded.

svbergerem:
It looks to me like this would fix #4875.

cmrd-senya:
Please review

svbergerem:
Looks good to me. Thank you. :)

cmrd-senya:
Yay! That was my longest opened PR I guess. Thanks!

PR #6407, opened 2015-09-17
#5397 proper links for mentions and tags in outgoing tumblr posts

Realtin:
We reformatted the tags and mentions to be links to the tag path and person's profile both formatted as markdown

### DIFF
--- a/app/models/services/tumblr.rb
+++ b/app/models/services/tumblr.rb
@@ -39,8 +39,23 @@ class Services::Tumblr < Service
     post.photos.each do |photo|
       html << "![photo](#{photo.url(:scaled_full)})\n\n"
     end
+    # save all the mentions within a post
+    @mentions = []
+    post.message.instance_values["options"][:mentioned_people].each do |o|
+      @person_url = o.url + "people/" + o.guid
+      @mentions << {name: o.name, url: @person_url}
+    end
+    # removes the hovercard link that gets added by default
     html << post.message.html(mentioned_people: [])
     html << "\n\n[original post](#{url})"
+    @mentions.each { |mention|
+      html.gsub!("#{mention[:name]}", "[#{mention[:name]}](#{mention[:url]})")
+    }
+    # the following regular expressions matches the tag with its surrounding html
+    # the two capture groups capture the tag with and without #-sign to build the
+    # markdown link
+    html.gsub!(/<a\s.+?>(#([\w|\d|_-|<3]+\b))<\/a>/, '[\1]' + "(#{AppConfig.pod_uri}tags/" + '\2)')
+    html
   end
 
   def delete_post(post)
@@ -63,4 +78,3 @@ class Services::Tumblr < Service
     @client ||= OAuth::AccessToken.new(@consumer, self.access_token, self.access_secret)
   end
 end
-
--- a/spec/models/services/tumblr_spec.rb
+++ b/spec/models/services/tumblr_spec.rb
@@ -2,7 +2,8 @@ require "spec_helper"
 
 describe Services::Tumblr, type: :model do
   let(:user) { alice }
-  let(:post) { user.post(:status_message, text: "hello", to: user.aspects.first.id) }
+  let(:test_string) { "@{#{bob.name}; #{bob.diaspora_handle}} can mention people like a pro #ilike" }
+  let(:post) { user.post(:status_message, text: test_string, to: user.aspects.first.id) }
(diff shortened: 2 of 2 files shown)


app/models/services/tumblr.rb
```diff
@@ -39,8 +39,23 @@ def tumblr_template(post, url)
...
+    # removes the hovercard link that gets added by default
     html << post.message.html(mentioned_people: [])
     html << "\n\n[original post](#{url})"
+    @mentions.each { |mention|
+      html.gsub!("#{mention[:name]}", "[#{mention[:name]}](#{mention[:url]})")
+    }
+    # the following regular expressions matches the tag with its surrounding html
+    # the two capture groups capture the tag with and without #-sign to build the
+    # markdown link
+    html.gsub!(/<a\s.+?>(#([\w|\d|_-|<3]+\b))<\/a>/, '[\1]' + "(#{AppConfig.pod_uri}tags/" + '\2)')
```
diaspora-code-review: Use `%r` around regular expression.

app/models/services/tumblr.rb
```diff
@@ -39,8 +39,23 @@ def tumblr_template(post, url)
     post.photos.each do |photo|
       html << "![photo](#{photo.url(:scaled_full)})\n\n"
     end
+    # save all the mentions within a post
+    @mentions = []
```
jhass: I don't think this needs to be an instance variable.

app/models/services/tumblr.rb
```diff
@@ -39,8 +39,23 @@ def tumblr_template(post, url)
...
+    post.message.instance_values["options"][:mentioned_people].each do |o|
```
jhass: We probably should add proper API for accessing it.

app/models/services/tumblr.rb
```diff
@@ -39,8 +39,23 @@ def tumblr_template(post, url)
...
+      @person_url = o.url + "people/" + o.guid
+      @mentions << {name: o.name, url: @person_url}
```
jhass: `o` is a person instance, right? Name it `person` and use `person.profile_url`

app/models/services/tumblr.rb
```diff
@@ -39,8 +39,23 @@ def tumblr_template(post, url)
...
+    end
```
jhass: This is a `map`

app/models/services/tumblr.rb
jhass: We really shouldn't duplicate the tag regexp here, rather add options to `MessageRenderer` and subsequently `Taggable` to output the links with the full URL.

spec/models/services/tumblr_spec.rb
```diff
@@ -49,6 +50,18 @@
...
 
+  describe "#tumblr_template" do
+    let(:tumblr_export) { service.tumblr_template(post, AppConfig.pod_uri) }
+
+    it "properly links mentions to the person's profile" do
+      expect(tumblr_export).to include("[#{bob.name}](#{bob.url}people/#{bob.guid})")
+    end
+
+    it "properly links tags to the pod's tag site" do
+      expect(tumblr_export).to include("[#ilike](#{AppConfig.pod_uri}tags/ilike)")
```
jhass: `AppConfig.url_to("/tags/ilike")`

zaziemo:
@jhass thanks for all the feedback. We will implement the changes!

app/models/services/tumblr.rb
jhass: Looking around there probably is a nice way we can reuse https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/taggable.rb#L62 (through MessageRenderer)

app/models/services/tumblr.rb
```diff
@@ -39,8 +39,23 @@ def tumblr_template(post, url)
...
+    # removes the hovercard link that gets added by default
     html << post.message.html(mentioned_people: [])
     html << "\n\n[original post](#{url})"
+    @mentions.each { |mention|
+      html.gsub!("#{mention[:name]}", "[#{mention[:name]}](#{mention[:url]})")
```
jhass: Mmmh, this will also replace occurrences of the name anywhere that aren't meant as  mentions. Say I mention a user named "commit" and post a link like https://github.com/foo/bar/commit/132345abdef ...

I'm afraid this likewise needs to add support to Mentionable for generating full URL links.

denschub:
Closing due to inactivity.

Issue #4601, opened 2013-11-23
Embed direct links to HTML5 media in <audio> / <video> tags


Bugsbane:
Currently if a user posts a link to a proprietary video source such as YouTube, the video gets embedded in their post for easy previewing and playing.

If someone posts a link to an open web HTML5 <video> tag compliant video (such as a webm or h264 mp4) it's left as just a link.

The same applies to audio (Soundcloud vs ogg/mp3)

As far as I can see, all this should require is looking for a link starting with http and ending with .ogg/.mp3/.mp4/.webm and if we find it, sticking it in html5 audio or video tags. From the sounds of it Friendica does this already. It would also help Diaspora's core values of encourage decentralization and increasing privacy for our users by removing YouTube / Vimeo / Soundcloud etc in some cases.

The main issue I can see is previews. If this is really an issue, then we could just not bother with them. Users could _additionally_ be allowed to set their own thumbnails in a way similar to posting pictures, eg:

```
+[Keyboard Cat](http://mydomain.com/awesumvids/thumnail2128.jpg)(http://mydomain.com/awesumvids/piano2128.webm)    
```

To embed a video called piano2128.webm with a preview image thumbnail2128.jpg.

Ideally though, people who don't know markup and just post a link would have it "just work"
##

cmrd-senya:
If nobody minds I'd like to work on it.

jhass:
Related discussion https://www.loomio.org/d/LROboQNj/html5-media-embed

jhass:
I think acceptable syntax would be to turn regular links and autolinks into embeds, so basically rendered `<a>` tags if their `href` ends in a matching extension.

Bugsbane:
> If nobody minds I'd like to work on it.

Please be my guest. This seems to be a remarkably simple change for the impact it would have. Real low hanging fruit.

svbergerem:
Another related discussion: http://talk.commonmark.org/t/embedded-audio-and-video/441

I think it would be great if we could implement this as a markdown-it plugin. There might be also other projects who are interested in this plugin. In the commonmark thread there is already one dev who started writing a markdown-it plugin. Maybe we can join forces.

Here are some things we should keep in mind / think about;
* Should we have a configurable list of sources or do we accept arbitrary sources?
* It would be great if the user has to activate the player before we load any files.
* What kind of information do we present our users before they activate the player?

Flaburgan:
>I think it would be great if we could implement this as a markdown-it plugin

+1

[...]

PR #6409, opened 2015-09-18
Enabling markdown-it-html5-embed plugin so user can embed audio and video

cmrd-senya:
Enabling markdown-it-html5-embed plugin so user can embed audio and video using markdown syntax in the HTML5 way (solves #4601).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -103,6 +103,7 @@ source "https://rails-assets.org" do
   gem "rails-assets-markdown-it--markdown-it-for-inline", "0.1.1"
   gem "rails-assets-markdown-it-sub",                     "1.0.0"
   gem "rails-assets-markdown-it-sup",                     "1.0.0"
+  gem "rails-assets-markdown-it-html5-embed",             "0.1.0"
   gem "rails-assets-highlightjs",                         "8.6.0"
   gem "rails-assets-typeahead.js",                        "0.11.1"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -579,6 +579,7 @@ GEM
     rails-assets-markdown-it (4.4.0)
     rails-assets-markdown-it-diaspora-mention (0.3.0)
     rails-assets-markdown-it-hashtag (0.3.1)
+    rails-assets-markdown-it-html5-embed (0.1.0)
     rails-assets-markdown-it-sanitizer (0.3.2)
     rails-assets-markdown-it-sub (1.0.0)
     rails-assets-markdown-it-sup (1.0.0)
@@ -877,6 +878,7 @@ DEPENDENCIES
   rails-assets-markdown-it--markdown-it-for-inline (= 0.1.1)!
   rails-assets-markdown-it-diaspora-mention (= 0.3.0)!
   rails-assets-markdown-it-hashtag (= 0.3.1)!
+  rails-assets-markdown-it-html5-embed (= 0.1.0)!
   rails-assets-markdown-it-sanitizer (= 0.3.2)!
   rails-assets-markdown-it-sub (= 1.0.0)!
   rails-assets-markdown-it-sup (= 1.0.0)!
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -31,6 +31,7 @@
 //= require markdown-it-sanitizer
 //= require markdown-it-sub
 //= require markdown-it-sup
+//= require markdown-it-html5-embed
 //= require highlightjs
 //= require clear-form
 //= require typeahead.js


jhass:
I think I'd prefer it to hook into `[]()` instead of ``. The disadvantage is that you can't link anymore without embedding, but the advantage of retaining backward compatibility with the mobile site (redcarpet) and tumblr outweights that in my opinion.

In any case we will need these to happen lazily, that is like the Youtube embeds for example the elements shouldn't even be in the DOM prior user consent of playing them. That will not only prevent wasting resources by prefetching things you never gonna play, but also should be a lot lighter when rendering the page.

Which brings me back to my idea of simply augmenting the `<a>`s generated by  `[]()`s with a class (`audio`, `video`) when the filename is recognized as embeddedable media. Then some CSS and a JS on click handler should be enough to implement the cover.

cmrd-senya:
So you mean leaving a ```<a>``` link in the post, and adding player to the end of the post just like it is done with youtube now?

jhass:
No I think I'd style the link such that it becomes a box with a play button, alike to the youtube player cover, that's then replaced on click with the actual embed.

cmrd-senya:
Ok, that will require some more work then.

But the embedding syntax you want contradicts with the commonmark guys vision [here](http://talk.commonmark.org/t/embedded-audio-and-video/441).

jhass:
I know, but the alternative is to also add the functionality to redcarpet at minimum, and probably have redcarpet's plaintext renderer convert those back to normal links.

Flaburgan:
I disagree about your syntax proposition @jhass, as linked in https://github.com/diaspora/diaspora/issues/4601#issuecomment-124288107 `[]()` is to do a link, `` is to embed an object, I really think we should not differ from the standard.

jhass:
@Flaburgan thanks for the feedback, to add something constructive, could you detail how you would tackle the issues I mentioned?

Flaburgan:
Well if I get it correctly the only point linked to the syntax is the compatibility with Redcarpet which has to be patched. The other points (don't embed the content immediately, wait for an user action first) look fine to me.

I never looked at redcarpet, I don't know if they easily accept pull requests and if it is hard to patch. But we already have several difference in features between desktop and mobile, it looks acceptable to me to say that this feature is only available on desktop at the moment.

cmrd-senya:
For tumblr, could we do some syntax conversion in [tumblr export code](https://github.com/diaspora/diaspora/blob/develop/app/models/services/tumblr.rb)?

jhass:
I don't like that so much, I made an effort to reduce the scathering and duplication of post processing around the codebase with #4836, also see my feedback in #6407

cmrd-senya:
I agree it is better to keep with the stadard, because:
1) The desicion is hard to undo.
2) In future there will be integration with some other federated networks (like friendica) with a different codebase, and it is more likely they'll keep to the standart. BTW, what does friendica have about embedding video?

cmrd-senya:
It's more important to accord with Friendica and other parts of the federation, than some external service.

jhass:
Well, while this sounds harsh, the Friendica interop is nothing that has explicit support from our side, while the Tumblr interop has.

cmrd-senya:
What do you mean by interop?
If I understand it correctly (looking at [the
page](http://the-federation.info/)), diaspora, friendica and redmatrix
exchange their posts. So if I embed video in Diaspora it should be
available in friendica. And it is more important to work ok, than export
service, because it is kind of network internals, working implicitly
without user notion.

jhass:
We don't implement Friendica's or Redmatrix's protocol, they implement ours. interop means inter operation.

cmrd-senya:
Ok, I see your point.

It makes sense for me, and I can implement it.

By I think we need to have a community consensus about the syntax. Maybe
do loomio voting then?

jhass:
I think it's a technical decision rather than one of preference. I agree that `` makes more sense semantically and given what CommonMark is heading at, however properly supporting it in all aspects (which I outlined above) is quite a bit more work. At the end of the day I leave that up to you, but I won't accept a patch that doesn't provide a solution that works well for all usecases, sorry.

jaywink:
I'd favour also the common standard, or not implementing at all. And not taking into account Friendica and RedMatrix "just because we don't implement them and they implement us" is just plain wrong IMHO.

I don't see why we should implement something differently from others just because it is extra work. Better not implement it at all than implement something that conflicts.

cmrd-senya:
@jhass, you wrote about doing things lazily, but it's hardly possible to make it lazier. When the video tag is rendered by browser, it doesn't prefetch the video, before user ever clicks the video box. Do we really need to do anything, but just add audio/video tag on a page? Would the usage of javascript onclick on box followed by replacement of content to the audio/video tag make any real improvents in perfomance or page render speed?

jhass:
I would expect it does, yes.

cmrd-senya:
What makes you think so?
We have cover for the youtube embeds, but youtube loads quite a big amount of html for each video. We have only a few lines of the ```<video>``` tag. What exactly do you think will consume resources? Drawing of video controls by the web browser?
Can you show some examples of web sites, who do it this way (puts video tag with javascript)? For example, tumblr places video tag right away at the page load:

denschub:
The browser may start loading the video right away since the spec does not provide rules for preloading stuff, it only issues a recommendation. To avoid automated preloading of the video file, I'd add `preload="metadata"` to the video tag just to make sure the browser will not download the file right away.

jhass:
I have a quite significant lag on this demo https://output.jsbin.com/makugi

denschub:
Well, true. A lot of video elements on the same page can cause slowness. Since we have an endless stream, I guess I have to change my mind and also prefer adding a onclick-element-loading like we have on YouTube elements. However, this should not be a big deal since we already have that code...

cmrd-senya:
It works fine in Chromium, but hangs for a while in Firefox. Looks like a browser-specific issue, even can be called a bug. I don't see any lags after page load, so for endless stream itself it won't be a problem, only if it loads a single post with many videos. In my opinion it's unlikely to happen too often, so not worth adding complexity to our code because of that. And the fact, that tumblr doesn't do it supports that idea.

I'll try to post it as a bug to Firefox's tracker, see what happen.

jhass:
Well, there's also to consider that a malicious user does it intentionally for a single post.

cmrd-senya:
I've [posted](https://bugzilla.mozilla.org/show_bug.cgi?id=1206758) a bug report to the Mozilla's bug tracker.

AugierLe42e:
>  but the advantage of retaining backward compatibility with the mobile site (redcarpet) and tumblr outweights that in my opinion.

Wouldn't the correct solution be to just drop RedCarpet and adopt Markdown-It both on mobile and desktop side?

Anyway, I agree with @Flaburgan and @jaywink. `[]()` syntax seems wrong and totally counter intuitive. And it would add another drawback : it takes aways the possibility for the user to chose between ambedding the video/audio or just make a link to it. That does not looks good to my eyes.

cmrd-senya:
>Wouldn't the correct solution be to just drop RedCarpet and adopt Markdown-It both on mobile and desktop side?
Yes, that's intenteresting for me as well.

At the moment, I've implemented both types of syntax in [the plugin](https://github.com/cmrd-senya/markdown-it-html5-embed). It can be changed in options on the plugin load.

After I worked on the issue for a while, I can say that embedding with link syntax doesn't feel so wrong to me as before. First, as far as I see, nobody really uses this syntax for embedding (so no compatibility with anyone so far). Second, almost every software piece that uses markdown, considers this syntax as image syntax, and using it differently would result in their attempts to interpret video/audio as image. This includes tumblr, but what is much more important - it includes older Diaspora pods. So for the upgrade period (and some podmins aren't very fast with updating) it will be usual, that new pods users use image syntax, and old pods users will see some crap like this in their stream:

It is not very pretty. For the link syntax it will just fall back to a link.

So though using `````` for an embed is more semantically correct, it's ruined by the fact, that all software (even ours) considers it as an image syntax, and audio/video can't be reasonably interpreted as an image, while can't be as a link.

AugierLe42e:
> Second, almost every software piece that uses markdown, considers this syntax as image syntax, and using it differently would result in their attempts to interpret video/audio as image.

Let me recall you that, if image interpretation fails, a text is still displayed.

Now, consider the following : if it is embedded under a link, but there are other links inside the post, we will have to choose between the audio/video and oEmbed which on to display at the end of the post.

Plus, if you choose this solution, you also take away the user the possibility to format the post himself because the embedded content will automatically be placed at the end of the post.

So, basically, this way of doing is only an ugly hack to sidestep problems. I don't feel like this is a correct way of doing it.

cmrd-senya:
@jhass, I noticed, that multiple videos render fine and fast in Firefox, if disable controls render ([the test](https://output.jsbin.com/wopiluwefe)). So we can implement it as tumblr does - they don't render controls before user clicks a video. They render only a big play button over a video.

But there is the same problem with a video poster (the image rendered before user clicks a video). If there are many videos, we experience a huge lag with ```preload="metadata"```, which adds a poster from a video. Tumblr caches an image and does ```poster="url/image_frame.jpg```. But to extract an image from a video and store it somewhere seems to me too complex.

jhass:
Yes, thus I still think looking into the reusablity of what we do for Youtube would be very well worth it. Just have a blackish/greyish stub of the size of the video embed, render a play button over it and replace on click.

cmrd-senya:
I think we could close it and move the following discussion to the #6418.

PR #6414, opened 2015-09-20
Remove unused social media logos

ghost:
Looks like Wordpress logo is the only one still used as image. The other are entypo fonts.


AugierLe42e:
Hmm... The images are still used on right sidebar.

Issue #4705, opened 2014-01-18
"Retina" resolution support


maliktunga:
Both the mobile and desktop versions lack high-DPI support for icons and images.

svbergerem:
I guess this should be fixed now. Please comment if there is still anything left to do.

PR #6244, opened 2015-07-24
Fix #6248

ghost:
Change `thumb_large` to resize width to 300px and crop height to 500px (or other value) if necessary (would solve #6248),

PR #6415, opened 2015-09-20
Replace mobile background image by a color

ghost:
I've taken the color from the image wit GIMP. It looks more or less the same than before:

screen shot 2015-09-20 at 17 29 41


Flaburgan:
Why would you want to do that?
And it doesn't look the same on pages where the content doesn't reach the bottom of the screen, like the publisher page.

AugierLe42e:
> Why would you want to do that?

Because the background image slowers content loading on slow networks (particularly in suburbs). I though that, with discussion on #6244, nobody would oppose removing useless content loading...

Flaburgan:
That was the funny part ^^ if you want to load 3MB pictures on mobile, you should not see any problem to use a 22kb pict as a background :)

AugierLe42e:
I want to load what's useful for. Images in streams are useful. Image background that can be replaced by a CSS rule doesn't really look useful ;)

svbergerem:
I guess this is the last open issue to close #4705.

AugierLe42e:
Feel free to merge, then ;)

app/assets/stylesheets/colors.scss
```diff
@@ -22,6 +22,7 @@ $light-green: lighten($green,20%);
 $red: #A80000;
 $blue: #3F8FBA;
 $dark-blue: darken(#0984C8,10%);
+$mobile-background: #E0DFDB;
```
diaspora-code-review: Color `#E0DFDB` should be written as `#e0dfdb`

svbergerem:
I think I'd even prefer the sidebars background that we also use for the desktop view. @Flaburgan what do you think?

AugierLe42e:
No problem for me. It adds consistency.

Flaburgan:
The lesser colors we have, the happier I am.

AugierLe42e:
Ok, here how it looks with new color:

screen shot 2015-10-02 at 10 28 27

app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -26,11 +26,7 @@ code {
   }
 
 body {
-  background: {
-    image: image-url("mobile/hatched-light.jpg");
-    /* scale background image down for iOS retina display */
-    size: 200px;
-  }
+  background-color: $sidebars-sub-background;
```
svbergerem: I was thinking about `$sidebars-background` which we also use for the desktop view as the background for the stream.

AugierLe42e: I tested it. I found it a bit light. I think mobile view needs more contrast than desktop, don't you?

AugierLe42e:
Hmm... I don't know. Maybe you're right, @svbergerem:

svbergerem:
Thank you!

PR #6418, opened 2015-09-22
Audio and video embedding in HTML5 way basic support

### LINKED ISSUE #4601: Embed direct links to HTML5 media in <audio> / <video> tags

cmrd-senya:
Enabling markdown-it-html5-embed plugin so user can embed audio and video using the markdown link syntax `[]()` in the HTML5 way.

Media files are embedded at the bottom of a post:

**To do:**
- [x] Update "How do I embed a video, audio or other multimedia content into a post?" in "Posts and posting" help section.

fixes #4601

Next steps after this one is merged:
- Support for covers for video files. We don't use HTML5 video preload attribute so in order to have covers we need to fetch them in the backend. Also we can fetch metadata to know about video size.
- Support for consequent audio files playback so people can share playlists.


app/assets/javascripts/app/helpers/text_formatter.js
```diff
@@ -80,6 +80,15 @@
 
     // Bootstrap table markup
     md.renderer.rules.table_open = function () { return "<table class=\"table table-striped\">\n"; };
+
+    var html5medialPlugin = window.markdownitHTML5Embed;
+    md.use(html5medialPlugin, { html5embed: {
+      use_link_syntax: true,
```
diaspora-code-review: Identifier 'use_link_syntax' is not in camel case.

app/assets/javascripts/app/helpers/text_formatter.js
```diff
@@ -80,6 +80,15 @@
...
+      is_allowed_mime_type: function(mimetype) {
```
diaspora-code-review: Identifier 'is_allowed_mime_type' is not in camel case.

app/assets/javascripts/app/helpers/text_formatter.js
```diff
@@ -80,6 +80,15 @@
...
+        var v = document.createElement(mimetype[1]);
+        return v.canPlayType && v.canPlayType(mimetype[0]) !== '';
```
diaspora-code-review: Strings must use doublequote.

cmrd-senya:
This is an alternative to #6409 that uses a link syntax for the embed.

AugierLe42e:
I guess this needs loomio.

app/assets/javascripts/app/helpers/text_formatter.js
```diff
@@ -80,6 +80,20 @@
...
+
+    var html5medialPlugin = window.markdownitHTML5Embed;
+    md.use(html5medialPlugin, { html5embed: {
+      use_link_syntax: true,
+      templateName: "inline-embed_tpl",
+      attributes : {
+        "audio": "controls preload=none",
+        "video": "preload=none"
+      },
+      is_allowed_mime_type: function(mimetype) {
```
diaspora-code-review: Identifier 'is_allowed_mime_type' is not in camel case.

app/assets/stylesheets/inline-embed.scss
```diff
@@ -0,0 +1,15 @@
+.inline-embed {
+  background-color: #dddddd;
```
diaspora-code-review: Color literals like `#dddddd` should only be used in variable declarations; they should be referred to via variable everywhere else.

app/assets/stylesheets/inline-embed.scss
```diff
@@ -0,0 +1,15 @@
...
+  .thumb {
```
diaspora-code-review: Rule declaration should be preceded by an empty line

app/assets/stylesheets/inline-embed.scss
```diff
@@ -0,0 +1,15 @@
...
+    @include video-overlay;
+  }
```
diaspora-code-review: Rule declaration should be followed by an empty line

app/assets/stylesheets/inline-embed.scss
```diff
@@ -0,0 +1,15 @@
...
+  .thumb video {
+    width: 100%;
+    min-height: 60%;
+    vertical-align: middle;
+  }
```
diaspora-code-review: Rule declaration should be followed by an empty line

app/assets/stylesheets/inline-embed.scss
diaspora-code-review: Color `#dddddd` should be written as `#ddd`

app/assets/stylesheets/inline-embed.scss
```diff
@@ -0,0 +1,15 @@
+.inline-embed {
+  background-color: #dddddd;
+  .thumb {
+    @include video-overlay;
+  }
+  .thumb video {
```
diaspora-code-review: Merge rule `.thumb video` with rule on line 3

app/assets/stylesheets/inline-embed.scss
```diff
@@ -0,0 +1,15 @@
...
+    width: 100%;
```
diaspora-code-review: Properties should be ordered min-height, vertical-align, width

app/assets/stylesheets/inline-embed.scss
```diff
@@ -0,0 +1,15 @@
...
+    min-height: 60%;
+    vertical-align: middle;
+  }
+  audio {
+    width: 100%;
```
diaspora-code-review: Properties should be ordered height, width

cmrd-senya:
@jhass, here I've added a cover for a video, please check, if it is OK.

I still need to write some testsuite. And we have to decide on embed syntax - we still don't have a consensus.

jhass:
Looks good I think, should add `opus` for the audio extensions ;)

Maybe some more margin on the thumb?

cmrd-senya:
I use [mimoza](https://github.com/nodeca/mimoza) for mime types resolution, and it uses [mime-db](https://github.com/jshttp/mime-db) in its order. So to add opus extension we have to PR them.

Do you mean to extend a grey colored field, or to add a margin around the field? It is the same width as a youtube embed at the moment, though a little bit higher.

jhass:
Around it, though if it matches youtube it's fine I guess.

features/step_definitions/inline_embed_steps.rb
```diff
@@ -0,0 +1,10 @@
+When /^I post an embeddable media file URL "([^"]*)"$/ do |text|
+  write_in_publisher("[title](" + text + ")")
+  submit_publisher_no_check
+end
+
+Then /^I should see a HTML5 (video|audio) player$/ do |type|
+  find(".post-content .inline-embed")
+  find(".stream_container").should have_css(".post-content .inline-embed " + type)
+end
+
```
diaspora-code-review: 1 trailing blank lines detected.

jaywink:
If the proposal regarding the link syntax passes, I strongly feel the code needs to be changed so the embed does not happen in place. Imagine people posting links to things and suddenly they start being embedded in the middle of the sentence? If we don't agree on this we should have another vote on whether the embed adds it in place or like YouTube embeds happen.

Thanks

gaffatape:
I think that this should be solved using OpenGraph meta-data for preview thumbnail information and finding out if the content is a video. It would take no more action than to change the OpenGraph parsing to look for og:type=video and check for an acceptable og:video:type={html5 acceptable mime-type}. Fx. look at YouTube's OpenGraph tags for any video.

If Diaspora begins to read all kinds of file types and parse them, I think the beast is out of the box and will become unmaintanable. Diaspora is a social network where you can share links using common standards. Video parsing is best left with the video sites.

AugierLe42e:
> I think that this should be solved using OpenGraph meta-data for preview thumbnail information and finding out if the content is a video. It would take no more action than to change the OpenGraph parsing to look for og:type=video and check for an acceptable og:video:type={html5 acceptable mime-type}. Fx. look at YouTube's OpenGraph tags for any video.

Nope. We are not talking about embedding web pages, but medias like audios or videos.

gaffatape:
Okay, just pretend what I said didn't have anything to do with embedding media, that's your problem.

My point was to generalize the existing code that is already used for embedding other content (also video and audio). It is not needed to introduce a new tag. That JavaScript code proposed by jhass sounds great if it improves performance, but if all links are already OpenGraph parsed, why is that not used as the standard for embedding videos? (i.e. fill out the video/audio tag).

cmrd-senya:
It's an offtopic then. For this PR. This PR is about embedding files.

AugierLe42e:
@cmrd-senya: Where are you with this? I'd love to get this feature ;)

cmrd-senya:
I beleive we stuck with not having consensus on embeding syntax type and the way we should embed - in-place or at the end.

cmrd-senya:
I want to revive this PR. The feature is cool and useful, many people would like to have it and half of the work is done and it's a shame it was suspended with the voting mess (which was unfortunately organized by myself).

So here is some of my assumptions on how it should be done:
- Pick the image embedding syntax.

  [Here](https://www.loomio.org/d/LROboQNj/proposal/5XzI7RDk) was the voting for the "link syntax". It formally has passed, but it has two "blocks" and only two "yes" ahead "no+blocks". So I guess it's not really a nice solution. I could start the "image syntax" voting, but, really I didn't hear too much opposition to it. If you think the voting is required, please tell me. Otherwise, I would like to implement the "image syntax" and finally get the issue solved.
- Redcarpet code adapted to the feature. 

  In #6409 @jhass said that there are a couple of issues with the image syntax - is that tumblr (and generally, services) and the mobile version code must still produce proper results when we enable media embedding. Both tumblr and the mobile version use Diaspora::MessageRenderer in order to render html from markdown (which relies on Redcarpet), so the task comes to supporting media embed with it.
- Lazy initialization of the media

  I guess that using preload="none" with the audio/video tags is  enough. That what tumblr does and I guess that complicated substitutions of html with the javascript are unnecessary. At first we can use just a black stub for the video. Later, we could implement some cover image fetching in backgroung but that is out of scope of my intentions for now.

- There could still be issues with the media type determination.

  Currently in this PR it is being determined by the extension of the file in the URL. Everything that has non-audio and non-video URL is considered an image. Some future background fetch may help to determine the media type in ambiguous situations, but I guess the way it is now it's okay for a while.

If the team or the community disagree on some of my points, please let's discuss them. Otherwise, please give me the greenlight signal so I could start working on the feature.

Flaburgan:
>I could start the "image syntax" voting, but, really I didn't hear too much opposition to it. If you think the voting is required, please tell me. Otherwise, I would like to implement the "image syntax" and finally get the issue solved.

+1 on this. I follow the [discussion on common mark](https://talk.commonmark.org/t/embedded-audio-and-video/441/84). Honestly, it's not going anywhere. No vote, no decision. It's unfortunately useless to wait for them. So, let's use `` which should be seen as "embedding" syntax and go ahead with that.

So, greenlight from me.

svbergerem:
@cmrd-senya 

> Pick the image embedding syntax.
>
> Here was the voting for the "link syntax". It formally has passed, but it has two "blocks" and only two "yes" ahead "no+blocks". So I guess it's not really a nice solution. I could start the "image syntax" voting, but, really I didn't hear too much opposition to it. If you think the voting is required, please tell me. Otherwise, I would like to implement the "image syntax" and finally get the issue solved.

I'll just cite what [I wrote on loomio](https://www.loomio.org/d/LROboQNj/comment/803051):

> We could use the link syntax until there is an approved standard. Switching to the standard later would display the embedded audio/video files as links which should be fine.

This wouldn't work with the image syntax so I'm against the image syntax and in favor of the link syntax as long as there is no approved standard for CommonMark.

SuperTux88:
I agree to @svbergerem here: We should use something that doesn't break if it doesn't become the standard (If we didn't guess the future standard correctly ;) ).

denschub:
Yeah, sorry, total ACK to @svbergerem and @SuperTux88 here.

Flaburgan:
So waiting for the finale embedding syntax to be decided, is the following behavior fine for the link syntax?
Display the markdown link `[]()` to a video file inside a text like every other markdown links but also add the embedding of the video at the very end of the post.

denschub:
That's what Steffen wrote.

Flaburgan:
@cmrd-senya I don't know the current state of this PR but it looks like a consensus has been found and the behavior should be: Link syntax, with the link displayed inside the post, and the video embedded at the bottom of the post. Please tell us if that's what this PR does, rebase it and it should be OK to be reviewed.

cmrd-senya:
No, that's not what the PR does. It embeds media in-place, not at the bottom.

Also, the solution you offer doesn't answer what should we do to audio embeds. If we do the same thing we'll be limited to only one embed per post, which is more or less fine for video, but feels like a weird limitation for audio. For instance, in vk.com it is common to share music albums as single posts. SoundCloud embeds in diaspora also allows embedding of playlists. So I foresee that will be the first thing people would ask about if this feature is merged in the form you say.

Also, what is the reason to restrict people from placing embeds where they want and force them to be at the bottom?

Flaburgan:
I don't see why bottom would mean only one media. They could all be at the bottom imo.

>Also, what is the reason to restrict people from placing embeds where they want and force them to be at the bottom?

Because the link syntax still is a link. It is mainly use in middle of paragraph. I feel like we shouldn't use the link syntax to embed in place.

SuperTux88:
> If we do the same thing we'll be limited to only one embed per post

If we handle that as a special case of oembed (or similar to how oembed is done now), you can embed add one html snipped at the end of the post, but that can also contain multiple audio-files.

> Also, what is the reason to restrict people from placing embeds where they want and force them to be at the bottom?

That would only be a limitation until there is an official spec for embedding audio/video inplace (like images now). But until then handling it similar to oembed sounds like the best solution. Because currently you can't embed audio/video at all, so it would at least be an improvement until there is an official spec.

Also: Using the link syntax would be backward compatible, old pods simply show the link, while new pods embed the media-files at the bottom.

cmrd-senya:
Ok, that makes sense. I'll come back to it and finish it as soon as I have some time for that.

cmrd-senya:
I'm on it.

BTW should be moved from "Needs discussion" to "Waiting on contributor".

Flaburgan:
When this will be merged, the "How do I embed a video, audio or other multimedia content into a post?" question in the "Posts and posting" section of the `/help` page will have to be updated.

SuperTux88:
Thanks @Flaburgan. I added that to the PR description, because it should then be directly included in this PR.

cmrd-senya:
I'd like to discuss the approach here. I modified https://github.com/cmrd-senya/markdown-it-html5-embed (didn't push it yet) to support "non-inline mode". In this mode the media embed goes to the place where a `[[html5media]]` clause is placed. I also modified the diaspora text formatter so it adds this clause to the bottom of the post. Besides it is possible for a user to add this clause themselves to any place of a post, in this case it is not added to the bottom, of course (but I'm not sure we can advertise such feature, because it won't work with other federation-compatible networks; we could even filter it out on save to keep users from using it).

Styles still need to be fixed, but here are some screenshots:

cmrd-senya:
Okay, after discussion with @Flaburgan at the IRC channel I decided to do the same thing without actually using intermediate `[[html5media]]` syntax.

cmrd-senya:
I found out that I can't perform any actual tests for this feature, because PhantomJS doesn't support `<video>`/`<audio>` tags. Otherwise the PR is ready, but what should I do about the tests? :(

cmrd-senya:
Can be reviewed now.

I also updated the top PR description post to represent it's actual state.

Flaburgan:
How is camo handling these changes? If it does not proxy video or audio in that pr it is probably not a problem but is it working with camo enabled ?

cmrd-senya:
I didn't check, but since the links aren't images, I don't think there can be any issues with Camo. We posted video links in diaspora before this PR and everything was okay.

SuperTux88 (changes requested):
Sorry, I completely forgot about this PR.

Overall the current state of the PR looks good to me (how it works), and as far as I can judge the code with my basic javascript knowledge it looks OK too (maybe somebody with more javascript knowledge wants to have a look at it?).

I have two little things that should be changed:

First: Below the audio player is a gray border (`#dddddd`), that's especially visible in firefox with the dark theme enabled, but is also there in chrome (I didn't test other browsers).



I think that should be removed.

Second: The embedded elements only work with https when CSP is enabled, because we allow `media-src` only with `https:`. So we either need to allow `*` or only embed medias when the URL starts with `https://` (personally I would prefer to keep only allowing `https:`, because otherwise we would have mixed content).

cmrd-senya:
Hmm, I released a new version of https://github.com/cmrd-senya/markdown-it-html5-embed, but I wasn't able to use the new version in diaspora, because of a bug with rails-assets.org: https://github.com/tenex/rails-assets/issues/418

It is kinda a blocker for now. I'm unsure what to do. The issue is opened almost for two months and has no reaction at all. Maybe I should built a ruby gem for my package instead of relying on rails-assets.org?

SuperTux88:
Hmm, that's bad. It looks like rails-assets has currently a lot of problems (there is also tenex/rails-assets#415 which is open for almost 3 months and still not solved, and the reason why we now added `--full-index` to our upgrade guides for the upcoming release, because without it, it either fails completely or takes a lot longer). So maybe packaging a gem for rubygems.org is an alternative? There are also other frontend-gems we use from rubygems.org, so I think that's OK?

cmrd-senya:
Okay, so I updated the PR.

* I created a ruby gem code at https://github.com/cmrd-senya/markdown-it-html5-embed/tree/rubygem. It seem to work fine so I think we can go this way. Do you think I should build and publish a proper gem to rubygems.org now?
* I changed styles so there is no gray line below the player anymore
* In the new version of markdown-it-html5-embed `http://` addresses are not embedded by default anymore.

SuperTux88:
Looks good and works 👍 

About the gem: I would name it with `-` instead of `_`: `markdown-it-html5-embed`. I know the naming conventions for gems is with `_`, but it's only a "recommendation". And it looks weird to me with `_` and the only other `markdown-it`-gem I can find on rubygems.org is with `-` too, and it's with `-` on rails-assets too. So I think it's more consistent to use `-` in this case. (You would also need to rename `lib/markdown_it_html5_embed.rb` with `-`)

Also I wouldn't duplicate the `markdown-it-html5-embed.js` in git, I would create a rake task that copies the file to the correct place before build.

But since it's your gem, that's everything up to you :)

cmrd-senya:
> Also I wouldn't duplicate the markdown-it-html5-embed.js in git, I would create a rake task that copies the file to the correct place before build.

I wrote it [this way](https://github.com/cmrd-senya/markdown-it-html5-embed/blob/rubygem/Rakefile) but it doesn't seem to work when I install the gem with bundler from git, because the file is not copied.

I think it's still fine because I can install the gem manually for testing but it's a shame to lose a possibility to use the gem from git. Don't you know if it is possible to run a rake task when one requests a gem from git with bundler? So that I can run `FileUtils.cp("dist/markdown-it-html5-embed.js", "lib/assets/javascripts")` after the `git clone` or `git pull` or whatever it runs?

SuperTux88:
Oh, yeah, you're right, I didn't thought about that you needed that because you installed the gem via git for testing from a temporary branch, my fault. And no, I don't know if that is possible, I think not. But I think it's OK when it doesn't work via git for this gem, because the repo is still primarily an npm package and not a gem, the gem is only a side-product.

cmrd-senya:
Updated

SuperTux88 (approved):

SuperTux88:
Merged as 3e31b91b1123fcace22736a737e17457372e6092 and b32c84431469ca4258723fc7e8d1afbfdb192b53

Thank you 🍪

Flaburgan:
Yay! Today is a big day, videos are coming to diaspora*! Thank you Senya for this work and being so patient.

PR #6336, opened 2015-08-22
Cleaning publisher

ghost:
Cleaning publisher template a bit + some responsive styling

screen shot 2015-08-28 at 11 58 51
screen shot 2015-08-28 at 12 02 52
screen shot 2015-08-28 at 13 44 07

PR #6435, opened 2015-09-28
Hide publisher cancel button on xsmall screens

Flaburgan:
As said on https://github.com/diaspora/diaspora/pull/6336 I think the "Remove all text" feature of the publisher is not that useful compare to the risk of a misclick on that very big button.

### DIFF
--- a/app/assets/stylesheets/publisher.scss
+++ b/app/assets/stylesheets/publisher.scss
@@ -48,15 +48,6 @@
         }
       }
 
-      $sm-xs-average: ($screen-sm + $screen-xs) / 2;
-      @media(max-width: $sm-xs-average) {
-        #hide_publisher {
-          display: block;
-          width: 100%;
-          margin-bottom: 5px;
-        }
-      }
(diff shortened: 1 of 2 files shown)


AugierLe42e:
As said on #6336, it'd be better to reduce it as a cross on a corner than simply remove it :D

Flaburgan:
Yeah, but waiting for someone to do it, I propose to hide it at the moment ;)

AugierLe42e:
Meh :(

Flaburgan:
The way I think about it is: I want people to love diaspora*, so to love their experience there. How frustrated would you be if you loose a big message you spent time to enter with your mobile VS how frustrated would you be if you have to reload the page / manually clean the publisher? The first is the worst in my opinion.

AugierLe42e:
True that.

Flaburgan:
The travis failure looks unrelated.

svbergerem:
Thanks!

Issue #5064, opened 2014-07-08
Keyboard navigation: 'j' sometimes not working for the last post


svbergerem:
When the last post in your stream is very long the infinite scroll won't load additional posts and so 'j' is not working to jump to the next post. In that case one has to scroll down manually to load more posts.

PR #6451, opened 2015-10-02
Improve infinite scroll triggering

ghost:
Infinite scroll triggered relative to last element top rather than window bottom

### DIFF
--- a/app/assets/javascripts/app/views/infinite_stream_view.js
+++ b/app/assets/javascripts/app/views/infinite_stream_view.js
@@ -95,12 +95,11 @@ app.views.InfScroll = app.views.Base.extend({
   },
 
   infScroll : function() {
-    var $window = $(window)
-      , distFromTop = $window.height() + $window.scrollTop()
-      , distFromBottom = $(document).height() - distFromTop
-      , bufferPx = 500;
+    var $window = $(window),
+        distFromBottom = $(document).height() - $window.height() - $window.scrollTop(),
+        elementDistance = this.$el.children().last().offset().top - $window.scrollTop() - 500;
 
-    if(distFromBottom < bufferPx) {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/app/views/infinite_stream_view.js
```diff
@@ -95,12 +95,11 @@ app.views.InfScroll = app.views.Base.extend({
   },
 
   infScroll : function() {
-    var $window = $(window)
-      , distFromTop = $window.height() + $window.scrollTop()
-      , distFromBottom = $(document).height() - distFromTop
-      , bufferPx = 500;
+    var $window = $(window),
+        elementTop = this.$el.children().last().offset().top,
+        elementDistance = elementTop - $window.scrollTop() - 500;
```
svbergerem: Since `$window` is used only once I think you can remove the variable.

svbergerem:
This PR should fix #5064.

svbergerem:
Could there be cases where the difference between the top of the last post and the top of the window is more than 500px but you are not able to scroll down anymore? (big screen and small last post?)

Maybe we should keep the current code and just add your code as a second condition for loading more posts?!

AugierLe42e:
The code should execute as soon as the post appears at the bottom of the page. I've tested it on a short last post.

svbergerem:
> The code should execute as soon as the post appears at the bottom of the page.

If the height of your screen is about 500px...

AugierLe42e:
Hmm... I've never really understood how lengths work. >.<
My screen is 1920*1808px. Anyway, give me 5 minutes to change this.

svbergerem:
Merged as 22882aa24d334646d675d71593712fe732ffd380

Thanks!

PR #6457, opened 2015-10-04
Add permalink for stream elements

### LINKED ISSUE #4975: Make more clear that the time ago is a link to the SPV
Many user don't get that we can click on the date of a post to display the post in the SPV.

One way to improve that situation would be to add a icon like this one next to the date:

Do you have any other ideas?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2415865-make-more-clear-that-the-time-ago-is-a-link-to-the-spv?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #4975.

### DIFF
--- a/app/assets/stylesheets/colors.scss
+++ b/app/assets/stylesheets/colors.scss
@@ -1,30 +1,8 @@
-$background-white: #FFFFFF;
-$background-grey: #EEEEEE;
-$background-blue: #E7F2F7;
+.gray {
+  color: $text-grey;
 
-$grey: #2B2B2B;
-$light-grey: #DDDDDD;
-
-$border-grey: #DDDDDD;
-$border-dark-grey: #999999;
-
-$link-grey: #777777;
-$link-disabled-grey: #999999;
-
-$text-grey: #999999;
-$text-dark-grey: #666666;
-$text: #333333;
-
-$white: white;
-$black: black;
-$green: #8EDE3D;
-$light-green: lighten($green,20%);
-$red: #A80000;
-$blue: #3F8FBA;
-$dark-blue: darken(#0984C8,10%);
-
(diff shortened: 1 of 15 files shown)


app/assets/stylesheets/colors.scss
```diff
@@ -1,30 +1,8 @@
...
-$blue: #3F8FBA;
-$dark-blue: darken(#0984C8,10%);
-
-$sidebars-background: #f0f0f0;
-$sidebars-sub-background: darken($sidebars-background, 12%);
-$left-navbar-drawer-background: darken($sidebars-background, 6%);
-
-$card-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.16), 0 2px 10px 0 rgba(0, 0, 0, 0.12);
+.gray {
+  color: $text-grey;
```
denschub: Nit: Maybe we should stick to AE or BE, not both.

svbergerem: AFAIK we stick to AE in most cases. (Bootstrap variables, color vs. colour, ...) Some of our color variables still use BE but replacing those would introduce a lot of conflicts (160 occurrences of “grey”) so I think it should be part of another PR. I spoke with @jhass about a similar PR which would do some changes like replacing “stream_element” with “stream-element” and we agreed that this should happen after the 0.5.4.0 release to reduce the amount of conflicts for the backport merges.

jhass:
Thanks!

PR #6378, opened 2015-09-03
Report tab; Bootstrap3 compatible and reported-user-detail button

ghost:
untitled

b3 hasn't triggered that.. more the new button.

It will search the guid via (admin) user_search.. This can be useful if the podmin
wants to lock, unlock or delete the account responsible for the reported post.

PR #6464, opened 2015-10-07
Uncapitalize type if trying to find translation-string

ghost:
related to #6378

### DIFF
--- a/app/mailers/report_mailer.rb
+++ b/app/mailers/report_mailer.rb
@@ -8,7 +8,7 @@ class ReportMailer < ActionMailer::Base
   def new_report(type, id, role)
     resource = {
       url:  report_index_url,
-      type: I18n.t("notifier.report_email.type." + type),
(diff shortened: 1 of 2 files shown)


jhass:
Ah, good catch. Let's just downcase the whole thing though.

Zauberstuhl:
w00t not sure why I made that so complicated ... maybe it was too early

jhass:
Cool, squash?

app/mailers/report_mailer.rb
```diff
@@ -8,7 +8,7 @@ def self.new_report(type, id)
   def new_report(type, id, role)
     resource = {
       url:  report_index_url,
-      type: I18n.t("notifier.report_email.type." + type),
+      type: I18n.t("notifier.report_email.type." + type.downcase),
```
jhass: Wait, actually update this to `"notifier.report_email.type.#{type.downcase}"`, then squash.

Zauberstuhl:
done

PR #6470, opened 2015-10-08
Scroll to top when reloading the stream

svbergerem:


jhass:
Merged as da9cb5200706f07cd09972391fd0029cc9a49d83

Thanks

PR #6471, opened 2015-10-08
Add DatabaseSizeControl sidetiq job

### LINKED ISSUE #4920: Database size control
So I had a look how stuff is stored in the database, in my relatively small pod (18 months online, 148 total users, 2K total posts locally).

The amount of space taken by different tables is ... interesting. And I'm no longer wondering about several podmins complaining the "database blows up".

Could we come up with ideas on how to make the diaspora* database a little leaner, and track them here? It surely cannot be very optimal considering most of the space is used by participations and likes - not actual content.

A little discussion here: https://iliketoast.net/posts/227708#c7131fbf3144b42d

My table size dump:

    +---------------------------+------------+
    | Tables                    | Size in MB |
    +---------------------------+------------+
    | participations            |     898.08 |
    | likes                     |     680.67 |
    | comments                  |     324.92 |
    | posts                     |     130.84 |
    | taggings                  |      83.39 |
    | people                    |      15.73 |
    | aspect_visibilities       |      12.75 |
    | o_embed_caches            |      11.03 |
    | photos                    |       9.88 |
    | open_graph_caches         |       8.52 |
    | profiles                  |       6.72 |
    | tags                      |       6.03 |
    | share_visibilities        |       2.75 |
    | messages                  |       2.55 |
    | notifications             |       2.55 |
    | notification_actors       |       2.52 |
    | users                     |       1.59 |
    | mentions                  |       0.52 |
    | aspect_memberships        |       0.23 |
    | contacts                  |       0.19 |
    | tag_followings            |       0.11 |
    | aspects                   |       0.09 |
    | invitations               |       0.06 |
    | conversation_visibilities |       0.06 |
    | services                  |       0.05 |
    | locations                 |       0.05 |
    | conversations             |       0.03 |
    | rails_admin_histories     |       0.03 |
    | post_reports              |       0.03 |
    | simple_captcha_data       |       0.03 |
    | schema_migrations         |       0.02 |
    | invitation_codes          |       0.02 |
    | roles                     |       0.02 |
    | blocks                    |       0.02 |
    | pods                      |       0.02 |
    | user_preferences          |       0.02 |
    | account_deletions         |       0.02 |
    +---------------------------+------------+
    37 rows in set (7.57 sec)

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1578037-database-size-control?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jaywink:
Cleans initially author signatures from old Like and Participations tables. These signatures take a huge amount of space in the database and their only function is to allow retraction of likes and participations. Allow podmins to define a set of days that these are kept, by default the cleaning is off with a default of 365 days.

Closes #4920

TODO:
- [ ] Do default terms need a if to print out some text regarding cleaning of data if enabled?

### DIFF
--- a/app/workers/database_size_control.rb
+++ b/app/workers/database_size_control.rb
@@ -1,0 +1,41 @@
+
+module Workers
+  class DatabaseSizeControl < Base
+    include Sidetiq::Schedulable
+
+    sidekiq_options queue: :maintenance
+
+    recurrence { daily }
+
+    def perform
+      return unless AppConfig.settings.maintenance.database_size_control.enable?
+      remove_signatures
+    end
+
+    def remove_signatures
+      return unless AppConfig.settings.maintenance.database_size_control.remove_old_participations_after?
+      remove_before_date = Time.zone.today -
+        (AppConfig.settings.maintenance.database_size_control.remove_old_participations_after.to_i).days
+      remove_like_sigs(remove_before_date)
+      remove_participation_sigs(remove_before_date)
+    end
+
+    def remove_participation_sigs(remove_before_date)
+      removals = Participation.where(
+        "created_at < ? and (author_signature is not null or parent_author_signature is not null)",
+        remove_before_date).update_all(
+          author_signature: nil, parent_author_signature: nil
+        )
+      logger.info "DatabaseSizeControl: nullified #{removals} signatures from the participations table"
+    end
+
+    def remove_like_sigs(remove_before_date)
+      removals = Like.where(
+        "created_at < ? and (author_signature is not null or parent_author_signature is not null)",
+        remove_before_date).update_all(
+          author_signature: nil, parent_author_signature: nil
+        )
+      logger.info "DatabaseSizeControl: nullified #{removals} signatures from the likes table"
+    end
+  end
+end
--- a/lib/tasks/maintenance.rake
+++ b/lib/tasks/maintenance.rake
@@ -36,7 +36,7 @@ RUBY
       puts "Could not install logrotate configs. Perhaps you should try running this task as root and ensuring logrotate is installed:\n#{logrotate_conf}"
     end
   end
-  
+
   desc "Queue users for removal"
   task :queue_users_for_removal => :environment do
     # Queue users for removal due to inactivity
@@ -44,5 +44,10 @@ RUBY
     # must still be enabled, this only bypasses
     # scheduling to run the queuing immediately
     Workers::QueueUsersForRemoval.perform_async
+  end
+
+  desc "Database size control"
+  task :database_size_control => :environment do
+    Workers::DatabaseSizeControl.perform_async
   end
 end
--- a/spec/workers/database_size_control_spec.rb
+++ b/spec/workers/database_size_control_spec.rb
@@ -1,0 +1,45 @@
+require "spec_helper"
+
+describe Workers::DatabaseSizeControl do
+  describe "database size control is active" do
+    before do
+      AppConfig.settings.maintenance.database_size_control.enable = true
+      AppConfig.settings.maintenance.database_size_control.remove_old_participations_after = 1
+    end
+
+    it "#removes signatures from likes older than set time" do
+      like = FactoryGirl.create(
+        :like, created_at: 2.days.ago, author_signature: "foobar", parent_author_signature: "barfoo"
+      )
+      like2 = FactoryGirl.create(
+        :like, created_at: Time.zone.today, author_signature: "foobar", parent_author_signature: "barfoo"
+      )
+      Workers::DatabaseSizeControl.new.perform
+      like = Like.find(like.id)
+      like2 = Like.find(like2.id)
+      expect(like.author_signature).to be_nil
+      expect(like2.author_signature).to eq("foobar")
+      expect(like.parent_author_signature).to be_nil
+      expect(like2.parent_author_signature).to eq("barfoo")
+    end
+  end
+end
+
+describe Workers::DatabaseSizeControl do
+  describe "database size control is inactive" do
+    before do
+      AppConfig.settings.maintenance.database_size_control.enable = false
+      AppConfig.settings.maintenance.database_size_control.remove_old_participations_after = 1
+    end
+
+    it "#does not remove any signatures" do
+      like = FactoryGirl.create(
+        :like, created_at: 2.days.ago, author_signature: "foobar", parent_author_signature: "barfoo"
+      )
+      Workers::DatabaseSizeControl.new.perform
+      like = Like.find(like.id)
+      expect(like.author_signature).to eq("foobar")
+      expect(like.parent_author_signature).to eq("barfoo")
+    end
+  end
+end
--- a/config/defaults.yml
+++ b/config/defaults.yml
@@ -154,7 +154,10 @@ defaults:
         enable: false
         after_days: 730
         warn_days: 30
-        limit_removals_to_per_day: 100
+        limit_removals_to_per_day: 100  
+      database_size_control:
+        enable: false
+        remove_old_participations_after: 365
     source_url:
     default_color_theme: "original"
   services:
--- a/config/diaspora.yml.example
+++ b/config/diaspora.yml.example
@@ -587,6 +587,20 @@ configuration: ## Section
         ## Limit queuing for removal per day.
         #limit_removals_to_per_day: 100
 
+      ## Database size control allows purging selected old data from the database
+      ## automatically.
+      database_size_control:
+
+        # Enable first database size control
+        #enable: true
+
+        # Like and Participations contain signatures that allow the like or participation
+        # to be retracted later. Unfortunately, these signatures make up most of the data
+        # in a regular pod. Removing the signatures will not remove the likes or participations,
+        # it will just make it impossible for remote users to remove them. Local users will still
+        # be able to remove their likes even after the signatures are removed.
+        #remove_old_participations_after: 365
+
     ## Source code URL
     ## URL to the source code your pod is currently running.
     ## If not set your pod will provide a downloadable archive.
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -56,6 +56,8 @@ FactoryGirl.define do
   factory :like do
     association :author, :factory => :person
     association :target, :factory => :status_message
+    author_signature "author_signature"
+    parent_author_signature "parent_author_signature"
   end
 
   factory :user do


app/workers/database_size_control.rb
```diff
@@ -0,0 +1,27 @@
+
+module Workers
+  class DatabaseSizeControl < Base
+    include Sidetiq::Schedulable
+
+    sidekiq_options queue: :maintenance
+
+    recurrence { daily }
+
+    def perform
```
diaspora-code-review: Assignment Branch Condition size for perform is too high. [24.27/20]

app/workers/database_size_control.rb
```diff
@@ -0,0 +1,27 @@
...
+      if AppConfig.settings.maintenance.database_size_control.enable?
```
diaspora-code-review: Use a guard clause instead of wrapping the code inside a conditional expression.

app/workers/database_size_control.rb
```diff
@@ -0,0 +1,27 @@
...
+        if AppConfig.settings.maintenance.database_size_control.remove_old_participations_after?
+          remove_before_date = Date.today-
```
diaspora-code-review: Do not use `Date.today` without zone. Use `Time.zone.today` instead.

diaspora-code-review: Surrounding space missing for operator `-`.

lib/tasks/maintenance.rake
```diff
@@ -45,4 +45,9 @@ RUBY
     # scheduling to run the queuing immediately
     Workers::QueueUsersForRemoval.perform_async
   end
+
+  desc "Database size control"
+  task :database_size_control => :environment do
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

spec/workers/database_size_control_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'spec_helper'
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/workers/database_size_control_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+describe Workers::DatabaseSizeControl do
+  describe 'database size control is active' do
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/workers/database_size_control_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      AppConfig.settings.maintenance.database_size_control.enable = true
+      AppConfig.settings.maintenance.database_size_control.remove_old_participations_after = 1
+    end
+
+    it '#removes signatures from likes older than set time' do
+      like = FactoryGirl.create(
+        :like, created_at: 2.days.ago, author_signature: "foobar", parent_author_signature: "barfoo"
+      )
+      like2 = FactoryGirl.create(
+        :like, created_at: Date.today, author_signature: "foobar", parent_author_signature: "barfoo"
```
diaspora-code-review: Do not use `Date.today` without zone. Use `Time.zone.today` instead.

spec/workers/database_size_control_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      like2 = Like.find(like2.id)
+      expect(like.author_signature).to be_nil
+      expect(like2.author_signature).to eq("foobar")
+      expect(like.parent_author_signature).to be_nil
+      expect(like2.parent_author_signature).to eq("barfoo")
+    end
+  end
+end
+
+
```
diaspora-code-review: Extra blank line detected.

spec/workers/database_size_control_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+describe Workers::DatabaseSizeControl do
+  describe 'database size control is inactive' do
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/workers/database_size_control.rb
```diff
@@ -0,0 +1,29 @@
...
+    sidekiq_options queue: :maintenance
+
+    recurrence { daily }
+
+    def perform
+      return unless AppConfig.settings.maintenance.database_size_control.enable?
+      remove_signatures
+    end
+
+    def remove_signatures
```
diaspora-code-review: Assignment Branch Condition size for remove_signatures is too high. [21.24/20]

jhass:
How do we communicate to the user that they  can no longer delete their likes and stop receiving notifications for a post?

jaywink:
> How do we communicate to the user that they can no longer delete their likes and stop receiving notifications for a post?

Oh forgot to mention, local likes *can* be removed even after signatures are removed. So it is only likes federated to remote pods.

I don't believe it is possible for us to somehow notify from the remote pods that a like or participation could not be removed from that particular location.

* So, say, user A, comments on a post from user B
* After a year, the signature is removed
* After this time the user clicks unlike

What are the possibilities? If we make the remote pod pop up a notification (each one, the like could be on 100 servers) by sending back a failure, that would be highly confusing to the user.

I believe we should solve a huge problem for *most users* out there (which is podmins struggling with database size, which in turn affects how much performance they can continue to give) and accept that interactions after X amount of time might not be possible to retract.

jhass:
We already have reports of "can't delete X", "X didn't disappear from all pods" etc, the current mechanism is not 100% reliable. How are we going to disambiguate those reports from observed behaviour of this "feature"? How are we going to explain users that they can unlike stuff only for a certain period of time, given that time is variable and might not apply at all if the podmin decides to keep the feature off?

I feel like this leaves many UX questions unanswered, but maybe I'm hypocritical so other opinions would be welcome, /cc @SuperTux88 @denschub @Raven24 @svbergerem. Also I wonder if this even affects database performance at all, I'd expect it only cuts down plain storage space.

jaywink:
> We already have reports of "can't delete X", "X didn't disappear from all pods" etc, the current mechanism is not 100% reliable. How are we going to disambiguate those reports from observed behaviour of this "feature"? How are we going to explain users that they can unlike stuff only for a certain period of time, given that time is variable and might not apply at all if the podmin decides to keep the feature off?

Exactly. I don't think it is possible for a pod where the user is on to reliably say if the like will be removed on *all* remote pods. Some pods already do this kind of cleanup anyway, by hand, I think we should just provide a reliable tool for it.

> I feel like this leaves many UX questions unanswered, but maybe I'm hypocritical so other opinions would be welcome, /cc @SuperTux88 @denschub @Raven24 @svbergerem. Also I wonder if this even affects database performance at all, I'd expect it only cuts down plain storage space.

Absolutely would be nice to get more opinions. There have been quite a few podmins requesting this though, like @diasporg and @koehn. I'll ask for more podmins to give their opinions.

jaywink:
Note also (after discussion with @Flaburgan ) that removing the signatures does not stop account removal to remove remote likes and participations. Likes are not currently removed AFAICT, but participations are. This is handled by the AccountDeleter message that is sent out to remote pods. To remove likes we just need to make sure the deleter does those too - there is no need to keep the signatures for that since they do have the author_id etc.

jhass:
Don't we need them to retract them from third party pods in any case?

* user@podA  likes post by user@podB
* podB federates like to recipients of post
* user@podA deletes account/retracts like
* podB receives account deletion/retraction
* podB relays/generates retraction for likes by user@podA to original recipients of the post
* podB deletes like locally

The question is whether any of those steps requires the data you want to drop. Did you verify that?

jaywink:
> The question is whether any of those steps requires the data you want to drop. Did you verify that?

Shouldn't we just relay the account deletion, not retract all the individual likes? Assuming you were talking about account deletion.

In the context of retracting a single like after signatures are removed, from third-party pods in the line - no I'm sure that wouldn't work, unless we change the retraction logic to just verify ownership by verifying the user only. But I'm assuming this is not how it is done at the moment.

koehn:
Does this actually reduce database size on Postgres and MySQL? Since we're purging the contents of a column I'm curious if those databases will reclaim disk storage. Obviously backups created with `*dump` will shrink.

Raven24:
It would be interesting to see what could be gained from this in a production-like environment.
Maybe one of the podmins for one of the larger pods could test this on a copied DB and report back the size improvements.
Also, [the MySQL docs](https://dev.mysql.com/doc/refman/5.7/en/optimize-table.html) talk about running `OPTIMIZE TABLE` after huge delete jobs ... maybe this could be applied here, too.

As for UX we should really make the user aware of what is essentially an intentional limitation in how the data is handled. For that it's necessary to know which interactions are affected in what way exactly, when the participation/like signatures are removed.

Also, IIRC the participations are more or less a cache of data that could be reconstructed with a few more costly DB queries - which is why it got introduced. So in theory this could be handled like a cache that might be re-calculated if needed? In that case it should be possible to drop the entire row instead of just the columns ... just some speculation here.

jhass:
For postgres you probably would need to run `VACUUM FULL ANALYZE`, which essentially repacks the data when encountering big holes.

jhass:
@Raven24 not really, it's used to determine whether to generate notifications or not I think, but possibly also to determine who to send comments etc to.

Flaburgan:
>Maybe one of the podmins for one of the larger pods could test this on a copied DB and report back the size improvements.

I'm downloading a dump of diaspora-fr.org database and I'll drop those columns to check the difference.

app/workers/database_size_control.rb
```diff
@@ -0,0 +1,41 @@
...
+        (AppConfig.settings.maintenance.database_size_control.remove_old_participations_after.to_i).days
+      remove_like_sigs(remove_before_date)
+      remove_participation_sigs(remove_before_date)
+    end
+
+    def remove_participation_sigs(remove_before_date)
+      removals = Participation.where(
+        "created_at < ? and (author_signature is not null or parent_author_signature is not null)",
+        remove_before_date).update_all(
+        author_signature: nil, parent_author_signature: nil
```
diaspora-code-review: Indent the first parameter one step more than the start of the previous line.

app/workers/database_size_control.rb
```diff
@@ -0,0 +1,41 @@
...
+      )
```
diaspora-code-review: Indent `)` the same as the start of the line where `(` is.

app/workers/database_size_control.rb
```diff
@@ -0,0 +1,41 @@
...
+      logger.info "DatabaseSizeControl: nullified #{removals} signatures from the participations table"
+    end
+
+    def remove_like_sigs(remove_before_date)
+      removals = Like.where(
+        "created_at < ? and (author_signature is not null or parent_author_signature is not null)",
+        remove_before_date).update_all(
+        author_signature: nil, parent_author_signature: nil
```
diaspora-code-review: Indent the first parameter one step more than the start of the previous line.

app/workers/database_size_control.rb
```diff
@@ -0,0 +1,41 @@
...
+      )
```
diaspora-code-review: Indent `)` the same as the start of the line where `(` is.

jhass:
btw you can run the code review locally with `bin/pronto -c develop`

jaywink:
Yes sorry I forgot that if I push changes code review attacks them immediately :) Used to requesting builds in our work repo...

jaywink:
Good points re table space. Yes, the space will not be freed from the operating system, but for example database dumps should be smaller. For iliketoast, I ran this and purged 600312 like signatures and 821242 participations signatures (over 365 days, default once enabled). In backup size, this changed my 3.4GB backup dump to 1.9GB, so a reduction of 44% in database size.

I tried `VACUUM ANALYZE` but as it should, it didn't reclaim space, only made it reusable. Still, that is an improvement since now afaict the next 1.4GB of signatures will not increase the table size?

`VACUUM ANALYZE FULL` should reclaim the space fully, will try that later after an announced service break and report how long it took to run - it locks the tables.

jaywink:
> VACUUM ANALYZE FULL should reclaim the space fully, will try that later after an announced service break and report how long it took to run - it locks the tables.

So, 16 days passed, finally got around to doing this. Top tables now look like this:

                          table_name                          | table_size | indexes_size | total_size 
--------------------------------------------------------------+------------+--------------+------------
 "public"."participations"                                    | 1243 MB    | 160 MB       | 1403 MB
 "public"."likes"                                             | 840 MB     | 161 MB       | 1001 MB
 "public"."comments"                                          | 795 MB     | 59 MB        | 854 MB
 "public"."posts"                                             | 292 MB     | 226 MB       | 518 MB
 "public"."taggings"                                          | 93 MB      | 201 MB       | 293 MB
 "public"."open_graph_caches"                                 | 56 MB      | 2808 kB      | 59 MB
 "public"."people"                                            | 33 MB      | 4776 kB      | 38 MB
 "public"."o_embed_caches"                                    | 31 MB      | 3848 kB      | 35 MB
 "public"."aspect_visibilities"                               | 13 MB      | 21 MB        | 34 MB
 "public"."photos"                                            | 23 MB      | 7472 kB      | 30 MB
 "public"."tags"                                              | 9528 kB    | 9864 kB      | 19 MB

Of course, I forgot to take the sizes before running full vacuum ;) But looking [at my DB 1,5y ago](https://github.com/diaspora/diaspora/issues/4920#issue-31261127), it's quite clear the proportion if data in participations and likes hasn't gone up by the same percentage due to cleaning the signatures. Note, I kept a years worth of the signatures. Keeping this feature on likely means my DB size will increase much slower now.

+---------------------------+------------+
| Tables                    | Size in MB |
+---------------------------+------------+
| participations            |     898.08 |
| likes                     |     680.67 |
| comments                  |     324.92 |
| posts                     |     130.84 |
| taggings                  |      83.39 |
| people                    |      15.73 |
| aspect_visibilities       |      12.75 |
| o_embed_caches            |      11.03 |
| photos                    |       9.88 |

Also, pod has been running and federating fine since these 16 days.

@Raven24,

> As for UX we should really make the user aware of what is essentially an intentional limitation in how the data is handled. For that it's necessary to know which interactions are affected in what way exactly, when the participation/like signatures are removed.

Interactions are not affected AFAICT, only retracting the old items from remote pods. The trouble with mentioning this in user documentation and especially the UI is that the limit is configurable per pod. It is just not possible to give any amount of time after which retractions are not possible. Due to the way federation is designed (contact based), I'm pretty sure interactions and retractions will fail and break more likely due to that than any housekeeping pods do **to serve current users better**.

I'd be willing to add to the mention on retracting remote data to the default terms, but I wouldn't add anything to the UX or help. Anything there would just confuse users as any guarantees cannot be kept even if this isn't implemented, currently.

So, to move this forward, 16 days I think is quite enough time to get more comments, what could I do? Is anyone in objection to merging this in and what is missing to make this happen?

@maxwell @rsofaer if you have a few minutes, would be nice to get your feedback too.

jaywink:
@cmrd-senya has a better PR for reducing DB size so closing this.

PR #6472, opened 2015-10-09
Remove old stream instance when reloading the stream

svbergerem:
Otherwise there would be multiple instances after some reloads which are all listening to scroll events and all instances except for the current one will try to load more posts every time which makes the page really slow and is bad for the UX.


Flaburgan:
Yeah I have users complaining about slow stream load since yesterday update. Thanks for that!

jhass:
Merged as dccad3f7e1a48eca46f81bb582e1cb20ac56f99b

Thanks

PR #6481, opened 2015-10-09
Improve federation regarding reshares and root post author

jaywink:
When author of the root post receives a reshare to it, no participation is added to the root author on the reshare. This causes any comments on the reshare on remote pods not to be sent to the author. Adding a participation should subscribe to the reshare and thus bring added comments back to the author.

Did some live testing with develop production pods to make sure of this case. Basically:
- User A posts a post on pod A
- User B, who follows user A, reshares it on pod B
- User C, who follows user B, comments on the reshare of B on pod C

Expected:
- Pod A should receive the reshare and the comment
- Pod B should receive the reshare and the comment
- Pod C should receive the reshare and the comment (has authored it)

What happens:
- Pod A receives only the reshare, not the comment
- Pod B receives the reshare and the comment
- Pod C receives the reshare and the comment (has authored it)

### DIFF
--- a/lib/federated/generator.rb
+++ b/lib/federated/generator.rb
@@ -12,9 +12,18 @@ module Federated
       relayable = build(options)
       if relayable.save!
         logger.info "user:#{@user.id} dispatching #{relayable.class}:#{relayable.guid}"
+        add_root_author(relayable)
         Postzord::Dispatcher.defer_build_and_post(@user, relayable, @dispatcher_opts)
         relayable
       end
+    end
+
+    def add_root_author(relayable)
+      return unless relayable.parent.respond_to?(:root) && relayable.parent.root
+      # Comment post is a reshare, include original author in subscribers
+      root_post = relayable.parent.root
+      @dispatcher_opts[:additional_subscribers] ||= []
+      @dispatcher_opts[:additional_subscribers] << root_post.author
     end
 
     def build(options={})
--- a/spec/lib/diaspora/federated/generator_spec.rb
+++ b/spec/lib/diaspora/federated/generator_spec.rb
@@ -1,0 +1,25 @@
+require "spec_helper"
+
+describe "adds root author on reshare" do
+  before do
+    @generator = Federated::Generator.new(double("user", id: 1), double)
+    @root_author = double("root_author")
+    root = double("root", author: @root_author)
+    parent = double("parent", root: root)
+    @relayable = double("relayable", parent: parent, class: "foo", guid: "123")
+  end
+
+  it "adds root to additional subscribers" do
+    @generator.add_root_author(@relayable)
+    additional_subscribers = @generator.instance_variable_get(:@dispatcher_opts)[:additional_subscribers]
+    expect(additional_subscribers).to include(@root_author)
+  end
+
+  it "calls add_root_author" do
+    allow(Postzord::Dispatcher).to receive(:defer_build_and_post).and_return(true)
+    allow(@generator).to receive(:build).and_return(@relayable)
+    allow(@relayable).to receive(:save!).and_return(true)
+    expect(@generator).to receive(:add_root_author)
+    @generator.create!
+  end
+end
--- a/app/models/reshare.rb
+++ b/app/models/reshare.rb
@@ -59,6 +59,7 @@ class Reshare < Post
   def receive(recipient, sender)
     local_reshare = Reshare.where(:guid => self.guid).first
     if local_reshare && local_reshare.root.author_id == recipient.person.id
+      recipient.participate! self
       return unless recipient.has_contact_for?(sender)
     end
     super(recipient, sender)
--- a/spec/models/reshare_spec.rb
+++ b/spec/models/reshare_spec.rb
@@ -51,6 +51,7 @@ describe Reshare, type: :model do
       receive_reshare
       expect(@root.resharers).to include(@reshare.author)
     end
+
     it "does not error if the root author has a contact for the resharer" do
       bob.share_with @reshare.author, bob.aspects.first
       expect {
@@ -58,6 +59,12 @@ describe Reshare, type: :model do
           receive_reshare # This doesn't ever terminate on my machine before it was fixed.
         end
       }.not_to raise_error
+    end
+
+    it "participates root author in the reshare" do
+      receive_reshare
+      participations = Participation.where(target_id: @reshare.id, author_id: @root.author_id)
+      expect(participations.count).to eq(1)
     end
   end
 


jaywink:
Also, the bug in action:

https://iliketoast.net/posts/775355 <-- my reshare of HQ post, has a comment from Sean
https://pod.diaspora.software/p/0746ec1050d10133bafa6c626dd55703 <-- my reshare on HQ, never got the comment

jaywink:
Hmm actually I need to continue testing this patch tomorrow...

jaywink:
Actually, this ended up not being so simple. I managed to (by moving the code to `lib.postzord.receiver.local_batch.receive!`) get the participations to be created, so for example a situation where:

* User A writes a post, doesn't follow anyone
* User B follows user A, reshares user A post
* User C follows user B, reshares user B reshare (ie user A post)

What happens now is that a participation is never created for user A on the reshared posts, which means that any replies will not arrive on user A pod. I managed to get that part working, ie participations are created to the root author always, independent of how far in the chain the resharing user is.

But, if user C comments on his/her reshare, the comment only comes to user B's pod, not user A's. I can't really figure out how to handle the situation, to make either user B pod relay the comment to user A or something else.

Anyway, I realized too that this changes quite a few things which maybe not everyone will like, though personally I would like. So, if this change was implemented, the **original root author would always be subscribed to comments and likes to reshares**. Of course, this is quite a big change in functionality, for me to the better..

What do people think - shall I continue working on fixing the relationships regarding participations in reshares?

If you are wondering what the problem is - it is basically this:

User A posts a post
User B reshares
User C reshares (from following user B)

--> all the **reshares** end up on user A pod, but **none** of the interactions to it. This is broken functionality where the post exists but not the participations. We should somehow fix it OR not deliver the reshare to the root author pod at all.

jaywink:
OK I've added functionality now to also push comments made to reshares to the original author. This, with the first comment, will cause also a notification to the reshare author, unless the "notifications for non-contacts" bug is hit.

I can't figure out how to write the test so that I can first build a comment, then except it in the dispatcher and then trigger everything so that I can check that `defer_build_and_dispatch` gets the correct `additional_subscribers`. Any pointers towards how to rewrite the test a bit for that?

jaywink:
For the changed functionality, would love some feedback whether it is acceptable. I guess it could be configurable ie "notify on all reshare activity" etc. The participation part could also be removed and only make comments federate better, I thought these are linked but not really in the end, so this PR kind of contains two changes. Can separate if wished.

Ping for opinions @Flaburgan @goobertron @svbergerem @jhass @denschub @AugierLe42e

AugierLe42e:
> OK I've added functionality now to also push comments made to reshares to the original author.

Do you think it is possible for the comments to be the same on all the reshares? Because doing this could cause the problem that otheer people answer on the original post to a comment that has been made on a reshare so the author of the comment does not know someone has answered him (because he is notified for the reshare and not the original post).

jaywink:
> Do you think it is possible for the comments to be the same on all the reshares?

Where the comments link hasn't been changed, the comment is just delivered to one extra place: the root authors pod.

jaywink:
I think @Flaburgan approved in principle in IRL discussion here at the hackathon, the adding participation between root author and reshare posts. Correct me if understood wrong :) I think it adds value to users to be notified when their posts are interacted with.

lib/federated/generator.rb
```diff
@@ -12,11 +12,21 @@ def create!(options={})
...
         logger.info "user:#{@user.id} dispatching #{relayable.class}:#{relayable.guid}"
+        add_root_author(relayable)
         Postzord::Dispatcher.defer_build_and_post(@user, relayable, @dispatcher_opts)
         relayable
       end
     end
 
+    def add_root_author(relayable)
+      # If comment post is a reshare, include original author in subscribers
+      if relayable.parent.respond_to?(:root) && relayable.parent.root
```
diaspora-code-review: Use a guard clause instead of wrapping the code inside a conditional expression.

jaywink:
Rebased, but still not really sure how to write the test. Any help appreciated.

Btw, it's awesome receiving notifications for other people commenting on my reshares. Would really love to get this in. Also the comment federation would be appreciated by many.

Anything I can do to move this?

Flaburgan:
Yep +1 from my side.

spec/lib/postzord/receiver/local_batch_spec.rb
```diff
@@ -90,4 +90,19 @@
...
   end
+
+  context "received reshare" do
+    before do
+      @reshare = FactoryGirl.create(:reshare, root:
+                                              FactoryGirl.build(:status_message, author: bob.person, public: true))
+      @root = @reshare.root
+    end
+
+    let(:receiver) { Postzord::Receiver::LocalBatch.new(@object, @ids) }
```
jhass: This line is not needed.

spec/lib/postzord/receiver/local_batch_spec.rb
```diff
@@ -90,4 +90,19 @@
...
+
+    it "creates a participation on the reshare for the root author" do
+      receiver
+      expect(@root.resharers.count).to eq(1)
```
jhass: This doesn't seem to even call your new code, so if the test passes, it doesn't assert anything about it.

spec/lib/postzord/receiver/local_batch_spec.rb
```diff
@@ -90,4 +90,19 @@
       receiver.perform!
     end
   end
+
+  context "received reshare" do
+    before do
+      @reshare = FactoryGirl.create(:reshare, root:
+                                              FactoryGirl.build(:status_message, author: bob.person, public: true))
```
jhass: The indentation is a bit weird here, just
```rb
@root = FactoryGirl...
@reshare = FactoryGirl.create(:reshare, root: @root)
```

lib/federated/generator.rb
```diff
@@ -12,11 +12,21 @@ def create!(options={})
...
+        root_post = relayable.parent.root
+        @dispatcher_opts[:additional_subscribers] ||= []
+        @dispatcher_opts[:additional_subscribers] << Person.find(root_post.author_id)
```
jhass: Just `root_post.author`

lib/postzord/receiver/local_batch.rb
```diff
@@ -25,6 +25,15 @@ def receive!
     # 09/27/11 this is slow
     notify_users
 
+    if @object.respond_to?(:type) && @object.type == "Reshare"
+      root_object = Post.find_by guid: @object.root_guid
```
jhass: Why refetch?

lib/postzord/receiver/local_batch.rb
```diff
@@ -25,6 +25,15 @@ def receive!
...
+      root_author = Person.find(root_object.author_id)
```
jhass: Ditto.

lib/postzord/receiver/local_batch.rb
```diff
@@ -25,6 +25,15 @@ def receive!
...
+      if root_author.local?
+        # create a participation to the user for this reshare, to federate comments to author
+        root_author.owner.participate! @object
+      end
```
jhass: Could use a modifer-if here, `@object.root.author.owner.participate! @object if @object.root.author.local?`

spec/lib/diaspora/federated/generator.rb
```diff
@@ -0,0 +1,16 @@
...
+
+describe "comment on reshare" do
+  before do
+    @status_message = FactoryGirl.build(:status_message, author: alice.person, public: true)
+    @reshare = FactoryGirl.build(:reshare, root: @status_message, author: bob.person)
+    @comment = FactoryGirl.build(:comment, post: @reshare, text: "reshare comment")
+  end
+
+  it "delivers comment also to root author" do
+    opts = {"additional_subscribers" => [alice]}
```
jhass: `alice` is a `User`, yet you add a `Person`, no idea how this ever passed, but it doesn't seem to test your addition.

lib/postzord/receiver/local_batch.rb
```diff
@@ -25,6 +25,15 @@ def receive!
...
+    end
+
```
jhass: The method is nicely cleaned up with higher level methods. Let's keep it that way and extract this to a nicely named method, instead of polluting it with lower level code. Keeping the same level of abstraction makes a method easier to follow.

lib/postzord/receiver/local_batch.rb
```diff
@@ -25,6 +25,15 @@ def receive!
     # 09/27/11 this is slow
     notify_users
 
+    if @object.respond_to?(:type) && @object.type == "Reshare"
```
jhass: We have the constructed object here, no? Why not duck type on `root` like below?

lib/postzord/receiver/local_batch.rb
jhass: On second though, how about moving the logic [here](https://github.com/diaspora/diaspora/blob/develop/app/models/reshare.rb#L59)?

lib/postzord/receiver/local_batch.rb
jaywink: Sorry I don't understand..

jhass: You use `@object.respond_to?(:root) && @object.root` elsewhere, why not here?

lib/postzord/receiver/local_batch.rb
jaywink: It would be more helpful to say what is the alternative instead of asking me the question of why. This way we only have an extra loop to go through.

lib/postzord/receiver/local_batch.rb
jaywink: So this is ok?

lib/postzord/receiver/local_batch.rb
jaywink: OK, thanks.

lib/postzord/receiver/local_batch.rb
jhass: I'd guess `@object.root` would work and probably be already loaded. But I didn't verify it and threw a debugger into that section just to check. Since you developed this code, you might have a good reason I'm overlooking, so I ask.

lib/postzord/receiver/local_batch.rb
jhass: I use ditto to say same as last comment. If something is ok I don't comment on it at all.

lib/postzord/receiver/local_batch.rb
jaywink: I'll have a look.

spec/lib/postzord/receiver/local_batch_spec.rb
jaywink: Sorry my bad, was waiting for input on how to write the other test, should not have committed this as a stub of copied from above.

lib/federated/generator.rb
jaywink: But it uses find by `id` here: https://github.com/jaywink/diaspora/blob/reshare-participate/app/workers/deferred_dispatch.rb#L28 ?

spec/lib/diaspora/federated/generator_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    @status_message = FactoryGirl.build(:status_message, author: alice.person, public: true)
+    @reshare = FactoryGirl.build(:reshare, root: @status_message, author: bob.person)
+    @comment = FactoryGirl.build(:comment, post: @reshare, text: "reshare comment")
+  end
+
+  it "delivers comment also to root author" do
+    # opts = {:additional_subscribers => [alice.person]}
+    allow_any_instance_of(Federated::Generator).to receive(:create!)
+    # expect(Postzord::Dispatcher).to receive(:defer_build_and_post).with(bob, @comment, opts)
+    expect(Postzord::Dispatcher).to receive(:defer_build_and_post)
```
jaywink: Thanks @jhass for the review. I still can't figure out this however, how to get `@comment.save!` (which I assume is the right way to go from here) to end up calling `defer_build_and_post`.

I assume I have to do some allow etc magic, but, as always, rspec got the better of me. I'll try reading up on it during the weekend, or if you have a quick hint, appreciated. Maybe the code doesn't work yet, but since the test doesn't even end up in that section (verified via debugger), difficult to say for sure.

lib/federated/generator.rb
```diff
@@ -12,11 +12,20 @@ def create!(options={})
...
         relayable
       end
     end
 
+    def add_root_author(relayable)
+      return unless relayable.parent.respond_to?(:root) && relayable.parent.root
+      # Comment post is a reshare, include original author in subscribers
+      root_post = relayable.parent.root
+      @dispatcher_opts[:additional_subscribers] ||= []
+      @dispatcher_opts[:additional_subscribers] << Person.find(root_post.author)
```
jhass: `author` should return a `Person` instance already, the `Person.find` is redundant.

spec/lib/diaspora/federated/generator_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+describe "comment on reshare" do
+  before do
+    @status_message = FactoryGirl.build(:status_message, author: alice.person, public: true)
+    @reshare = FactoryGirl.build(:reshare, root: @status_message, author: bob.person)
+    @comment = FactoryGirl.build(:comment, post: @reshare, text: "reshare comment")
+  end
+
+  it "delivers comment also to root author" do
+    # opts = {:additional_subscribers => [alice.person]}
+    allow_any_instance_of(Federated::Generator).to receive(:create!)
```
jhass: This stubs out any call to `create!`, the original code is no longer called after this and so isn't your new code then.

jaywink: OK, I misunderstood `allow` then. Anyway, nothing happens on `@comment.save!` :angry: I'm having difficulty following what should happen but apparently `save!` is not the way. I'll try to find another place to trigger this.

jaywink: OK this triggers the code, now just need to find a way to verify the result...

    bob.comment!(@reshare, "reshare comment", {})

spec/lib/diaspora/federated/generator_spec.rb
jhass: Let's trace the call chain. Creating a public comment via the UI:

* https://github.com/diaspora/diaspora/blob/develop/app/controllers/comments_controller.rb#L15
* https://github.com/diaspora/diaspora/blob/develop/app/services/comment_service.rb#L15
* https://github.com/diaspora/diaspora/blob/develop/app/models/user/social_actions.rb#L4
* https://github.com/diaspora/diaspora/blob/develop/app/models/comment.rb#L94
* https://github.com/diaspora/diaspora/blob/develop/lib/federated/generator.rb#L11
*  https://github.com/diaspora/diaspora/blob/develop/lib/postzord/dispatcher.rb#L50
* https://github.com/diaspora/diaspora/blob/develop/app/workers/deferred_dispatch.rb#L16

So no, `save!` doesn't trigger a callback that calls `create!`, rather it's called explicitly and is what calls `save!`. Let's pause here for a second and look at ...

Receiving a public comment via federation:

* https://github.com/diaspora/diaspora/blob/develop/app/controllers/publics_controller.rb#L24
* https://github.com/diaspora/diaspora/blob/develop/app/workers/receive_unencrypted_salmon.rb#L12
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver.rb#L14
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver/public.rb#L24
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver/public.rb#L54
* https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/parser.rb#L13
* Effectively `Comment.from_xml`, `from_xml` is ROXML magic (the `xml_attr` stuff in the models). No callbacks.
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver/public.rb#L59
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver/public.rb#L67 (We're persisted to the DB now)
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver/public.rb#L32-L33 (https://github.com/diaspora/diaspora/blob/develop/app/models/comment.rb#L10, https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/relayable.rb#L43)
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver/public.rb#L45
* https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/relayable.rb#L89

Now we're at `Postzord::Dispatcher.build(...).post` again, so the code paths converge here.

* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/dispatcher.rb#L34
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/dispatcher.rb#L66
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/dispatcher.rb#L85
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/dispatcher.rb#L126
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/dispatcher.rb#L139
* https://github.com/diaspora/diaspora/blob/develop/app/workers/receive_local_batch.rb#L11-L12
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver/local_batch.rb#L19
* https://github.com/diaspora/diaspora/blob/develop/lib/postzord/receiver/local_batch.rb#L36
* https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/relayable.rb#L69 (the other codepath now, where we own the parent).

That should show all relevant "callback"/"hook" situations, the last part rather being for completeness.

It's probably fair if you just the testcase for `Federated::Generator` directly, using a test subclass that returns stub (`double`) values. Or you argue that it's opaque behaviour of the `CommentsController` or `CommentService` even, though that's a bit on the far edge of unit testing.

spec/lib/diaspora/federated/generator_spec.rb
jhass: See above for why it does.

spec/lib/diaspora/federated/generator_spec.rb
jaywink: Thanks, now my head hurts ;) Will continue later with something..

jaywink:
Hmmm odd thing with the tests. One of the builds passes (mysql 2.2), but all three others fail?

jaywink:
Final stage:

* Send comment to reshare author when commenting on reshare. As posts are always delivered also to reshare root, comments should also be delivered to reshare root, for concistency.
* Add participation to root.author on receiving reshare. When author of the root post receives a reshare to it, no participation is added to the root author on the reshare. This causes any comments on the reshare on remote pods not to be sent to the author. Adding a participation should subscribe to the reshare and thus bring added comments back to the author.

I have nothing more to add to this so unless review gives anything and if tests pass, ready for mergy. @denschub - could this be included in tomorrows 0.5 freeze since it improves federation?

denschub:
No objections, merged as 9a35a0d8dd27f142095ff02fd698f83f9f16668f.

TheRadialActive:
Thanks @jaywink for your work. It really saves me a lot of time :)

Issue #6326, opened 2015-08-17
500 logging out


Flaburgan:
``
[2015-08-17T09:25:54] FATAL PID-27479 TID-53576292760 Rails: 
NoMethodError (undefined method `reset_authentication_token!' for nil:NilClass):
  app/controllers/sessions_controller.rb:10:in `reset_authentication_token'
  lib/rack/chrome_frame.rb:39:in `call'
  lib/unicorn_killer.rb:35:in `call'
``

Just encountered this error logging out of diaspora*. Last develop code. F5 (request resent), same error. Going back to /stream, the user session is still active. Clicking again on Sign out: logging out normally, back to the pod page. No idea what happened.

svbergerem:
I was able to reproduce the bug:

1. Log in, open the stream in 2 tabs
2. Log out in one tab and log in again with the same user
3. Use the second tab to log out

satishck1992:
I am thinking on working on this bug, how about a long running HTTP request which notifies when a user has logged out ?

jhass:
We should just not 500, I think then it'll solve itself.

satishck1992:
good idea :+1:

marienfressinaud:
No news from 25 Aug, I'm working on it.

Basically, I fixed that with:

```ruby
current_user.reset_authentication_token! unless current_user.nil?
```

But now when I log out in the second tab, I'm redirected to stream page and a notification tell me that I'm already logged in. Quite strange isn't it?

PR #6484, opened 2015-10-10
Test if user is nil in #reset_authentication_token

marienfressinaud:
Actually it redirects to stream page and says that user is already logged in. Is it a problem?

Fix https://github.com/diaspora/diaspora/issues/6326


jhass:
Mh it does? I thought it happens because we no longer have a session, why does it still display the stream then?

marienfressinaud:
The second step of @svbergerem is "Log out in one tab and log in again with the same user" so you are still connected. I guess second tab remembers stuff about first session and tries to logout with this session... but you know probably better Rails than me :)

jhass:
Right, I should've reviewed that again but misremembered it, sorry.

Issue #6391, opened 2015-09-07
Allow administrators to set user roles in the admin panel


Flaburgan:
As an administrator, I should be able to add / remove the moderator, administrator (?) or spotlight role to a user directly from the admin panel.
##

Flaburgan:
Looks like we forgot to close that issue ;)

PR #6486, opened 2015-10-10
[WIP] An admin can set user roles

marienfressinaud:
At the moment, it only shows role of users. I'm not sure of the way to support role changes: my first idea was to put role form inside a modal but it is not as simple as it seems to me first :).

See https://github.com/diaspora/diaspora/issues/6391

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -10,7 +10,7 @@ class Role < ActiveRecord::Base
   scope :admins, -> { where(name: "admin") }
   scope :moderators, -> { where(name: %w(moderator admin)) }
 
-  def self.is_admin?(person_id)
+  def self.admin?(person_id)
     exists?(person_id: person_id, name: "admin")
   end
 
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -26,6 +26,10 @@ class Role < ActiveRecord::Base
     find_or_create_by(person_id: person.id, name: "moderator")
   end
 
+  def self.spotlight?(person)
+    exists?(person_id: person.id, name: "spotlight")
+  end
+
   def self.add_spotlight(person)
     find_or_create_by(person_id: person.id, name: "spotlight")
   end
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -10,24 +10,24 @@ class Role < ActiveRecord::Base
   scope :admins, -> { where(name: "admin") }
   scope :moderators, -> { where(name: %w(moderator admin)) }
 
-  def self.is_admin?(person)
-    exists?(person_id: person.id, name: "admin")
+  def self.is_admin?(person_id)
+    exists?(person_id: person_id, name: "admin")
   end
 
   def self.add_admin(person)
     find_or_create_by(person_id: person.id, name: "admin")
   end
 
-  def self.moderator?(person)
-    moderators.exists?(person_id: person.id)
+  def self.moderator?(person_id)
+    moderators.exists?(person_id: person_id)
   end
 
   def self.add_moderator(person)
     find_or_create_by(person_id: person.id, name: "moderator")
   end
 
-  def self.spotlight?(person)
-    exists?(person_id: person.id, name: "spotlight")
+  def self.spotlight?(person_id)
+    exists?(person_id: person_id, name: "spotlight")
   end
 
   def self.add_spotlight(person)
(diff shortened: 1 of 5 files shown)


app/views/admins/_user_entry.haml
```diff
@@ -37,6 +37,11 @@
           %dd= user.username
           %dt= t('.email')
           %dd= user.email
+          %dt= t('.role.label')
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/views/admins/_user_entry.haml
```diff
@@ -37,6 +37,11 @@
...
+          %dd
+            = t('.role.spotlight') if Role.spotlight? user.person
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/views/admins/_user_entry.haml
```diff
@@ -37,6 +37,11 @@
...
+            = t('.role.moderator') if Role.moderator? user.person
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/views/admins/_user_entry.haml
```diff
@@ -37,6 +37,11 @@
...
+            = t('.role.admin') if Role.is_admin? user.person
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/role.rb
```diff
@@ -26,6 +26,10 @@ def self.add_moderator(person)
     find_or_create_by(person_id: person.id, name: "moderator")
   end
 
+  def self.spotlight?(person)
+    exists?(person_id: person.id, name: "spotlight")
```
jhass: No need to fetch the `person`, just use `person_id`..

app/views/admins/_user_entry.haml
```diff
@@ -37,6 +37,11 @@
           %dd= user.username
           %dt= t('.email')
           %dd= user.email
+          %dt= t(".role.label")
+          %dd
+            = t(".role.spotlight") if Role.spotlight? user.person
+            = t(".role.moderator") if Role.moderator? user.person
+            = t(".role.admin") if Role.is_admin? user.person
```
jhass: Join with `,` perhaps?

app/models/role.rb
marienfressinaud: Yes probably but should I do that in this PR or should be done in another one?

jhass: Okay enough drive by change IMO, you know "always leave the place cleaner  than you found it" ;)

marienfressinaud: Fine for me so

marienfressinaud:
Do you have any idea on how I should render "change role" form? Modal seems to be the cleaner solution but I don't know how to manage with BackboneJS

app/models/role.rb
```diff
@@ -10,22 +10,26 @@ class Role < ActiveRecord::Base
   scope :admins, -> { where(name: "admin") }
   scope :moderators, -> { where(name: %w(moderator admin)) }
 
-  def self.is_admin?(person)
-    exists?(person_id: person.id, name: "admin")
+  def self.is_admin?(person_id)
```
diaspora-code-review: Rename `is_admin?` to `admin?`.

app/models/role.rb
```diff
@@ -10,22 +10,26 @@ class Role < ActiveRecord::Base
   scope :admins, -> { where(name: "admin") }
   scope :moderators, -> { where(name: %w(moderator admin)) }
 
-  def self.is_admin?(person)
-    exists?(person_id: person.id, name: "admin")
+  def self.admin?(person_id)
+    exists?(person_id: person_id, name: "admin")
```
jhass: On second thought, maybe they should handle both for now, `id = person.is_a?(Person) ? person.id : person`

marienfressinaud: I realised if you pass a person instead of an id it works too (at least, rspec tests passes with both `person` and `person.id`). And the Rails magic happened!

jhass: Mmh, that's interesting.

Flaburgan:
Closing in favor of https://github.com/diaspora/diaspora/pull/7868

Issue #6062, opened 2015-06-04
Add links to the empty stream message


Flaburgan:
When the stream is empty, it displays "you could search or invite contacts".
"search" should link to the search page and "invite contacts" to the invite modal (the same than in the left side bar of the /contacts page)

giniouxe:
Hi. I'd like to work on this one if it's OK.

jhass:
@giniouxe of course it's ok ;)

Flaburgan:
@giniouxe I started to work on this but will not have time to finish it soon, feel free to base your work on this branch: https://github.com/Flaburgan/diaspora/commit/e72399128a2a0f76e172995faf13e21308203aea

giniouxe:
ok :)

A little question though : I can't run a `bundle install` since I updated the develop branch : I got an `There is an error in your Gemfile` message. Any idea of a recent change that may cause this problem?

jhass:
Make sure your Bundler version is 1.10.0 or later.

giniouxe:
@jhass That was the problem, thanks.

giniouxe:
https://github.com/diaspora/diaspora/pull/6189

One question though : is it user friendly to link to a search page where the is no actual search field, and, since you did'nt actually search anything by clicking, you're welcomed by a message saying you should actually search for something to get some results ?

jhass:
Yeah, maybe don't link the search page, it should be obvious enough that you can search in the header.

giniouxe:
OK

marienfressinaud:
I started to continue on this in another branch https://github.com/marienfressinaud/diaspora/tree/6062-add-search-and-invte-links-when-no-contacts

Feel free to use my work if you want to continue yourself :)

PR #6189, opened 2015-07-12
6062 add search and invite links when no contacts

giniouxe:

PR #6492, opened 2015-10-11
Add invite links when no contacts

marienfressinaud:
Fix https://github.com/diaspora/diaspora/issues/6062
Continue work from https://github.com/diaspora/diaspora/pull/6189

### DIFF
--- a/app/views/shared/_invitations.haml
+++ b/app/views/shared/_invitations.haml
@@ -1,15 +1,9 @@
-= t('.share_this')
+= t(".share_this")
 = invite_link(current_user.invitation_code)
-#invitations-button.btn.btn-link{ "data-toggle" => "modal"}
+.invitations-link.btn.btn-link#invitations-button{"data-toggle" => "modal"}
   = t(".by_email")
 
-= render 'shared/modal',
+= render "shared/modal",
     path: new_user_invitation_path,
-    id: 'invitationsModal',
-    title: t('invitations.new.invite_someone_to_join')
-
--# This will have to be extracte in a JS view
-:javascript
-  $("#invitations-button").click(function(){
-    app.helpers.showModal("#invitationsModal");
-  });
+    id: "invitationsModal",
+    title: t("invitations.new.invite_someone_to_join")
--- a/app/assets/javascripts/app/views/stream_view.js
+++ b/app/assets/javascripts/app/views/stream_view.js
@@ -11,6 +11,7 @@ app.views.Stream = app.views.InfScroll.extend({
     this.setupNSFW();
     this.setupInfiniteScroll();
     this.markNavSelected();
+    this.initInvitationModal();
   },
 
   postClass : app.views.StreamPost,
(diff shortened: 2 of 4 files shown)


marienfressinaud:
Finally I removed link for searching since it was not relevant to be redirected to an empty search page.

jhass:
Could you squash your commits into one and reword Fla's?

```sh
git remote add upstream https://github.com/diaspora/diaspora.git
git fetch upstream
git checkout 6062-add-search-and-invte-links-when-no-contacts
# Leave the first commit at pick, choose squash or fix for the others.
# If there are any conflicts git will inform you about that, resolve them for
# example with git mergetool or manually and with git add, then run
# git rebase --continue, again  watch the output since that might happen for
# each commit.
git log # Look at the updated commit history
git status # should report you're back on your branch
# We need to force push since a rebase edits the commits, effectively
# creating new ones. This is okay to do in a feature branch, the pull
# request will update.
git push -f origin 6062-add-search-and-invte-links-when-no-contacts # Assuming you cloned from your fork
```

Hint: You can also edit your last commit with `git commit --amend`, since that also effectively creates a new commit, you to need to force push afterwards and again you should only do it in feature branches.

marienfressinaud:
Done

jhass:
https://travis-ci.org/diaspora/diaspora/jobs/84776533#L444

marienfressinaud:
I plan to change test selectors by `.btn.invitations-link` which is working but I'm not sure it's good practice: would you prefer I add two different `id` (one on the stream link, one on the sidebar link) and create cucumber tests for both of them?

app/views/shared/_invitations.haml
```diff
@@ -1,15 +1,9 @@
-= t('.share_this')
+= t(".share_this")
 = invite_link(current_user.invitation_code)
-#invitations-button.btn.btn-link{ "data-toggle" => "modal"}
+.invitations-link.btn.btn-link{"data-toggle" => "modal"}
```
svbergerem: You could just add the id `invitations-button` here while keeping the `.invitations-link` class.

jhass:
Updating the selectors for the existing test should be fine I think.

app/views/shared/_invitations.haml
```diff
@@ -1,15 +1,9 @@
-= t('.share_this')
+= t(".share_this")
 = invite_link(current_user.invitation_code)
-#invitations-button.btn.btn-link{ "data-toggle" => "modal"}
+#invitations-button.invitations-link.btn.btn-link{"data-toggle" => "modal"}
```
diaspora-code-review: Classes should be listed before IDs (.invitations-link should precede #invitations-button)

marienfressinaud: Rhaaa, sorry ><

jhass: You can check those locally with `bin/pronto -c develop` btw ;)

jhass:
Mmh, I'd still kept that in the same commit :)

app/views/shared/_invitations.haml
marienfressinaud: I know but I don't think to it each time. `git commit ; git push` is quicker :)

denschub:
Merged into `develop`. Merging into stable would, unfortunately, be a bit too much effort and it looks like it may break things.

PR #6493, opened 2015-10-11
Fetch notification counts

### LINKED ISSUE #6473: Notifications are not updated when stream is reloaded
With new stream loading, notifications are not shown when I reload.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27285613-notifications-are-not-updated-when-stream-is-reloaded?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jaywink:
Add route for polling unread notification counts. Add polling to JS every 30 seconds to refresh the count.

Fixes #6473 and #5780 
Also partly addresses #3890

TODO's
- [ ] Jasmine tests
- [ ] Waiting for additional input from @svbergerem re code placement
- [ ] Also support conversation counts
- [ ] Make polling inactive when tab is not visible

---

This change is [Reviewable](https://reviewable.io/reviews/diaspora/diaspora/6493)

### DIFF
--- a/app/assets/javascripts/app/router.js
+++ b/app/assets/javascripts/app/router.js
@@ -188,6 +188,14 @@ app.Router = Backbone.Router.extend({
     $("#main_stream").html(app.page.render().el);
     $("#selected_aspect_contacts .content").html(streamFacesView.render().el);
     this._hideInactiveStreamLists();
+    this._ensureUpdateHeaderCounts();
+  },
+
+  _ensureUpdateHeaderCounts: function() {
+    // Update header counts if they are not polled automatically
+    if (gon.appConfig.settings.notifications.polling_interval === 0) {
+      app.header.notificationDropdown.updateHeaderCounts();
+    }
   }
 });
 // @license-end
--- a/app/assets/javascripts/app/views/notification_dropdown_view.js
+++ b/app/assets/javascripts/app/views/notification_dropdown_view.js
@@ -17,7 +17,9 @@ app.views.NotificationDropdown = app.views.Base.extend({
     this.ajaxLoader = this.dropdown.find(".ajax-loader");
     this.perfectScrollbarInitialized = false;
     this.updateHeaderCounts();
-    setInterval(this.updateHeaderCounts, 30000);
+    if (gon.appConfig.settings.notifications.polling_interval > 0) {
+      setInterval(this.updateHeaderCounts, gon.appConfig.settings.notifications.polling_interval*1000);
+    }
   },
 
   toggleDropdown: function(evt){
@@ -35,6 +37,7 @@ app.views.NotificationDropdown = app.views.Base.extend({
   showDropdown: function(){
     this.resetParams();
     this.ajaxLoader.show();
+    this.updateHeaderCounts();
     this.dropdown.addClass("dropdown-open");
     this.updateScrollbar();
     this.dropdownNotifications.addClass("loading");
@@ -107,7 +110,6 @@ app.views.NotificationDropdown = app.views.Base.extend({
         }
       });
     });
-    this.updateHeaderCounts();
 
     this.hideAjaxLoader();
 
--- a/app/assets/javascripts/app/views/notification_dropdown_view.js
+++ b/app/assets/javascripts/app/views/notification_dropdown_view.js
@@ -16,6 +16,8 @@ app.views.NotificationDropdown = app.views.Base.extend({
     this.dropdownNotifications = this.dropdown.find(".notifications");
     this.ajaxLoader = this.dropdown.find(".ajax-loader");
     this.perfectScrollbarInitialized = false;
+    this.updateHeaderCounts();
+    setInterval(this.updateHeaderCounts, 30000);
   },
 
   toggleDropdown: function(evt){
@@ -105,6 +107,7 @@ app.views.NotificationDropdown = app.views.Base.extend({
         }
       });
     });
+    this.updateHeaderCounts();
 
     this.hideAjaxLoader();
 
@@ -131,6 +134,22 @@ app.views.NotificationDropdown = app.views.Base.extend({
       this.dropdownNotifications.perfectScrollbar("destroy");
       this.perfectScrollbarInitialized = false;
     }
+  },
+
+  updateHeaderCounts: function() {
+    $.getJSON(Routes.notificationsCounts(), function(data) {
+      var headerBadge = $(".notifications-link .badge"),
+        markAllReadLink = $("a#mark_all_read_link");
+      if (data.notifications > 0) {
+        headerBadge.html(parseInt(data.notifications));
+        headerBadge.removeClass("hidden");
+        markAllReadLink.removeClass("enabled");
+      } else {
+        headerBadge.addClass("hidden");
+        headerBadge.html(0);
+        markAllReadLink.removeClass("disabled");
+      }
+    });
   }
 });
 // @license-end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -150,7 +150,10 @@ class ApplicationController < ActionController::Base
   def gon_set_appconfig
     gon.push(appConfig: {
                chat:     {enabled: AppConfig.chat.enabled?},
-               settings: {podname: AppConfig.settings.pod_name},
+               settings: {
+                 podname: AppConfig.settings.pod_name,
+                 notifications: AppConfig.settings.notifications
+               },
                map:      {mapbox: AppConfig.map.mapbox}
              })
   end
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -5,6 +5,8 @@
 class NotificationsController < ApplicationController
   before_action :authenticate_user!
 
+  respond_to :json, only: :counts
+
   def update
     note = Notification.where(:recipient_id => current_user.id, :id => params[:id]).first
     if note
@@ -78,4 +80,9 @@ class NotificationsController < ApplicationController
 
   end
 
+  def counts
+    render json: {
+      notifications: current_user.unread_notifications.count
+    }
+  end
 end
--- a/spec/controllers/notifications_controller_spec.rb
+++ b/spec/controllers/notifications_controller_spec.rb
@@ -181,4 +181,19 @@ describe NotificationsController, :type => :controller do
       expect(response).not_to be_redirect
     end
   end
+
+  describe "counts" do
+    it "succeeds" do
+      get :counts
+      expect(response).to be_success
+    end
+
+    it "returns unread notifications count" do
+      post = FactoryGirl.create(:status_message)
+      FactoryGirl.create(:notification, recipient: alice, target: post)
+      FactoryGirl.create(:notification, recipient: alice, target: post, unread: false)
+      get :counts
+      expect(response.body).to eq('{"notifications":1}')
+    end
+  end
 end
--- a/app/assets/templates/header_tpl.jst.hbs
+++ b/app/assets/templates/header_tpl.jst.hbs
@@ -9,7 +9,7 @@
             <span class="icon-bar"></span>
             <span class="icon-bar"></span>
           </button>
-          <a href="/stream" class="navbar-brand" data-stream-title="{{t "my_stream"}}">
+          <a href="/stream" rel="backbone" class="navbar-brand" data-stream-title="{{t "my_stream"}}">
             {{ podname }}
           </a>
           <ul class="nav nav-badges visible-sm visible-xs">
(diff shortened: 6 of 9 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -78,4 +80,9 @@ def read_all
 
   end
 
+  def counts
+    render :json => {
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

spec/controllers/notifications_controller_spec.rb
```diff
@@ -181,4 +181,19 @@
...
+
+  describe "counts" do
+    it "succeeds" do
+      get :counts
+      expect(response).to be_success
+    end
+
+    it "returns unread notifications count" do
+      post = FactoryGirl.create(:status_message)
+      FactoryGirl.create(:notification, :recipient => alice, :target => post)
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

spec/controllers/notifications_controller_spec.rb
```diff
@@ -181,4 +181,19 @@
...
+      FactoryGirl.create(:notification, :recipient => alice, :target => post, :unread => false)
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

jhass:
30 seconds seems a bit much, maybe make configurable? Could also be interesting to detect user activity (tab in background or not) and adjust the interval.

I don't agree this fixes #6473, it should still do an explicit fetch on stream reload too and I think #5780 was meant to update the count immediately.

app/assets/javascripts/app/views/notification_dropdown_view.js
```diff
@@ -131,6 +134,22 @@ app.views.NotificationDropdown = app.views.Base.extend({
       this.dropdownNotifications.perfectScrollbar("destroy");
       this.perfectScrollbarInitialized = false;
     }
+  },
+
+  updateHeaderCounts: function() {
+    $.getJSON("/notifications/counts", function(data) {
```
jhass: `Routes.notifications_count_path()` or something.

jaywink: Thanks! Was looking for something like that but failed to find it somehow :(

app/assets/javascripts/app/views/notification_dropdown_view.js
```diff
@@ -16,6 +16,8 @@ app.views.NotificationDropdown = app.views.Base.extend({
     this.dropdownNotifications = this.dropdown.find(".notifications");
     this.ajaxLoader = this.dropdown.find(".ajax-loader");
     this.perfectScrollbarInitialized = false;
+    this.updateHeaderCounts();
```
svbergerem: This will be initialized on pageload so I think we can remove this call here.

jaywink:
> 30 seconds seems a bit much, maybe make configurable? Could also be interesting to detect user activity (tab in background or not) and adjust the interval.

Seems a bit heavy to put in settings but can do, maybe allow disabling polling totally too by setting the value to 0.

> I don't agree this fixes #6473, it should still do an explicit fetch on stream reload too

I'll move the logic to some kind of helper and call it in more places. 

> and I think #5780 was meant to update the count immediately.

What do you mean immediately? The route is so light weight it is pretty much instant as it can be :) I don't think we should duplicate logic to many places just to save a few ms.

app/assets/javascripts/app/views/notification_dropdown_view.js
```diff
@@ -105,6 +107,7 @@ app.views.NotificationDropdown = app.views.Base.extend({
         }
       });
     });
+    this.updateHeaderCounts();
```
svbergerem: I'd like to see this in `showDropdown` instead.

jhass:
Well, currently it doesn't update if you mark notifications as read and open the dropdown. I didn't check carefully if this fixes the latter tbh but I didn't see anything for the former (which could be done without server interaction).

jaywink:
(btw, as talked with Fla IRL, the *real* solution would be to use pushes via websockets or something, but that is a whole another discussion and issue to deal with Just in case someone asks "why not websockets". Ajax polling is really the best for now)

app/assets/javascripts/app/views/notification_dropdown_view.js
```diff
@@ -131,6 +134,22 @@ app.views.NotificationDropdown = app.views.Base.extend({
...
+      if (data.notifications > 0) {
+        headerBadge.html(parseInt(data.notifications));
+        headerBadge.removeClass("hidden");
+        markAllReadLink.removeClass("enabled");
+      } else {
+        headerBadge.addClass("hidden");
+        headerBadge.html(0);
+        markAllReadLink.removeClass("disabled");
+      }
+    });
```
svbergerem: We currently handle these kind of changes in the notification view. I think it is a good idea to have them in the notification dropdown view. Maybe write a helper to set the header count which you can call with the number of unread notifications. Then you can use this function from the notifications view. The notification dropdown view should be always accessible under `app.header.notificationDropdown`.

jaywink:
> Well, currently it doesn't update if you mark notifications as read and open the dropdown.

Hmm for me at leas that happens, if you mean the `Mark all as read`. It does it per notification here https://github.com/jaywink/diaspora/blob/4364a80330f82998e2df053e3d30f7789d55633f/app/assets/javascripts/app/views/notifications_view.js#L71-71

app/assets/javascripts/app/views/notification_dropdown_view.js
svbergerem: I'd also like to see some jasmine tests.

app/assets/javascripts/app/views/notification_dropdown_view.js
```diff
@@ -16,6 +16,8 @@ app.views.NotificationDropdown = app.views.Base.extend({
     this.dropdownNotifications = this.dropdown.find(".notifications");
     this.ajaxLoader = this.dropdown.find(".ajax-loader");
     this.perfectScrollbarInitialized = false;
+    this.updateHeaderCounts();
+    setInterval(this.updateHeaderCounts, 30000);
```
AugierLe42e: 30 seconds seems a bit long. I'd go for 10.

denschub:
Please revert aabe6951e5f4626b1a518fb096ced2a9a1e91c47 in this pull request when you are done.

app/assets/javascripts/app/views/notification_dropdown_view.js
jaywink: I started working on this PR again, but now reading your comments above I'm kinda confused on what to do. The change is already in the notification dropdown view, or did you mean notification view?

jaywink:
I don't actually see why the notifications updating needs to be fetched when clicking different streams and stuff. Yes it is a changed, but this is also changing in the fact that notification counts are fetched even if the user doesn't click things. Why fetch separately when clicking stream links when there is a constant fetch ongoing, unless maybe the setting is off?

app/assets/javascripts/app/views/notification_dropdown_view.js
```diff
@@ -16,6 +16,10 @@ app.views.NotificationDropdown = app.views.Base.extend({
     this.dropdownNotifications = this.dropdown.find(".notifications");
     this.ajaxLoader = this.dropdown.find(".ajax-loader");
     this.perfectScrollbarInitialized = false;
+    this.updateHeaderCounts();
+    if (gon.appConfig.settings.notifications.polling_interval > 0) {
```
diaspora-code-review: Identifier 'polling_interval' is not in camel case.

app/assets/javascripts/app/views/notification_dropdown_view.js
```diff
@@ -16,6 +16,10 @@ app.views.NotificationDropdown = app.views.Base.extend({
...
+      setInterval(this.updateHeaderCounts, gon.appConfig.settings.notifications.polling_interval*1000);
```
diaspora-code-review: Identifier 'polling_interval' is not in camel case.

app/controllers/application_controller.rb
```diff
@@ -150,7 +150,10 @@ def current_user_redirect_path
   def gon_set_appconfig
     gon.push(appConfig: {
                chat:     {enabled: AppConfig.chat.enabled?},
-               settings: {podname: AppConfig.settings.pod_name},
+               settings: {
+                 podname: AppConfig.settings.pod_name,
```
diaspora-code-review: Align the elements of a hash literal if they span more than one line.

app/assets/javascripts/app/router.js
```diff
@@ -188,6 +188,14 @@ app.Router = Backbone.Router.extend({
     $("#main_stream").html(app.page.render().el);
     $("#selected_aspect_contacts .content").html(streamFacesView.render().el);
     this._hideInactiveStreamLists();
+    this._ensureUpdateHeaderCounts();
+  },
+
+  _ensureUpdateHeaderCounts: function() {
+    // Update header counts if they are not polled automatically
+    if (gon.appConfig.settings.notifications.polling_interval == 0) {
```
diaspora-code-review: Identifier 'polling_interval' is not in camel case.

diaspora-code-review: Expected '===' and instead saw '=='.

app/assets/javascripts/app/views/notification_dropdown_view.js
jaywink: It's now configurable.

jaywink:
Currently afaict todo:
* [ ] Jasmine tests
* [ ] Waiting for additional input from @svbergerem re code placement

So currently you can configure the interval, in seconds. If it is 0, then any page or stream reload will trigger the fetch. There really shouldn't be need to trigger it on every page or stream reload if it is being done in the background anyway.

jhass:
I've been thinking, can we detect whether the page is in view or in background and adjust the poll interval accordingly? If it's in background it would be fine to poll once a minute or maybe even less, when it's focused 10-15 seconds might be more preferred.

jaywink:
> I've been thinking, can we detect whether the page is in view or in background and adjust the poll interval accordingly?

Probably, though I'm guessing different browsers might have different support and might require even hacks. I remember doing something like this in one project at work, detecting when the user is actually doing something.

The question is why? This is minor optimization and the real solution anyway is to not use ajax polling, but at the moment it works and is easy to throw in.

I'd rather get this in and optimize later if needed, but surely the 30s miniature size pings are not going to be much weight on top of what the app already has to do? So I'm not prepared to spend a lot of time hacking on a solution that might not even work on all browsers we aim to support.

denschub:
> I don't actually see why the notifications updating needs to be fetched when clicking different streams and stuff.

The mentioned commit reloads the stream asynchronously, thus not triggering a header reload, thus not refreshing the counter. As a result, the user will never see the header update unless he does a refresh. Since this PR will implement a background loading, we can load our streams asynchronously again.

> Probably, though I'm guessing different browsers might have different support and might require even hacks. I remember doing something like this in one project at work, detecting when the user is actually doing something

No need to detect user activity, there is `document.hidden` and a `visibilitychange` event with "good enough" browser support (e.g. just ignore and assume the site is visible when unsupported), [see](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API).

denschub:
Ping?

jaywink:
Sorry I haven't had time to come back to this yet. I've moved the todo's to the top and hopefully will be able to finish the items in coming weeks. I've had this current branch on my pod and it works super well - but I agree to the changes that are needed (except @svbergerem needs to tell me what he means by his comment :)).

app/assets/javascripts/app/views/notification_dropdown_view.js
svbergerem: > We currently handle these kind of changes in the notification view.

https://github.com/diaspora/diaspora/blob/4364a80/app/assets/javascripts/app/views/notifications_view.js#L53

> I think it is a good idea to have them in the notification dropdown view.

Which is what you did here.

> Maybe write a helper to set the header count which you can call with the number of unread notifications.

So instead of fetching and changing the unread counts in the same function move the changes based on the unread count to a helper function which one can call with the number of unread notifications and it does all the UI changes. `updateHeaderCounts` would then just fetch the current counts and call the helper function with those.

> Then you can use this function from the notifications view.

Instead of having there some similar code you can just call your new helper function with the unread count.

jaywink:
It's not likely I'll be finishing this at the moment, feel free to orphan, continue it or close it.

denschub:
Erm, #6952.

PR #6499, opened 2015-10-13
Remove the lines from inlined_jobs.rb which never get called.

cmrd-senya:
Tests result doesn't change with removing of this lines.

With this lines if the test fails with NoMethodError calls given block for the second time which won't help and will confuse a tester.


jhass:
Merged as 70b5d863864a9357063b5da247b2a431db49b826

Issue #5947, opened 2015-05-11
Sidekiq fail on Heroku


roymckenzie:
Sorry to come here for this issue, but I was not receiving any support on IRC and this place seems to get more traffic and is less ephemeral.

I am running the latest master and things are running smoothly save for Sidekiq on Heroku. Every time I try to start the process or Heroku starts it through the Procfile I get this error:

```
No such file or directory - 
/app/vendor/ruby-2.0.0/lib/ruby/2.0.0/logger.rb:601:in `initialize'
/app/vendor/ruby-2.0.0/lib/ruby/2.0.0/logger.rb:601:in `open'
/app/vendor/ruby-2.0.0/lib/ruby/2.0.0/logger.rb:601:in `create_logfile'
/app/vendor/ruby-2.0.0/lib/ruby/2.0.0/logger.rb:596:in `open_logfile'
/app/vendor/ruby-2.0.0/lib/ruby/2.0.0/logger.rb:551:in `initialize'
/app/vendor/ruby-2.0.0/lib/ruby/2.0.0/logger.rb:318:in `new'
/app/vendor/ruby-2.0.0/lib/ruby/2.0.0/logger.rb:318:in `initialize'
/app/vendor/bundle/ruby/2.0.0/gems/sidekiq-3.3.3/lib/sidekiq/logging.rb:31:in `new'
/app/vendor/bundle/ruby/2.0.0/gems/sidekiq-3.3.3/lib/sidekiq/logging.rb:31:in `initialize_logger'
/app/vendor/bundle/ruby/2.0.0/gems/sidekiq-3.3.3/lib/sidekiq/cli.rb:344:in `initialize_logger'
/app/vendor/bundle/ruby/2.0.0/gems/sidekiq-3.3.3/lib/sidekiq/cli.rb:39:in `parse'
/app/vendor/bundle/ruby/2.0.0/gems/sidekiq-3.3.3/bin/sidekiq:7:in `<top (required)>'
/app/vendor/bundle/ruby/2.0.0/bin/sidekiq:23:in `load'
/app/vendor/bundle/ruby/2.0.0/bin/sidekiq:23:in `<main>'
```

My Procfile is untouched from the master branch Procfile.

Thanks for any advice you can offer.

roymckenzie:
OK so I found a "solution though it may be a hack. I had a guess that there was a problem creating the `sidekiq.log` file so I commented out line 4 in the `sidekiq.yml` file:

[...]

PR #6500, opened 2015-10-14
Don't set key for logfile on Heroku

lstoll:
This fixes the issue reported in #5947. Sidekiq will only log to stdout when this key evals to false, so this can't be forced with an empty string. This change eliminates the key altogether when running on Heroku.


jhass:
Merged as 6ac6dc1059758efe0ae07c9e68cff0d4f064206b

Thanks!

PR #6503, opened 2015-10-15
Display publisher on user profile path

### LINKED ISSUE #6501: Can't create a new post from my profile page depending of the link
We have two different link associated with our accounts: "https://pod.url/u/username" and "https://pod.url/people/4l0t0fnumb3rs4ndl3tt3rs".
If I try to create a new post from my "https://pod.url/u/username" profile URL I can't, but if I do it from the "https://pod.url/people/4l0t0fnumb3rs4ndl3tt3rs" link then there's no problem. 

I noticed this in two different servers for months and now I'm seeing that a new big release is coming and the error is not fixed.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27418362-can-t-create-a-new-post-from-my-profile-page-depending-of-the-link?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #6501.

### DIFF
--- a/app/views/people/show.html.haml
+++ b/app/views/people/show.html.haml
@@ -23,7 +23,7 @@
 
       .stream_container
 
-        -if user_signed_in? && current_page?(person_path(current_user.person))
+        - if user_signed_in? && current_user.person == @person
           = render 'publisher/publisher', publisher_aspects_for(nil)
 
         .stream.clearfix#main_stream
--- a/spec/integration/profile_spec.rb
+++ b/spec/integration/profile_spec.rb
@@ -1,0 +1,72 @@
+require "spec_helper"
+require "requests_helper"
+
+describe PeopleController, type: :request do
+  context "for the current user" do
+    before do
+      login alice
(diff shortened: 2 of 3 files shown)


denschub:
Merged as 995ce18c377df01e4e7620d55d9c36ce56cc40aa, thanks.

app/views/people/show.html.haml
```diff
@@ -23,7 +23,7 @@
 
       .stream_container
 
-        -if user_signed_in? && current_page?(person_path(current_user.person))
+        - if user_signed_in? && current_user.person == @person
```
denschub: Comparing `.id` could be a faster way, but this looks good to me.

Issue #6506, opened 2015-10-15
Receiving notifications for own comment on own reshare


AugierLe42e:
Related to #6498

jhass:
@cmrd-senya want  to look into this?

cmrd-senya:
Oops, sorry, I'll take a look!

jaywink:
This should be a blocker for 0.5.4 imho if it is confirmed - 6 weeks is a long time to live with this ;)

Awesome work @cmrd-senya btw, but a little surprised the change was pushed to stable *after* code freeze.

jhass:
@jaywink stable targets 0.5.5 already, the freeze creates a release branch.

jaywink:
Ha, sorry, my bad ;)

AugierLe42e:
Ok, it looks like I receive notifications for *some of my original posts* too. That disturbing. I have the impression of talking to myself.

cmrd-senya:
Weird, but the problem reproduces randomly, mostly not reproducing. How about totally blocking a possibility to notify oneself?

I'll post a PR including tests and the block.

cmrd-senya:
@Flaburgan, @svbergerem, any of those who manage to reproduce the issue.

Could you please apply a patch for enabling extended logging: 5ce13b6b4d6f0455af10c4a3808ababe7451dfed

Then, reproduce the issue and send me logs?

cmrd-senya:
I managed to reproduce it locally.
It is a federation issue. It reproduces only when a user sharing with someone on a different pod. At the same time, it is not necessary to share a post of a remote user.

cmrd-senya:
#6512 should fix it.
Please test.

cmrd-senya:
I think it would have been nice if some of the reporters have tested it on their pods, before merge, but ok.

jhass:
Well, it's a valid edge case to guard against, I wouldn't revert it if it doesn't resolve the problem and we can always reopen the issue.

svbergerem:
#6512 fixed this for me. Let's see if @AugierLe42e is still able to reproduce this since he had more tests.

Flaburgan:
FTR this is merged in d-fr since 9AM this morning, UTC+1. If anyone wants to test it again there.

cmrd-senya:
Looks like it's fine now.

AugierLe42e:
NTR on my side. Thx.

PR #6508, opened 2015-10-15
Disable a self-notification possibility.

cmrd-senya:
a fix suggestion for #6506.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -15,7 +15,7 @@ class Notification < ActiveRecord::Base
   end
 
   def self.notify(recipient, target, actor)
-    return nil unless target.respond_to? :notification_type
+    return nil unless target.respond_to? :notification_type || recipient != actor
 
     note_type = target.notification_type(recipient, actor)
     return nil unless note_type
--- a/features/desktop/notifications.feature
+++ b/features/desktop/notifications.feature
@@ -142,6 +142,37 @@ Feature: Notifications
     And I should see "also commented on"
     And I should have 3 email delivery
 
+  Scenario: comment on my own reshare
+    Given "alice@alice.alice" has a public post with text "check this out!"
+    When I sign in as "bob@bob.bob"
+    And I am on "alice@alice.alice"'s page
+    And I follow "Reshare"
+    And I confirm the alert
+    And I am on "bob@bob.bob"'s page
+    And I focus the comment field
+    And I fill in the following:
+        | text        | hello, myself!    |
+    And I press "Comment"
+    Then I should see "less than a minute ago" within ".comment"
+    When I follow "Notifications" in the header
+    Then the notification dropdown should be visible
+    And I should not see "also commented on"
+    And I should have 1 email delivery
+
+  Scenario: comment on my own post
+    Given "alice@alice.alice" has a public post with text "check this out!"
+    When I sign in as "alice@alice.alice"
+    And I am on "alice@alice.alice"'s page
+    And I focus the comment field
+    And I fill in the following:
+        | text        | hello, myself!    |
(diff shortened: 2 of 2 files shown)


app/models/notification.rb
```diff
@@ -15,7 +15,7 @@ def self.for(recipient, opts={})
   end
 
   def self.notify(recipient, target, actor)
-    return nil unless target.respond_to? :notification_type
+    return nil unless target.respond_to? :notification_type || recipient != actor
```
diaspora-code-review: Use parentheses in the method call to avoid confusion about precedence.

jhass: I'm afraid that's not changing anything, you got to retest.

``rb
[50] pry(main)> puts "foo" || "bar"
foo
=> nil
[51] pry(main)> puts("foo") || "bar"
foo
=> "bar"
```

Comment (cmrd-senya) [inline app/models/notification.rb]: Well, that just means it doesn't reproduce for me. I had only one reproduction and then everything worked fine every time.

Is it fine if I do parenthesis:
```ruby
-    return nil unless target.respond_to? :notification_type || recipient != actor
+    return nil unless (target.respond_to?(:notification_type) || recipient != actor)
```

jhass: It would be very good to have a test that fails without the change :/

cmrd-senya: This new tests don't fail for me without a change.

@AugierLe42e, can you please help us reproducing the issue?

AugierLe42e: What the f? I didn't get that mention. The power of zoning randomly. What do you need exactly, @cmrd-senya ?

cmrd-senya: Not only we who have notification problems :)

Well, you can check if this tests fail in your setup. For me #6506
doesn't reproduce, so I need some steps to reproduce it.

AugierLe42e: You mean you want me to merge this PR on my pod and see if solves #6506?

jhass: @AugierLe42e no, we'd like a failing spec for the issue.

cmrd-senya: @AugierLe42e, are you able to certainly reproduce it?

AugierLe42e: Really, I don't understand what you are waiting for me. You want me to run these tests on my computer ?

Flaburgan: @cmrd-senya is not able to reproduce the bug, @AugierLe42e you're the one who reported it so he's asking for a way to reproduce it. The perfect way to give the steps to reproduce would be a test. But you can also describe it here.

cmrd-senya: Yes, that what I was going to answer :)

I believe the issue exists, I think I saw it once myself, but never
managed to reproduce.

AugierLe42e: Just reshare any public post and comment it. You should see a notification saying that you commented your own reshare. Reproduced on diaspora-fr.org. diaspote is not up to date.

AugierLe42e: It also reproduces on original posts but not all of them.

cmrd-senya: I don't have an account on diaspora-fr, and if I had, I doubt @Flaburgan
would let me tamper with the running pod to debug the issue.

I need a way to reproduce it on a fresh Diaspora installation.

Flaburgan: Ping me on IRC if you need an account, no problem with that.

@AugierLe42e the "comment on my own reshare" is tested in this PR. As you said, it's not the only case.

AugierLe42e: Yep. It also does it on some regular posts. No pattern found yet.

Issue #6479, opened 2015-10-09
Implement support for federation testing in the Diaspora automated testing system


cmrd-senya:
Capybara [has](https://groups.google.com/forum/#!searchin/ruby-capybara/sessions%7Csort:date/ruby-capybara/W4rXBp7x2yc/G7G4Q4GRCQAJ) a possibility to run multiple sessions.

We could employ this feature and run two instances of Diaspora application within capybara/cucumber tests, run separate sessions for each of them and test the pods interconnection this way.

Thus we can do automatic tests for federation, which will improve our software quality.

jhass:
However this will probably be fairly slow, we should look into getting a secondary CI server/service for them so we can run them in parallel to the existing builds.

cmrd-senya:
Oh, I thought it was closed!

For the record, so, generally, my https://github.com/cmrd-senya/diaspora_federation_behavior project and it's test [deployment](http://190.15.209.124:8010/grid) is a step to this issue solution.

cmrd-senya:
[diaspora_federation_behavior](https://github.com/cmrd-senya/diaspora_federation_behavior) consists of  several parts. It uses a modifed version of [diaspora-replica](https://github.com/cmrd-senya/diaspora-replica/tree/testfarm), it contains some rake tasks to control pods produced by diaspora-replica and it contains rspec tests.

I think, that having this project as a separate is likely to result in its abandonment. And nobody would really want to use it in order to test some federation changes. On the other hand, I see that it's possible to integrate these tests into the main diaspora tree. It still can't be easily integrated into travis due to use of vagrant, but at least developers could use it locally or bring up another CI system.

Another positive thing about having diaspora-replica integrated is that we will have an official and reproductible way of instantiating a pod, that will make it easier to try diaspora as podmin or even maintain it.

I'm interested in implementing this, but, however I want to be sure that it is something that is interesting for other project members to have in, so that my job is not wasted.

Issue #5114, opened 2014-08-12
Extract federation code into a gem and integrate it in diaspora


Flaburgan:
@Raven24 did an awesome work rewritting all the federation code of diaspora* in a gem available on his repo: 
\- https://github.com/Raven24/diaspora-federation 
This gem was tested by @carolinagc and @juliaguar and worked pretty nicely.
\- https://github.com/Team-D/federation-testbed

The next step is to integrate it inside the diaspora* code and then remove all the code which does not use this gem. Once done, it will be way more easier to debug, test and improve the federation, and we will be sure that the gem doesn't have bugs, so we can promote it to the world.

* [x] dig out the testbed project, verify it still runs (optional)
* [x] document how to run it + extend to more test cases, eventually covering all federation events (optional)
* [x] use the gem for generating user discovery documents (hostmeta, webfinger, hcard) - this should be relatively easy
* [x] use the gem for parsing user discovery documents - this could mean some bigger changes to the core code. we may need some additional test coverage in this area.
* [x] use the gem for generating federation messags (salmon envelopes w/o encryption, payloads) - this will touch some of the oldest code in this project. creating a pull request as early as possible is strongly recommended. this part requires regression testing with an unmodified d* server.
* [x] use the gem for parsing federation messsages - same as above
* [x] publish the gem as 0.0.1(?) and move the repo to the d* organization on GitHub

dimaursu:
Some of the specs do not belong in the main repository (and so do the same libraries): 
https://github.com/diaspora/diaspora/blob/develop/spec/lib/hcard_spec.rb, for example.
I started some work, but unfortunately there aren't that many tests for the federation itself - it's hard to tell if I broke it or it's still working. I'll work on the tests the following days.

jaywink:
@Raven24 and others: @SuperTux88 is working [on a diaspora_federation gem also](https://github.com/diaspora/diaspora/pull/6151). Unless the plan has "changed" and raven24's gem isn't going to be used and supertux's gem will be used (between you guys), I'm worried there could be lots of effort lost from someone, so just calling out here for possibility for some work-together in this area?

Also, there is a $185 bounty to the person who integrates raven24's gem. I'm going to throw in the assumption that we can rename the issue to "implement diaspora federation through a gem" etc.

Raven24:
we're on it: SuperTux88/diaspora_federation#1
;)

Flaburgan:
[...]

PR #6526, opened 2015-10-28
A prototype for the federation message generation automated tests.

cmrd-senya:
This is a prototype made in purpose to try and implement an automated
way of testing federation. This patch contains a test for an outgoing
federation message.

The test case loads [federation-testbed](https://github.com/cmrd-senya/federation-testbed/tree/shape-into-gem) as a gem. It uses an embedded
Sinatra web application of the testbed and stubs outgoing federated
calls to it. The testbed application process queries as if it was a real
Diaspora pod and fails a test in case when the received message
is inconsistent.

This is an early version used to highlight difficulties and issues
of the approach and implementation. Development versions of the
involved gems are used.

refs #5114 #6479

**For some reason `bin/rake spec` gets broken with this. Help wanted!**

While the newly added test runs fine if run alone with `bin/rspec spec/federation_messages/federation_messages_spec.rb`, the whole thing gets broken - running `bin/rake spec` produces [weird output](https://gist.github.com/cmrd-senya/cf18acdc42b0c5849874#file-rake-spec-log-L64). The highlighted line points to some bundler message. So that is probably just some problem cause by using git versions of gems. I am pretty sure, that it's not my fault!

### DIFF
--- a/app/models/pod.rb
+++ b/app/models/pod.rb
@@ -25,7 +25,9 @@ class Pod < ActiveRecord::Base
 
   class << self
     def find_or_create_by(opts) # Rename this method to not override an AR method
-      u = URI.parse(opts.fetch(:url))
+      url = opts.fetch(:url)
+      return if url.nil?
+      u = URI.parse(url)
       find_or_initialize_by(host: u.host).tap do |pod|
         unless pod.persisted?
           pod.ssl = (u.scheme == "https")
--- a/spec/federation_messages/federation_messages_spec.rb
+++ b/spec/federation_messages/federation_messages_spec.rb
@@ -1,0 +1,42 @@
+require "spec_helper"
+require "federation-testbed"
+
+describe "Generation and dispatch of federation messages" do
+  before :all do
+    @testbed_pod = FactoryGirl.build(:pod).host
+    FederationTestbed.config.configure(
+      "testbed_config" => {"host" => @testbed_pod, "method" => "http"},
+      "test_target"    => {
+        "method" => "http",
+        "host"   => "localhost:3000",
+        "user"   => "alice"
+      }
+    )
+    u = FederationTestbed::User.new
+    u.id = alice.diaspora_handle
+    u.public_key = alice.public_key
+    u.private_key = nil
+    FederationTestbed.test_data.push(u)
+    rack_app = FederationTestbed::App.new
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/.well-known/host-meta")).to_rack(rack_app)
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/webfinger?q=acct:test@" + @testbed_pod))
+      .to_rack(rack_app)
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/hcard/users/test@" + @testbed_pod))
+      .to_rack(rack_app)
+    stub_request(:post, Addressable::Template.new(@testbed_pod + "/receive/users/{guid}")).to_rack(rack_app)
+  end
+  before do
+    allow_any_instance_of(Postzord::Dispatcher::Private).to receive(:deliver_to_remote).and_call_original
+    allow_any_instance_of(Postzord::Dispatcher::Public).to receive(:deliver_to_remote).and_call_original
+  end
+  after :all do
+  end
+
+  describe "user share request" do
+    it "works" do
+      friend = Person.find_or_fetch_by_identifier("test@" + @testbed_pod)
+      expect(friend).not_to be_nil
+      expect(alice.share_with(friend, alice.aspects.first)).not_to be_falsy
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,8 @@ gem "unicorn", "4.9.0", require: false
 
 # Federation
 
-gem "diaspora_federation-rails", "0.0.8"
+gem 'diaspora_federation-rails', :git => "https://github.com/SuperTux88/diaspora_federation.git", :branch => "salmon"
+
 
 # API and JSON
 
@@ -282,6 +283,8 @@ group :test do
   gem "timecop",            "0.8.0"
   gem "webmock",            "1.21.0", require: false
   gem "shoulda-matchers",   "2.8.0", require: false
+
+  gem "federation-testbed", :git => "https://github.com/cmrd-senya/federation-testbed.git", :branch => "shape-into-gem"
 end
 
 group :development, :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,26 @@
+GIT
+  remote: https://github.com/SuperTux88/diaspora_federation.git
+  revision: cc63cfe285a55212b40ddffc43279db0afbf83f8
+  branch: salmon
+  specs:
+    diaspora_federation (0.0.8)
+      faraday (~> 0.9.0)
+      faraday_middleware (~> 0.10.0)
+      nokogiri (~> 1.6, >= 1.6.6)
+      typhoeus (~> 0.7)
+      valid (~> 1.0)
+    diaspora_federation-rails (0.0.8)
+      diaspora_federation (= 0.0.8)
+      rails (~> 4.2)
+
+GIT
+  remote: https://github.com/cmrd-senya/federation-testbed.git
+  revision: f712789aa6da3e8ff440cd9d4f72fdad9e42d676
+  branch: shape-into-gem
+  specs:
+    federation-testbed (0.1.0)
+      rest-client
+
 GEM
   remote: https://rubygems.org/
   remote: https://rails-assets.org/
@@ -154,15 +177,6 @@ GEM
       eventmachine (~> 1.0.8)
       http_parser.rb (~> 0.6)
       nokogiri (~> 1.6)
-    diaspora_federation (0.0.8)
-      faraday (~> 0.9.0)
-      faraday_middleware (~> 0.10.0)
-      nokogiri (~> 1.6, >= 1.6.6)
-      typhoeus (~> 0.7)
-      valid (~> 1.0)
-    diaspora_federation-rails (0.0.8)
-      diaspora_federation (= 0.0.8)
-      rails (~> 4.2)
     diff-lcs (1.2.5)
     docile (1.1.5)
     domain_name (0.5.24)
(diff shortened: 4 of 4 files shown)


cmrd-senya:
Travis-ci seem to be happy with that at least in one test, so that probably my local issue, but still I need help investigating.

jhass:
Try `bundle exec rake spec`.

cmrd-senya:
Yes, that helped.

Any comments on the patch itself? I'm pretty sure there are many to improve in my Ruby code.

denschub:
Mh, using the HEAD of @SuperTux88 federation gem feels wrong. There is heavy development going on and the releases are there for a reason.

Also, I wonder if this actually should be part of the federation gem? I am aware that the gem is not a server implementation, but it feels like diaspora should not test the federation itself, it should "trust" the federation gem testsuite and test if the gem gets called correctly.

I have not looked at your testbed gem but are you able to integrate it into the federation gem rather than diasporas core itself?

Also, this is just my very unsure opinion here. Let's see what others have to say.

cmrd-senya:
Of course, I don't propose to pull development version of any gem! That's just the only way it works for a moment, beacuse this is a part of heavy development you talk about.

The federation gem has its own testsuit and the testbed gem is the user of the federation gem itself.

If you look on what we test, you'll see that we call functions of Diaspora internal API:
```rb
      friend = Person.find_or_fetch_by_identifier("test@" + @testbed_pod.host)
      expect(friend).not_to be_nil
      expect(alice.share_with(friend, alice.aspects.first)).not_to be_falsy
```

For that test it is indifferent if we use the federation gem or not, this is a black box for us. We check, that if we ask Diaspora to fetch a remote user and setup a sharing relationships, then we will see that Diaspora federates correct messages for that. This test routes outgoing calls to the federation testbed and its responsibility is to raise an exception in case of any message inconsistency. Maybe I should also add a feedback from a testbed here to test, what exactly happened, because, for example, something could be broken so, that Diaspora sends a correctly composed federation message, but in wrong order.

This will help to find regressions while changing the federation related code, mainly while moving message generation features to the new federation gem. Also this will help to catch regressions in case of wrong usage of the federation gem.

app/models/pod.rb
```diff
@@ -25,7 +25,9 @@ class Pod < ActiveRecord::Base
 
   class << self
     def find_or_create_by(opts) # Rename this method to not override an AR method
-      u = URI.parse(opts.fetch(:url))
+      url = opts.fetch(:url)
```
jhass: `fetch` will raise on a missing key.

spec/federation_messages/federation_messages_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    @testbed_pod = FactoryGirl.build(:pod).host
+    FederationTestbed.config.configure(
+      "testbed_config" => {"host" => @testbed_pod, "method" => "http"},
+      "test_target"    => {
+        "method" => "http",
+        "host"   => "localhost:3000",
+        "user"   => "alice"
+      }
+    )
+    u = FederationTestbed::User.new
```
jhass: Please try to avoid single letter variables.

spec/federation_messages/federation_messages_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    u.id = alice.diaspora_handle
+    u.public_key = alice.public_key
+    u.private_key = nil
+    FederationTestbed.test_data.push(u)
+    rack_app = FederationTestbed::App.new
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/.well-known/host-meta")).to_rack(rack_app)
```
jhass: Prefer `"string #{interpolation}"` over `"string " + concatenation`

spec/federation_messages/federation_messages_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/webfinger?q=acct:test@" + @testbed_pod))
+      .to_rack(rack_app)
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/hcard/users/test@" + @testbed_pod))
+      .to_rack(rack_app)
+    stub_request(:post, Addressable::Template.new(@testbed_pod + "/receive/users/{guid}")).to_rack(rack_app)
+  end
+  before do
```
jhass: Add an empty line between such calls.

spec/federation_messages/federation_messages_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    allow_any_instance_of(Postzord::Dispatcher::Private).to receive(:deliver_to_remote).and_call_original
+    allow_any_instance_of(Postzord::Dispatcher::Public).to receive(:deliver_to_remote).and_call_original
+  end
+  after :all do
+  end
```
jhass: No need to include an empty hook.

spec/federation_messages/federation_messages_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "spec_helper"
+require "federation-testbed"
+
+describe "Generation and dispatch of federation messages" do
```
jhass: These are ultimately integration tests, so how about a subfolder in `spec/integration`?

spec/federation_messages/federation_messages_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/.well-known/host-meta")).to_rack(rack_app)
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/webfinger?q=acct:test@" + @testbed_pod))
+      .to_rack(rack_app)
+    stub_request(:get, Addressable::Template.new(@testbed_pod + "/hcard/users/test@" + @testbed_pod))
+      .to_rack(rack_app)
+    stub_request(:post, Addressable::Template.new(@testbed_pod + "/receive/users/{guid}")).to_rack(rack_app)
+  end
+  before do
+    allow_any_instance_of(Postzord::Dispatcher::Private).to receive(:deliver_to_remote).and_call_original
+    allow_any_instance_of(Postzord::Dispatcher::Public).to receive(:deliver_to_remote).and_call_original
```
jhass: These are no-ops. Did you mean to use `expect`?

app/models/pod.rb
cmrd-senya: I check this not because the key could be missing, but because it could present but be nil.
When called from [here](https://github.com/diaspora/diaspora/blob/develop/lib/hydra_wrapper.rb#L87) while running the testsuite response.effective_url in nil. That's why I had to add these lines.

spec/federation_messages/federation_messages_spec.rb
cmrd-senya: I would prefer spec/integration/federation or something like this. "testbed" is a tool to test federation. And the folder name should represent what we test but not the tool we use to test.

spec/federation_messages/federation_messages_spec.rb
cmrd-senya: Well, probably I just don't need these lines. I copied these from somewhere and thought they affect some call stubbing.

app/models/pod.rb
cmrd-senya: Maybe then I can add a check for Rails.env.test? here? And exit silently on url.nil? if we are on test?

denschub:
Thanks for the clarification, works for me! :)

app/models/pod.rb
jhass: I would be fine with raising `ArgumentError` when `nil` is passed explicitly, but not silently ignoring it. But then the validations should catch it. In any case, either the specs should stub out behaviour that differs in the test environment to match the behaviour in the production environment, or if the value might end up `nil` in the production environment too, the call should be fixed to not happen in that case.

spec/federation_messages/federation_messages_spec.rb
jhass: Sure, whatever.

app/models/pod.rb
cmrd-senya: After the latest fixes (probably moving to net-http?) it doesn't happend anymore, so I'll just remove these lines.

spec/federation_messages/federation_messages_spec.rb
cmrd-senya: These aren't no-ops. Without these lines no outgoing calls to ```"#{@testbed_pod}/receive/users/{guid}"``` are performed.

jhass: That seems unlikely.

cmrd-senya: Here diffs are provided against the present PR.
1) https://gist.github.com/cmrd-senya/37614dd5789421e0f2d9
Fails because tries to perform an outgoing call that we don't stub (the stub call commented out).
2) https://gist.github.com/cmrd-senya/5b1d01a005ff590f5de4
Comment out the lines we talk about. No fails anymore => an outgoing call doesn't performed.

Is it enough as a prove?

app/models/pod.rb
cmrd-senya: And bringing back the lines about Postzord has brought back this problem as well.

spec/federation_messages/federation_messages_spec.rb
jhass: Oh well, it overrides https://github.com/diaspora/diaspora/blob/develop/spec/spec_helper.rb#L85

cmrd-senya: So it is fine to leave it as is?

app/models/pod.rb
cmrd-senya: This is generally a webmock problem. It doesn't stub effective_url in https://github.com/bblimke/webmock/blob/master/lib/webmock/http_lib_adapters/typhoeus_hydra_adapter.rb.

So to match the behavior in the production environment we have to patch it.

cmrd-senya: [This thing](https://github.com/bblimke/webmock/pull/540) helps.

spec/federation_messages/federation_messages_spec.rb
jhass: For now, I'm not sure if we want to keep these global stubs, but that's out of scope here anyway.

cmrd-senya:
The PR now lives at #6531. Please do any further discussions there.

PR #6530, opened 2015-10-30
Consolidate amount of sidekiq queues from 13 to 5

### LINKED ISSUE #5571: Consolidate some sidekiq queues
Sidekiq wiki says:

> I don't recommend having more than a handful of queues. Sidekiq is not designed to work well with dozens of queues.

https://github.com/mperham/sidekiq/wiki/Advanced-Options

We have 13 at the moment. And should probably consolidate some if there is no huge reason not to.

We could also optionally weight some of the queues much higher than the low prio ones which probably are getting too much attention now, like maintenance and default..

Probably good idea to add to readme a note to run sidekiq queue empty for at least removed queues.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7835581-consolidate-some-sidekiq-queues?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jaywink:
Sidekiq documentation says 'Sidekiq is not designed to work well with dozens of queues.'. Having the amount of queues we have at the moment brings no anyway.

Closes #5571

If this is accepted, should put a note in Changelog, in relevant version to be included in, for podmins to run their queues empty on update.

---

This change is [Reviewable](https://reviewable.io/reviews/diaspora/diaspora/6530)

### DIFF
--- a/app/workers/export_photos.rb
+++ b/app/workers/export_photos.rb
@@ -5,7 +5,7 @@
 
 module Workers
   class ExportPhotos < Base
-    sidekiq_options queue: :export
+    sidekiq_options queue: :low
 
     def perform(user_id)
       @user = User.find(user_id)
--- a/app/workers/export_user.rb
+++ b/app/workers/export_user.rb
@@ -5,7 +5,7 @@
 
 module Workers
   class ExportUser < Base
-    sidekiq_options queue: :export
+    sidekiq_options queue: :low
 
     def perform(user_id)
       @user = User.find(user_id)
--- a/app/workers/fetch_public_posts.rb
+++ b/app/workers/fetch_public_posts.rb
@@ -4,7 +4,7 @@
 
 module Workers
   class FetchPublicPosts < Base
-    sidekiq_options queue: :http_service
+    sidekiq_options queue: :medium
 
     def perform(diaspora_id)
       Diaspora::Fetcher::Public.new.fetch!(diaspora_id)
--- a/app/workers/http_multi.rb
+++ b/app/workers/http_multi.rb
@@ -4,7 +4,7 @@
 
 module Workers
   class HttpMulti < Base
-    sidekiq_options queue: :http
+    sidekiq_options queue: :medium
 
     MAX_RETRIES = 3
     ABANDON_ON_CODES=[:peer_failed_verification, # Certificate does not match URL
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -6,16 +6,8 @@
 <% end %>
 :concurrency: <%= AppConfig.environment.sidekiq.concurrency.to_i %>
 :queues:
-  - socket_webfinger
-  - photos
-  - http_service
-  - dispatch
-  - mail
-  - delete_account
-  - receive_local
-  - receive
-  - receive_salmon
-  - http
-  - export
-  - maintenance
+  - urgent
+  - high
+  - medium
+  - low
   - default
--- a/spec/controllers/publics_controller_spec.rb
+++ b/spec/controllers/publics_controller_spec.rb
@@ -42,7 +42,7 @@ describe PublicsController, :type => :controller do
       post :receive, "guid" => @user.person.guid.to_s, "xml" => xml
     end
 
-    it 'unescapes the xml before sending it to receive_salmon' do
+    it "unescapes the xml before sending it to queue" do
       aspect = @user.aspects.create(:name => 'foo')
       post1 = @user.post(:status_message, :text => 'moms', :to => [aspect.id])
       xml2 = post1.to_diaspora_xml
--- a/spec/workers/receive_salmon_spec.rb
+++ b/spec/workers/receive_salmon_spec.rb
@@ -12,7 +12,7 @@ describe Workers::ReceiveEncryptedSalmon do
       end
     }
   end
-  it 'calls receive_salmon' do
+  it "calls receiver" do
     zord = double
 
     expect(zord).to receive(:perform!)
--- a/Changelog.md
+++ b/Changelog.md
@@ -30,6 +30,16 @@ bind to an UNIX socket at `unix:tmp/diaspora.sock`. Please change your local
 
 With the port to Bootstrap 3, app/views/terms/default.haml has a new structure. If you have created a customised app/views/terms/terms.haml or app/views/terms/terms.erb file, you will need to edit those files to base your customisations on the new default.haml file.
 
+## Sidekiq queue changes
+
+We've decreased the amount of sidekiq queues from 13 to 5 in PR [#6530](https://github.com/diaspora/diaspora/pull/6530).
+The new queues are organized according to priority for the jobs they will process. When upgrading please make sure to 
+empty the sidekiq queues before shutting down the server for an update.
+
+If you run your sidekiq with a custom queue configuration, please make sure to update that for the new queues.
+
+The new queues are: `urgent, high, medium, low, default`.
+
 ## Refactor
 * Improve bookmarklet [#5904](https://github.com/diaspora/diaspora/pull/5904)
 * Update listen configuration to listen on unix sockets by default [#5974](https://github.com/diaspora/diaspora/pull/5974)
@@ -59,6 +69,7 @@ With the port to Bootstrap 3, app/views/terms/default.haml has a new structure. 
 * Replace mobile background with color [#6415](https://github.com/diaspora/diaspora/pull/6415)
 * Port flash messages to backbone [#6395](https://github.com/diaspora/diaspora/6395)
 * Change login/registration/forgot password button color [#6504](https://github.com/diaspora/diaspora/pull/6504)
+* Consolidate sidekiq queues [#6530](https://github.com/diaspora/diaspora/pull/6530)
 
 ## Bug fixes
 * Destroy Participation when removing interactions with a post [#5852](https://github.com/diaspora/diaspora/pull/5852)
--- a/app/workers/clean_cached_files.rb
+++ b/app/workers/clean_cached_files.rb
@@ -2,7 +2,7 @@ module Workers
   class CleanCachedFiles < Base
     include Sidetiq::Schedulable
 
-    sidekiq_options queue: :maintenance
+    sidekiq_options queue: :low
 
     recurrence { daily }
 
--- a/app/workers/deferred_dispatch.rb
+++ b/app/workers/deferred_dispatch.rb
@@ -4,7 +4,7 @@
 
 module Workers
   class DeferredDispatch < Base
-    sidekiq_options queue: :dispatch
+    sidekiq_options queue: :high
 
     def perform(user_id, object_class_name, object_id, opts)
       user = User.find(user_id)
--- a/app/workers/delete_account.rb
+++ b/app/workers/delete_account.rb
@@ -5,7 +5,7 @@
 
 module Workers
   class DeleteAccount < Base
-    sidekiq_options queue: :delete_account
+    sidekiq_options queue: :low
     
     def perform(account_deletion_id)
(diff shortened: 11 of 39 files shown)


spec/controllers/publics_controller_spec.rb
```diff
@@ -42,7 +42,7 @@
       post :receive, "guid" => @user.person.guid.to_s, "xml" => xml
     end
 
-    it 'unescapes the xml before sending it to receive_salmon' do
+    it 'unescapes the xml before sending it to receive' do
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/workers/receive_salmon_spec.rb
```diff
@@ -12,7 +12,7 @@
       end
     }
   end
-  it 'calls receive_salmon' do
+  it 'calls receive' do
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/controllers/publics_controller_spec.rb
jaywink: Weird, I ran `bin/pronto run -c upstream/develop` and it didn't give me anything :P

config/sidekiq.yml
```diff
@@ -6,16 +6,8 @@
...
+  - http
   - mail
-  - delete_account
-  - receive_local
   - receive
-  - receive_salmon
-  - http
-  - export
-  - maintenance
-  - default
```
jhass: We definitely should keep polling `default`.

jaywink: Why if we don't push anything to it? Or at least I couldn't find anything..

jhass:
That's a bit aggressive, have a look at https://github.com/diaspora/diaspora/blob/joindiaspora/Procfile for example.

We definitely should keep webfinger jobs at the top priority and slow http_multi jobs at the least priority before maintenance.

jaywink:
I don't think joindiaspora.com configuration is a good way to structure things on, judging by how bad it mostly behaves ;)

I like the idea of priority queues, but then our default way of running sidekiq should somehow use them. If we have a queue called `urgent` and a queue called `low`, it is just misleading to then treat them the same. Or at least I can't figure out whether we prioritize queues currently, except the way joindiaspora is doing.

Single pods can always prioritize things however they want. But if we do it in the core software, our default configuration should behave like the prioritization has some meaning.

Separating webfinger jobs back into their own queue wont make them run any faster with the normal configuration?

jhass:
The order in sidekiq.yml does weight the queues: https://github.com/mperham/sidekiq/blob/b0d606929d42fb380da771ae094621305840ed1b/lib/sidekiq/cli.rb#L395-L400

jaywink:
Ah, fair enough then. Priority queues would make sense then! I'll make a suggestion on what could go to which queue, and then adjust.

But why do you want to keep polling default?

jhass:
https://github.com/mperham/sidekiq/wiki/Advanced-Options#queues Here's where I read it.

`default` is what's used if we forget to specify a queue for a job or the specification breaks for some reason. That way impact is limited.

denschub:
Just for the record: I currently run workers in groups of

- `receive`, `receive_salmon`, `receive_local`, `dispatch`
- `http`, `http_service`, `socket_webfinger`
- `export`, `maintenance`, `delete_account`
- `photos`, `mail`

in order to keep the job latency as low as possible.

jaywink:
Okey dokey, priority queues now. What do you think of the queue I assigned each worker to?

app/workers/export_photos.rb
```diff
@@ -5,7 +5,7 @@
 
 module Workers
   class ExportPhotos < Base
-    sidekiq_options queue: :export
+    sidekiq_options queue: :low
```
jhass: I'd give this a `medium`, since while long running, it is an action on which completion the user actively waits for.

app/workers/export_user.rb
```diff
@@ -5,7 +5,7 @@
 
 module Workers
   class ExportUser < Base
-    sidekiq_options queue: :export
+    sidekiq_options queue: :low
```
jhass: Ditto.

app/workers/fetch_public_posts.rb
```diff
@@ -4,7 +4,7 @@
 
 module Workers
   class FetchPublicPosts < Base
-    sidekiq_options queue: :http_service
+    sidekiq_options queue: :high
```
jhass: I'd downrate this to `medium` or even `low`, having some public posts when viewing a profile the first time is nice, but not mandatory.

jhass:
Looks like you missed a few? :) Noticed `HttpMulti` missing at least.

jaywink:
Going through them now... Of course I only did the ones in the previous commit...

app/workers/fetch_public_posts.rb
jaywink: I'd like to keep this `medium` then. At least personally I quite often add a profile and then refresh for the posts a little while.

app/workers/http_multi.rb
```diff
@@ -4,7 +4,7 @@
 
 module Workers
   class HttpMulti < Base
-    sidekiq_options queue: :http
+    sidekiq_options queue: :high
```
jhass: These jobs are extremely long running, they shouldn't block other jobs, so this should be `medium` or even `low`.

denschub:
I added a list of workers and their assigned jobs to the PR text above so it's a bit nicer to discuss about it.

Good job so far. I agree to move all `maintenance` jobs into the `low` queue since nobody cares at all. I'd love to have a queue that, in theory, could be turned off for some hours without anyone noticing for scaling reason. Imagine there are a ton of incoming messages and you'd be able to turn off the `low` worker and add more resources to process the incoming requests, that would be quite nice.

Therefore, I'd move all `Mail` workers and `ResendInvitation` up to the medium stage while I'd move `DeleteAccount`, `ExportPhotos` and `ExportUser` down to the low queue.

In my opinion, `ResetPassword` is not more important than other mails, so I'd keep those in the same queue. There will be a delay because of slow email servers, graylisting, ... nonetheless so I do not think we need to process those jobs super fast.

I do agree that `FetchWebfinger` should have a high priority since this is the only job interacting in real time with the user, but wasting an entire queue (and thus, in my setup, a whole bunch of workers) just for profile fetching seems a bit odd.

On the other hand, `FetchProfilePhoto`, `GatherOEmbedData`, `GatherOpenGraphData` and `HttpMulti` in the `high` queue could possible take a huge amount of time to process, thus blocking the Receive jobs from being processed. Maybe move the receive jobs to the urgent queue as well so we're able to process incoming stuff as fast as possible? After all, they are not doing that much.

I do think OEmbed and OpenGraph data is not worth having a high priority since they do not really add much, so I'd move them down to the medium queue. However, `NotifyLocalUsers` should get a bit more attention.

jaywink:
> Therefore, I'd move all Mail workers and ResendInvitation up to the medium stage

I don't really see mail jobs as being more than `low`, except for password reset. IMHO everything else should be more important and done first.

> while I'd move DeleteAccount, ExportPhotos and ExportUser down to the low queue.

I think DeleteAccount should definitely not be in low. But I can move it there if Jonne agrees also. But Jonne wanted the export jobs higher than low, where I initially placed them. Which will it be? I'd go for low myself. These are the least urgent jobs to me.

Rationale for `reset_password` being urgent: if a user needs to login via resetting password, they want the password reset link immediately. When I myself do a password reset, I immediately bang fetch in my email client until it arrives. If it doesn't, I go do something else and say fuck that stupid site. Also, it can't be a heavy job.

After the other changes mentioned, the queues now look like this:

Urgent:
* fetch_webfinger
* receive*
* reset_password

High:
* deferred_dispatch
* delete_post_from_service
* notify_local_users

Medium:
* delete_account
* export_photos
* export_user
* fetch_profile_photo
* fetch_public_posts
* gather_oembed_data
* gather_open_graph_data
* http_multi
* post_to_service
* publish_to_hub

Low:
* Mails (except reset password)
* clean_cached_files
* process_photo
* queue_users_for_removal
* recurring_pod_check
* remove_old_user
* resend_invitation

denschub:
> I think DeleteAccount should definitely not be in low.

Why not? The account is locked anyway and the job could possible take forever.

If you feel strong about the reset password mails, fine. Everything else looks good to me.

jaywink:
> Why not? The account is locked anyway and the job could possible take forever.

OK fair point. Moved to low.

jaywink:
@jhass ok with moving ExportPhotos and ExportUser down to low or shall we just use this? Shall I add a readme to the changelog now or later? I doubt this should be included in the minor.

jhass:
Yeah I guess let's try it out for a while, changelog would be nice, make sure to mention that custom worker configurations need to be redone. At the end of the day I'm just guessing what would be good values here too.

jaywink:
I've moved the exports to low then, as Dennis also agreed on that. I don't think they should hold any jobs in medium up since they are nothing to do with things like pushing messages out and so on, which is more important.

I've added a changelog entry, hopefully with sane sounding text. Need to remind of course on actual 0.6 release separately for podmins to check this.

Mergy mergy?

jaywink:
Just pulled into my pod and turns out migrating the queues is kinda annoying, since we have scheduled stuff in queues `http` and `maintenance` - at least.

So talked on IRC that maybe a rake task to run a sidekiq with the old queues that podmins can fire off to flush the queues whenever the scheduled jobs accumulate there. I think this is acceptable since this PR is 0.6 stuff anyway. I'd rather not go towards hacking any redis migrations.

So no mergy yet.

denschub:
++ to the dedicated worker triggered by a rake task.

denschub:
Ping.

jaywink:
It's not likely I'll be finishing this at the moment, feel free to orphan, continue it or close it.

Flaburgan:
So what's missing here? Only the rake task to clean the old queues? Is there anything else left?

cmrd-senya:
So if we apply this patch, and then run sidekiq with the old queues, how will it manage to pick proper workers for the old jobs scheduled in redis? Is the worker name stored for each job schedule? So it is as simple as running siekiq with the old queues names? And what if we leave the old queues names in the config/sidekiq.yml along with the new ones for a migration period? Will it work? Or it would bloat the memory? Maybe it's possible to set sidekiq to run this queues only when there are some jobs left for them? Because there won't be any new jobs in these queues anyway.

denschub:
> how will it manage to pick proper workers for the old jobs scheduled in redis?

You can specify the queue via the workes cli arguments, `-q [queue name]`. Since queues are only identified by their name without any magic, I see no issues here.

> And what if we leave the old queues names in the config/sidekiq.yml along with the new ones for a migration period? Will it work? Or it would bloat the memory?

No idea. Sidekiq claims it's not designed for "more than a handful of queues", but who knows what would happen. I doubt the `sidekiq.yml` will be an issue when specifying the queue manually as mentioned above, so removing them in the config should not cause any issues. Obviously, this has to be tested, i'm replying right off the top of my head here.

Flaburgan:
So, can we simply ask podmins to manually run `bin/bundle exec sidekiq -q maintenance dispatch delete_account http http_service export photos socket_webfinger mail receive_local receive receive_salomon` once the new queues have been created?

jhass:
We can but it's bug prone, writing a rake task seems simple enough.

jhass:
Superseded by #6950

PR #6533, opened 2015-11-04
Catch Diaspora::NotMine on Post Controller

### LINKED ISSUE #6532: Inconsistent Error 500 when it should be 403
I'm busy writing a [diaspora\* API class for WordPress](https://github.com/gutobenn/wp-to-diaspora/blob/master/lib/class-api.php) and have come across inconsistent error code responses.
When trying to delete a comment that I'm not authorised to, I correctly get a "403 Forbidden" response.
If I try the same with a post, I get a "500 Internal Server Error". This should also return "403 Forbidden" in my view.

This behaviour is forcing me to tweak my class in an illogical way, so I'm hoping this can be fixed at the source.

Looking at the diaspora\* code, `comments_controller.rb` has a [`respond_destroy_error`](https://github.com/diaspora/diaspora/blob/develop/app/controllers/comments_controller.rb#L66) method, which handles the issue correctly when a [`destroy`](https://github.com/diaspora/diaspora/blob/develop/app/controllers/comments_controller.rb#L28 "method in comments_controller.rb") fails.
Looking at the [`destroy`](https://github.com/diaspora/diaspora/blob/develop/app/controllers/posts_controller.rb#L51 "method in posts_controller.rb") method in `posts_controller.rb`, there is no error handling in case of a failure, which is why I assume an internal server error occurs.

P.S. Maybe this would be an issue to get me started into developing for diaspora\* and finally getting into RoR. If you feel this is a newbie issue, I'd love to try and give it a shot :smiley: 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27945770-inconsistent-error-500-when-it-should-be-403?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

KentShikama:
Fixes #6532 

@noplanman Thanks for the catch! The `posts_controller.rb` was indeed missing a `rescue_from` clause.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -21,6 +21,10 @@ class PostsController < ApplicationController
     end
   end
 
+  rescue_from Diaspora::NotMine do
+    render text: "You are not allowed to do that", status: 403
+  end
+
   def show
     post_service = PostService.new(id: params[:id], user: current_user)
     post_service.mark_user_notifications
--- a/spec/controllers/posts_controller_spec.rb
+++ b/spec/controllers/posts_controller_spec.rb
@@ -135,5 +135,15 @@ describe PostsController, type: :controller do
       message = alice.post(:status_message, text: "hey", to: alice.aspects.first.id)
       delete :destroy, format: :js, id: message.id
     end
+
+    context "when Diaspora::NotMine is raised by retract post" do
+      it "will respond with a 403" do
+        expect(post_service_double).to receive(:retract_post).and_raise(Diaspora::NotMine)
+        message = alice.post(:status_message, text: "hey", to: alice.aspects.first.id)
+        delete :destroy, format: :js, id: message.id
+        expect(response.body).to eq("You are not allowed to do that")
+        expect(response.status).to eq(403)
+      end
+    end
   end
 end


app/controllers/posts_controller.rb
```diff
@@ -21,6 +21,10 @@ class PostsController < ApplicationController
     end
   end
 
+  rescue_from Diaspora::NotMine do
+    render text: "You are not allowed to do that", status: 403
```
jhass: Shouldn't we `I18n` this?

theworldbright: Along with these I suppose?

https://github.com/diaspora/diaspora/blob/develop/app/controllers/aspect_memberships_controller.rb#L71
https://github.com/diaspora/diaspora/blob/develop/app/controllers/aspect_memberships_controller.rb#L75
https://github.com/diaspora/diaspora/blob/develop/app/controllers/aspect_memberships_controller.rb#L79
https://github.com/diaspora/diaspora/blob/develop/app/controllers/invitation_codes_controller.rb#L5

jhass: I guess so, maybe in a separate PR

jhass:
Merged as a054a35863e68015b54f8f029866e5e753592e6e

Thanks!

Issue #6444, opened 2015-10-01
close the new gallery when clicking outside an image


kas70:
for now it's only closable by pressing ESC or clicking on the X.

AugierLe42e:
This is unhopfully impossible due to the way the new gallery manages image dispaying. The image is set as background image of the gallery so it's not possible to detect outside clicking as there is technically no outside of the image.

kas70:
is it possible to recognize clicking inside the image? i think this could also be an option.

svbergerem:
This is working in the [blueimp gallery demo](https://blueimp.github.io/Gallery/).

kas70:
ok, i think we should adopt the behaviour: 
-  clicking outside closes the gallery
- clicking inside fades out the navigation

AugierLe42e:
> This is working in the blueimp gallery demo.

That's right. They have probably released an update.

svbergerem:
Not really. The current version is 2.16.0 and we already use that version.

AugierLe42e:
Weird. Our implementation uses image as background on a `div`. There must be a configuration to do somewhere, then.

PR #6536, opened 2015-11-07
Close gallery on outside click

ghost:
Solves #6444

### DIFF
--- a/app/assets/javascripts/app/views/gallery_view.js
+++ b/app/assets/javascripts/app/views/gallery_view.js
@@ -18,8 +18,10 @@ app.views.Gallery = app.views.Base.extend({
   preventHideControls: function(){
     var lightbox = $("#blueimp-gallery");
     var onEvent = function(ev){
-      ev.preventDefault();
-      ev.stopPropagation();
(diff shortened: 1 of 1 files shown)


AugierLe42e:
The second commit let the gallery display the original photo instead of the thumb large.

AugierLe42e:
This is ready to merge.

svbergerem:
I'm not sure about your second commit. The file size of the original image could be a few MB.

AugierLe42e:
On most pods, though, it'll be limited in size, like, for instance to 4MB on d-fr. I don't think this is really a big problem since it only affects desktop version.

jhass:
Please just stop sneaking potentially controverse changes into otherwise unrelated PRs, you're harming yourself.

AugierLe42e:
I think you're misinterpreting me, @jhass. If I wanted to "sneak" something, I couldn't be doing worse than in a such tiny PR. It's a bit ridiculous saying that. I sincerely didn't think that could be any controversial (and I highly regret there are so many controversial things). But if that's what you want, I open a separate PR. It has zero cost for me.

svbergerem:
Thanks! :)

PR #6539, opened 2015-11-10
Implement integration tests for the federation messages receive feature

cmrd-senya:
This is some initial tests, more to come.

refs #5114 #6479


spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
+require 'spec_helper'
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+require File.dirname(__FILE__) + "/factories"
+
+describe Workers::ReceiveEncryptedSalmon do
+  def generate_xml(entity)
+    xml = DiasporaFederation::Salmon::EncryptedSlap.generate_xml(@remote_user.diaspora_handle,
```
diaspora-code-review: Useless assignment to variable - `xml`.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+      OpenSSL::PKey::RSA.new(@remote_user.encryption_key),
```
diaspora-code-review: Align the parameters of a method call if they span more than one line.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+      entity,
```
diaspora-code-review: Align the parameters of a method call if they span more than one line.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+      OpenSSL::PKey::RSA.new(@user.encryption_key)
```
diaspora-code-review: Align the parameters of a method call if they span more than one line.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+    )
```
diaspora-code-review: Align `)` with `(`.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  before do
+    @user = alice
+    allow(User).to receive(:find){ |id|
```
diaspora-code-review: Space missing to the left of {.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+      if id == @user.id
+        @user
+      else
```
diaspora-code-review: Redundant `else`-clause.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+    expect(@user.contacts.count).to eq(2)
+    new_contact = @user.contacts.last
+    expect(new_contact).not_to be_nil
+    expect(new_contact.sharing).to eq(true)
+    expect(new_contact.person.diaspora_handle).to eq(@remote_user.diaspora_handle)
+  end
+
+  it "doesn't save the status message if there is no sharing" do
+    entity = FactoryGirl.build(:status_message_entity,
+      diaspora_id: @remote_user.diaspora_handle,
```
diaspora-code-review: Align the parameters of a method call if they span more than one line.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+      public: false
```
diaspora-code-review: Align the elements of a hash literal if they span more than one line.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+    )
```
diaspora-code-review: Align `)` with `(`.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+    before do
+      @author = Person.find_or_fetch_by_identifier(@remote_user.diaspora_handle)
+      contact = @user.contacts.find_or_initialize_by(person_id: @author .id)
+      contact.sharing = true
+      contact.save
+    end
+
+    it "treats status message receive correctly" do
+      entity = FactoryGirl.build(:status_message_entity,
+        diaspora_id: @remote_user.diaspora_handle,
```
diaspora-code-review: Align the parameters of a method call if they span more than one line.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+        public: false
```
diaspora-code-review: Align the elements of a hash literal if they span more than one line.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+      )
```
diaspora-code-review: Align `)` with `(`.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,93 @@
...
+
```
diaspora-code-review: Trailing whitespace detected.

cmrd-senya:
This new version of the patch utilizes the latest changes in the diaspora_federation gem and does an integration test on comments receive. More tests to come.

cmrd-senya:
@jhass, this contains fix to the security issue with signatures we discovered before. I put it here, because here I test it using the integration tests.

I propose to merge it all together, what do you think?

There is to major things to do, before we can merge it:
- [ ] Do something to factories.rb in this patch
- [ ] Release the diaspora_federation gem with the latest changes

factories.rb here merged from the diaspora_federation gem and has some changes.

cmrd-senya:
@jhass, this contains fix to the security issue with signatures we discovered before. I put it here, because here I test it using the integration tests.

I propose to merge it all together, what do you think?

There is to major things to do, before we can merge it:
- [ ] Do something to factories.rb in this patch
- [ ] Release the diaspora_federation gem with the latest changes

factories.rb here merged from the diaspora_federation gem and has some changes. I think it is bad idea to duplicate code, so it would be nice if we could managed to reuse it somehow. Maybe we can ship this factory definitions with the diaspora_federation gem? What do you think, @SuperTux88?

config/initializers/diaspora_federation.rb
```diff
@@ -61,5 +35,26 @@
 
       person_entity.save!
     end
+
+    on :fetch_private_key_by_id do |diaspora_id|
+      Person.find_by(diaspora_handle: diaspora_id).owner.encryption_key
```
jhass: We shouldn't need to allocate AR objects here, the right combination of `pluck` and `joins` should do it.

config/initializers/diaspora_federation.rb
```diff
@@ -61,5 +35,26 @@
...
+    end
+
+    on :fetch_private_key_by_post_guid do |guid|
+      author_user = Post.find_by(guid: guid).author.owner
+      author_user.encryption_key unless author_user.nil?
```
jhass: Same.

config/initializers/diaspora_federation.rb
```diff
@@ -61,5 +35,26 @@
...
+    on :fetch_public_key_by_id do |diaspora_id|
+      Person.find_by(diaspora_handle: diaspora_id).author.public_key
+    end
+
+    on :fetch_public_key_by_post_guid do |guid|
+      Post.find_by(guid: guid).author.public_key
+    end
+
+    on :post_author_is_local? do |guid|
+      Post.find_by(guid: guid).author.local?
```
jhass: Is the post guid really all the information we have at this point in the gem? I think this hook is highly confusing, I'd prefer much much more to yield the diaspora id of the author. Actually same applies to the other two hooks.

spec/integration/federation/factories.rb
```diff
@@ -0,0 +1,219 @@
+require "factory_girl"
+
+def r_str
+  SecureRandom.hex(3)
+end
```
jhass: I don't think extracting this simple call into a method is doing any good.

spec/integration/federation/factories.rb
```diff
@@ -0,0 +1,219 @@
...
+#
+# Sort hash according to an entity class's property sequence.
+# This is used for rspec tests in order to generate correct input hash to
+# compare results with.
+#
+def sort_hash(data, klass)
+  klass.class_props.map { |prop|
+    [prop[:name], data[prop[:name]]] unless data[prop[:name]].nil?
+  }.compact.to_h
+end
```
jhass: The name of the method feels a bit too generic for this project. Also note that hashes compare order independent, `{a: 1, b: 2} ==  {b: 2, a: 1 } #=> true`

spec/integration/federation/factories.rb
```diff
@@ -0,0 +1,219 @@
...
+
+def relayable_attributes_with_signatures(entity_type)
+  DiasporaFederation::Entities::Relayable.update_signatures!(
+    sort_hash(FactoryGirl.attributes_for(entity_type), FactoryGirl.factory_by_name(entity_type).build_class)
+  )
+end
```
jhass: The method seems unused?

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
+require "spec_helper"
+require File.dirname(__FILE__) + "/factories"
```
jhass: Rather than this, place the factories into the correct folder, `spec/support/federation_integration_factories.rb` or something.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+  )
+end
+
+shared_examples_for "it deals correctly with a relayable" do
+  before do
+    allow(DiasporaFederation.callbacks).to receive(:trigger)
+                                              .with(
+                                                :save_person_after_webfinger,
+                                                kind_of(DiasporaFederation::Entities::Person)
+                                              ).and_call_original
```
jhass: I'm not sure why you do this. `allow` doesn't raise if the call doesn't happen and you don't return a mock.

spec/integration/federation/receive_federation_messages.rb
cmrd-senya: If it is possible, I'd prefer to use factories shipped from the diaspora_federation gem, instead of copying it here.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+      diaspora_id:      remote_user.person.diaspora_handle,
+      author_signature: "I'm a wrong signature"
+    )
+
+    expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                              .with(:fetch_private_key_by_post_guid, kind_of(String))
+                                              .and_call_original
+
+    Workers::ReceiveEncryptedSalmon.new.perform(user.id, generate_xml(entity, remote_user, user))
+    expect(klass.find_by(guid: entity.guid)).to be_nil
```
jhass: `expect(klass.exists?(...).to be_false` might be a bit clearer.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+
+    expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                              .with(:fetch_private_key_by_id, remote_user.person.diaspora_handle)
+                                              .once
+                                              .and_return(remote_user.encryption_key)
+
+    Workers::ReceiveEncryptedSalmon.new.perform(user.id, generate_xml(entity, remote_user, user))
+    expect(klass.find_by(guid: entity.guid)).to be_nil
+  end
+end
```
jhass: This should probably go into its own file.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+
+describe Workers::ReceiveEncryptedSalmon do
+  before do
+    @user = alice
+    allow(User).to receive(:find) { |id|
```
jhass: This again doesn't really do anything.

spec/integration/federation/receive_federation_messages.rb
cmrd-senya: When DiasporaFederation.callbacks is called with some arguments which weren't specified by expect/allow it fails, that's why. I put expect for a different callback, and because of this all other original callbacks fail to call.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+      .to receive(:hcard) {|instance|
+        [@remote_user, @remote_user2].find {|user| user.diaspora_handle == instance.diaspora_id }.person.hcard
+      }
+  end
+
+  it "treats sharing request recive correctly" do
+    entity = FactoryGirl.build(:request_entity, recipient_id: @user.diaspora_handle)
+
+    allow(Diaspora::Fetcher::Public).to receive(:queue_for)
+    expect(Diaspora::Fetcher::Public).to receive(:queue_for).exactly(1).times
```
jhass: No need for the `allow` here.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+
+    Workers::ReceiveEncryptedSalmon.new.perform(@user.id, generate_xml(entity, @remote_user, @user))
+
+    expect(@user.contacts.count).to eq(2)
+    new_contact = @user.contacts.last
```
jhass: Not guaranteed to be stable, add an `order` clause.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+      entity = FactoryGirl.build(
+        :status_message_entity,
+        diaspora_id: @remote_user.diaspora_handle,
+        public:      false
+      )
+
+      Workers::ReceiveEncryptedSalmon.new.perform(@user.id, generate_xml(entity, @remote_user, @user))
+
+      new_status_message = StatusMessage.find_by(guid: entity.guid)
+      expect(new_status_message).not_to be_nil
```
jhass: No real need for the local and again just use `exists?`

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+    entity = FactoryGirl.build(
+      :status_message_entity,
+      diaspora_id: @remote_user.diaspora_handle,
+      public:      false
+    )
+
+    Workers::ReceiveEncryptedSalmon.new.perform(@user.id, generate_xml(entity, @remote_user, @user))
+
+    new_status_message = StatusMessage.find_by(guid: entity.guid)
+    expect(new_status_message).to be_nil
```
jhass: Ditto.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,224 @@
...
+        @local_message = FactoryGirl.create(:status_message, author: @user.person)
+        @remote_message = FactoryGirl.build(
+          :status_message_entity,
+          diaspora_id: @remote_user.diaspora_handle,
+          public:      false
+        )
+        Workers::ReceiveEncryptedSalmon.new.perform(@user.id, generate_xml(remote_message, @remote_user, @user))
+      end
+
+      it_behaves_like "it deals correctly with a relayable" do
```
jhass: Let's wrap these into `context "with comment"` etc.

spec/integration/federation/receive_federation_messages.rb
jhass: So this should still be an `expect`?

cmrd-senya: I just checked, it works without it now, because I use the same remote user everywhere. Earlier it was different.

config/initializers/diaspora_federation.rb
cmrd-senya: Well, yes, we know only parent post guid. Look at the [comment structure](https://github.com/SuperTux88/diaspora_federation/blob/salmon/lib/diaspora_federation/entities/relayable.rb). Instead I can propose to add a hook ```fetch_diaspora_id_by_post_guid``` then and call two hooks consequently to receive information about parent author.

spec/integration/federation/receive_federation_messages.rb
```diff
@@ -0,0 +1,216 @@
...
+                                              .once
+                                              .and_return(remote_user.encryption_key)
+    expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                              .with(:fetch_private_key_by_post_guid, kind_of(String))
+                                              .and_call_original
+
+    Workers::ReceiveEncryptedSalmon.new.perform(user.id, generate_xml(entity, remote_user, user))
+    received_entity = klass.find_by(guid: entity.guid)
+    expect(received_entity).not_to be_nil
+    expect(received_entity.author.diaspora_handle).to eq(remote_user.person.diaspora_handle)
```
cmrd-senya: @jhass, do I need to avoid AR object here?

jhass: I'd say it's okay if you access multiple properties. It's more critical in non test code anyway.

config/initializers/diaspora_federation.rb
jhass: Mmh, doing two queries is worse, so I guess we keep it for now.

cmrd-senya:
I'm kinda finished, it's left to make a release of the diaspora_federation.

It covers only a few cases of federation receive, but I think it's better to merge it and then add more tests in further commits, so we don't stay with the opened vulnerability.

config/initializers/diaspora_federation.rb
```diff
@@ -61,5 +35,30 @@
...
+    end
+
+    on :fetch_public_key_by_post_guid do |guid|
+      key = Post.where(guid: guid).joins(:author).pluck(:serialized_public_key).first
+      OpenSSL::PKey::RSA.new key unless key.nil?
+    end
+
+    on :post_author_is_local? do |guid|
+      id_array = Post.where(guid: guid).joins(:author).pluck(:owner_id)
+      !id_array.first.nil? unless id_array.empty?
```
jhass: Wouldn't something like `Post.joins(:author).where(guid: guid).not(owner_id: nil).exists?` work?

spec/integration/federation/shared_receive_relayable.rb
```diff
@@ -0,0 +1,94 @@
...
+      parent_author_signature: "I'm a wrong signature"
+    )
+
+    expect(DiasporaFederation.callbacks).to receive(:trigger)
+                                              .with(:fetch_private_key_by_id, remote_user.person.diaspora_handle)
+                                              .once
+                                              .and_return(remote_user.encryption_key)
+
+    Workers::ReceiveEncryptedSalmon.new.perform(user.id, generate_xml(entity, remote_user, user))
+    expect(klass.exists?(guid: entity.guid)).to be(false)
```
jhass: `be_true` and `be_false` should be things.

config/initializers/diaspora_federation.rb
cmrd-senya: Look at the [spec](https://github.com/cmrd-senya/diaspora/commit/cc821e19744ef57a8155763bb2e5bee06f0b2699#diff-89118032a5e2b114d8286b81e0db3b02R235). There is three different cases, and I made it to return nil if guid is not found. With what you offered it would be false in that case, which isn't true strictly.

jhass: Then this shouldn't be a predicate (`foo?`), they should only return `true` or `false`, but rather `post_author_type` with the possible outcomes `:post_not_found`, `:local`, `:remote` and theoretically `:author_not_found`. The third possibility would be to raise in that case.

I think I do prefer to consider the post not found case `false`.

cmrd-senya: Ok, that would be simpler then.

spec/integration/federation/shared_receive_relayable.rb
cmrd-senya: When I try this instead I get [the following](https://gist.github.com/cmrd-senya/1c1b1d880682ae471b88).

jhass: Meh, nvm then I guess.

cmrd-senya:
I realized that some of situations tested here are artificial. In real life, it's not possible, or just doesn't make sense, that sender makes right signature for Salmon and then fails to make a signature for the payload using the same key. Thus, I'll rewrite tests in consideration that we only test situations where the both signatures (Salmon and in-payload, like in relayables) are correct or both signatures are wrong.

Flaburgan:
Also consider the situation of a malicious user and pod owner ;)diaspora/diaspora a écritI realized that some of situations tested here are artificial. In real life, it's not possible, or just doesn't make sense, that sender makes right signature for Salmon and then fails to make a signature for the payload using the same key. Thus, I'll rewrite tests in consideration that we only test situations where the both signatures (Salmon and in-payload, like in relayables) are correct or both signatures are wrong.

—Reply to this email directly or view it on GitHub.


  
  


Envoyé depuis Firefox OS

cmrd-senya:
Sure, it is considered, it is the only reason why we need signatures at
all :)

ATM we have a situation where entity in payload, for example comment
gets signed by the key of a user, and then packed in Salmon, and Salmon
gets signed with the same key again. There isn't much sense to check
cases, when one signature is ok and the other is broken (moreover, we
probably just don't need these duplicating signatures at all, see #6566).

The fun is that the only case when a signature in the entity is in fact
useful
([this](https://github.com/diaspora/diaspora/pull/6539/files#diff-a1ee075c06bc50b99a8bc09ea96aae48R61)
is the one where we don't check it now :)

I guess that there maybe was some idea behind it, like they wanted, for
example, to make pods exchange messages in P2P style, like pod A posted
a status, pod B fetched the status and resent it to pod C, which resent
it to pod D. That would explain why they duplicated signatures in the
payloads but not only in signed envelopes, and that would explain why
they wanted to keep signatures in the database. But we don't do that, so
most of this signatures code is redundant for us.

cmrd-senya:
Please, review.

cmrd-senya:
diaspora_federation 0.0.9 is released, I guess this one is therefore ready.

jhass:
Merged as 4e41b8dc86dceaf43c154e718969e87de3c1a0f6 08b910bd8851839745cdf41fb6d78c9fd8e3baef 922d26f9768b206127b13642f476ca959e9e2dc9 f0fc62e94d27d27e9fc6bc48455b415d555f5922

Thank you!

cmrd-senya:
Wow, cool!

PR #6549, opened 2015-11-16
Using original size for pictures in gallery on desktop view

ghost:
The counterpart of abandonning `stretchImages` option in  #6536 is that pictures are now really small in gallery. I think the gallery on desktop view should display the photo in its original size.

### DIFF
--- a/app/assets/templates/photo_tpl.jst.hbs
+++ b/app/assets/templates/photo_tpl.jst.hbs
@@ -21,7 +21,7 @@
     {{/if}}
 
     <div class="thumbnail img-thumbnail">
-      <a href="{{sizes.large}}" class="photo-link gallery-picture">
+      <a href="{{sizes.original}}" class="photo-link gallery-picture">
         <img src="{{sizes.large}}" class="photo big_photo">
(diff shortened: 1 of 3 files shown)


AugierLe42e:
I've open a Loomio thread to discuss this PR: https://www.loomio.org/d/4U2vwQgW/ideas-regarding-pictures

AugierLe42e:
Ping?

svbergerem:
I agree that the large thumbnail is too small in some cases but the original image could be huge and the user won't be able to find out before they open the gallery.

AugierLe42e:
So a solution could be to add another image size.

svbergerem:
Someone on the IRC mentioned that fixing #6649 first could be a nice compromise. Then users are still able to see the large thumbnail and after clicking on the big image they will see the original size.

svbergerem:
I just checked this again and we don't use the large thumbnails, but `scaled_full` which is resized to `[700, nil]`. I think that's good enough and will close this PR. Whoever disagrees with me can still open a new issue.

PR #6552, opened 2015-11-18
Do not add participation for comment if comment validation failed

manuelVo:
Previously the server created a participation for a comment before validating the comment itself. Thus if a client sent an invalid comment to the server (e.g. consists only of spaces) a participation would be created while the comment was refused. This pull request changes the order of the operations so that comment is validated before the participation gets created.

I also noticed that the other social actions had the same odd behavior so i changed them too.


jhass:
Merged as 0925a26506ae58c49af72258316ec45d0824915c

Good catch, thank you!

Flaburgan:
Thank you!

PR #6572, opened 2015-12-01
sexe -> genre

stragu:
"sexe" is too specific and refers only to the biological trait; it does not include the cultural and social aspect of the English term "gender".


jhass:
Thank you! Unfortunately we can't accept this pull request due to technical reasons. Please have a look at https://wiki.diasporafoundation.org/Contribute_translations

stragu:
Sure, thank you for the feedback, @jhass ! I will change it at WebTranslateIt. Cheers

Issue #4826, opened 2014-03-04
Incorrect time stamp in notifications


goobertron:
I've noticed a case in which the timeago stamp in notifications is incorrect.

The case is that in which two or more people have commented on a post, and one or more of them has commented more than once since you last looked at your account. In this case, the time stamp is the first instance in which the last of those people commented (i.e. the oldest time at which all of the people mentioned in the notification commented).

For example,

Fred and Ginger have had a conversation on your post since you last looked in:

Fred commented 15 hours ago
Ginger commented 14 hours ago
Fred commented 8 hours ago
Ginger commented 3 hours ago
Fred commented 2 hours ago.

The notification will read 'Fred and Ginger commented on your post 14 hours ago'. It would (to me) make more sense if it gave the time of the most recent comment, i.e. '2 hours ago' in this instance.

Have noticed this on two separate posts (and two separate notifications), so it's not a one-off.

PR #6573, opened 2015-12-01
Fix_notifications_timestamp_issue #4826

apsc92:
- Showing updated_at instead of created_at to show latest time.
  For e.g there is a 'like' notification and another likes the same post then time shown should be the time when second person liked the post.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -64,6 +64,8 @@ private
       begin
         n.actors = n.actors | [actor]
         n.unread = true
+        # Explicitly touch the notification to update updated_at whenever new actor is inserted in notification.
+        n.touch
         n.save!
       rescue ActiveRecord::RecordNotUnique
         nil
--- a/app/views/notifications/_notification.haml
+++ b/app/views/notifications/_notification.haml
@@ -10,4 +10,4 @@
   .media-body
     = notification_message_for(note)
     %div
-      = timeago(note.created_at)
+      = timeago(note.updated_at)
--- a/spec/controllers/notifications_controller_spec.rb
+++ b/spec/controllers/notifications_controller_spec.rb
@@ -46,7 +46,7 @@ describe NotificationsController, :type => :controller do
   describe '#index' do
     before do
       @post = FactoryGirl.create(:status_message)
-      FactoryGirl.create(:notification, :recipient => alice, :target => @post)
+      @notification = FactoryGirl.create(:notification, recipient: alice, target: @post)
     end
 
     it 'succeeds' do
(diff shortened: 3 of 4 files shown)


app/views/notifications/_notification.haml
```diff
@@ -10,4 +10,4 @@
   .media-body
     = notification_message_for(note)
     %div
-      = timeago(note.created_at)
+      = timeago(note.updated_at) # To show latest time
```
jhass: I don't think we need the comment here :)

app/models/notification.rb
```diff
@@ -64,6 +64,8 @@ def self.concatenate_or_create(recipient, target, actor, notification_type)
       begin
         n.actors = n.actors | [actor]
         n.unread = true
+        # I have to specially touch it to update updated_at whenever new actor is inserted in notification.
```
jhass: Don't use first person in comments. Either "We need to explicitly ..." or just  "Explicitly touch the notification ..."

app/models/notification.rb
```diff
@@ -64,6 +64,8 @@ def self.concatenate_or_create(recipient, target, actor, notification_type)
...
+        n.touch
```
jhass: That seems a bit odd, we gonna need a test that fails without this line.

jhass:
We're going to need at least two specs for this, one ensuring `updated_at` is used in the template and as already mentioned one ensuring `updated_at` is updated when amending a notification.

spec/controllers/notifications_controller_spec.rb
```diff
@@ -46,7 +46,7 @@
   describe '#index' do
     before do
       @post = FactoryGirl.create(:status_message)
-      FactoryGirl.create(:notification, :recipient => alice, :target => @post)
+      @notification = FactoryGirl.create(:notification, :recipient => alice, :target => @post)
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

spec/controllers/notifications_controller_spec.rb
```diff
@@ -56,8 +56,15 @@
     end
 
     it 'succeeds for notification dropdown' do
+      Timecop.travel(6.seconds.ago) do
+        @notification.touch
+      end
       get :index, :format => :json
       expect(response).to be_success
+      note_html = JSON.parse(response.body)[0]['also_commented']['note_html']
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/controllers/notifications_controller_spec.rb
```diff
@@ -56,8 +56,15 @@
...
+      note_html = Nokogiri::HTML(note_html)
+      timeago_content = note_html.css('time')[0]["data-time-ago"]
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

apsc92:
@jhass I have added some test cases in this pull request.Please have a look.

jhass:
Merged as fdb3ae5cb8da29dd2f4c4605b50756cfc8f4c2d9

Thank you!

Flaburgan:
Thank you!

PR #6586, opened 2015-12-12
Remove parent author signature for relayables from the DB

### LINKED ISSUE #4920: Database size control

cmrd-senya:
Remove author and parent author signatures for relayables from the DB since they are considered redundant.

This not strictly closes #4920, but makes it less tense.

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -21,11 +21,6 @@ class Message < ActiveRecord::Base
                    # inside, which would cause an infinite recursion
     #sign comment as commenter
     self.author_signature = self.sign_with_key(self.author.owner.encryption_key) if self.author.owner
-
-    if self.author.owns?(self.parent)
-      #sign comment as post owner
-      self.parent_author_signature = self.sign_with_key(self.parent.author.owner.encryption_key) if self.parent.author.owner
-    end
     self.save!
     self
   end
--- a/db/migrate/20151210213023_remove_signatures_from_relayables.rb
+++ b/db/migrate/20151210213023_remove_signatures_from_relayables.rb
@@ -1,0 +1,9 @@
+class RemoveSignaturesFromRelayables < ActiveRecord::Migration
+  def change
+    remove_column :comments, :parent_author_signature, :text
+    remove_column :poll_participations, :parent_author_signature, :text
+    remove_column :messages, :parent_author_signature, :text
+    remove_column :participations, :parent_author_signature, :text
+    remove_column :likes, :parent_author_signature, :text
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151003142048) do
+ActiveRecord::Schema.define(version: 20151210213023) do
 
   create_table "account_deletions", force: :cascade do |t|
     t.string   "diaspora_handle", limit: 255
@@ -92,7 +92,6 @@ ActiveRecord::Schema.define(version: 20151003142048) do
     t.integer  "author_id",               limit: 4,                      null: false
     t.string   "guid",                    limit: 255,                    null: false
     t.text     "author_signature",        limit: 65535
-    t.text     "parent_author_signature", limit: 65535
     t.datetime "created_at",                                             null: false
     t.datetime "updated_at",                                             null: false
     t.integer  "likes_count",             limit: 4,     default: 0,      null: false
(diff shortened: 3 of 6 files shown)


cmrd-senya:
I have no idea how to run a migration so that it doesn't put unrelated shit to the db/schema.rb. Suggestions welcome.

SuperTux88:
As I already said in https://github.com/diaspora/diaspora/issues/4920#issuecomment-163794165:

You need the `author_signature` column to store the signature if you are the author of the parent, because you need to federate this to other pods. You can store `NULL` if you are not the author of the parent, but you need the column.

cmrd-senya:
Hmm, I suppose this is true only for a situation when we have to resend a relayable after a period. Because if we resend it immedeately we still have the signature federated from the other pod.

SuperTux88:
Yes, but I think an author of a post should always be able to resend the whole thread. Currently there is only fetching of posts, but I think it is a good idea to also add the comments to this posts. This is not possible if you delete this info now, because if its gone, its gone, you can not recover them.

cmrd-senya:
Fair enough, I'll bring author_signature back

cmrd-senya:
Here it is.

cmrd-senya:
@SuperTux88, I thought, that if we want to fetch comments we could ask the original pod to resend the signature. This way we could reduce the database size in a price of increasing traffic volume between pods. We could also implement some caching and set maximal cache size. In case we don't have author signature cached we refetch it from the original pod. That is complex, but probably the best compromise.

cmrd-senya:
Please tell me, how do I make a sane ```db/schema.rb``` file?

jhass:
You would have to remigrate from a clean MySQL DB. Since that's annoying I just `git commit -p` it most of the time :/

SuperTux88:
@cmrd-senya I still think we should keep the `author_signature` if we are the owner of the post, because it would make the fetching so much more complex and prone to errors if we need to fetch from many pods. You can still set if to `NULL` if you are not the owner of the parent, which should save a lot space for small pods.

Zauberstuhl:
Hi, I am currently working with @maxwell on the JD migration and this PR would be pretty important for us.

Please correct me if I am wrong..

* We can get rid of `parent_author_signature` without second thoughts
* Keep `author_signature` could come in handy at some point in time

Cheers,
Lukas

SuperTux88:
Yes, `author_signature` shouldn't be deleted if the parent (the post) is from a user from JD. But you can delete the `author_signature` for all relayables where you are **not** the owner of the parent.

denschub:
Ping?

cmrd-senya:
That's ready for maintainer's review.

denschub:
I cannot make out any issues and the tests are green. Let's see if we break stuff.

Flaburgan:
Did someone apply this patch in a production pod yet? Any feedback?

Flaburgan:
So, I applied this on d-fr, fyi, the pg_dump before was 12273534975 and now is 7279783047. Nice size improvement!

Issue #6520, opened 2015-10-22
warn user if he leaves profile editing page with unsaved changes


kas70:
##

SuperTux88:
Fixed in #7473

PR #6589, opened 2015-12-15
Warn user if he leaves profile editing page #6520

atulonruby:


svbergerem:
I think a cucumber test for the new behavior would be great.

denschub:
Ping.

PR #6597, opened 2015-12-23
Adjust stream loading

lu-x:
Don't reload the whole site, when clicking 'Stream' or 'My Activity' in the Header.


svbergerem:
I already implemented that and we reverted the change, since the notification and conversation counts don't update anymore if you don't reload the whole website. We need to implement asynchronous updates for those first.

jaywink:
Yep sorry, the pull is #6493 - still waiting for me to write tests :(

denschub:
Agreed, this already will be done in #6493, but thank you for your work!

PR #6585, opened 2015-12-10
Update gems

jhass:
#3855

PR #6600, opened 2015-12-26
Bump omniauth-facebook to latest version (v3.0.0)

balasankarc:
I was able to post to Facebook with the latest version of omniauth-facebook (v3.0.0)


denschub:
This Gem uplift is already included in #6585, sorry.

balasankarc:
Ah, ok. I missed that. My bad.

PR #6615, opened 2016-01-03
Wait for ajax to complete before checking for posts in the stream

manuelVo:
Previously cukes using the line "Then I should not see any posts in my stream" would always succeed. Because the assert_selector's condition is already fulfilled while the stream is loading the command doesn't wait for the stream to load on its own leading to immediate success of the condition (wether the stream is empty or not).

This PR introduces a loop waiting for all ajax requests to complete (and thus the stream to load) before checking if any posts are in the stream.

### DIFF
--- a/features/step_definitions/posts_steps.rb
+++ b/features/step_definitions/posts_steps.rb
@@ -15,6 +15,7 @@ Then /^I should not see an uploaded image within the photo drop zone$/ do
 end
 
 Then /^I should not see any posts in my stream$/ do
+  page.assert_selector("#paginate .loader", visible: :hidden)
   page.assert_selector(".stream_element", count: 0)
 end
 


jhass:
I'm not particularly fond of this. We had a construct like it previously and the cukes were littered with calls to it everywhere, instead of properly waiting for DOM changes using Capybara's inbuilt mechanism. Ridding the suite of it was a lot of effort but actually increased the stability in the end.

Can't we assert for the spinner to disappear or something?

manuelVo:
The test now waits for the spinner to disappear (instead of the ajax call to complete)

jhass:
https://travis-ci.org/diaspora/diaspora/jobs/100001042#L3054 :)

manuelVo:
I already noticed that ;)

Seems like my assumption the second method would have the same bug (with the same solution) was wrong. I already updated the commit (the commit id didn't change though - thats the reason there is no note about that here).

jhass:
Merged as a4257a70e5491f70d85b39ae2bb532824a7a3108

Thanks!

Issue #6538, opened 2015-11-09
Replace Sidetiq


jhass:
Sadly it looks like Sidetiq turned into abandonware and is preventing Sidekiq upgrades now, in particular it it will prevent upgrading to the highly anticipated Sidekiq 4.0, which drops the Celluloid dependency.

So we need to find and adopt a replacement.

Candidates:

* [sidekiq-dejavu](https://github.com/felixbuenemann/sidekiq-dejavu)
* [sidekiq-cron](https://github.com/ondrejbartas/sidekiq-cron)
* [sidekiq-repeat](https://github.com/FlavourSys/sidekiq-repeat)
* [sidekiq-scheduler](https://github.com/Moove-it/sidekiq-scheduler)

AugierLe42e:
Do you have propositions in mind?

jhass:
I didn't do an extensive research yet, else I would've included that.

AugierLe42e:
What about replacing Sidekiq itself? @jaywink talked about [goworkers](https://www.goworker.org/)?

Raven24:
@AugierLe42e That's not a simple drop-in replacement. It's a background-worker system implemented in Go, where we can't re-use any of the existing ruby/rails models - we'd have to reimplement them in Go. I'm not saying that's necessarily a showstopper, but someone would have to do the porting of everything worker-related.

jaywink:
I don't think I proposed it, it was more like "hey look goworker" just like I say "hey look neat" about 100 other things during a week :)

Anyway, we could always just use crons + rake tasks :P That could be kept as one option, but would require additional configuration for podmins.

AugierLe42e:
@Raven24: From what I see on the web site, the workers can be called from ruby code. So only the workers would have to be rewritten, right? I don't know the amount of work this represent. Anyway, it's just some idea.

@jaywink : Hmm, I don't think this is a good idea to increase the amount of work for podmins :/

jhass:
Yeah, I would strongly prefer to not use cron.

AugierLe42e:
A quick research gave me three options:

 * [sidekiq-cron](https://github.com/ondrejbartas/sidekiq-cron),
 * [sidekiq-repeat](https://github.com/FlavourSys/sidekiq-repeat),
 * [sidekiq-scheduler](https://github.com/Moove-it/sidekiq-scheduler).

balasankarc:
GitLab uses sidekiq-cron without any issues I think.

[...]

PR #6616, opened 2016-01-04
Replace sidetiq with sidekiq-cron

cmrd-senya:
ref #6538


Flaburgan:
The upgrade of Sidekiq to the [really improved](http://www.mikeperham.com/2015/10/14/optimizing-sidekiq/) version 4 was blocked by the abandoned sidetiq. On the [readme](https://github.com/ondrejbartas/sidekiq-cron) of sidekiq-cron you can read "Sidekiq 4 or greater is required (for Sidekiq < 4 use version sidekiq-cron 0.3.1)". If the reason you choose to lock sidekiq-cron to 0.3.1 was because of Sidekiq 3.x used in diaspora*, then maybe you can try to directly use sidekiq-cron 0.4.2 and bump sidekiq to the last 4 version in the same time?

cmrd-senya:
I've bumped it to sidekiq 4.

The problem is that it works fine locally and passes all tests with ```bundle exec ./script/ci/build.sh```, but it fails on Travis :( I'm unsure how to deal with that.

denschub:
Sure, why not. I don't quite like the job initialization in the bottom of those class files. In my opinion, it would be a bit nicer to have a `config/schedule.yml` file as [shown in sidekiq-crons README](https://github.com/ondrejbartas/sidekiq-cron#adding-cron-job).

denschub:
I feel like the travis failure was a temporary issue with travis. We'll see if the tests are green when you push the changes.

cmrd-senya:
Done.

pravi:
@denschub anything more to be done?

denschub:
So, Sidekiq 4 removes the ability to use namespaces (which is [discouraged anyway](http://www.mikeperham.com/2015/09/24/storing-data-with-redis/)) and if you want to use namespaces, you need to add the `redis-namespace` gem now.

I tend to deprecating custom namespaces for the reasons mentioned in the blog above. I've [just publiched a post](https://pod.diaspora.software/posts/84064) asking our podmins if someone is using custom namespaces. If not, I'd simply remove that.

jaywink:
Well, some way to configure a pod to exist with another pod on the same server is needed - be it namespaces, databases or redis url. As far as we support that, good riddance namespaces. And AFAICT, pods can already [specify a separate redis instance or db](https://github.com/diaspora/diaspora/blob/develop/config/diaspora.yml.example#L52) - so no need for namespaces?

denschub:
I've merged this as ed61ef6d6a33b46be60ff3aaf21c010bd3f206a5. In addition, with 5f986464daa6b66550770c788cebc276749f1912, I've removed support for custom redis namespaces and added a Changelog entry.

Funny enough, `redis-namespaces` is a dependency of `sidekiq-cron`, but since there is support for up to 8 databases in Redis, there is no need to actually continue supporting this.

cmrd-senya:
Don't forget to close #6538

Flaburgan:
Cool! Thank you @denschub and @cmrd-senya!

jaywink:
Awesome! So, I've got the `pod.diaspora.software` pod running on a custom namespace. I guess the transition is?

* Run down sidekiq queue (if any)
* Change setting to point to a separate redis db
* Reboot

denschub:
That is correct.

jaywink:
Well, `environment.redis` seems to work, was a bit nervous updating :P

selection_214

*so empty*

jaywink:
I have no idea whether this is related, but since I upgraded the HQ pod, sidekiq RAM usage has been crazy, between 12G and 15G (see post https://iliketoast.net/posts/892040#80744b20b1f0013389526c626dd55703).

I did no other changes to the settings except change the redis url. Sidekiq version is 4.1.0. My other pod, iliketoast.net, runs stable at ~2GB using the old sidekiq.

> The pod runs on 0.5.99.0. The pods current commit is https://github.com/diaspora/diaspora/commit/0c05b474c5563adfca75455957379b061413c374. The last update was around 2016/02/05 10:30 UTC.

I restarted it approx 10 hours ago and it is now "only" at 4.6G on a 5 concurrency. Before the restart it was at 12G, the pattern has been ongoing a few times now.

Flaburgan:
dfr is running with 24GB of RAM, I upgraded to sidekiq 4 the 30th of Januray, I also do see a change in memory usage:

capture du 2016-02-10 08-49-53

No settings were changed at all.

dadall:
I restart diaspote every night but I've seen no changes in memory usage.

jaywink:
@Flaburgan did you do a restart when the RAM drops between the 5th-6th and 9th-10th?

Sounds like there is some kind of issue going on if so.

Flaburgan:
@jaywink yes I restarted to upgrade to the last develop pod. To be sure to have nothing left, I usually go with the hard way: `killall ruby` :p

denschub:
Which ruby version are you guys using?

jaywink:
ruby-2.2.1 [ x86_64 ] for both pods, from rvm (both just take whatever is in .ruby-version at the time).

denschub:
Just for the sake of having a clean test environment, could you please upgrade to 2.2.4? Not that it should matter (well, except for the fact that 2.2.2 was an openssl-security release), but I'd like to have that sorted out.

Flaburgan:
I use ruby-2.2.4 myself.

jaywink:
Decreased concurrency to 3, turned off stacktrace and both of the debug logging settings - still going up to 12G within a few days.

selection_220

And before you say, no I haven't tried 2.2.4 yet, will try when time.

cmrd-senya:
Could some kind of memory usage profiling help there to track down the issue?

Flaburgan:
@cmrd-senya what can I provide to you other than the graph above?

jaywink:
I already rolled back to a commit before sidekiq 4 - the pod ate my server, which means it used not only approx 16G ram, but 12G swap too.

Will try again at some point in the future, maybe with daily restarts..

jaywink:
Yeah and rolling back did stabilize the memory usage, has been between 2-4GB, with 5 concurrency. So either sidekiq 4 or some other bumped gem maybe, has caused something to leak badly.

Flaburgan:
Here is an updated graph:
capture du 2016-02-24 14-54-31

I restarted the server quite often because of the new changes in develop. I'll try to not restart it for a week or two and see what's happening.

CSammy:
@Flaburgan Interestingly enough, this looks not like a traditional _leak_ since it appears to stabilize at some point, though this point is nowhere near acceptable when compared to previous performance. I know you want to run the Pod for a week to take a look at the long-time performance, but would you like to try this (or have you already, and what were your findings)? [Profile memory usage](https://gist.github.com/krasnoukhov/a036bb7bc8c8416e9481)
The [Troubleshooting](https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting) page of sidekiq indicates this issue may be due to unoptimized AR use which in my opinion, after having seen the database and related code, is not unlikely. Maybe some corner case is showing after the update.

@jaywink It's good to know that reverting to a previous commit makes the issue disappear. The [sidekiq Changelog for version 4.0](https://github.com/mperham/sidekiq/blob/master/4.0-Upgrade.md#upgrade) states that we did not go the recommended way of updating to version >4. Would you be willing to try the recommended way? If not (which I'd completely understand), what would be the recommended way to test this? Apparently, version <4 gives deprecation warnings which might help to track down problems.

CSammy:
@Flaburgan Just a hunch: Did you make use of the namespace feature of redis before?

cmrd-senya:
I tried injecting the profiling code as stated in https://gist.github.com/krasnoukhov/a036bb7bc8c8416e9481.

Here are my changes: https://github.com/cmrd-senya/diaspora/commit/c32809956224804930615df8055c25b83ded2bc9.

And that lead the ruby interpreter to segfault on running our rspec tests!

[Here](https://gist.github.com/cmrd-senya/23d616ab2aeb6b79cbbf) is the output of ```bundle exec rspec```.

It doesn't happen everytime, probably depends on the random seed. There was no segfault with 9169 and there was with 35881.

Flaburgan:
I can tell you, sidekiq is killing my server:
htop

capture du 2016-03-01 10-30-30
capture du 2016-03-01 10-30-48

So, no need to let it go forward anymore, it just fulled everything. I'm restarting the server.

denschub:
@Flaburgan please read and follow @CSammy's advise...

cmrd-senya:
Hmm, I can't get the profiler to generate ```heap.json```. Look like the [call](https://github.com/cmrd-senya/diaspora/commit/c32809956224804930615df8055c25b83ded2bc9#diff-fd9e64e4a154ab77edea808fa4285141R35)  method from the Sidekiq middleware doesn't get called. I tried manually issuing cron jobs and doing profile fetching from the UI. What am I doing wrong?

denschub:
@cmrd-senya Did you start sidekiq with `PROFILE=1` set?

cmrd-senya:
@denschub, I ran it without ```if ENV["PROFILE"]``` clause: https://github.com/cmrd-senya/diaspora/commit/c32809956224804930615df8055c25b83ded2bc9

Issue #6584, opened 2015-12-10
Visibility of hidden post on profile author's page is impossible


juliaviluhina:
Steps
1. Hide some post
2. Post is not shown in stream
3. Go to the profile page of author of hidden post (It is  <pod>/people/<id>, isn't it?)

Expected result
according to https://wiki.diasporafoundation.org/FAQ_for_users#How_do_I_hide_a_post.3F_.2F_How_do_I_stop_getting_notifications_about_a_post_that_I_commented_on.3F
"You can still see the post if you visit the profile page of the person who posted it."

Actual result
I don't see hidden post in profile page of post author.

How can I see hidden post or cancel hiding post?

PR #6617, opened 2016-01-04
Show posts of ignored users on their profile page

manuelVo:
Currently the profile page of ignored users shows the text "You are ignoring all posts from %user". This gives the user no possibility to view the post of an ignored user except logging out.

This PR displays the posts of a user on the profile page even if he is ignored giving users the possibility to view the posts of a ignored user if they wish to do so.

### DIFF
--- a/app/assets/javascripts/app/pages/profile.js
+++ b/app/assets/javascripts/app/pages/profile.js
@@ -75,13 +75,6 @@ app.pages.Profile = app.views.Base.extend({
     if(!this.model.has("profile")){
       return false;
     }
-    if( this.model.isBlocked() ) {
-      $("#main_stream").empty().html(
-        '<div class="dull">'+
-        Diaspora.I18n.t("profile.ignoring", {name: this.model.get("name")}) +
-        "</div>");
-      return false;
-    }
 
     // a collection is set, this means we want to view photos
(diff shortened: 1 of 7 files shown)


Flaburgan:
I agree. I guess it would be nice to modify the posts rendering to not display the `hide` and `ignore` buttons though.

manuelVo:
The `hide`-button is needed to un-hide posts (at least when #6584 is fixed).

I agree about removing the `ignore` button but i think it should be done in a separate PR.

Flaburgan:
>The hide-button is needed to un-hide posts

Oh, right!

config/locales/javascript/javascript.br.yml
```diff
@@ -174,4 +173,4 @@ br:
       reshare: "Rannañ pelloc'h"
       reshared: "Rannet pelloc'h"
       stop_following_post: "Paouez da heuliañ an embannadenn-mañ"
-      unlike: "Displijus"
\ No newline at end of file
+      unlike: "Displijus"
```
jhass: Please never touch locales besides `en`.

manuelVo:
All non `en` locales are now removed from the commit

denschub:
Merged. Thanks!

PR #6402, opened 2015-09-14
Original white background - Color Theme

jpope777:
This adds a color theme which has the all white background similar to the previous D\* css.

PR #6631, opened 2016-01-09
White color theme

svbergerem:
Supersedes #6402.

white color theme

### DIFF
--- a/app/assets/stylesheets/color_themes/_color_theme_override_origwhite.scss
+++ b/app/assets/stylesheets/color_themes/_color_theme_override_origwhite.scss
@@ -1,0 +1,23 @@
+/* Raw CSS */
+body {
+  #left-navbar .hoverable { border-bottom: none; }
+
+  #left-navbar-fixed-sub-background { background: $sidebars-background; }
+  #left-navbar-fixed-background {
+    background: $sidebars-background;
+    border-right: 1px solid $sidebars-background;
+  }
+  #left-navbar {
+    background: $sidebars-background;
+    border-right: 1px solid $sidebars-background;
+  }
+  #right-sidebar-fixed-sub-background { background: $sidebars-background; }
+  #right-sidebar-fixed-background {
+    background: $sidebars-background;
+    border-left: 1px solid $sidebars-background;
+  }
+  .rightbar {
+    background: $sidebars-background;
+    border-left: 1px solid $sidebars-background;
+  }
+}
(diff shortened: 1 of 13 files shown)


Flaburgan:
Thanks for that! I'm sure many are not going to want the new design, to provide them an easy way to keep the old design is really needed.

Is it hard to re-add a left grey border to the stream? It looks a bit weird without it when you compare to the current 0.5 design.

Flaburgan:
Also, the background of the chat is still grey, I don't know if we can control that from the diaspora* code or if it's in JSXC.

svbergerem:
> Is it hard to re-add a left grey border to the stream? It looks a bit weird without it when you compare to the current 0.5 design.

Done and just updated the screenshot.

> Also, the background of the chat is still grey, I don't know if we can control that from the diaspora* code or if it's in JSXC.

Chat changes won't be part of this PR.

Flaburgan:
The border is still missing on the profile and tag page. Looks great otherwise, tested with both chromium and Firefox.

svbergerem:
> The border is still missing on the profile and tag page. 

Done.

app/assets/stylesheets/color_themes/_color_theme_override_origwhite.scss
```diff
@@ -0,0 +1,26 @@
...
+    border-bottom: 1px solid $border-grey;
+  }
+
+  .left-navbar .hoverable { border-bottom: 0; }
+
+  .profile-sidebar,
+  .left-navbar-fixed-background,
+  &.page-tags .left-navbar-fixed-background,
+  .left-navbar {
+    border-right: 1px solid $border-grey;
```
Flaburgan: You should put a border-left on the content instead. With that, the border stop at the end of the left elem which is usually pretty short (look at a profile page with only an avatar for example).

svbergerem: I tried that but it looked weird without changing some padding/margin of the stream / the profile header. I'm not planning to change this kind of stuff since these changes are much more likely to break stuff later when we adjust the original "theme". I'd also like to emphasis that we are talking about a "color theme" here.

jaywink:
Is this default?

svbergerem:
> Is this default?

No, it is the second color theme in the list. We have a new default design and this color theme is for those who don't like it.

> Ps. Some description would be nice in PR's ;)

What kind of information is missing? The title says that this is a color theme and the screenshot shows how it will look like. If there is anything missing then I'll add it to the description.

jaywink:
> No, it is the second color theme in the list. We have a new default design and this color theme is for those who don't like it.

Awesome :+1: 

> What kind of information is missing? The title says that this is a color theme and the screenshot shows how it will look like. If there is anything missing then I'll add it to the description.

Just a general note, sorry. I think it is good practise, towards other people, to give a description, for even obvious things, like whether this is default.

jhass:
LGTM, thank you. We can always touch it up with further PRs.

PR #6638, opened 2016-01-15
Removal of pre-filling message box - Do not merge. Not complete

ticoombs:
Upon requesting facebook integration for 'publish_actions' (As set out in the [dispora wiki](https://wiki.diasporafoundation.org/Integrating_other_social_networks) ) we actually break one of the rules which is pre-filling a message field. 

Facebook's policy example is here: https://developers.facebook.com/docs/apps/review/prefill
(While it basically says nothing)

How to produce this issue:
- Signup to a pod with a new account.
- Login with a new account
- Message box is now Pre-Filled with "Hello everyone, I'm #newhere"
- Example image:
  messagebox-example

My proposed fix is as follows:
- Replace the pre-filling of the post upon users first login, with an updated introduce_yourself h3.
- I removed the no-wrap so even on a smaller screen all text is viewable. 

messagebox-1920x

message

Thoughts?

### DIFF
--- a/app/assets/stylesheets/getting-started.scss
+++ b/app/assets/stylesheets/getting-started.scss
@@ -54,7 +54,6 @@
     color: $white;
     margin: 0;
     text-overflow: ellipsis;
-    white-space: nowrap;
   }
 
   #gs-skip-x {
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -243,7 +243,7 @@ en:
       people_sharing_with_you: "People sharing with you"
 
       welcome_to_diaspora: "Welcome to diaspora*, %{name}!"
-      introduce_yourself: "This is your stream.  Jump in and introduce yourself."
+      introduce_yourself: "This is your stream.  Why not let everyone know by creating a post with the #newhere tag to get started"
 
       new_here:
         title: "Welcome new users"
--- a/features/desktop/signs_up.feature
+++ b/features/desktop/signs_up.feature
@@ -39,23 +39,6 @@ Feature: new user registration
     Then I should be on the stream page
     And I close the publisher
 
-  Scenario: new user without any tags posts first status message
-    When I follow "awesome_button"
-    And I confirm the alert
-    Then I should be on the stream page
-    When I submit the publisher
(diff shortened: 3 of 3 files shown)


jhass:
What other changes are required to get into full compliance with Facebook?

ticoombs:
We also need to implement:

>For the login dialog, please use the JavaScript SDK without specifying an explicit display type. If you have special requirements precluding use of the SDK, set height and width on your window.open() call to properly size the dialog. People must have the complete option to change their audience settings. Learn more about [https://developers.facebook.com/docs/facebook-login/web](Facebook Login). 


Just realised I didn't update all other locale files that have the "new_user_prefill" variable being imported. As well as the other functions in `lib/stream/multi.rb` which actually do all the prefill variables (sorry for the pre-emptive PR, still getting my head around the codebase)

jhass:
No worries. Please only edit the English locales.

So we don't need to offer "sign in with Facebook"?

jaywink:
I don't feel good about bending over to facebook regarding removing something that is considered a feature for us - ie the prefilling of the newhere post. You could just blank the input field with a comment to the user when they click the Facebook icon - if they haven't actually typed anything yet.

ticoombs:
@jaywink Any pre-filling is technically a policy violation. Including clicking a prompt to auto pre-fill.

I just had a thought, if we change the pre-fill to only happen if the user has not auth'd via Facebook would this be an acceptable middle ground? (compared to removing all pre-fill)

Of course I'm happy to test and run with this idea, and see what the Facebook monkeys say. Just want to get an idea of where we should go with this.

jaywink:
> I just had a thought, if we change the pre-fill to only happen if the user has not auth'd via Facebook would this be an acceptable middle ground? (compared to removing all pre-fill)

That sounds like an acceptable solution to me at least.

Flaburgan:
I second Jason on this: it can be nice to support Facebook but it's definitely not a must-have in my opinion: I don't want anything to support Facebook to degrade user experience of *every* diaspora* user.

denschub:
Ping.

denschub:
Superseded by #6778.

PR #6535, opened 2015-11-07
Design of stream pages

Flaburgan:
I didn't like the dark grey background we currently have on the stream pages with a wide screen. I have several ideas I want to try about what the stream pages can become. They can be controversial though so I wanted to first open this PR with changes which could be accepted by everybody.

**Reminder: the current stream view looks like that on the develop branch:**
current-stream

This pull request removes the right side bar, moving the content in the left side bar. That way, 80% of the screen is available to the stream. It also moves the user avatar and name at the left of the publisher, which helps users to know with which account they are connected and avoid a too big width for the publisher.

Here is how it looked (at the beginning of this PR, this is an old screenshot, kept for history, see the up-to-date screenshot at the end of this post) for a 1440px width screen:

capture d ecran 2015-11-07 a 16 45 46

I took care of mobile in this PR even if this is the desktop version. I know @AugierLe42e  will be happy about that ;) The content of what was the right side bar disappears on small screen to directly access the stream and not have to scroll down a lot. The avatar and the name next to the publisher also disappear on xs and sm screens.

**Here is how the stream currently looks with this PR:**

PR #6639, opened 2016-01-15
Fix broken chat view in original white theme

ghost:
That will remove the gray border to the bottom on the left navbar.
This is also the same design like in the profile view and SPV.

untitled

### DIFF
--- a/app/assets/stylesheets/color_themes/_color_theme_override_origwhite.scss
+++ b/app/assets/stylesheets/color_themes/_color_theme_override_origwhite.scss
@@ -12,10 +12,13 @@ body {
   .left-navbar .hoverable { border-bottom: 0; }
 
   .profile-sidebar,
-  .left-navbar-fixed-background,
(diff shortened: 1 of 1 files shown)


AugierLe42e:
My poor @Zauberstuhl. Any time the design changes, you have to repaire the chat :p

Flaburgan:
Let's make it work first and then when 0.6 will be ready and will not change I will help you to integrate it in the design ;)

Zauberstuhl:
@Flaburgan why isn't it working for you?

svbergerem:
In 0.5.x.x the border is associated with the stream and thus will use the full height in most cases. This is also what the white theme currently does. Since the chat also breaks the other themes I think we should have something that fixes it for all themes at the same time.

> This is also the same design like in the profile view and SPV.

With the default theme the borders next to the main stream also look different from those in the profile and the SPV. I guess we'll need a plan how to unify that for all themes.

svbergerem:
Superseded by #6535.

PR #6644, opened 2016-01-17
Add link to public

dmorley:
Nice to have /public via the mobile UI also.


dmorley:
Ah I see the more button in current state on the mobile ui jumps from /public to /activity so this is more than adding a link

PR #6651, opened 2016-01-23
update federation gem to prevent locale leakage in receive requests

SuperTux88:


denschub:
Merged as ebfb901b3997256b02ffa8821518169289c0f3a5, thanks.

PR #6654, opened 2016-01-24
CSS: set "resize: vertical" as default for textarea

korun:
Hi there!

I think we should set default `resize` param to make textareas looks pretty, on the conversations pages, or `/profile/edit`...

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -20,6 +20,10 @@ textarea {
 }
 // scss-lint:enable QualifyingElement
 
+textarea {
+  resize: vertical;
+}
+
 // Forms described here are only used on the public pages at the moment
 .block-form {
   margin: 20px auto;
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -32,6 +32,10 @@ body {
   padding: 0;
 }
 
+textarea {
+  resize: vertical;
+}
+
 h3 {  margin-top: 0; }
 .clear { clear: both; }
 #main { padding: 56px 10px 0 10px; }


Flaburgan:
Yeah, why not.

More about resize [on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/resize)

Flaburgan:
Don't worry about the CI, it's a random failure.

svbergerem:
For the publisher and the comment textarea we set `resize` to `none` and use [autosize](http://www.jacklmoore.com/autosize/) to automatically change the height based on the input. Maybe this is something we could use for all textareas?

korun:
> Don't worry about the CI, it's a random failure.

Oh, thanks! :)

> For the publisher and the comment textarea we set `resize` to `none` and use [autosize](http://www.jacklmoore.com/autosize/) to automatically change the height based on the input. Maybe this is something we could use for all textareas?

Maybe, but I think that we also need to have a clear default value.

svbergerem:
> Maybe, but I think that we also need to have a clear default value.

This was meant to be a clear default value:

> use for all textareas

`textarea { resize: none; }` and `autosize($("textarea"))` for both desktop and mobile.

korun:
I think `resize: vertical` is provide more functional, for more browsers and versions.

Of course, I can change property to `none` if needed. Should I also add autosize js in this PR?

svbergerem:
Autosize changes the height on input so having `resize: vertical` doesn't sound really useful to me. It might be also confusing when you are able to resize the textarea and with the next character you enter into the textarea it resizes automatically.

But if anyone has some good reasons why we should keep `resize: vertical` that would be also ok for me.

> Should I also add autosize js in this PR?

I think that would be great but maybe wait for a second opinion before changing anything.

denschub:
Actually, I have no issues with `resize: vertical` if it looks okay. Github, for example, is also autosizing their textearas but they still allow the users to resize them vertically. I like that, but I'll not fight for it.

++ to adding the `autosize` library on text areas where it is not active yet.

svbergerem:
Alright, so let's stick with `resize: vertical`. @korun Would you like to add autosize in this PR or should it happen in another one?

korun:
I think autosize will require more checks in different situations, so it is better to add it in a separate PR.

svbergerem:
Thank you! :)

korun:
Thank you all, too)

PR #6658, opened 2016-01-27
let mention-regex only match usable strings

Faldrian:
Fixed regex for finding mentions. Previous to this a mention like `@{...}` would throw an exception and make the post and the stream unloadable, since it could not be parsed into display name and handle.
The specific syntax helps to avoid processing those syntax experiments.

### DIFF
--- a/spec/models/mention_spec.rb
+++ b/spec/models/mention_spec.rb
@@ -9,7 +9,6 @@ describe Mention, :type => :model do
     before do
       @user = alice
       @aspect1 = @user.aspects.create(:name => 'second_aspect')
-
     end
 
     it 'notifies the person being mentioned' do
--- a/app/controllers/status_messages_controller.rb
+++ b/app/controllers/status_messages_controller.rb
@@ -65,6 +65,7 @@ class StatusMessagesController < ApplicationController
   end
 
   def handle_create_error(error)
+    logger.debug error
     respond_to do |format|
       format.html { redirect_to :back }
       format.mobile { redirect_to stream_path }
--- a/lib/diaspora/mentionable.rb
+++ b/lib/diaspora/mentionable.rb
@@ -5,7 +5,7 @@ module Diaspora::Mentionable
   # ex.
   #   "message @{User Name; user@pod.net} text"
   #   will yield "User Name" and "user@pod.net"
-  REGEX = /(@\{([^\}]+)\})/
+  REGEX = /(@\{(.+?; [^\}]+)\})/
 
   # class attribute that will be added to all mention html links
   PERSON_HREF_CLASS = "mention hovercardable"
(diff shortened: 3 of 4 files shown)


spec/models/mention_spec.rb
```diff
@@ -25,6 +24,25 @@
     end
   end
 
+  describe "#mention_syntax" do
+    before do
+      @user = alice
+      @aspect1 = @user.aspects.create(:name => 'second_aspect')
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/models/mention_spec.rb
```diff
@@ -25,6 +24,25 @@
...
+    end
+
+    it 'should not try to build a mention for broken syntax' do
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/models/mention_spec.rb
```diff
@@ -25,6 +24,25 @@
...
+      sm = @user.build_post(:status_message,
+                            :text => "look at @{...} and @{blalbubb} and @{stuff; badhandle} and @{bad ;semicolon} - this is wrong syntax", :to => @user.aspects.first)
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

diaspora-code-review: Line is too long. [167/120]

spec/models/mention_spec.rb
```diff
@@ -25,6 +24,25 @@
...
+      expect(sm.mentioned_people).to be_empty
+    end
+
+    it 'should recognize correctly mentioned people' do
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/models/mention_spec.rb
```diff
@@ -25,6 +24,25 @@
...
+      sm = @user.build_post(:status_message,
+                            :text => "hi @{#{bob.name}; #{bob.diaspora_handle}} ho", :to => @user.aspects.first)
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

Faldrian:
As always, travis failed 2/8 but it should work and is covered by tests. I have modified an existing test to include the offending syntax examples, so the bug that got me started on this is covered by a test (as well as the general syntax of the mention in all allowed forms).

jhass:
Merged as 8e3816e64e1ee91568978b4b5569a2a8e1db9c9e

Thanks!

PR #6659, opened 2016-01-28
fix NameError: uninitialized constant User::Zip by requiring the ruby-zip gem

visualsayed:


app/models/user.rb
```diff
@@ -2,6 +2,7 @@
 #   licensed under the Affero General Public License version 3 or later.  See
 #   the COPYRIGHT file.
 
+require 'zip'
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

jhass:
Given we directly use the gem by now, we should just add it to the Gemfile instead. Then the `Bundler.require` call will also load it.

visualsayed:
@jhass I was reasoning about it and other options as well trying to achieve a lazy require until I read the following posts:
1- http://myronmars.to/n/dev-blog/2012/12/5-reasons-to-avoid-bundler-require 
2- https://twitter.com/justinko/status/281103551227379713
3- https://bugs.ruby-lang.org/issues/5653

I have chosen to go with the simple require as a quick fix but if you like the Bundler.require it'll be my pleasure to commit again.

visualsayed:
Oops! actually the [rubyzip] gem is in the Gemfile and should be loaded by by Bundler. I’m investigating this issue right now...

jhass:
Rails calls `Bundler.require` by default and kind of relies on doing so, in any case moving away from it now would be entirely unreasonable at this point for diaspora.

visualsayed:
@jhass I agree with you and still trying to find out why Bundler.require doesn't load rubyzip gem?

jhass:
Maybe the gem doesn't provide the right filename for bundler to do the autorequire? Figuring out why is kinda what it takes for solving the issue, if I just knew I'd just fix it.

visualsayed:
@jhass Actually the rubyzip gem file wasn't loaded by Bundler cause it's main file name (zip) is different than  it's name (rubyzip) and this piece of information was specified in the fem file so modifying the Gemfile with:
```
 gem "rubyzip", "1.1.7", require: "zip"
```
will solve this issue, I'll commit it right now.

Thanks,
Sayed

visualsayed:
@jhass hopefully done :)

I found out two more bugs while debugging this, I'll report them soon then send the PRs.

Thanks again,
Sayed

jhass:
Please squash, one commit is enough for this change.

visualsayed:
squashed :)

jhass:
Merged as 8c0d12b6d3dce76356845d91b6f439c5266d3d3e

Thanks!

Flaburgan:
Thank you!

Flaburgan:
Hm, I'm now wondering, what about the users who requested the export before this patch? Shouldn't we add a db migration to it, to reset everyone to "not requested" and allow them to retry?

PR #6660, opened 2016-01-28
fix queues for cron jobs

SuperTux88:
remove the queues from the schedule.yml so the configured queue in the worker is used


jhass:
Thanks1

denschub:
Good catch, thank you!

PR #6666, opened 2016-02-01
Fix blueimp gallery indicators

svbergerem:
The blueimp gallery repo refactored the bower files and it looks like the indicators are no longer included. I don't know if this is a blueimp gallery issue or a rails-assets issue but this PR manually loads the missing files and fixes the gallery indicators (thumbnails).

Fixes a regression from #6665.


jhass:
Thanks!

Issue #6662, opened 2016-01-29
Routing Error when clicking Refresh my photos link


visualsayed:
To reproduce this:

1- Navigate to: pod/user/edit

2- Click [Request my photos] link if you didn't request it before otherwise click [Refresh my photos] link.

You will get a Routing Error: No route matches [GetT] "/user/export_photos"

The reason behind this is that we didn't defined a post route only while in views/users/_edit.haml partial we use the [link_to] helper method to create a link tag without {method: :post} option which dynamically creates an HTML form and submit it using the HTTP get verb.

Adding the {method: :post} should fix this issue which makes rails create and submit the form using an HTTP post method.

I'll managed to check and complete the cucumber scenario at features/desktop/download_photos.feature then send a PR.

Thanks,
Sayed

svbergerem:
Looks like this has been fixed in #6667.

PR #6667, opened 2016-02-01
Issue #6662: fix refresh my photos routing error

visualsayed:


jhass:
Thanks! Again a regression on develop, so I merged only the cuke to stable.

PR #6670, opened 2016-02-04
Fix horizontal scrolling on mobile profile

Flaburgan:


svbergerem:
So we are dropping the Bootstrap grid in the mobile view? Currently we use it for [the photos page](https://github.com/diaspora/diaspora/blob/816092c/app/views/photos/_index.mobile.haml) and maybe others, too. (I didn't go through our repo)

Flaburgan:
Well I looked how main stream was displayed to fix this one and yeah it doesn't use bootstrap for that. Which doesn't block us to use it inside the stream elements.

svbergerem:
What I was trying to say is: If we are using the Bootstrap grid then I think we should use it properly:
```haml
.container-fluid
  .row
    .col-*
```
In the main stream we don't use the grid but on the profile page we use it for the photo stream. I think we should either use the grid properly on the profile page (and remove those margins / paddings that interfere with the grid) or we should completely drop the grid from the mobile view.

Flaburgan:
I removed the bootstrap classes for the mobile photos page.

svbergerem:
Thank you.

PR #6671, opened 2016-02-04
Check for collapsible content_view again after images are loaded

### LINKED ISSUE #3307: reshared posts dont get shrinked when they're too long
no beautiful "show more" :(

here's an example: https://despora.de/posts/426898

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/332-reshared-posts-dont-get-shrinked-when-they-re-too-long?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Faldrian:
I added a check for collapseOversized after each image in a post content have finished loading.
This was the main reason for posts with long / big images were not collapsed, since the checking code was run after the page was finished loading, but the images finished loading later.

I'm not certain how to write a test for this, since it's a countermeasure to some sort of timing problem. There are already tests for collapsing long posts, but they deal with text content only.

This will fix https://github.com/diaspora/diaspora/issues/3307

### DIFF
--- a/app/assets/javascripts/app/views/content_view.js
+++ b/app/assets/javascripts/app/views/content_view.js
@@ -72,6 +72,21 @@ app.views.Content = app.views.Base.extend({
 
   postRenderTemplate : function(){
     _.defer(_.bind(this.collapseOversized, this));
+
+    // run collapseOversized again after all contained images are loaded
+    var self = this;
+    _.defer(function() {
+      self.$("img").each(function() {
+        this.addEventListener("load", function() {
+          // only fire if the top of the post is in viewport
+          var rect = self.el.getBoundingClientRect();
+          if(rect.top > 0) {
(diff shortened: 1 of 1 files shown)


Faldrian:
I may add a thing that @SuperTux88 just mentioned... I need to implement a check for scrollposition, so I do not collapse posts that are already scrolled above the current viewport, since this would make the current position in the stream jump.

Faldrian:
Implemented the check. :)

denschub:
This works and I have no idea to write clean tests for this on top of my head. May be added later as well, so I've merged it. Thanks.

PR #6672, opened 2016-02-06
Use Hamlit to render Haml

k0kubun:
[Hamlit](https://github.com/k0kubun/hamlit) is faster implementation of Haml language. I use it for production and some of OSS apps like [ManageIQ](https://github.com/manageiq/manageiq) uses it for performance.

This pull request changes Haml implementation to Hamlit. I confirmed that all tests except already failing ones passed on my local machine.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -175,7 +175,7 @@ gem "typhoeus",           "0.8.0"
 # Views
 
 gem "gon",                     "6.0.1"
-gem "haml",                    "4.0.7"
+gem "hamlit",                  "2.2.1"
 gem "mobile-fu",               "1.3.1"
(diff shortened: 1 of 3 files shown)


denschub:
Thanks! We'll need some time to evaulate hamlit and potential issues, but we'll get back to you soonish! For now, could you please squash your commits into a single one?

k0kubun:
I'm looking forward to hearing the result of evaluation. Squashed the commits.

denschub:
Sorry it took so long. I was unsure whether we should add another native dependency or not. Anyway, I've merged this into our upcoming 0.6.0.0 release. Thanks!

k0kubun:
Thank you for merging!

PR #6675, opened 2016-02-06
Set maxwidth for container-fluid and unify settings UI

svbergerem:
This PR sets the max-width for `.container-fluid` to 1200px which makes sure that streams, settings etc. don't look bad on wide screens. The max-width is still the same for the different stream pages and the tags view since they would need some redesign first and this is done in #6535. The width of the profile page however changes with this PR.

Since the settings pages were now really too wide I used the same `list-group` that we already have on the contacts and conversations page. This only affects the desktop settings pages.

I split these changes into two commits. There is a third commit since scss-lint didn't enforce comments before temporary linter changes. I can move that to another PR or just push the commit to the develop branch if anyone has a problem with the commit being part of this PR.
### settings page

settings page
### conversations page

conversations page
### profile page

profile page

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -35,7 +35,7 @@ linters:
     enabled: true
 
   DisableLinterReason:
-    enabled: false
+    enabled: true
 
   DuplicateProperty:
     enabled: true
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -3,6 +3,13 @@ body {
   padding: none;
 }
 
+.container-fluid { max-width: $screen-lg-min; }
+
+.page-streams,
+.page-tags.action-show {
+  > .container-fluid { max-width: 100%; }
+}
+
 // Overflow
 h1,
 h2,
--- a/app/assets/stylesheets/conversations.scss
+++ b/app/assets/stylesheets/conversations.scss
@@ -191,11 +191,12 @@
   }
 }
 
-@media (max-width: 1354px) {
-  #left_pane #conversation_inbox .pagination ul > li > a {
-    padding: 4px 7px;
-  }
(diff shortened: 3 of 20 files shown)


denschub:
This is pretty. Another step to a unified UI, thank you **very much**.

Flaburgan:
Thank you for this work, a good base to continue to hack on the design :)

Flaburgan:
This also drastically impacted the SPV, which now look like that:
capture du 2016-02-09 15-02-26

denschub:
yes? Having a global max width was the purpose of this?

AugierLe42e:
It's not really pretty :/

denschub:
Neither is the full width. A full-width SPV will create long lines, which decreases readability (as discussed and agreed in #6535, there's no difference between the stream and the SPV in that regards).

Flaburgan:
@denschub I wanted to point it, @svbergerem said that the profile page changed, he didn't say anything about the SPV. I also want to point that the profile page content has a col9 which means 870px width for the content. The SPV now has 570px when you remove padding. I don't say I don't like it or we should not do that, just that it was not introduced so we didn't think about it.

svbergerem:
@Flaburgan I guess there was some misunderstanding. This PR set the max-width for all pages except for the stream views that use the main stream template (stream, activity, aspects, etc.) and the tags stream. The max-width is consistent for all other pages including settings, profiles, SPV, contacts, notifications, conversations, etc.

AugierLe42e:
> Do you have an idea how to "make it prettier" without loosing readability?

Previous SPV was ok to me.

AugierLe42e:
Yeah, I confirm: This was a very dumb move. Now I have to revert these chenges by hand on diaspote. Why hasn't this been polled before being merged?

PR #6677, opened 2016-02-07
Fix mobile registration layout after failed registration

svbergerem:
Fixes #6634.

PR #6678, opened 2016-02-07
Fix mobile registration design

svbergerem:
### before

before
### after

after
#### failed registration with #6677

after with correct layout


denschub:
Merged, sänk juh!

PR #6679, opened 2016-02-07
Add reason for post report to email sent to admins

### LINKED ISSUE #4711: Polish post report
I just tested the recently merged post report feature https://github.com/diaspora/diaspora/pull/4517. It's nice but can be improved.

* [x] It should be possible to report comments
* [x] There should be a confirmation message when the user validates the pop up with the indication why he reports the post
* [ ] The e-mail received by the podmin should contain the reason given by the user
* [x] The e-mail received by the podmin should contain a link to /post_report
* [x] The tooltip should have the good design
* [x] The report should not be sent even if the user cancels (see https://github.com/diaspora/diaspora/issues/4720)
* [x] The e-mail received by the podmin should be localized for the podmin - or be in English
* [x] There should be a link to the reported comment in the admin interface
* [x] The "Delete elem" button should be hidden when the post has already been deleted (by the user or any other way)

And (I don't know if this is the wanted behavior or not), but when a user report a post which has already been reported and marked as reviewed by the podmin, the podmin doesn't receive a second e-mail, and the post is not listed a second time in the /post_report page.

I guess this is nice to avoid spamming the podmin, but maybe he wants to know that another user reported this post... I don't know..

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1387933-polish-post-report?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #4711.

### DIFF
--- a/app/mailers/report_mailer.rb
+++ b/app/mailers/report_mailer.rb
@@ -1,15 +1,17 @@
 class ReportMailer < ActionMailer::Base
   default from: AppConfig.mail.sender_address
 
-  def self.new_report(type, id)
-    Role.moderators.map {|role| super(type, id, role) }
+  def self.new_report(report_id)
+    report = Report.find_by_id(report_id)
+    Role.moderators.map {|role| super(report.item_type, report.item_id, report.text, role) }
   end
 
-  def new_report(type, id, role)
+  def new_report(type, id, reason, role)
     resource = {
-      url:  report_index_url,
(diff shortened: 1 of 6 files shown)


denschub:
Yay, thanks.

PR #6686, opened 2016-02-08
Remove stream faces

### LINKED ISSUE #4661: Hovercards exceed the right border of the page.

### LINKED ISSUE #6008: Thumbnails in right bar should lead to associated posts in the stream instead of people's page
The thumbnails at the top of the right bar are convenient to see in one glance who is present in our current stream. When you click on it, you get the profile of the person, which may not to be the expected behaviour (you presumably want to see what the person has posted, not his whole activity). Furthermore, being sent to the profile page of the person takes time since a new page has to be loaded, and you need to go back to the stream view after it.

In accordance with the "Stream" heading above this area, I think it should be better to remain in the stream view and simply get a focus on the related post. This should be smoother and lighter.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/18710422-thumbnails-in-right-bar-should-lead-to-associated-posts-in-the-stream-instead-of-people-s-page?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #6008 and I think that this also fixes  #4661.


denschub:
If you're happy and you've merged it, clap your hands! :clap: :clap:

Flaburgan:
This is okay with [the poll I posted on diaspora](https://diaspora-fr.org/posts/1742990) though I feel like this could have been discussed a bit more :p 

Btw I saw in https://github.com/diaspora/diaspora/pull/5502 that the legend came from the contact_title method (see https://github.com/diaspora/diaspora/blob/develop/lib/stream/aspect.rb#L91 for example). I started to unify those in that PR, now that the feature is dropped, I guess those files and the language files should be cleaned too?

svbergerem:
Good catch! `contacts_title` doesn't seem to be used anywhere right now. Will remove them.

PR #6687, opened 2016-02-08
Remove contacts_title method and translations

svbergerem:


Flaburgan:
:)

denschub:
Merged as ac5a7746e28327dae898413913ed770841bc5e8d

Flaburgan:
Hm, I just saw with the freeze that this was merged in 0.5.7 when https://github.com/diaspora/diaspora/pull/6686 is in 0.6, if the stream faces are still in 0.5.7 then contacts_title should not be removed there.

svbergerem:
#6686 didn't remove any calls for `contacts_title`. It was just an obsolete function that we never used.

PR #6691, opened 2016-02-09
Remove stream-title data attribute

Flaburgan:
Following #6686 and https://github.com/diaspora/diaspora/commit/ac5a7746e28327dae898413913ed770841bc5e8d it looks like we don't need data-stream-title anymore, `.stream_title` now only _really_ refers to the stream title (the legend of avatars in the right side column was previously also changed by this code). The only place where we still set data-stream-title instead of using the text of the link was the pod name in the header, which launches a refresh of the page anyway (so the title of the stream is not loaded with JS).

### DIFF
--- a/app/assets/javascripts/app/app.js
+++ b/app/assets/javascripts/app/app.js
@@ -103,12 +103,7 @@ var app = {
 
       evt.preventDefault();
       var link = $(this);
-      if(link.data("stream-title") && link.data("stream-title").length) {
-        $(".stream_title").text(link.data("stream-title"));
-      } else {
-        $(".stream_title").text(link.text());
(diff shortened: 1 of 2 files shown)


svbergerem:
https://github.com/diaspora/diaspora/commit/aabe6951e5f4626b1a518fb096ced2a9a1e91c47 removed the asynchronous reloading as a temporary fix which should become obsolete with #6493. Then we will need this again so I'd like to keep it.

Flaburgan:
Yeah I was wondering about that. Maybe it would even be cleaner to add it to every links which open a stream and use that instead of `.text()`. Closing for the moment.

PR #6708, opened 2016-02-21
Activate "Stream::Person returns the most recent posts" spec

svbergerem:
#4020. I ran this a few times and it always succeeded.


denschub:
Sure, why not.

PR #6709, opened 2016-02-21
Adjust the SPV design

svbergerem:
spv1
spv2
spv3

### DIFF
--- a/app/assets/javascripts/app/views/single-post-viewer/single_post_comment_stream.js
+++ b/app/assets/javascripts/app/views/single-post-viewer/single_post_comment_stream.js
@@ -20,7 +20,7 @@ app.views.SinglePostCommentStream = app.views.CommentStream.extend({
 
   postRenderTemplate: function() {
     app.views.CommentStream.prototype.postRenderTemplate.apply(this);
-    this.$(".new_comment_form_wrapper").removeClass('hidden');
+    this.$(".new-comment-form-wrapper").removeClass("hidden");
     _.defer(this.highlightPermalinkComment);
   },
 
--- a/app/assets/javascripts/app/views/single-post-viewer/single_post_content_view.js
+++ b/app/assets/javascripts/app/views/single-post-viewer/single_post_content_view.js
@@ -7,6 +7,7 @@ app.views.SinglePostContent = app.views.Base.extend({
 
   templateName: "single-post-viewer/single-post-content",
   tooltipSelector: "time, .post_scope",
+  className: "framed-content",
 
   subviews : {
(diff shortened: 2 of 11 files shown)


Flaburgan:
Maybe the reshares / likes and comments counts should also be wrapper in a `.framed-content`?

Also, I'm wondering about the size of each column. In the other views, the post is in a col9 where here it only has a col6, which is weird because the SPV is where you go when you want to focus on a special post. So to have more space for it when it's lost in the stream than when you focus it seems wrong. In another hand, we sometime have small post but big comments, so to reduce the comment is maybe not the solution... I'm kinda lost here.

AugierLe42e:
> Maybe the reshares / likes and comments counts should also be wrapper in a .framed-content?

+1

denschub:
> In another hand, we sometime have small post but big comments, so to reduce the comment is maybe not the solution... I'm kinda lost here.

Yeah, right. I still feel like 50/50 is a somewhat suitable approach. Maybe something like 60/40 on large screens.

> Maybe the reshares / likes and comments counts should also be wrapper in a .framed-content?

Actually, I've tried that locally with Steffens changes and we had a small chat about this. Thouse counters will look pretty lost in their own white box, which does not look nice either. I also had the idea of "reusing" the notification badge in some way, but that's not possible since we would have no way of showing the faces.

Not too sure about this. While this PR is something I could live with, I have a hard time finding a "*perfect*" solution...

svbergerem:
spv1_2
spv4_2

denschub:
Hm. Looks cleaner. Other opinions?

Flaburgan:
I'd say go for it and we'll see if better ideas pop out :)

AugierLe42e:
I'm ok with that.

denschub:
Common consent, what a rare occasion. Merged. Thanks everybody!

Flaburgan:
Just merged it in d-fr. It looks like we miss some padding-top:
capture du 2016-02-23 12-10-49
I would add at least put a 15px padding-top to `.head` to have the same space than the left one. And same opinion about the interaction div.

Also, I feel like the interaction avatars should not go under the like / reshare counts but stay aligned:
capture du 2016-02-23 12-11-18

denschub:
Padding [is fine on a clean develop branch](https://cloud.githubusercontent.com/assets/344777/13250113/678c4e74-da28-11e5-8f00-6ff88afde9a6.png) so that's probably your issue. Good idea regarding the faces, though.

svbergerem:
@Flaburgan Like this?
SPV alignment

Flaburgan:
@svbergerem yes :)

PR #6711, opened 2016-02-23
Only mark unread notifications as read

skyschub:
Query before:

``` sql
UPDATE notifications SET unread=false WHERE recipient_id = 42;
```

Query after:

``` sql
UPDATE notifications SET unread=false WHERE recipient_id = 42 AND unread=true;
```


svbergerem:
Merged as 9f2e5b1868aa436c44f60c75c9e8998bfc369b46. Thank you!

denschub:
Actually, thanks to @CSammy for making the suggestion.

PR #5729, opened 2015-03-04
Change back-to-top icon design

SansPseudoFix:
Screenshot : 
back-to-top4

Global screenshot : 
back-to-top_hover_big

PR #6713, opened 2016-02-23
Change back-to-top icon design

ghost:
New design for back-to-top arrow originally from @SansPseudoFix. Supersedes #5729
How it looks on diaspote.org:

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -68,26 +68,45 @@ pre { word-wrap: break-word; }
 }
 
 .back-to-top {
-  background-color: $border-dark-grey;
-  border-radius: 10px;
-  bottom: 20px;
-  color: $white;
   display: block;
-  font-size: 2.9em;
-  line-height: 1.5;
-  opacity: 0;
-  padding: 0 12px;
+  color: $white;
+  width: 50px;
+  height: 50px;
   position: fixed;
-  right: 54px;
   z-index: 49;
+  right: 54px;
+  bottom: 20px;
+  opacity: 0;
+  font-size: 2.9em;
+  background-color: $border-dark-grey;
+  border-radius: 0;
+  box-shadow: $card-shadow;
+  line-height: 1.5;
+  transition: box-shadow .3s, background-color .3s;
 
-  &:hover,
-  &.visible:hover {
-    color: $white;
+  &:hover {
+    background-color: lighten($border-dark-grey, 5%);
+    box-shadow: $card-shadow-hover;
+  }
+
+  &.visible,
+  &:hover {
     opacity: .85;
   }
 
-  &.visible { opacity: .5; }
+  &:before {
+    border-style: solid;
+    border-width: .25em .25em 0 0;
+    color: $white;
+    content: '';
+    display: inline-block;
+    height: 25px;
+    margin: 0 12px;
+    position: relative;
+    transform: rotate(-45deg);
+    vertical-align: middle;
+    width: 25px;
+  }
 }
 
 // general purpose classes
--- a/app/assets/stylesheets/color-variables.scss
+++ b/app/assets/stylesheets/color-variables.scss
@@ -30,3 +30,4 @@ $sidebars-sub-background: darken($sidebars-background, 12%) !default;
 $left-navbar-drawer-background: darken($sidebars-background, 6%);
 
 $card-shadow: 0 1px 2px 0 rgba(0, 0, 0, .16), 0 2px 10px 0 rgba(0, 0, 0, .12) !default;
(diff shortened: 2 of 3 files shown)


denschub:
Is there any reason not to use `entypo-chevron-up` or even `entypo-chevron-thin-up`? Those lines in the top of that arrow are rather distracting.

In an unmodified `develop` source, the icon is not centered.

Also, I do not like that the initial opacity/color changed and it now gets lighter on hover. Right now, the icon is noticeable lighter and gets darker on hover, which makes more sense.

AugierLe42e:
> Also, I do not like that the initial opacity/color changed and it now gets lighter on hover. Right now, the icon is noticeable lighter and gets darker on hover, which makes more sense.

Because right now the color is the only thing that changes. In this PR, the shadow also gets more diffuse just as if the button was getting higher. It would be weird that the icon gets darker as it gets higher.

denschub:
I see no reason why that should be weird and I really do not like the back to top-icon to be that visible. After all, it's a very insignificant UI element and there is absolutely no reason to have it attract that much attention in my opinion. @svbergerem, can I have a second opinion here?

AugierLe42e:
> I see no reason why that should be weird

Because this is not how light reacts most of the time in real life. Objetcs rarely get darker as they get closer to you ;)

Flaburgan:
>Because this is not how light reacts most of the time in real life. Objects rarely get darker as they get higher in elevation ;)

That's called a shadow. :p 
Majority of elements are designed like that, have a look at how our own buttons react on hover.
To remove the text underline and to change the ugly arrow feels enough to me here.

AugierLe42e:
> That's called a shadow. :p 

I don't speak about the shadow. I speak about the color of the object.

denschub:
UI elements with little contrast: not important, inactive. UI elements with high contrast: imprtant, active. I do not want to discuss with you about how light works since that's not topic of this discussion. We're talking about basic UI design principles here.

AugierLe42e:
Hey, calm down. You're asking why the button gets lighter on hover, I'm just explaining why I did this choice. Its ok to change but then I have to remove the shadow, that's it. There's really no need to get angry for a color...

denschub:
Chill, angry messages look very different. ;) I don't think changing the shadow at all is necessary. Just leaving the shadow as it initially is but making the color darker on hover would work visually and user experience wise. Obviously, removing the shadow is also a solution I'd not object, after all, that would create less noise. Do whatever you want, but I don't like the current state.

goobertron:
Yes, I also would like this element only just visible when not focused, and becoming more visible when hovered/focused.

What exactly is wrong with the current back-to-top icon? I prefer it visually to this new, rather angular chevron. Could you at least add the rounded corners used by SansPseudoFix in his version? That (along with making the icon lower-contrast when not focused) would help soften it. Thanks in advance.

svbergerem:
```bash
app/views/people/index.html.haml:41:          &#8679;
app/views/people/contacts.haml:27:        &#8679;
app/views/people/show.html.haml:37:        &#8679;
app/views/tags/show.haml:40:        &#8679;
```

AugierLe42e:
@goobertron: I dropped the rounded corners to unify the styling with the new design scheme of diaspora. No other element in the design is rounded so it would feel weird.

svbergerem:
> No other element in the design is rounded so it would feel weird.

Buttons, avatars, badges, ...

AugierLe42e:
> Buttons, avatars, badges, ...

Not that much. Border radius of these elements is 4px. Plus, that back-to-top arrow is neither a badge, a button, nor an avatar. It is displayed just next to posts cards wich have square angles.

denschub:
You can click on it and it does stuff after you clicked it. Can we agree that this is a button?

AugierLe42e:
I meant: visually. But I can also make it blue with a text like the other buttons ;)

goobertron:
> > Buttons, avatars, badges, ...
> 
> Not that much. Border radius of these elements is 4px.

According to @SansPseudoFix in [this comment](https://github.com/diaspora/diaspora/pull/5729#issuecomment-77138961), the radius on the back-to-top button in his PR was also `4px`.

AugierLe42e:
Ok, so, if I sum up: change the gray on hover, remove shadow and add 4px border radius. I.e, what @SansPseudoFix did in #5729. Now, the question is: why #5729 has never been merged if it was ok?

denschub:
> why #5729 has never been merged if it was ok?

Because there were unfinished discussions, questions and apparently nobody thought a change is needed. And because you superseded it.

goobertron:
Re the light/dark grey on unfocused/focused: rather than using different shades of grey, could this be done using opacity, as it done currently? See this example from diaspora-fr.org:

arrow

I don't know what opacities are used currently, but I'd think something like 40% when unfocused and 80% when focused would work (mock-ups below).

arrow2

denschub:
Superseded by #6782.

PR #6715, opened 2016-02-24
Remove #publisher_photo_upload from haml

cmrd-senya:
It seems like #publisher_photo_upload is relict of blueprint.
It was removed from CSS in 0c7d319fdc31ea6de8c6f30ae7badf89c9d5aef8 and
probably was unused even before. So remove it from publishers.


svbergerem:
Thanks!

PR #6697, opened 2016-02-13
Unify contacts, notifications and settings UI

svbergerem:
This PR uses the design of the profile and conversations page for contacts, notifications and settings. The redesign of the stream pages happens in #6535. I also refactored some code of the conversations and profile page so all pages now share the same code for the sidebar.

Here are some screenshots. I also added one of the profile and one of the conversations page although I only refactored the code.

contacts
conversations
notifications
privacy
profile
profile_edit

PR #6717, opened 2016-02-24
Fix style regression on contacts page from #6697 

svbergerem:
for notification dropdown and no contacts well

dropdown
well


denschub:
Merged, thanks.

PR #6718, opened 2016-02-24
Fix notification count change on contacts page

svbergerem:
Before it also changed the contacts count.


denschub:
Merged, thanks.

PR #6719, opened 2016-02-25
Include a private key to the user data export archive

cmrd-senya:
In order to allow a user to notify his contacts of his migration with a trustworthy signature the old private key must be available to the new pod where we import data.

### DIFF
--- a/app/serializers/export/user_serializer.rb
+++ b/app/serializers/export/user_serializer.rb
@@ -4,6 +4,7 @@ module Export
                :email,
                :language,
                :username,
+               :serialized_private_key,
                :disable_mail,
                :show_community_spotlight_in_stream,
                :auto_follow_back,


Flaburgan:
Does this key have a passphrase? Is it the user password?

denschub:
> Does this key have a passphrase?

No.

cmrd-senya:
No there is no passphrase ATM. That's bad, yeah. We probably need one. However there are some other sensitive data in the archive, so if encrypt here, we should probably encrypt the whole archive then.

Flaburgan:
I would say let's at least set the user's password as the passphrase of the private key. Or encrypt everything with the user's password.

cmrd-senya:
I doubt we have a user password in a clear text to make that possible. It's getting to complicated then, it's better to define and apporve new archive format at once then.

denschub:
Okay, I've merged this in for 0.6.0.0. But there are some notes I want to add.

First, if you do further work on the user serializer, please consider adding a test, especially if you rely on the format. See `spec/serializers` for examples.

Next, a warning. It feels like you are going to use this for your work in your crowdfunding campaign. I am not going to give you any plans or ideas on how to do that, that's entirely not my point and you actually claimed that project yourself, which is fine. However, please note that I strongly recommend you to do some planning work before starting the development and get feedback on those plans before implementing it. I'm mainly thinking about format- and exchange-protocol-specifications or even a rough draft of those. After all, all your work depends outside the core team and if you want to get your stuff merged after you are done, I really encourage you to discuss your plans. In the worst case, you could end up developing something that no one on the core team likes and you could be forced to re-engineer large portions of your work. That would be very unfortunate.

Flaburgan:
I still feel uncomfortable seeing a private key leaving the server without protection...

cmrd-senya:
I hope I'll change that before this goes to release.

denschub:
The private key is leaving the server via TLS which is exactly the same way the users password gets transferred to the server. The user has to be logged in to export the data, so if you suggest a password based encryption (which is practically impossible since we don't store the plain text password obviously), the password would be lost anyway if you consider TLS unsecure.

And if the user is not able to keep their private data private (that is, storing the exported data), the password is also not stored private, and thus the access to the user profile is not private. The user is the only one who can export the data, so what's the loss?

cmrd-senya:
@denschub, I have a draft JSON schema of the archive in [here](https://gitlab.com/TheFederation/backup-restore/blob/80e80f9d69304d00461c309b48657135c55daac5/archive-format.schema).

It mainly differs from the present archive schema by names of some parameters and I also removed some fields which are considered diaspora specific. Implementation could still have them, but the schema is supposed to be more common, so only more or less common options were left.

Where do I put it for the discussion with the core team?

Flaburgan:
Though https is not an obligation to set up a pod at the moment, but yeah, it's the recommended set up.
My point is more about the user not knowing the importance of the archive actually. He could choose to download it at work for example just to have a look, without being aware of the presence of the private key in it. So maybe a checkbox "include my private key" unchecked by default or something like this?

denschub:
@cmrd-senya I'm not only talking about the format but also about the general flow and implementation ideas. Like how the import could look like and stuff like that. Right now, there is absolutely no usable plan. I've somewhere heard you think about writing a devise extension. A "handcrafted import" could also work, without having the dependency to our authentication framework. There are a lot of questions like these. Usually, Loomio is a good place for discussions...

@Flaburgan The user not being aware that the archive contains all the private information when the user requests to export all the private information? Ah, I'm not sure, I don't think people start randomly sharing all their data with everyone, do you?

Krassmus:
Fla, although this key is technically the same as a PGP-private key it is not as sensitive, since it's not used for end-to-end-encryption. The key is practical to identify the user on the new pod (when or if the other ticket of Senya will be merged). But it is not more or less sensitive as the password to the account, which is also encrypted with SSL or not encrypted at all at non-SSL-pods. Maybe we should just add a warning when downloading the archive "warning: other people could hijack your account if they get a copy of this archive".

cmrd-senya:
I've created [a proposal](https://www.loomio.org/d/qsVJ2K1t/backup-and-restore?proposal=z9BZQnMD) on loomio on the topic of archive encryption in the manual export case. Please, share your opinions.

PR #6723, opened 2016-02-27
Link share visibilities directly with users

SuperTux88:
Share visibilities were linked with contacts, but these are lose objects which can be deleted/created and have side-effects. I linked the share visibilities directly to the user, which doesn't change. Also this makes the Stream-Queries simpler, because we need fewer joins.

Also changed:
- remove all share visibilities where the shareable is already deleted
- remove all share visibilities where the user is already deleted
- remove the evil-query for person-stream and replace it with one simpler query.
- improve multi-/aspect-stream queries and use fewer queries
- don't add share visibilities for the last 100 public posts if a person if he shares with someone. This fails anyway because share visibilities are only for private posts.
- delete share visibilities when the shareable is deleted.
- remove an old spec for migrations, because it tries to rollback all other migrations which will fail, because the new code doesn't work with older db-schema-versions.
- remove `created_at` and `updated_at` from share visibilities because this was not used and makes the DB bigger.


Flaburgan:
Wow, that's a big work! Thanks a lot for doing that, it was really needed.

lib/diaspora/federated/shareable.rb
```diff
@@ -101,18 +101,18 @@ def update_existing_sharable(shareable)
...
+        def receive_shareable_visibility(user, shareable)
+          user.receive_shareable(shareable)
           user.notify_if_mentioned(shareable)
           logger.info "event=receive payload_type=#{self.class} status=complete " \
-                      "sender=#{diaspora_handle} receiver=#{person.diaspora_handle} guid=#{shareable.guid}"
+                      "sender=#{diaspora_handle} receiver=#{user.diaspora_handle} guid=#{shareable.guid}"
         end
 
-        def receive_non_persisted(user, person)
+        def receive_non_persisted(user)
```
diaspora-code-review: Assignment Branch Condition size for receive_non_persisted is too high. [23.43/20]

spec/models/post_spec.rb
```diff
@@ -148,6 +148,22 @@
           Post.for_visible_shareable_sql(Time.now + 1, "created_at")
         end
 
+        context "with two posts with the same timestamp" do
+          before do
+            aspect_id = alice.aspects.where(name: "generic").first.id
+            Timecop.freeze Time.now do
```
diaspora-code-review: Do not use `Time.now` without zone. Use one of `Time.zone.now`, `Time.now.current`, `Time.now.in_time_zone`, `Time.now.utc`, `Time.now.getlocal`, `Time.now.iso8601`, `Time.now.jisx0301`, `Time.now.rfc3339`, `Time.now.to_i`, `Time.now.to_f` instead.

spec/models/post_spec.rb
```diff
@@ -148,6 +148,22 @@
...
+              alice.post(:status_message, text: "first", to: aspect_id)
+              alice.post(:status_message, text: "second", to: aspect_id)
+            end
+          end
+
+          it "returns them in reverse creation order" do
+            posts = Post.for_visible_shareable_sql(Time.now + 1, "created_at")
```
diaspora-code-review: Do not use `Time.now` without zone. Use one of `Time.zone.now`, `Time.now.current`, `Time.now.in_time_zone`, `Time.now.utc`, `Time.now.getlocal`, `Time.now.iso8601`, `Time.now.jisx0301`, `Time.now.rfc3339`, `Time.now.to_i`, `Time.now.to_f` instead.

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
```
denschub: I'll ignore this for now, but please use `git commit -p` or similar in the future to avoid mass-changing of whitespace in the schema files.

db/migrate/20160225232049_link_share_visibilities_with_user.rb
```diff
@@ -0,0 +1,65 @@
...
+    add_foreign_key :share_visibilities, :contacts, name: :post_visibilities_contact_id_fk, on_delete: :cascade
+
+    add_column :share_visibilities, :created_at, :datetime
+    add_column :share_visibilities, :updated_at, :datetime
+  end
+
+  private
+
+  def cleanup_deleted_share_visibilities
+    ShareVisibility.joins("LEFT OUTER JOIN posts ON posts.id = share_visibilities.shareable_id")
```
jhass: Just to be on the safe side, let's use [migration models](http://blog.makandra.com/2010/03/how-to-use-models-in-your-migrations-without-killing-kittens/) here.

denschub: Eh. Happy conflicts.

We discussed about that via IRC decided to avoid allocating tons of classes. We should not run into any left-over childs since those get cascaded away. Why would you prefer models?

denschub:
okay, I've merged this without knowing that @jhass was reviewing at the same time. Happy time conflicts. There is one discussion left, but that could be done in a follow-up.

db/migrate/20160225232049_link_share_visibilities_with_user.rb
jhass: I'm not saying don't use the query fragment here, migration models are about not depending on the code in `app/` when running migrations, to maximize the lifetime of a migration and have a higher chance that the migration still works after the code in `app/` changed.

denschub: Hm. Right, good point, thanks. @SuperTux88, please follow up here. Thanks, Jonne!

cmrd-senya:
Look like DB migration fails after the PR was merged:
```
git checkout 75c3e6068ca1ed19ab9eb8710927cc9159c2bedc
RAILS_ENV=development bin/rake db:drop db:create db:schema:load
git checkout develop
bin/rake db:migrate
```
This produces [this output](https://gist.github.com/cmrd-senya/9ed26b8430233b1d14c0).

75c3e6068ca1ed19ab9eb8710927cc9159c2bedc is last commit before the PR

denschub:
Argh, crap. I've done some other things locally so my database actually skipped the migration and MySQL went well. There is https://github.com/rails/rails/issues/13496 which is causing the issue. Here is the query generated by AR on Postgres:

```sql
UPDATE
  "share_visibilities"
SET
  share_visibilities.user_id = contacts.user_id
WHERE
  "share_visibilities"."id" IN (
    SELECT
      "share_visibilities"."id"
    FROM
      "share_visibilities"
    INNER JOIN contacts ON share_visibilities.contact_id = contacts.id
  )
```

The join gets placed inside the subselect. Very funny. Reverted with 84cfbd22fc49c9f8d96f4f78f9dd455be2a658c0 as this needs more time to investigate. Grrrr.

denschub:
Fixed in #6736. Also, @jhass, are you okay with the models in that PR? I had to merge that in as soon as possible in case somebody already pulled it...

jhass:
Yes, that's perfect.

PR #6726, opened 2016-02-29
Update the user data export archive format.

### LINKED ISSUE #6138: user export: conversations not included
Related to issue #6115 

The export of posts does not contain the complete message threads, but only the top level post and any comments which you made personally. This makes it impossible for an import utility to reconstruct the conversation, especially since due to issue #6115 there is no identifying information for the post so you can't even look it up on the original site (which could very well be dead and possibly is the reason you need to relocate).  We know from experience that trying to re-create a conversation from a federated site is likely to be incomplete, so all the more reason to include all the comments and likers in the export file. 
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/22873219-user-export-conversations-not-included?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #6780: Maybe move test data seeding from AccountDeletion to FixtureBuilder?
[There](https://github.com/diaspora/diaspora/blob/develop/spec/integration/account_deletion_spec.rb) is a plenty of test data is generated in order to test the account delection feature. I added some more in my #6750. However this data are just the database filling and are nothing special to the account deletion. They may be used in different tests. It is just happened that the account deletion feature requires to have all the kinds of the relations in the database to test the correctness of the deletion. So does the migration (and it's partial implementation in #6750).

I would like to propose to move this test data generation to the [fixture builder](https://github.com/diaspora/diaspora/blob/develop/spec/support/fixture_builder.rb), so they are available for the whole test suite. Also, the generated fixtures may be used later to seed a database in the development mode to do some manual testing with comprehensive data.

Please, give me feedback on the idea.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32483019-maybe-move-test-data-seeding-from-accountdeletion-to-fixturebuilder?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cmrd-senya:
This PR introduces changes to the user data export archive format.

Names of some parameters are changed to be less diaspora-specific. This was originally proposed by @jaywink in order to implement a common data format for The Federation at whole. Also the commit adds the archive format json schema. ATM it is used in automatic tests only, but in future it will also be used to validate incoming archives.

fixes #6138 
fixes #6780


Gemfile
```diff
@@ -19,6 +19,7 @@ gem "diaspora_federation-rails", "0.0.12"
 gem "acts_as_api", "0.4.2"
 gem "json",        "1.8.3"
 gem "json-schema", "2.6.0"
+gem "json-schema-rspec", "0.0.4"
```
denschub: That gem is not needed in production.

app/assets/schemas/archive-format.json
```diff
@@ -0,0 +1,203 @@
+{
```
denschub: I wonder if we want to have the schemas inside the assets folder, since it's technically not a frontend asset. Do you have a better idea?

app/assets/schemas/archive-format.json
```diff
@@ -0,0 +1,203 @@
...
+              },
+              "text": {
+                "id": "https://the-federation.info/specs/backup-restore#archive_format/user/posts/131/text",
+                "type": "string"
+              },
+              "public": {
+                "id": "https://the-federation.info/specs/backup-restore#archive_format/user/posts/131/public",
+                "type": "boolean"
+              },
+              "thefederation_uid": {
```
denschub: Maybe not for discussion here, but I find that name rather weird. I'd prefer something more neutral like `accountIdentifier` or similar. What do you think?

app/assets/schemas/archive-format.json
```diff
@@ -0,0 +1,203 @@
...
+            "properties": {
+              "sharing": {
+                "id": "https://the-federation.info/specs/backup-restore#archive_format/user/contacts/185/sharing",
+                "type": "boolean"
+              },
+              "following": {
+                "id": "https://the-federation.info/specs/backup-restore#archive_format/user/contacts/185/receiving",
+                "type": "boolean"
+              },
+              "person_thefederation_uid": {
```
denschub: Maybe not for discussion here, but I find that name rather weird. I'd prefer something more neutral like `accountIdentifier` or similar. What do you think?

cmrd-senya: I think "account identifier" is fine. But the snake_case is used in this document, so it probably should be ```account_identifier``` or ```account_id```.

denschub: Agreed. Was just brianstorming there on the go.

denschub:
https://travis-ci.org/diaspora/diaspora/builds/112495051

app/assets/schemas/archive-format.json
cmrd-senya: How about ```app/schemas``` then?

cmrd-senya:
I would like to link #6138 which I've just noticed because it's related to the export archive.

I haven't checked it closely, but probably it's a good idea to take into account the #6138 in that PR as well. Also because the point of this PR is to make data export more platform agnostic and the issue is from @redmatrix.

cmrd-senya:
BTW, the tests was fixed after rebasing on the newer develop.

denschub:
Actually, the tests were failing because of the leap day... I'll submit a PR to fix this very edgy edge case soon and maybe even write a blog post about it, since that was pretty funny...

cmrd-senya:
Whoa, cool! :D
Looking forward to see it.

cmrd-senya:
c22d2146cf3c6b67eb06f90803c95d4c8ea3aa76 fixes #6138.

Maybe add more tests?

app/schemas/archive-format.json
```diff
@@ -0,0 +1,222 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://the-federation.info/specs/backup-restore#archive_format",
```
denschub: So, what's the status of this document? Is it even maintained? I see a lot of opened issues in the gitlab and nobody seems to care.

app/serializers/export/user_serializer.rb
```diff
@@ -17,8 +17,13 @@ class UserSerializer < ActiveModel::Serializer
     has_many   :comments, each_serializer: Export::CommentSerializer
 
     def comments
-      object.person.comments
+      Comment
```
denschub: Why should we export of other users if they've been made to the export users posts?

cmrd-senya: I guess it should be discussed at #6138

denschub: Agreed, so please don't change it here unless we're sure what we want to do. This would export other peoples comments below limited posts as well and I don't feel comfortable with exporting them unless it's clear that everybody wants that. Maybe we should move this onto Loomio since I really feel like this could violate other users data safety.

denschub:
I'm somewhat worried about the lifetime/maintenance of the spec you're using and you did not comment on that one so far. Also, please notice my latest comment. Other than that, it looks nice!

Sorry for the long delay, but I've been incredibly busy. :)

cmrd-senya:
Yep, I still have to update the spec to some more complete form. I guess, #6138 must be decided before, since it affects the data format as well.

cmrd-senya:
About the specification: the scope of this PR is the archive format. In the spec that was written originally by @jaywink the archive format was a part of a larger document about all the migration/backup thing. I guess it's not necessary to make such document at once (I'm more comfortable to create the documentation along with the code, not before). @denschub, if I build a separate document about the archive format and leave aside all the other migration stuff for a while, will it be enough for this PR? I see the document as a json schema of the archive with comments about fields meaning.

app/schemas/archive-format.json
cmrd-senya: This document is likely to be discontinued and the ID should be changed. But I'm not sure how to make a new one.

jhass: One should host the exact document at that URL. We can host stuff on diaspora.software, perhaps a hostname pointing to some Github pages repo or just on the diasporafoundation.org server.

cmrd-senya: What document is expected to be at the URL? Does it have to be something official? I'm writing something like an implementation plan&details for the account migration. But that though related it doesn't specify the exact format of the archive (but rather relies on the schema). Do I need to have a separate document defining the archive format? To me it seemed, that I could avoid having a document for the archive specification since there is a schema (it's kinda self-documenting).

jhass: The very exact schema file referencing it.

cmrd-senya: Ah, I see, thanks for clarification

lib/diaspora/exporter.rb
```diff
@@ -6,22 +6,23 @@ module Diaspora
 
   class Exporter
 
-    SERIALIZED_VERSION = '1.0'
+    SERIALIZED_VERSION = "1.1"
```
diaspora-code-review: Freeze mutable objects assigned to constants.

cmrd-senya:
In the latest commit I published the updated [schema](https://github.com/cmrd-senya/diaspora/blob/8979e1e364ebcfc9bf2617057edabf95bf5ef822/app/schemas/archive-format.json).

[Here](https://gist.github.com/cmrd-senya/821c5db956b7fd5b866df5c8c5e87f4b) is an example of file conforming to the schema.

cmrd-senya:
I'd like to ask for a preliminary review of the PR. There are still some TODOs in the code, mostly related to specs and docs. Before finishing them and polishing the PR I'd like to get green light about the approach and general feedback on the code.

The PR is currently rebased on `SuperTux88/federation-update` branch since it requires the development version of `diaspora_federation`.

cmrd-senya:
Also ping @annando (as promised), so that you can check the archive format changes for something relevant for Friendica.

annando:
We aren't currently importing this data from Diaspora, so I'm happy with every change. However, I always wanted to write some import.

cmrd-senya:
Also I've set up this PR branch at [troublesome.space](https://troublesome.space) so if anyone wants to test it there, welcome.

cmrd-senya:
I rebased the PR on the latest develop, so it should be ready for the (preliminary ) review now.

SuperTux88 (commented):
OK, here is some feedback. The direction looks good.

The main thing is about `participations`. They are currently used for two completely different things:
1. for which posts does the user want notifications
2. which other users/pod subscribed to a post

I planned to refactor/split these things up in the database, but I think we should already keep this clean in the export. So we should export the two different things that are currently in the `participations` table separately.

So we should have:
1. A list of posts which the exported user wants notifications
2. A list of subscribers for posts of the exported user:
    * A list of pods which are subscribed to a public post (we don't need the persons here)
    * A list of persons who received a private post (you export currently aspects here, but we need people here for federation)

I planned to change 2. in the database too, because federation of private relayables is currently broken because we only save aspects.

spec/spec_helper.rb
```diff
@@ -133,6 +133,20 @@ def client_assertion_with_nonexistent_client_id_path
   end
 
   config.include FactoryGirl::Syntax::Methods
+
+  config.include JSON::SchemaMatchers
+  config.json_schemas[:archive_schema] = "lib/schemas/archive-format.json"
+
+  # rubocop:disable Rails/DynamicFindBy
+  gemdir = Gem::Specification.find_by_name("diaspora_federation-json_schema").gem_dir
```
SuperTux88: I just created diaspora/diaspora_federation#70 and I think that's a cleaner solution for that, because you can just do this now:
```ruby
  JSON::Validator.add_schema(
    JSON::Schema.new(
      DiasporaFederation::Schemas.federation_entities,
      Addressable::URI.parse(DiasporaFederation::Schemas::FEDERATION_ENTITIES_URI)
    )
  )
```

What do you think about it?

cmrd-senya:
Thanks for the review, @SuperTux88! It must have been very time consuming. I'll keep on working considering your notes and will post any updates.

app/serializers/serializer_post_processing.rb
```diff
@@ -0,0 +1,18 @@
...
+  # TODO: docs
+  def serializable_object(options={})
+    super.tap {|hash|
+      modify_serializable_object(hash)
+    }
+  end
+
+  # TODO: docs
+  def modify_serializable_object(hash)
+    # TODO: having this method as an abstract overridable method is not very good,
```
cmrd-senya: There is a problem that `serializable_object` method is a part of undocumented API of `ActiveModel::Serializer` and it is likely to change (in fact in AMS 0.10 it changes it's name to `serializable_hash`). That's why I think it is a good idea to keep the notion of this method and the behavior we implement on top of it separately and explicitly, so that we don't have to hunt for references to the abandoned method once we've moved to another version.

But maybe I can do something like the following here:
```
  def serializable_object(options={})
     modify_serializable_object(super)
   end

def modify_serializable_object(hash)
  hash
end
```

What do you think?

SuperTux88: OK, I think I understand that better now. And I agree, we should rely on internal API as little as possible. And where you do, it's good when you document it.

Also as you mention it here: I tried to upgrade AMS in #7500, and "everything" broke when I used 0.10. I didn't have the time to have a closer look, so I just upgraded to the latest 0.9 version. But since you did more with serializers, you maybe understand the problems better than I and could have a look at the upgrade to 0.10?

cmrd-senya: AMS upgrade is pain, because so many things have changed there. It is even possible that some features which are required for us are missing or impossible in a newer version. Upgrade `0.9` => `0.10` requires a lot of work. There is no migration guide, there is no proper documentation for `0.10`, and so many things are non-obvious. Moreover it is possible that there will be `1.0` release and it'll make some breaking changes again and will require another painful migration from `0.10` to `1.0` (see [Status of AMS](https://github.com/rails-api/active_model_serializers#status-of-ams)).

I tried to start the upgade, but I can see that it'll take a lot of time, so if we block this PR for AMS upgrade, it'll delay that for a month at least. I think that I can work on AMS upgrade at the meantime, but using 0.9 is totally okay for now -- there is no deprecation announced for it and it's not clear if it will happen anytime soon.

SuperTux88: I didn't want to say that it's blocker for this PR, and I would keep 0.9 for now. This PR is a blocker for 0.7 and I'd like to have this finished first. But since you have worked with serializers it would be cool when you could try to upgrade it sometime later (probably after 0.7 is released).

spec/integration/account_deletion_spec.rb
```diff
@@ -1,128 +1,46 @@
...
+        account_removal_method
+      }.to change(nil, "user preferences empty?") { UserPreference.where(user_id: user.id).empty? }.to(be_truthy)
+        .and(change(nil, "notifications empty?") { Notification.where(recipient_id: user.id).empty? }.to(be_truthy))
+        .and(change(nil, "blocks empty?") { Block.where(user_id: user.id).empty? }.to(be_truthy))
+        .and(change(nil, "services empty?") { Service.where(user_id: user.id).empty? }.to(be_truthy))
+        .and(change(nil, "share visibilities empty?") { ShareVisibility.where(user_id: user.id).empty? }.to(be_truthy))
+        .and(change(nil, "aspects empty?") { user.aspects.empty? }.to(be_truthy))
+        .and(change(nil, "contacts empty?") { user.contacts.empty? }.to(be_truthy))
+        .and(change(nil, "tag followings empty?") { user.tag_followings.empty? }.to(be_truthy))
+        .and(change(nil, "clearable fields blank?") {
```
diaspora-code-review: Align `.and` with `.to` on line 17.

spec/shared_behaviors/account_deletion.rb
```diff
@@ -2,32 +2,27 @@
...
+      .and(change(nil, "participations empty?") { Participation.where(author_id: person.id).empty? }.to(be_truthy))
+      .and(change(nil, "roles empty?") { Role.where(person_id: person.id).empty? }.to(be_truthy))
+      .and(change(person, :closed_account).to(be_truthy))
+      .and(change(nil, "first name is blank?") { person.profile.first_name.blank? }.to(be_truthy))
+      .and(change(nil, "last name is blank?") { person.profile.last_name.blank? }.to(be_truthy))
+      .and(change(nil, "conversation visibilities empty?") {
+        ConversationVisibility.where(person_id: person.id).empty?
+      }.to(be_truthy))
+      .and(remain(nil, "conversations empty?") { Conversation.where(author: person).empty? }.from(be_falsey))
+      .and(remain(nil, "conversation visibilities of other participants empty?") {
```
diaspora-code-review: Align `.and` with `.to` on line 11.

cmrd-senya:
@SuperTux88, I corrected most of your remarks (except the one touching the federation gem, since we need another release with https://github.com/diaspora/diaspora_federation/pull/70 merged so I can fix it).

Could you please check this main commit with changes according to your notes: https://github.com/diaspora/diaspora/pull/6726/commits/5feac2775901e878740ce0b4ed6547a02fc876f6?

As soon as you approve it I'll start polishing and finishing the PR.

spec/integration/account_deletion_spec.rb
cmrd-senya: This and another message of rubocop is inadequate and maybe errorneous. I don't know how to fix it, but maybe will come with some idea later.

SuperTux88 (commented):
👍 I only reviewed the code changes from the last commit (I didn't review the tests yet), but it looks good. Only found some small things or questions.

spec/serializers/export/own_post_serializer_spec.rb
```diff
@@ -1,12 +1,54 @@
...
+    let(:object) { create(:status_message_with_photo, author: author) }
+  end
+
+  let(:json) { Export::OwnPostSerializer.new(post, root: false).to_json }
+
+  context "with private post" do
+    let(:post) { create(:status_message_in_aspect, author: author) }
+
+    before do
+      expect(post.subscribers).not_to be_empty
```
SuperTux88: I would just move this to the `includes remote people` test. Same for `with public post`.

spec/serializers/federation_entity_serializer_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def enable_test_entity_builder
+    allow(Diaspora::Federation::Mappings).to receive(:builder_for).with(object.class).and_return(:test_entity)
+    allow(Diaspora::Federation::Entities).to receive(:test_entity).with(object) {
+      TestEntity.new(test: "asdf")
+    }
+  end
+
+  before do
+    enable_test_entity_builder
```
SuperTux88: Why did you extract this to a method? If it's only used here I would just move the code here.

spec/serializers/federation_entity_serializer_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  it_behaves_like "it is federation entity serializer"
```
SuperTux88: Nit: "a federation entity serializer", results in:
* `shared_examples_for "a federation entity serializer"`
* `it_behaves_like "a federation entity serializer"`

spec/lib/diaspora/exporter_spec.rb
```diff
@@ -10,6 +17,14 @@
       user,
       %i[generic_user_data activity status_messages_flavours work_aspect]
     )
+
+    # verifications of data generation (protect us from possible false positives in case of poor data preset)
```
SuperTux88: I'm not sure about these "tests for tests". We want to test the code here, not the test. This `DataGenerator` is maybe an exceptions, so maybe we want to add some basic tests for the `DataGenerator` to test that it actually generates the data we want (but only the basics, not 100 test cases ;) ). But that should be somewhere else and not in the `exporter_spec`.

SuperTux88 (commented):
Also reviewed the tests now.

spec/serializers/federation_entity_serializer_spec.rb
cmrd-senya: Well, I followed the idea of self-description here: instead of writing a comment explaining what the specific piece of code does I extract it to a method, which name explains what it does. But if you think I should change it, I'm okay with it.

SuperTux88: I think a small comment is better here (it's only one extra line and is simpler), instead of creating an extra method (which is 3 lines + one empty line, and is more complex, because I first see the method, and then need to check where it's used and see that it's the `before` step).

lib/diaspora/exporter/posts_with_activity.rb
```diff
@@ -0,0 +1,68 @@
...
+        relayable_activity(participation_subquery)
+      end
+
+      def polls_activity
+        StatusMessage.where.not(author_id: person_id).joins(:poll_participations)
+                     .where("poll_participations.author_id = #{person_id}")
+      end
+
+      def relayable_activity(subquery)
+        Post.where.not(author_id: person_id).joins(subquery)
```
cmrd-senya: Maybe it's then better to implement a scope `with_activity` for the `Post` model instead of having a separate class for that?

spec/serializers/export/own_post_serializer_spec.rb
cmrd-senya: Well, I put  this string here for "tests for test" purpose again. If we don't do that, then we should remove it. Otherwise, it makes sense for both test since `doesn't include remote pods subscriptions` test theoretically could be false positive in case when there is no subscribers.

spec/lib/diaspora/exporter_spec.rb
cmrd-senya: We have a `spec/misc_spec.rb` with "tests for tests" now. I can put it to this file or create something else (another file or folder)?

cmrd-senya: I didn't fix this one yet, I have no idea why it was marked as "outdated"

lib/diaspora/shareable.rb
```diff
@@ -50,6 +50,14 @@ def subscribers
       end
     end
 
+    # Remote pods which are known to be subscribed to the post. Must include all pods which received the post in the
+    # past.
+    #
+    # @return [Array<String>] The list of pods' URIs
+    def subscribed_pods_uris
+      subscribers.map {|subscriber| subscriber.pod.try(:url_to, "") }.compact.uniq
```
cmrd-senya: This will not work for public posts since for public posts `subscribers` return an array of people, not a relation.

lib/diaspora/exporter/posts_with_activity.rb
SuperTux88: Hmm, since we only need the combined query here, I think we should keep it here (and combine it like in `all_activities`), and use the existing scopes from the Post model.

lib/diaspora/shareable.rb
SuperTux88: You're right. But then `subscribers.select(&:remote?).map(&:url).uniq` would work. I also thought about getting all unique `pod_id`s and then get all pods and only build the url once per pod.

spec/serializers/export/own_post_serializer_spec.rb
SuperTux88: About false positives: When there are subscribers for one test, then they are also there for the other test, so there is no need to check it on every test. I also don't like `expect` stuff outside of the actual test.

About "tests for test": I think the "better" solution would be, to check the resulting json has the key "subscribed_users_ids" and it isn't empty, in addition to the `include_json` check. That way we would "cleaner" test the result and not the test. But since `expect(post.subscribers).not_to be_empty` in addition with the `include_json` does the same, I think that's also a way to achieve the same result.

spec/lib/diaspora/exporter_spec.rb
SuperTux88: Hmm, I think I would create a `spec/spec/` folder for specs about specs ;) I didn't expect a file called `misc_spec.rb` to test the tests.

(And I don't know why GitHub marked it as outdated)

cmrd-senya: So maybe we should move `misc_spec.rb` to `spec/spec` sometime in future?

cmrd-senya:
Okay I fixed the notes so far. I'll go on with completing all TODOs.

lib/diaspora/exporter/posts_with_activity.rb
```diff
@@ -0,0 +1,51 @@
...
+      def likes_activity
+        other_people_posts.liked_by(person)
+      end
+
+      def comments_activity
+        other_people_posts.commented_by(person)
+      end
+
+      def subscriptions
+        return nil if person.owner.nil?
```
SuperTux88: This should never be nil, because we can only export a users data. And I think I would just change `person` to `user` in the constructor to make that clear.

Gemfile
```diff
@@ -13,8 +13,8 @@ gem "unicorn-worker-killer", "0.4.4"
 
 # Federation
 
-gem "diaspora_federation-json_schema", "0.2.0", github: "SuperTux88/diaspora_federation", branch: "json-schemas-helper"
```
SuperTux88: That's probably because I deleted that branch after you merged the PR. So you should use `develop` here

SuperTux88 (commented):
👍

cmrd-senya:
It also looks like I forgot to add reshares to the `PostsWithActivity`, so I'm going to do that.

cmrd-senya:
I covered most of TODOs now, I'll need to complete some spec still, but otherwise it is mostly finished.

cmrd-senya:
Okay, I guess it's now ready for the final review round.

Gemfile
```diff
@@ -13,7 +13,8 @@ gem "unicorn-worker-killer", "0.4.4"
 
 # Federation
 
-gem "diaspora_federation-rails", "0.2.0"
+gem "diaspora_federation-json_schema", "0.2.0", github: "diaspora/diaspora_federation", branch: "develop"
```
SuperTux88: Version 0.2.1 is now released :)

(that also probably breaks some tests, because nil elements are now missing)

lib/diaspora/exporter.rb
```diff
@@ -6,22 +6,23 @@ module Diaspora
 
   class Exporter
 
-    SERIALIZED_VERSION = '1.0'
+    SERIALIZED_VERSION = "2.0.dev".freeze
```
SuperTux88: I think we should change that to `2.0` for the 0.7.0.0 release.

features/desktop/change_settings.feature
```diff
@@ -53,3 +53,7 @@ Feature: Change settings
     And I go to the stream page
     And I expand the publisher
     Then I should see "Public" within ".aspect_dropdown"
+
+  Scenario: exporting profile data
+    When I click on the first selector "#account_data a"
+    Then I should see "Download my profile"
```
SuperTux88: Maybe also check to receive the notification mail.

app/serializers/export/own_post_serializer.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    private
+
+    def subscribed_pods_uris
+      object.subscribed_pods_uris.push(AppConfig.pod_uri.to_s)
+    end
+
+    def subscribed_users_ids
+      object.subscribers.map(&:diaspora_handle)
```
SuperTux88: I just saw that `subscribers` returns the person multiple times, when you have the person in multiple aspects.

That's because [Person.in_aspects](https://github.com/diaspora/diaspora/blob/9a2798331361954c259657a96d300fbe59ad5590/app/models/person.rb#L78-L81) which is used in [User.people_in_aspects](https://github.com/diaspora/diaspora/blob/d3edbfd82962bc5cf6a995085d3b2a537d5db84f/app/models/user/querying.rb#L54) returns the person multiple times. So we should filter that somewhere instead of adding the diaspora ID multiple times to the export. (That is also broken in the federation, we encrypt the message multiple times, but send it only once, because the target-url is used as a key)

SuperTux88 (changes requested):
I only reviewed the new diffs, but looks good. I found some small things, can you please fix them and then squash them and I'll do a final complete review then.

app/serializers/export/own_post_serializer.rb
Flaburgan: Are you still able to reinsert the contact in the correct aspects if you do that?

SuperTux88: Yes, only the diaspora ID is used here. For the contacts the aspect names are exported [here](https://github.com/cmrd-senya/diaspora/blob/2a97a49a6f0c4a6af989deb006ae77c9347d72b9/app/serializers/export/contact_serializer.rb#L20), so that are two different things.

Gemfile
cmrd-senya: Is it okay to bump to 0.2.1 within this PR?

app/serializers/export/own_post_serializer.rb
cmrd-senya: So I think it should be okay to add `distinct` to the query in the `Person.in_aspects` scope.

Gemfile
SuperTux88: Yes, please do that 👍 (You can do it in a separate commit)

app/serializers/export/own_post_serializer.rb
SuperTux88: Yes, that's also what I would do. I didn't check if anything breaks, but I think it should work.

Gemfile
cmrd-senya: We add some new gems to the gemfile here also. How do I split it?

I can do one commit with the version bump and then put all new gem additions to the https://github.com/diaspora/diaspora/pull/6726/commits/4ff4b297f748b3008a45258f5973f424b3d133c1 (Update the user data export archive format).

SuperTux88: I don't mind how you split the commits, as long as it makes sense and commit messages match, it's OK for me. I would probably do the version-bump in a separate commit and add the additional gem together in the commit where you use it.

cmrd-senya:
I squashed and rebased everything that was done before your notes. For your latest notes I added changes as fixups so you could check want I changed exactly.

SuperTux88:
fixups look good 👍 

One little thing: I try to start the commit message with a capital letter, but at least do it the same for every commit (`New Exporter::OthersRelayables class` and next commit `new Exporter::NonContactAuthors class`). So since you need to rebase again anyway, can you please change this commit message and also write `Bump` with a capital letter.

I'll do a complete review now again, because I only viewed the diffs until now, but you can already rebase and squash, so I can merge it afterwards (I expect that everything will be fine ;) )

cmrd-senya:
Done

app/serializers/export/contact_serializer.rb
```diff
@@ -1,12 +1,28 @@
 module Export
   class ContactSerializer < ActiveModel::Serializer
     attributes :sharing,
-               :receiving,
+               :following,
```
SuperTux88: I just realized, that you renamed that, probably because it is also named `following` in the federation gem. But these are different things:

* `sharing`: I share with that person and the person can view my private profile, send me private messages, ...
* `following`: I have this persons posts in my stream (currently always true for diaspora, the plan with adding `sharing` and `following` to the `contact` entity in the protocol was, that you can only follow the posts, without sharing your information, or you maybe want to share with the person, but this person posts much, so you don't want that in your stream)
* `receiving`: The other person shares with me (other direction of `sharing`)

So the diaspora DB currently only has `sharing` and `receiving`.

I think we should rename `following` back to `receiving` here. And we can add a new `following` that's hardcoded `true` here (because for data currently exported it is `true`).

We can also add a `subscribed` that is always `true`, for the other direction of `following` (so when the other person subscribed to us). We can then replace the hardcoded `true` with the correct value when the diaspora DB supports `following` and `subscribed`.

And we should definitely document that somewhere, what `sharing`, `following`, `receiving` and `subscribed` means.

app/serializers/export/contact_serializer.rb
```diff
@@ -1,12 +1,28 @@
...
                :person_guid,
                :person_name,
-               :person_first_name,
-               :person_diaspora_handle
+               :account_id,
+               :serialized_public_key
```
SuperTux88: Oh, another little thing: You renamed `serialized_private_key` to `private_key` for the users own private key, So I think this should be renamed to just `public_key`.

app/serializers/export/person_metadata_serializer.rb
```diff
@@ -0,0 +1,13 @@
+module Export
+  class PersonMetadataSerializer < ActiveModel::Serializer
+    attributes :guid,
+               :account_id,
+               :serialized_public_key
```
SuperTux88: And here too: `public_key`

SuperTux88:
Looks good 👍 I found still two things that I missed when only reviewing the single commits. But they are only a naming issues with the export-format (which should be fixed before "releasing" the 2.0 format), so that should be fixable quick :) I commented it as single comments (so you don't need to wait until I completed the whole PR, but I'm finished now), so you probably already saw them.

And I promise that it's ready to merge this time after you fixed these two (three) things :)

app/serializers/export/contact_serializer.rb
cmrd-senya: Do you think we should federate this `subscribed` info? What user see in his stream feels like a local setting, rather than network-wise. So `sharing`/`receiving` is a representation of exchange relationships in federation, while `following` is just a setting of composing the stream. We can include `following` since user will want to preserve this through migration, while `subsribed` is something meaningful only for the remote person, not for the local one. Right?

SuperTux88: No, when you can follow without sharing (for example follow a newsfeed, but don't want to share the profile to the feed), then you need the information federated. So when `following` in the `contact` message is true, you need to send all (public) posts to this user. That's why we currently federate always true for `following`.

So in the export we need `sharing` and `following` for the users setting of the contact, and `receiving` and `subscribed` for if we are allowed to display the profile, and send private messages, and if we need to send new posts to that contact.

So we should export `sharing` and `receiving` from the database, and export `following` with the same value as `sharing` and `subscribed` with the same value as `receiving` (instead of always `true` what I wrote above, because it's wrong). That represents the way how diaspora currently works and also what we currently federate with the `contact` message.

cmrd-senya: Okay, so in terms of the federation:
`sharing` - I send limited posts and profile to the contact
`following` - I request contact to send me public posts
`receiving` - Contact sends me limited posts and profile (when `subscribed` is implemented that is only needed for the green check mark on profile view page actually)
`subscribed` - I send public posts to the contact

Actually this vocabulary feels quite confusing. It's not clear who is `following` who and who is `subscribed` for what out of this words. `following` and `subscribed` are synonyms and if we just swap these two names it won't look like anything changed at all.

How about using the word `followed` instead of `subscribed`? It's better because:
1) It's obvious that it's an opposite of `following` this way
2) It gives grammatical hint: "I am following" -> I read public posts. "I am followed" -> I send public posts

So then we have a simple relation:
`sharing` - I send private messages
`receiving` - I receive private messages
`followed` - I send public messages (I am followed)
`following` - I receive public message

SuperTux88: Yes, that makes sense 👍 (nevertheless we need to document that somewhere, but that's out of scope for this PR).

(for federation it's clear, because it's only what the sender does, but for export we need to export both sides, so `following` and `followed` makes it clearer :) )

cmrd-senya: We can add some description to the schemas actually since JSON schemas are designed to be self-descriptive (and there is a `"description"` key for comments). But I'm not sure that everybody who wants to find this information will manage to guess it's there.

SuperTux88: You can add a `"description"` (doesn't hurt), but we should still document the whole migration feature somewhere. We can do that as part of the federation documentation or do an extra page at http://diaspora.github.io/migration-documentation

cmrd-senya:
I added a diff for your notes. I didn't add a `"descirption"` since I don't have time for that right now. Also I didn't run all the tests locally, so maybe I missed something.

SuperTux88 (approved):

SuperTux88:
Thanks 👍 The diff looks good, so I approved the PR. Let's see what travis says (travis is currently busy anyway :/ ), so if it's green, I'll merge it. (If you want, you can squash it, if not, I'll do that before merging).

And we can do the documentation later, that's not a blocker for 0.7.0.0 :)

cmrd-senya:
Squashed!

SuperTux88:
Merged, thanks for your work! ❤️

app/serializers/export/contact_serializer.rb
SuperTux88: What just came to my mind: We shouldn't trust `receiving` when importing (so that is something we need to remember when adding the import functionality), because otherwise it is a privacy issue, when somebody adds contacts with `receiving` true to their export. But we still need it in the export (so the new format is OK), because we need to send the `account_migration` message to all contacts, even when only `receiving` is true.

For the import, the contact should resend a `contact` message when receiving the `account_migration` message, so we can recreate the contact with `receiving = true`. That's something that can already be added to #6750 at the end of `AccountMigration.perform!`: If new person is remote, then send new `contact` for all users who are sharing with the old person.

cmrd-senya: Could you please decribe in more details what is the issue with privacy when someone sets `receiving` to true before import? I didn't quite get it.

SuperTux88: When one sets `receiving` to true, one could see the private profile of the person (when it's already on the pod for another contact), even when the person never shared the profile with the importing user. And the only way we can be sure one is allowed to see the private profile of a person is, when we ask the person, and not trust the person who says "I'm allowed to see that, trust me!"

So we need it in the export, so we can send the `account_migration` to them, but we should import it with `receiving` set to `false` until the person sends us a `contact` message again, as response to the `account_migration` message.

cmrd-senya: Yes, actually we need some mechanism to sync data between pods, since it is currenlty possible to have non-synchronized contact states accross pods. If one pod misses `Contact` message and then sends `Contact`  message itself for the same users , then this pod will see only `sharing` with `receiving` false, while actually it is `receiving` also. What do you think, maybe we should implement fetching for `Contact`s instead? So that it was possible to update `Contact` state on demand.

Also imagine that we import old archive, and `Contact` state has changed in the federation. Then it is possible that we have `sharing=true`, `receiving=false` in the archive, when the remote pod has `sharing=false`, `receiving=false`. Just setting  `sharing` to `true` would mean that we make the state non-synchronized. So generally we should request all the contacts for the migrating user, not only for `receiving=true`. But also it means that we can't just create them as is in case of mismatch. Because if user has deleted the contact but imports the old archive it doesn't mean he wants to create the contact again. So we should make some UI, that will report the user that there are some mismatches between you and the remote contact and let the user decide what to do: add contact or remove it again (and resend `Contact` if the state has changed). 

Also, since it is possible, that this `Contact` state mismatch happens during the normal operation, not only the migration, we allow that procedure to be called on demand (at least by a podmin). Therefore I think that fetching of `Contacts` from the remote pod for this purposes is a better way than using push model.

SuperTux88: Fetching would need access control, currently it's only possible to fetch public stuff in the protocol. So that's totally out of scope here (and for 0.7.0.0).

We can ask the user which contacts to add again on import, that would fix changed `sharing` states since the backup was created. We can then send new `contact` messages for all contacts from the archive with `sharing`/`following` to `true` or `false`, depending on what the user selected. But that's something for when the import is done, not now.

The only thing we need to do now (for 0.7.0.0) is send `contact` messages again after receiving a `account_migration` for all contacts still `sharing=true` with the migrated user to update the `receiving=true` on the imported users contact. That should be done as part of #6750.

Issue #6696, opened 2016-02-12
API privilege scope escalation


denschub:
I just discovered a rather odd OpenID Connect implementation flaw that allows applications to escalate their permission scope after they were authenticated once.

STR:

1. Generate a authorization URL without any scopes, for example `/api/openid_connect/authorizations/new?response_type=id_token+token&redirect_uri=http%3A%2F%2Fexample.com%2F&nonce=foo&client_id=foo`
2. The user authorizes the application that "requires no permissions" and the app receives the access token.
3. Build another authorization URL, this time including all possible scopes: `/api/openid_connect/authorizations/new?response_type=id_token+token&scope=aud+name+nickname+openid+picture+profile+read+sub+write&redirect_uri=http%3A%2F%2Fexample.com%2F&nonce=foo&client_id=foo`

Expected behavior:

The user receives another authorization dialog, asking for authorizing the new permissions.

Actual behavior:

The app immediately receives a new `access_token` with additional privileges. Looking at `/api/openid_connect/user_applications` does confirm that the app has extended its privileges.

Since the API is not in any stable release and there are no real endpoints, I've decided to make this issue public. @theworldbright, could you take a look?

theworldbright:
Yup thanks for the catch! Sending in a fix right now.

PR #6729, opened 2016-03-02
#6696 - Fix API scope escalation

KentShikama:
See #6696

### DIFF
--- a/app/controllers/api/openid_connect/authorizations_controller.rb
+++ b/app/controllers/api/openid_connect/authorizations_controller.rb
@@ -20,7 +20,8 @@ module Api
       before_action :auth_user_unless_prompt_none!
 
       def new
-        auth = Api::OpenidConnect::Authorization.find_by_client_id_and_user(params[:client_id], current_user)
+        auth = Api::OpenidConnect::Authorization.find_by_client_id_user_and_scopes(params[:client_id],
+                                                                                   current_user, params[:scope])
         reset_auth(auth)
         if logged_in_before?(params[:max_age])
           reauthenticate(params)
@@ -224,7 +225,8 @@ module Api
       def handle_prompt_with_signed_in_user
         client_id = params[:client_id]
         if client_id
-          auth = Api::OpenidConnect::Authorization.find_by_client_id_and_user(client_id, current_user)
+          auth = Api::OpenidConnect::Authorization.find_by_client_id_user_and_scopes(client_id,
+                                                                                     current_user, params[:scope])
           if auth
             process_authorization_consent("true")
           else
--- a/spec/controllers/api/openid_connect/authorizations_controller_spec.rb
+++ b/spec/controllers/api/openid_connect/authorizations_controller_spec.rb
@@ -4,7 +4,6 @@ describe Api::OpenidConnect::AuthorizationsController, type: :controller do
   let!(:client) { FactoryGirl.create(:o_auth_application) }
   let!(:client_with_xss) { FactoryGirl.create(:o_auth_application_with_xss) }
   let!(:client_with_multiple_redirects) { FactoryGirl.create(:o_auth_application_with_multiple_redirects) }
-  let!(:auth_with_read) { FactoryGirl.create(:auth_with_read) }
 
   before do
     sign_in :user, alice
@@ -98,7 +97,7 @@ describe Api::OpenidConnect::AuthorizationsController, type: :controller do
       end
 
       context "when redirect URI does not match pre-registered URIs" do
-        it "should return an invalid request error", focus: true do
+        it "should return an invalid request error" do
           post :new, client_id: client.client_id, redirect_uri: "http://localhost:2000/",
                response_type: "id_token", scope: "openid", nonce: SecureRandom.hex(16)
           expect(response.body).to include("Invalid client id or redirect uri")
@@ -184,10 +183,10 @@ describe Api::OpenidConnect::AuthorizationsController, type: :controller do
     end
 
     context "when already authorized" do
-      let!(:auth) {
-        Api::OpenidConnect::Authorization.find_or_create_by(o_auth_application: client, user: alice,
-                                                            redirect_uri: "http://localhost:3000/", scopes: ["openid"])
-      }
+      before do
+        Api::OpenidConnect::Authorization.create!(
+          o_auth_application: client, user: alice, redirect_uri: "http://localhost:3000/", scopes: ["openid"])
+      end
 
       context "when valid parameters are passed" do
         before do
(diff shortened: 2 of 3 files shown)


app/controllers/api/openid_connect/authorizations_controller.rb
```diff
@@ -20,7 +20,7 @@ class AuthorizationsController < ApplicationController
       before_action :auth_user_unless_prompt_none!
 
       def new
-        auth = Api::OpenidConnect::Authorization.find_by_client_id_and_user(params[:client_id], current_user)
+        auth = Api::OpenidConnect::Authorization.find_by_client_id_user_and_scopes(params[:client_id], current_user, params[:scope])
```
diaspora-code-review: Line is too long. [132/120]

app/controllers/api/openid_connect/authorizations_controller.rb
```diff
@@ -224,7 +224,7 @@ def handle_prompt_none
       def handle_prompt_with_signed_in_user
         client_id = params[:client_id]
         if client_id
-          auth = Api::OpenidConnect::Authorization.find_by_client_id_and_user(client_id, current_user)
+          auth = Api::OpenidConnect::Authorization.find_by_client_id_user_and_scopes(client_id, current_user, params[:scope])
```
diaspora-code-review: Line is too long. [125/120]

app/presenters/user_applications_presenter.rb
```diff
@@ -4,10 +4,12 @@ def initialize(user)
...
-    @applications ||= @user.o_auth_applications.map do |app|
-      authorization = Api::OpenidConnect::Authorization.find_by_client_id_and_user(app.client_id, @user)
-      UserApplicationPresenter.new app, authorization.scopes, authorization.id
-    end
+    @applications ||= Array(@user.o_auth_applications.map do |app|
+      authorizations = Api::OpenidConnect::Authorization.find_all_by_client_id_and_user(app.client_id, @user)
+      authorizations.map {|authorization|
+        UserApplicationPresenter.new app, authorization.scopes, authorization.id
+      }
+    end.flatten!)
```
diaspora-code-review: Avoid chaining a method call on a do...end block.

db/migrate/20150708153926_create_authorizations.rb
```diff
@@ -3,7 +3,7 @@ def change
     create_table :authorizations do |t|
       t.belongs_to :user, index: true
       t.belongs_to :o_auth_application, index: true
-      t.string :refresh_token
+      t.string :refresh_token, index: {unique: true, length: 191}
```
jhass: This should cause some `db/schema.rb` change to be included, no?

jhass:
Shouldn't creating the new authorization destroy/remove the old one?

app/presenters/user_applications_presenter.rb
```diff
@@ -4,10 +4,13 @@ def initialize(user)
...
-      authorization = Api::OpenidConnect::Authorization.find_by_client_id_and_user(app.client_id, @user)
-      UserApplicationPresenter.new app, authorization.scopes, authorization.id
+    application_array = @user.o_auth_applications.map do |app|
+      authorizations = Api::OpenidConnect::Authorization.find_all_by_client_id_and_user(app.client_id, @user)
+      authorizations.map {|authorization|
+        UserApplicationPresenter.new app, authorization.scopes, authorization.id
+      }
     end
+    @applications ||= Array(application_array.flatten!)
   end
```
jhass: This should be possible with a single query actually, something like `Api::OpenidConnect::Authorization.joins(:o_auth_application).where(user: @user, o_auth_application: {user: @user}).map` perhaps?

db/migrate/20150708153926_create_authorizations.rb
SuperTux88: And I think this shouldn't be edited here, because every pod that has updated already wouldn't run this again and this index would be missing on these pods.

theworldbright:
@jhass I was thinking that it would be handy for a developer to test out the various scopes. But I suppose they could just create multiple users?

db/migrate/20150708153926_create_authorizations.rb
theworldbright: Yeah I forgot we aren't in the middle of that massive PR where I was free to redo the migrations each time :+1:

jhass:
My concern is that it might be quite confusing to the user when one application appears multiple times.

denschub:
Agreed, I'd drop existing authorizations.

db/migrate/20150708153926_create_authorizations.rb
theworldbright: I think I'll just take this out for this PR. I don't think we need to enforce `SecureRandom.hex(32)` to be unique.

app/presenters/user_applications_presenter.rb
theworldbright: I assume this will not apply once creating the new authorization destroys the old one?

theworldbright: I'm about to send in an update to this PR.

spec/controllers/api/openid_connect/authorizations_controller_spec.rb
```diff
@@ -184,10 +183,9 @@
     end
 
     context "when already authorized" do
-      let!(:auth) {
-        Api::OpenidConnect::Authorization.find_or_create_by(o_auth_application: client, user: alice,
-                                                            redirect_uri: "http://localhost:3000/", scopes: ["openid"])
-      }
+      before do
+        Api::OpenidConnect::Authorization.create!(o_auth_application: client, user: alice, redirect_uri: "http://localhost:3000/", scopes: ["openid"])
```
diaspora-code-review: Line is too long. [150/120]

spec/controllers/api/openid_connect/authorizations_controller_spec.rb
```diff
@@ -231,6 +229,23 @@
...
+              scope: "openid read", nonce: 413_093_098_3, state: 413_093_098_3
+        end
+
+        it "should receive another authorization request" do
+          expect(response.body).to match("Diaspora Test Client")
+        end
+
+        it "should overwrite old authorization scope after approval" do
+          post :create, approve: "true"
+          authorization_with_old_scope = Api::OpenidConnect::Authorization.find_by_client_id_user_and_scopes(client.client_id, alice, ["openid"])
```
diaspora-code-review: Line is too long. [145/120]

jhass:
Thank you!

PR #6732, opened 2016-03-03
Cleanup aspect visibilities

SuperTux88:
This is based on #6723, because I needed the new stream-query-refactorings.

This cleans up most of the `aspect_visibilities` table:
- with nerdpol.ch database:
  - before: 560k rows
  - after: 32k rows

The `aspect_visibilities` for public posts didn't work anyway, because after creating a new aspect, all old public posts were missing in this aspect-stream

### DIFF
--- a/app/controllers/photos_controller.rb
+++ b/app/controllers/photos_controller.rb
@@ -150,10 +150,12 @@ class PhotosController < ApplicationController
     @photo = current_user.build_post(:photo, params[:photo])
 
     if @photo.save
-      aspects = current_user.aspects_from_ids(params[:photo][:aspect_ids])
 
       unless @photo.pending
-        current_user.add_to_streams(@photo, aspects)
+        unless @photo.public?
+          aspects = current_user.aspects_from_ids(params[:photo][:aspect_ids])
+          current_user.add_to_streams(@photo, aspects)
+        end
         current_user.dispatch_post(@photo, :to => params[:photo][:aspect_ids])
       end
 
--- a/app/controllers/reshares_controller.rb
+++ b/app/controllers/reshares_controller.rb
@@ -11,7 +11,6 @@ class ResharesController < ApplicationController
     end
 
     if @reshare.save
-      current_user.add_to_streams(@reshare, current_user.aspects)
       current_user.dispatch_post(@reshare, :url => post_url(@reshare), :additional_subscribers => @reshare.root_author)
       render :json => ExtremePostPresenter.new(@reshare, current_user), :status => 201
     else
--- a/app/models/user/querying.rb
+++ b/app/models/user/querying.rb
@@ -134,7 +134,10 @@ module User::Querying
     query = opts[:klass].where(conditions)
 
     unless opts[:all_aspects?]
-      query = query.joins(:aspect_visibilities).where(aspect_visibilities: {aspect_id: opts[:by_members_of]})
+      query = query.with_aspects.where(
+        AspectVisibility.arel_table[:aspect_id].in(opts[:by_members_of])
+          .or(opts[:klass].arel_table[:public].eq(true))
+      )
     end
 
     ugly_select_clause(query, opts)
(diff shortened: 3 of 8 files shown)


lib/diaspora/federated/shareable.rb
```diff
@@ -101,18 +101,18 @@ def update_existing_sharable(shareable)
...
+        def receive_shareable_visibility(user, shareable)
+          user.receive_shareable(shareable)
           user.notify_if_mentioned(shareable)
           logger.info "event=receive payload_type=#{self.class} status=complete " \
-                      "sender=#{diaspora_handle} receiver=#{person.diaspora_handle} guid=#{shareable.guid}"
+                      "sender=#{diaspora_handle} receiver=#{user.diaspora_handle} guid=#{shareable.guid}"
         end
 
-        def receive_non_persisted(user, person)
+        def receive_non_persisted(user)
```
diaspora-code-review: Assignment Branch Condition size for receive_non_persisted is too high. [23.43/20]

spec/models/post_spec.rb
```diff
@@ -148,6 +148,22 @@
           Post.for_visible_shareable_sql(Time.now + 1, "created_at")
         end
 
+        context "with two posts with the same timestamp" do
+          before do
+            aspect_id = alice.aspects.where(name: "generic").first.id
+            Timecop.freeze Time.now do
```
diaspora-code-review: Do not use `Time.now` without zone. Use one of `Time.zone.now`, `Time.now.current`, `Time.now.in_time_zone`, `Time.now.utc`, `Time.now.getlocal`, `Time.now.iso8601`, `Time.now.jisx0301`, `Time.now.rfc3339`, `Time.now.to_i`, `Time.now.to_f` instead.

spec/models/post_spec.rb
```diff
@@ -148,6 +148,22 @@
...
+              alice.post(:status_message, text: "first", to: aspect_id)
+              alice.post(:status_message, text: "second", to: aspect_id)
+            end
+          end
+
+          it "returns them in reverse creation order" do
+            posts = Post.for_visible_shareable_sql(Time.now + 1, "created_at")
```
diaspora-code-review: Do not use `Time.now` without zone. Use one of `Time.zone.now`, `Time.now.current`, `Time.now.in_time_zone`, `Time.now.utc`, `Time.now.getlocal`, `Time.now.iso8601`, `Time.now.jisx0301`, `Time.now.rfc3339`, `Time.now.to_i`, `Time.now.to_f` instead.

Flaburgan:
>all old public posts were missing in this aspect-stream

You mean when the user select the new aspect in the left column, there are no posts in it? And we would have expected his prrevious public posts to be there?

SuperTux88:
When the user posts a public post, then creates a new aspect, and selects only this aspect in the aspect-stream, the public post is not shown in this stream. This is fixed now, all public posts from the user are in all aspect-streams no matter when they were created.

Flaburgan:
>all public posts from the user are in all aspect-streams

I was wondering if this is the wanted behaviour.

SuperTux88:
The aspect-stream shows all posts from the contacts in this aspect, and it shows all posts from you, that were shared with this aspect. As public posts are visible from persons in all aspects (even new ones), they should appear in all aspect-streams.

Flaburgan:
Yeah I get that logic, but I can also imagine another point of view where the user is only interested by the posts he explicitly made in that specific aspect and doesn't want to be flooded by all this public posts and reshares. I don't know what's the best, so I just wanted to point the other possibility which could also be a valid use case.

SuperTux88:
yeah, but then the old logic was also wrong, because it showed public posts, but not in all aspects ;) I think this solution is the best, most consistent and the closest to the old logic.

SuperTux88:
rebased and tested with postgres and mysql :)

jhass:
Thanks!

PR #6735, opened 2016-03-04
Update gems

skyschub:
Mainly to get rid of the insecure Actionpack version. The issue is not relevant to us, but having insecure versions in our gemset is not nice becaue it makes our badge so red...
#3855


jhass:
Merged, thanks!

PR #6740, opened 2016-03-05
Refactor invitations and public stream cukes

svbergerem:
We had two files for invitation cukes which I merged into one. The public stream cuke now only checks that there is a post on the public stream. The visibility of posts is checked by rspec tests.


denschub:
Every removed cuke is a good cuke. Thanks.

Issue #4493, opened 2013-09-01
@mentions not working for names in Malayalam language


balasankarc:
Cannot mention someone using @mentions if they are using their name in Malayalam Language (language code : ml) .. For eg: I cannot mention @ബിപിൻദാസ് in Diaspora, because his name is not written in English.

cmrd-senya:
This is probably a blocker for the issue: https://github.com/podio/jquery-mentions-input/issues/117

svbergerem:
This might have been fixed on the develop branch in #6728. Would be great if someone could try this again.

denschub:
Closing this. Please reopen if this is still an issue (which I doubt :))

PR #6741, opened 2016-03-05
Fix typeahead suggestions with nonlatin chars

### LINKED ISSUE #6242: Mention popup doesn't work for Cyrillic names
````@````-completion doesn't work for contacts who have Cyrillic names for me on 0.5.1.1-p9bef1053, only names with Latin characters appear.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24883097-mention-popup-doesn-t-work-for-cyrillic-names?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #6242 and #4493. Before we were using [`Bloodhound.tokenizer.nonword`](https://github.com/twitter/typeahead.js/blob/250a5dc/src/bloodhound/tokenizers.js#L24) which splits the username at all chars in `\W`.

### DIFF
--- a/app/assets/javascripts/app/views/search_base_view.js
+++ b/app/assets/javascripts/app/views/search_base_view.js
@@ -10,13 +10,18 @@ app.views.SearchBase = app.views.Base.extend({
     if(options.autoselect) { this.setupAutoselect(); }
   },
 
+  bloodhoundTokenizer: function(str) {
+    if(typeof str !== "string") { return []; }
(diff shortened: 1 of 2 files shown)


denschub:
Nice, thanks.

PR #6727, opened 2016-02-29
Improve pod check

SuperTux88:
This improves the check added with #6290 and is a preparation to fix #6220 with the new federation.
- add a port to the pod, so we can check pods with ports
- link the people with pods, so the url is not stored redundant. My pod had many inconsistent people (from the same pod, some with http and some with https).
- use the /etc/hosts for checks. I had my local test-pods in the `/etc/hosts`, everything (federation) worked, but the check failed. Now the check works the same as the federation.

I also fixed some "false up"s:
- Some old pods now host a website or a dummy page, so they respond `200 OK` on `<pod-url>/`. I added an additional test for `<pod-url>/.well-known/host-meta`. This path is needed for diaspora. This way we can check if the domain really hosts a working pod. We can remove this check once every pod has NodeInfo.
- Some old pods redirected to other domains. We can consider these pods as down, because the domain can't be changed.

PR #6743, opened 2016-03-06
fix pod table migration if someone deleted a user (owner) manually

SuperTux88:
@jpope777 deleted a user manually in the database, so the owner of the person was nil and the migration created a `Pod` for his own pod and linked this with all other users. This is now fixed.

Follow up for #6727


jhass:
Thank you!

jpope777:
w00t! Thanks. :D

PR #6744, opened 2016-03-07
Refactor services

SuperTux88:
I unified the services a bit and make them more reusable:
- put view stuff back to the controllers (parameters, presenters).
- don't save states in the service. A service is created for a user but has no other state.
- parameters for the actions are at the method for the action now. So you don't need to know what you want to do with the service when you create it.
- and I wrote specs for all the services.

Also removed some old unused code with this:
- the `iframe` route was never accessed on my production pod in the last days. The `oembed` route returns the iframe.
- remove "not public" error page. Logged in users see the normal 404 and not logged in users are redirected to the login page.

Also fixed some other stuff:
- mark attached photos as public if the StatusMessage is also public. So they would federation with the correct public flag.
- move the id/guid switch to 16 chars ( < 16 id, >= 16 guid). I don't know where the 8 char limit comes from, but the used `hex(8)` is 16 chars long and I have no guid shorter than 16 chars in my database.
- when a user opens the SPV for a post, all notifications are marked as read. I optimised the query a bit and I don't update the `updated_at` field, because this is displayed in the frontend as the date, when the notification was made (or when the last actor was added). So it is weird if it updates when I read the notification. (this is also at more places in the code, but I only fixed the SVP for now)


denschub:
Merged, thank you very much! :cake:

theworldbright:
@SuperTux88 Learned how a code a bit more cleanly from this PR :+1: Thanks for the cleaning up the services.

PR #6745, opened 2016-03-07
Remove diaspora_handle from shareables

cmrd-senya:
We can determine diaspora_handle from a relation with people for the
shareables (posts, photos). So we don't need to store diaspora_handle in
the DB. Also remove tmp_old_id from photos which is not refenrenced anywhere.


denschub:
Fair enough, merged, thanks!

PR #6747, opened 2016-03-09
Fix usage of hard-coded pod URI in rspec testsuite

cmrd-senya:


jhass:
for the path helper cases, doesn't using `_uri` instead of `_path` work? For the other cases, we got `AppConfig.url_to`.

cmrd-senya:
```post_url(@post)``` returns a value with ```test.host``` as a host value, and the test therefore fails.

**Comment by cmrd-senya** [comment]
I rebased it upon the latest develop branch and the tests are green.

File: spec/helpers/posts_helper_spec.rb [hunk 1]
```diff
@@ -32,7 +32,7 @@
     end
 
     it "returns an iframe containing the post" do
-      expect(post_iframe_url(@post.id)).to include "src='http://localhost:9887#{post_path(@post)}'"
+      expect(post_iframe_url(@post.id)).to include "src='#{AppConfig.pod_uri.to_s.chop}#{post_path(@post)}'"
```
jhass: Why not `AppConfig.url_to` here too?

spec/lib/diaspora/markdownify_email_spec.rb
```diff
@@ -8,12 +8,12 @@
 
     it 'should autolink a hashtag' do
       markdownified = @html.preprocess("#tag")
-      expect(markdownified).to eq("[#tag](http://localhost:9887/tags/tag)")
+      expect(markdownified).to eq("[#tag](#{AppConfig.url_to('tags/tag')})")
```
jhass: Maybe update those to use `tag_path` while at it?

cmrd-senya: It gives me
```
     NoMethodError:
       undefined method `tag_path' for #<RSpec::ExampleGroups::DiasporaMarkdownifyEmail::Preprocess:0x00000009d073a8>
```
for that file.

I suppose that happens because it is not a controller and not the Rails application context.

jhass: Doing `include Rails.application.routes.url_helpers` in the toplevel `describe` block should make it available.

cmrd-senya:
@jhass, changes applied, test are green.

cmrd-senya:
So I can't replace ```AppConfig.url_to(*_path())``` with ```url_to()``` because the latter uses ```test.host``` as a host value instead of our real url. I wonder if it worth investigation.

PR #6748, opened 2016-03-09
https oembed

SuperTux88:
Use `AppConfig.url_to` instead of `_url` which didn't generate the pod-url with https if the pod has https.

Also fixed the `jsxc_helper_spec.rb`. It changed the `AppConfig.environment.url` permanently for all following specs. Mock it temporarily now.

### DIFF
--- a/spec/helpers/jsxc_helper_spec.rb
+++ b/spec/helpers/jsxc_helper_spec.rb
@@ -2,20 +2,19 @@ require 'spec_helper'
 
 describe JsxcHelper, :type => :helper do
   before do
-    AppConfig.chat.server.bosh.proxy = false
     AppConfig.chat.server.bosh.port = 1234
-    AppConfig.chat.server.bosh.bind = '/bind'
-    AppConfig.environment.url = "https://localhost/"
-    AppConfig.instance_variable_set(:@pod_uri, nil)
+    AppConfig.chat.server.bosh.bind = "/bind"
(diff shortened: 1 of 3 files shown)


cmrd-senya:
Aha, I see, that is what breaks the testsuite in #6747. Must be merged
first then.

spec/helpers/jsxc_helper_spec.rb
```diff
@@ -2,20 +2,20 @@
...
   describe "#get_bosh_endpoint" do
     it "using http scheme and default values" do
+      AppConfig.chat.server.bosh.proxy = false
       expect(helper.get_bosh_endpoint).to include %Q(http://localhost:1234/bind)
     end
 
     it "using https scheme and no port" do
       AppConfig.chat.server.bosh.proxy = true
+      allow_any_instance_of(Configuration::Methods).to receive(:pod_uri)
+        .and_return(Addressable::URI.parse("https://localhost/"))
```
jhass: `AppConfig` should be the only instance, so stubbing it directly should be fine.

SuperTux88: Fixed.

denschub:
Merged as 6ea8cd546e08fbcdfe46576b9e210ac1533eb36f 67fbd02ae045ffd12f338eb7b0ff4aeec20adc6f

PR #6750, opened 2016-03-14
Account migration model/message

cmrd-senya:
This PR introduces support for AccountMigration federation message receive. It covers the cases when the new home pod for a user is remote respective to the recepient pod of the message. It also allows to initiate migration locally by a podmin from the rails console. This will give the pods a possibility to understand the account migration event on the federation level and thus future version which will implement migration will be backward compatible with the pods starting from this PR.

TODO:
- [x] Add the required test cases for every migration scenario
- [x] Add `AccountMigration`message dispatch feature for locally initiated migration scenarios
- [x] Refactor account_migration_spec.rb to a less mad structure
- [x] Add contact update feature for remotely inititated migration scenarios
- [x] Unit tests for new modules
- [x] Fix account_deleter_spec.rb (see travis)

### DIFF
--- a/app/models/account_migration.rb
+++ b/app/models/account_migration.rb
@@ -1,0 +1,165 @@
+class AccountMigration < ApplicationRecord
+  include Diaspora::Federated::Base
+
+  belongs_to :old_person, class_name: "Person"
+  belongs_to :new_person, class_name: "Person"
+
+  validates :old_person, uniqueness: true
+  validates :new_person, uniqueness: true
+
+  after_create :lock_old_user!
+
+  attr_accessor :old_private_key
+
+  def receive(*)
+    perform!
+  end
+
+  def public?
+    true
+  end
+
+  def sender
+    @sender ||= old_user || ephemeral_sender
+  end
+
+  # executes a migration plan according to this AccountMigration object
+  def perform!
+    raise "already performed" if performed?
+
+    ActiveRecord::Base.transaction do
+      account_deleter.tombstone_person_and_profile
+      account_deleter.close_user if user_left_our_pod?
+      account_deleter.tombstone_user if user_changed_id_locally?
+
+      update_all_references
+    end
+
+    dispatch if locally_initiated?
+    dispatch_contacts if remotely_initiated?
+  end
+
+  def performed?
+    old_person.closed_account?
+  end
+
+  # We assume that migration message subscribers are people that are subscribed to a new user profile updates.
+  # Since during the migration we update contact references, this includes all the contacts of the old person.
+  # In case when a user migrated to our pod from a remote one, we include remote person to subscribers so that
+  # the new pod is informed about the migration as well.
+  def subscribers
+    new_user.profile.subscribers.remote.to_a.tap do |subscribers|
+      subscribers.push(old_person) if old_person.remote?
+    end
+  end
+
+  private
+
+  # Normally pod initiates migration locally when the new user is local. Then the pod creates AccountMigration object
+  # itself. If new user is remote, then AccountMigration object is normally received via the federation and this is
+  # remote initiation then.
+  def remotely_initiated?
+    new_person.remote?
+  end
+
+  def locally_initiated?
+    !remotely_initiated?
+  end
+
+  def old_user
+    old_person.owner
+  end
+
+  def new_user
+    new_person.owner
+  end
+
+  def lock_old_user!
+    old_user&.lock_access!
+  end
+
+  def user_left_our_pod?
+    old_user && !new_user
+  end
+
+  def user_changed_id_locally?
+    old_user && new_user
+  end
+
+  # We need to resend contacts of users of our pod for the remote new person so that the remote pod received this
+  # contact information from the authoritative source.
+  def dispatch_contacts
+    new_person.contacts.sharing.each do |contact|
+      Diaspora::Federation::Dispatcher.defer_dispatch(contact.user, contact)
+    end
+  end
+
+  def dispatch
+    Diaspora::Federation::Dispatcher.build(sender, self).dispatch
+  end
+
+  EphemeralUser = Struct.new(:diaspora_handle, :serialized_private_key) do
+    def id
+      diaspora_handle
+    end
+
+    def encryption_key
+      OpenSSL::PKey::RSA.new(serialized_private_key)
+    end
+  end
+
+  def ephemeral_sender
+    raise "can't build sender without old private key defined" if old_private_key.nil?
+    EphemeralUser.new(old_person.diaspora_handle, old_private_key)
+  end
+
+  def update_all_references
+    update_person_references
+    update_user_references if user_changed_id_locally?
+  end
+
+  def person_references
+    references = Person.reflections.reject {|key, _|
+      %w[profile owner notifications pod].include?(key)
+    }
+
+    references.map {|key, value|
+      {value.foreign_key => key}
+    }
+  end
+
+  def user_references
+    references = User.reflections.reject {|key, _|
+      %w[
+        person profile auto_follow_back_aspect invited_by aspect_memberships contact_people followed_tags
+        ignored_people conversation_visibilities pairwise_pseudonymous_identifiers conversations o_auth_applications
+      ].include?(key)
+    }
+
+    references.map {|key, value|
+      {value.foreign_key => key}
+    }
+  end
+
+  def update_person_references
+    logger.debug "Updating references from person id=#{old_person.id} to person id=#{new_person.id}"
+    update_references(person_references, old_person, new_person.id)
+  end
+
+  def update_user_references
+    logger.debug "Updating references from user id=#{old_user.id} to user id=#{new_user.id}"
+    update_references(user_references, old_user, new_user.id)
+  end
+
+  def update_references(references, object, new_id)
+    references.each do |pair|
+      key_id = pair.flatten[0]
+      association = pair.flatten[1]
+      object.send(association).update_all(key_id => new_id)
+    end
+  end
+
+  def account_deleter
+    @account_deleter ||= AccountDeleter.new(old_person)
+  end
+end
--- a/db/migrate/20170730154117_create_account_migrations.rb
+++ b/db/migrate/20170730154117_create_account_migrations.rb
@@ -1,0 +1,14 @@
+class CreateAccountMigrations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :account_migrations do |t|
+      t.integer :old_person_id, null: false
+      t.integer :new_person_id, null: false
+    end
+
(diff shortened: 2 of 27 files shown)


cmrd-senya:
I've added some frontend

cmrd-senya:
depends on #6780 #6783

cmrd-senya:
Rspec has failed because of `ApplicationRecord`, but that's what RuboCop told me to do. I'm not guilty!

SuperTux88:
Yes, saw that. `ApplicationRecord` will come with the rails5 PR, so you are already prepared 👍

SuperTux88 (commented):
Some small comments, otherwise it looks OK so far. I also added a todo in the PR description for the currently failing travis tests.

app/models/account_migration.rb
```diff
@@ -33,6 +45,17 @@ def perform!
 
       update_all_references
     end
+
+    dispatch if locally_initiated?
+  end
+
+  # TODO: docs
+  def subscribers
+    raise "You shouldn't federate AccountMigration if it was remotely initiated" if remotely_initiated?
```
SuperTux88: I don't think we need this. Our code will never do this, and when somebody wants to trigger that manually with rails console, then a pod can only do it when having both private keys anyway, which is only when locally initiated. The old home pod has only the private key of the old user and can't create a valid message.

db/migrate/20170730154117_create_account_migrations.rb
```diff
@@ -0,0 +1,11 @@
+class CreateAccountMigrations < ActiveRecord::Migration[5.1]
+  def change
+    create_table :account_migrations do |t|
+      t.integer :old_person_id, null: false
+      t.integer :new_person_id, null: false
+    end
+
+    add_foreign_key :account_migrations, :people, column: :old_person_id
+    add_foreign_key :account_migrations, :people, column: :new_person_id
+  end
```
SuperTux88: I think you should also add unique indexes here. Even when it should never happen (because rails validates it), we had that already in the past, somehow the message was received twice and while validating it was still unique both times, but then both saved at the same time. So I trust the database more than rails ;)

SuperTux88 (commented):
Some more feedback :) Still haven't looked at `account_migration_spec.rb` because you want to refactor that.

cmrd-senya:
Okay, it looks like it is functionally complete. I also need to make some fixes and refactorings to the tests, but otherwise it's ready for testing.

db/migrate/20170730154117_create_account_migrations.rb
```diff
@@ -0,0 +1,13 @@
...
+  def change
+    create_table :account_migrations do |t|
+      t.integer :old_person_id, null: false
+      t.integer :new_person_id, null: false
+    end
+
+    add_foreign_key :account_migrations, :people, column: :old_person_id
+    add_foreign_key :account_migrations, :people, column: :new_person_id
+
+    add_index :account_migrations, %i[old_person_id new_person_id], unique: true
```
SuperTux88: There should also be an index for only `old_person_id`, I don't know if we want to support merging multiple accounts. Currently that wouldn't work (needs more work, merging contacts for example), and is out of scope for now, but that's something that's maybe possible later when doing the import. If not we need a unique index for `new_person_id`, if yes this combined index is enough, because every combination needs to be unique.

cmrd-senya: Since we don't support this yet then we should probably add another unique index for `new_person_id`. We can remove this index if we introduce this feature afterwards, right? Also we have separate `unique` validation for new and old person in Rails now.

SuperTux88: > We can remove this index if we introduce this feature afterwards, right?

Yes, we need to remove the index when we add support for receiving merge migrations. Import an export to an existing profile would again be a second step for that. But as said, that's something for later. If we already know we want to support merging later, we can omit the index here, because the rails validation should do it for now. (then we don't need to add the index here and remove it later again).

> Also we have separate `unique` validation for new and old person in Rails now.

As said, I don't trust rails validations ;) They do a select and see if the object exists, and then save it. We already had duplicates with rails validation in the past, because things happened in parallel, both did the select, saw that there is nothing there, and saved it ;) So every unique validation in rails should be covered by the database (even when the risk for this to happen is really small, because it is federated via public receiver, so only once per pod). I'd only make an exception if we know we want to remove that index again. That way support for receive merging could be added in a minor without the need of a migration.

app/models/account_migration.rb
```diff
@@ -0,0 +1,155 @@
...
+    end
+
+    dispatch if locally_initiated?
+    dispatch_contacts if remotely_initiated?
+  end
+
+  # TODO: docs
+  def subscribers
+    subscribers = new_user.profile.subscribers.remote
+    return subscribers if old_person.local?
```
SuperTux88: Nit: I think with a tap (similar like [here](https://github.com/diaspora/diaspora/blob/d0b634852b440a8c3deac9734270d8577ef5a06b/app/models/post.rb#L151-L153)) this is more readable.

spec/integration/federation/receive_federation_messages_spec.rb
```diff
@@ -21,6 +21,27 @@
...
+
+        new_person = Person.find_by(diaspora_handle: new_person_entity.diaspora_id)
+        expect(new_person).not_to be_nil
+        expect(AccountMigration.exists?(old_person_id: sender.person.id, new_person_id: new_person.id)).to be_truthy
+      end
+
+      it "rejects account deletion with wrong diaspora_id" do
+        delete_id = Fabricate.sequence(:diaspora_id)
+        post_message(generate_payload(DiasporaFederation::Entities::AccountDeletion.new(diaspora_id: delete_id), sender))
+
```
SuperTux88: missing `end`

SuperTux88 (commented):
Some small things again, otherwise the code itself looks good, so waiting for you to finish the tests 👍

cmrd-senya:
Okay this must be ready for review now

spec/support/data_generator.rb
```diff
@@ -38,6 +38,7 @@ def generic_user_data
     private_post_as_receipient
     tag_following
     generic_person_data
+    remote_mutual_friend
```
SuperTux88: Thanks, no need to depend on `conversations` anymore to create a mutual contact ❤️

spec/models/account_migration_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    it "calls AccountDeleter#tombstone_person_and_profile" do
+      expect(embedded_account_deleter).to receive(:tombstone_person_and_profile)
+      account_migration.perform!
+    end
+
+    # TODO: This test is very slow. Also it is duplicated in the integration test.
+    # TODO: Should we remove it or leave? It's possible to optimize the test in a manner
+    # TODO: like account deletion integration test work. It'll make it a slightly faster.
+    # it_behaves_like "it updates person references" do
```
SuperTux88: I don't think we need this twice, but shoud we do a faster test and mock `update_references` and test if it's called once per reference? Similar to something like this:

https://github.com/diaspora/diaspora/blob/b0797ff9f47d2b6b49fcf3a0595c34e814ad9ce3/spec/lib/account_deleter_spec.rb#L75-L91

cmrd-senya: Well, `update_references` is private and it's generally not very good to mock private methods. We could do that if it was public or if it was incapsulated in another class. It doesn't make too much sense to make it public for `AccountMigration` but we could actually extract it somewhere, but where?

spec/integration/account_migration_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+    end
+    inlined_jobs { run_migration }
+  end
+end
+
+shared_examples_for "migration scenarios initiated locally" do
+  it "dispatches account migration message to the federation" do
+    expect(DiasporaFederation::Federation::Sender).to receive(:public) do |sender_id, obj_str, urls, xml|
+      if old_user.person.remote?
+        expect(sender_id).to be_nil
```
SuperTux88: Didn't realize that while reading the code, but reading the test makes it clearer: sender_id is only used for logging, so nil is OK, but how about using the old diaspora ID here to have a more useful logging?

spec/models/account_migration_spec.rb
cmrd-senya: Though I think I have a better idea about it

SuperTux88: But `expect(account_migration).to receive(:update_references).with(...)` should work for private too? I also only would do one test, and then expecting it to be called for every reference. But that would be a more unit-test-like way instead of an integration test.

But if you think it isn't worth it, you can just delete it, it doesn't make much sense to do the same (slow) test again. (Maybe also move the shared example to the integration test then).

SuperTux88: I don't think `model.find(old_id).send(association).update_all(key_id => new_id)` should be public? This is a very special case and should only be used here.

spec/integration/federation/receive_federation_messages_spec.rb
```diff
@@ -29,6 +29,80 @@
...
+        end
+      end
+
+      context "with discovered profile" do
+        let(:new_user) { create_remote_user("example.org") }
+
+        it "updates person profile with data from entity" do
+          new_user.profile.bio = "my updated biography"
+          expect(entity.profile.bio).to eq("my updated biography")
+          new_user.profile.reload
```
SuperTux88: Maybe replace this line with `expect(new_user.profile.reload.bio).not_to eq("my updated biography")` to ensure it's saved nowhere.

spec/models/account_migration_spec.rb
cmrd-senya: Well, I think it is a better approach anyway, than mocking a private method, since tests shouldn't rely on private methods. Private methods are supposed to be allowed to change freely the way they work and named. While there is nothing wrong in having a very special public method (we have many very special scopes which are used only once). And this will allow us to write a proper test without relying on private methods and checking attributes directly as in the integration test.

I think that we could make a helper module encapsulating this `model.find(old_id).send(association).update_all(key_id => new_id)` behavior, and include it to our `AccountMigration` model.  We can make this module publiclty accessible via a scope inside `AccountMigration` model to give an idea that this behavior is account migration specific and isn't expected to be used outside. Something like `AccountMigration::ReferenceUpdater`. What do you think?

SuperTux88 (approved):
Thank you very much, that was a lot of work! ❤️ 

Here, have some cookies: 🍪 🍪 🍪 🍪 🍪 🍪 🍪 🍪 🍪 🍪

cmrd-senya:
Thank you too for a lot of work and cookies from your side! I'll celebrate this with a cup of tea :tea:

Issue #6557, opened 2015-11-21
Upgrade to jQuery 2


AugierLe42e:
diaspora currently relies on jQuery 1.11.2. Nothing stops us to upgrade to jQuery 2.1.4 except diaspora_jsxc which requires exactly 1.11.2. I just tested which a patched version of the diaspora_jsxc locally, everything seems to work well. @Zauberstuhl, do you have an opposition onn patching diaspora_jxsc so we can upgrade to jQuery 2?

svbergerem:
> jQuery 2.x has the same API as jQuery 1.x, but does not support Internet Explorer 6, 7, or 8. [...] Since IE 8 is still relatively common, **we recommend using the 1.x version unless you are certain no IE 6/7/8 users are visiting the site**.

Is there any reason why we should switch to jQuery 2.x?

Flaburgan:
@svbergerem we don't actually support IE 6/7/8 at all. I don't know about the current code but 3 years ago diaspora* wasn't working at all with those browsers and we didn't work on this since that moment, so I assume this is not a blocker.

AugierLe42e:
> Is there any reason why we should switch to jQuery 2.x?

jQuery 2.x code is lighter so I suppose that optimises performance. Originally, this is related to #6555 whiwh uses Hammer.JS. The jQuery plugin depends on jQuery 2. This is not really blocking as I've just copied the file in `vendor/assets/javascripts` instead of using th rails-asset gem, but I though it could be interresting to raise this issue to discuss the topic.

Zauberstuhl:
that could hurt. but I can try it in the next view days. also I have to talk with upstream about it

diaspora/jsxc#134

Zauberstuhl:
//adding @pravi 

he is currently checking if we can use jquery 2 out of the box with the debian package

AugierLe42e:
Ping.

pravi:
@AugierLe42e https://github.com/diaspora/jsxc/issues/134#issuecomment-160160358

svbergerem:
Fixed by #6944.

PR #6751, opened 2016-03-14
remove ChromeFrame

SuperTux88:
Google Chrome Frame was a plugin for Internet Explorer, but is no longer
supported. And it was included with http and not https which triggered a
warning on most pods.

Also set the minimum version to 9 to support #6557


config.ru
```diff
@@ -6,14 +6,14 @@
 
 require ::File.expand_path('../config/environment',  __FILE__)
 require ::File.expand_path('../lib/unicorn_killer',  __FILE__)
-require ::File.expand_path('../lib/rack/chrome_frame',  __FILE__)
+require ::File.expand_path('../lib/rack/internet_explorer_version', __FILE__)
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

lib/rack/internet_explorer_version.rb
```diff
@@ -0,0 +1,41 @@
...
+#   the COPYRIGHT file.
+
+module Rack
+  class InternetExplorerVersion
+    def initialize(app, options={})
+      @app = app
+      @options = options
+    end
+
+    def call(env)
```
diaspora-code-review: Method has too many lines. [22/20]

spec/lib/rack/internet_explorer_version_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  before :all do
+    @app = Rack::Builder.parse_file(Rails.root.join("config.ru").to_s).first
+  end
+
+  subject { get_response_for_user_agent(@app, ua_string) }
+
+  context "non-IE browser" do
+    let(:ua_string) { "another browser chromeframe" }
+
+    it "shouldn't complain about the browser and shouldn't have chrome frame" do
```
svbergerem: > shouldn't have chrome frame

You don't test this here so I'd drop that part of the description.

jhass:
Merged as d97448e5523f1a8614fe9297356ecfabbc4d1360

Thanks!

PR #6764, opened 2016-03-21
Unify mobile and desktop registration and session forms

svbergerem:
### before

before sign up
before sign in
### after

after sign up
after sign in


svbergerem:
I now also increased the space between the footer links a bit:

sign in after footer links

Flaburgan:
Looks cleaner, thank you :)

denschub:
Thanks!

PR #6768, opened 2016-03-25
Use Poltergeist instead of Selenium

svbergerem:
This PR uses [poltergeist](https://github.com/teampoltergeist/poltergeist/) for most cukes and updates Selenium. I kept selenium for the following cukes:
- <s>post preview with location, because poltergeist doesn't support geolocation (yet)</s>
- <s>mobile and desktop sign up with bogus data, because poltergeist doesn't support html5 form validation (yet)</s>
- <s>drawer cukes, because they where much more reliable with selenium</s>
- <s>download photos cuke, because poltergeist doesn't support file downloads (yet)</s>

Because poltergeist behaves different in some situations I also refactored some cukes and tried to improve them while doing that.

All cukes were green on my laptop. Let's see how this works on Travis.


Flaburgan:
Just out of curiosity, what are the advantages to use poltergeist instead of selenium?

svbergerem:
@Flaburgan Poltergeist uses PhantomJS and I hope that the cukes will be faster and have less random fails with PhantomJS instead of Firefox.

denschub:
Poltergeist is headless.

denschub:
Just for the record (I assume you already figured that out): some cukes are burning on Travis. However, the run time is impressive!

denschub:
Nice! Finally, this passes. Thanks for your work and wheeee, fast cukes!

PR #6769, opened 2016-03-26
new fetch routes

SuperTux88:
I added the new fetch routes described in SuperTux88/diaspora_federation#31. This only adds the new routes (fetching will be added with receiving with the federation-gem), but this should be backportable so that the routes are already there when fetching is added.


config/initializers/diaspora_federation.rb
```diff
@@ -106,7 +106,20 @@
...
+      # TODO
+    end
+
+    on :fetch_public_entity do |entity_type, guid|
+      entity = entity_type.constantize.find_by(guid: guid, public: true)
+      Diaspora::Federation.post(entity) if entity.is_a? Post
+    end
+
+    on :fetch_person_url_to do |diaspora_id, path|
+      Person.find_by(diaspora_handle: diaspora_id).send(:url_to, path)
```
SuperTux88: This is only for backporting and can be changed to
```
Pod.joins(:people).find_by(people: {diaspora_handle: diaspora_id}).url_to(path)
```
in `develop`-branch later.

spec/federation_callbacks_spec.rb
```diff
@@ -296,4 +296,58 @@
...
+      expect(
+        DiasporaFederation.callbacks.trigger(:fetch_public_entity, "Post", "unknown-guid")
+      ).to be_nil
+    end
+  end
+
+  describe ":fetch_person_url_to" do
+    it "returns the url with with the pod of the person" do
+      pod = FactoryGirl.create(:pod)
+      person = FactoryGirl.create(:person, pod: pod)
```
SuperTux88: when backporting this needs to be changed to `FactoryGirl.create(:person, url: "https://example.org")` and `#{pod.host}` on line 350 needs to be changed to `example.org` in `stable`-branch.

In `develop`-branch this need to stay as it is here.

fbe:
Gut gemacht!

netinetwalker:
Gut gemacht!

Gemfile
```diff
@@ -168,7 +168,7 @@ gem "addressable",        "2.3.8", require: "addressable/uri"
 gem "faraday",            "0.9.2"
 gem "faraday_middleware", "0.10.0"
 gem "faraday-cookie_jar", "0.0.6"
-gem "typhoeus",           "0.8.0"
+gem "typhoeus",           "1.0.1"
```
denschub: <3

config/initializers/diaspora_federation.rb
denschub: I assume you'll change that in a further PR.

denschub:
Merged as 69f08a4a7eb09ee7f12398ff201ec5abbd0dd425. Thanks for the hints for backporting, much appreciated.

Eh. Gut gemacht!

config/initializers/diaspora_federation.rb
SuperTux88: yes

PR #6770, opened 2016-03-26
Toggle action isn't idempotent, so it can't be PUT-action

VorontsovIE:
It looks that toggle actions have wrong HTTP-method semantics. PUT method is for idempotent actions, but toggle performed twice is not equal to toggle performed once. POST method should be more appropriate.

### DIFF
--- a/app/views/contacts/_header.html.haml
+++ b/app/views/contacts/_header.html.haml
@@ -4,14 +4,14 @@
       - if @aspect.contacts.size > 0 && @aspect.contacts.size < 20
         = start_a_conversation_link(@aspect, @aspect.contacts.size)
 
-      = link_to aspect_toggle_contact_visibility_path(@aspect), id: "contacts_visibility_toggle", class: "contacts_button", method: :put, remote: true do
+      = link_to aspect_toggle_contact_visibility_path(@aspect), id: "contacts_visibility_toggle", class: "contacts_button", method: :patch, remote: true do
         -if @aspect.contacts_visible?
           %i.entypo-lock-open.contacts-header-icon{title: t("aspects.edit.aspect_list_is_visible")}
         -else
           %i.entypo-lock.contacts-header-icon{title: t("aspects.edit.aspect_list_is_not_visible")}
 
       -if AppConfig.chat.enabled?
-        = link_to aspect_toggle_chat_privilege_path(@aspect), id: "chat_privilege_toggle", class: "contacts_button", method: :put, remote: true do
+        = link_to aspect_toggle_chat_privilege_path(@aspect), id: "chat_privilege_toggle", class: "contacts_button", method: :patch, remote: true do
           -if @aspect.chat_enabled?
             %i.entypo-chat.enabled.contacts-header-icon{title: t("aspects.edit.aspect_chat_is_enabled")}
           -else
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -58,8 +58,8 @@ Diaspora::Application.routes.draw do
   get "aspects" => "streams#aspects", :as => "aspects_stream"
 
   resources :aspects do
-    put :toggle_contact_visibility
-    put :toggle_chat_privilege
+    patch :toggle_contact_visibility
+    patch :toggle_chat_privilege
     collection do
       put "order" => :update_order
     end


jhass:
Well, that's pure definition. If we take [Wikipedia](https://en.wikipedia.org/wiki/Representational_state_transfer):

* *PUT*: Replace the addressed member of the collection, or if it does not exist, create it.
* *POST*: Not generally used. Treat the addressed member as a collection in its own right and create a new entry in it.

Or the [Rails guide](http://guides.rubyonrails.org/routing.html#crud-verbs-and-actions):

* *POST*: create a new photo
* *PUT/PATCH*: update a specific photo

So first of all neither really matches for what the action does inside REST. The question is, what's closer and I'd argue updating the resource is closer than creating a new one here. Furthermore one could argue that always the same action happens, no matter how many requests you do, just that the resulting state is undefined.

VorontsovIE:
@jhass RFC 2616 states (https://www.w3.org/Protocols/rfc2616/rfc2616-sec9#sec9.1.2) the following:

> 9.1.2 Idempotent Methods
> Methods can also have the property of "idempotence" in that (aside from error or expiration issues) the side-effects of N > 0 identical requests is the same as for a single request. The methods GET, HEAD, PUT and DELETE share this property. Also, the methods OPTIONS and TRACE SHOULD NOT have side effects, and so are inherently idempotent.

PUT action is intended to look like `make_contact_visible` or `make_contact_invisible`, which can be repeated as many times as you wish, but not for toggle action.
May be PATCH method (https://tools.ietf.org/html/rfc5789) serves better than both POST and PUT?

> The difference between the PUT and PATCH requests is reflected in the
> way the server processes the enclosed entity to modify the resource
> identified by the Request-URI.  In a PUT request, the enclosed entity
> is considered to be a modified version of the resource stored on the
> origin server, and the client is requesting that the stored version
> be replaced.  With PATCH, however, the enclosed entity contains a set
> of instructions describing how a resource currently residing on the
> origin server should be modified to produce a new version.  The PATCH
> method affects the resource identified by the Request-URI, and it
> also MAY have side effects on other resources; i.e., new resources
> may be created, or existing ones modified, by the application of a
> PATCH.
> PATCH is neither safe nor idempotent as defined by RFC2616

app/views/contacts/_header.html.haml
```diff
@@ -4,14 +4,14 @@
       - if @aspect.contacts.size > 0 && @aspect.contacts.size < 20
         = start_a_conversation_link(@aspect, @aspect.contacts.size)
 
-      = link_to aspect_toggle_contact_visibility_path(@aspect), id: "contacts_visibility_toggle", class: "contacts_button", method: :put, remote: true do
+      = link_to aspect_toggle_contact_visibility_path(@aspect), id: "contacts_visibility_toggle", class: "contacts_button", method: :patch, remote: true do
```
diaspora-code-review: Line is too long. [155/120]

app/views/contacts/_header.html.haml
```diff
@@ -4,14 +4,14 @@
...
         -if @aspect.contacts_visible?
           %i.entypo-lock-open.contacts-header-icon{title: t("aspects.edit.aspect_list_is_visible")}
         -else
           %i.entypo-lock.contacts-header-icon{title: t("aspects.edit.aspect_list_is_not_visible")}
 
       -if AppConfig.chat.enabled?
-        = link_to aspect_toggle_chat_privilege_path(@aspect), id: "chat_privilege_toggle", class: "contacts_button", method: :put, remote: true do
+        = link_to aspect_toggle_chat_privilege_path(@aspect), id: "chat_privilege_toggle", class: "contacts_button", method: :patch, remote: true do
```
diaspora-code-review: Line is too long. [148/120]

denschub:
> With PATCH, however, the enclosed entity contains a set of instructions describing how a resource currently residing on the origin server should be modified to produce a new version.

We do not have any payload in our request so `PATCH` is not right for sure. In our yet-to-be-implemented API, [we do have a `PATCH` method for aspects](https://diaspora.github.io/api-documentation/routes/aspects.html#edit-an-aspect) that would actually follow the rules, which is the thing we *should* do here as well. But since this is a private API and is likely to be replaced with something API compatible in the future, why even bother changing the method at this point?

denschub:
No feedback received, and I see no reason to change something here. Closing for now.

PR #6771, opened 2016-03-27
account_deletions get stuck while selecting participations

ghost:
Following query takes over 20 minutes and fills up the database:
SELECT \* FROM participations WHERE author_id = XXXXX

On servers with a huge participations-table this is a serious
performance issue which is easily fixed by the author_id index

@SuperTux88 thanks for your help on this

### DIFF
--- a/db/migrate/20160327103605_add_author_id_index_to_participations.rb
+++ b/db/migrate/20160327103605_add_author_id_index_to_participations.rb
@@ -1,0 +1,5 @@
+class AddAuthorIdIndexToParticipations < ActiveRecord::Migration
+  def change
+    add_index :participations, :author_id, :using => :btree
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160307142216) do
+ActiveRecord::Schema.define(version: 20160327103605) do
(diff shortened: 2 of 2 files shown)


denschub:
Thank you!

Flaburgan:
Can this be backported to the stable branch?

Zauberstuhl:
that was already asked on IRC. In short no, but you can add the index manually if you insert the migration version into the schema_migration-table. Then you can be certain that it will not result in any db:migration issues in future

SuperTux88:
I simply `cherry-pick`ed this commit, so I have the migration and didn't need to edit the `schema_migration` table by hand. Every podmin can do the same, if he wants this on his pod, but we don't want big (long running) migrations in stable.

Flaburgan:
As long as it doesn't require any manual actions, I don't see the running time as a problem, but as you want.

denschub:
> Triggers to disqualify stuff to be included:
> 
> [...]
> 
> * Stuff that requires long migrations.

([src](https://www.loomio.org/d/mq6HoprD/fixed-schedule-for-bugfix-releases))

SuperTux88:
I'm not sure what an "average sized pod" is, the migration took 32s on my pod, which is below 5 minutes. So I don't know if I'm "average sized"?

denschub:
Fine, you win. Backported as a3ac1e3072293ab5ce5efcda214019a9daffc04e.

PR #6778, opened 2016-03-31
Changes for my successful facebook reviews

ghost:
I had two facebook reviews in the past days
Removing the display type and adding a explicit
'Share to' provider tag did the trick

Even if I had the impression that the reviews were pure arbitrariness..
I think it will not hurt for future reviews

### DIFF
--- a/app/helpers/publisher_helper.rb
+++ b/app/helpers/publisher_helper.rb
@@ -12,9 +12,10 @@ module PublisherHelper
   end
 
   def service_button(service)
+    share_to = I18n.t "services.index.share_to"
     content_tag :div,
                 class:   "btn btn-link service_icon dim",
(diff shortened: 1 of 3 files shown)


app/helpers/publisher_helper.rb
```diff
@@ -12,9 +12,10 @@ def all_aspects_selected?(selected_aspects)
   end
 
   def service_button(service)
+    share_to = I18n.t "services.index.share_to"
     content_tag :div,
                 class:   "btn btn-link service_icon dim",
-                title:   "#{service.provider.titleize} (#{service.nickname})",
+                title:   "#{share_to} #{service.provider.titleize} (#{service.nickname})",
```
jhass: Don't use concatenation here, that assumes a certain grammar, use interpolation.

denschub:
Merged, thanks.

PR #6779, opened 2016-03-31
Configuration wrapper for Prosody

ghost:
With the wrapper it is possible to start Prosody via rails-runner.

Why should we use this instead of vines? Currently the S2S patch isn't working for every server and since I am a one-men-dev-team that could take a while.. Also the implementation of XEP would be obsolete if we use prosody.

Gem -> https://github.com/Zauberstuhl/gem_diaspora-prosody-config
### How does it work?
- Install prosody on your system
- Enable the chat feature with server.enabled set to `true`
- Start diaspora

In the background the wrapper maps specific diaspora-configuration to a temporary prosody-config
and starts the server with the mentioned configuration.
### But you cannot map all parameters, right?

For this case you can a) disable the wrapper or b) use the template in `config/prosody.cfg.lua`!
If you modify it the wrapper will use it after a server restart or if you hit

```
rails runner  config/initializers/prosody.rb
eye restart xmpp
```


Gemfile
```diff
@@ -1,5 +1,9 @@
 source "https://rubygems.org"
 
+
```
diaspora-code-review: Extra blank line detected.

Gemfile
```diff
@@ -1,5 +1,9 @@
...
+gem "prosody2gem", git: "https://github.com/Zauberstuhl/prosody2gem"
+
+
```
diaspora-code-review: Extra blank line detected.

config/initializers/prosody.rb
```diff
@@ -0,0 +1,3 @@
+Prosody.config({
```
diaspora-code-review: Redundant curly braces around a hash parameter.

config/initializers/prosody.rb
```diff
@@ -0,0 +1,3 @@
...
+  file: "config/prosody.cfg.lua"
```
diaspora-code-review: Use 2 spaces for indentation in a hash, relative to the first position after the preceding left parenthesis.

config/initializers/prosody.rb
```diff
@@ -0,0 +1,3 @@
...
+})
```
diaspora-code-review: Indent the right brace the same as the first position after the preceding left parenthesis.

denschub:
Cheez. Please run your code style checks locally before submitting the PR. ;)

I don't quite get why that wrapper is needed? Why can't we simply depend on a installed prosody, which is installed via the systems package manager and configured with configuration examples we provide in the wiki? I don't see the reason to maintain a dedicated gem that builds and configures prosody.

Zauberstuhl:
> I don't see the reason to maintain a dedicated gem that builds and configures prosody.

automatism and its more user friendly.  Imagine you hit bundler once and you are ready to go. Also it could be a benefit for the debian package?!

denschub:
> automatism and its more user friendly

So you want to maintain your gem for the rest of your life? Imagine there is a prosody security hotfix (which, eh, happens.). When we use a gem that bundles prosody, 

* you'd need to publish a new Gem version
* we'd need to freeze a new version inside the Gemfile
* we'd need to release a diaspora\* hotfix release

If we'd depend on a prosody installed via the systems package manager, the podmins just need to run `pacman -Syu` or whatever and they are done. I don't really feel great with adding that much complexity to a security update (which *will* happen) and some additional days of fixing delay.

> Also it could be a benefit for the debian package

The debian folks are not going to package your gem anyway because they actually have a rule for using system packages when possible. If they did not, we'd never have any issues since they could just ship a diaspora distribution with all the gems installed inside `vendor`, which they don't. 

I'd love to hear other opinions here but I do not really see the gain at the moment.

Zauberstuhl:
>  you want to maintain your gem for the rest of your life?

If I have to decide between the whole vineserver and a simple wrapper gem then I choose the gem, yes.

> Imagine there is a prosody security hotfix (which, eh, happens.). When we use a gem that bundles prosody

Also true, but increasing a version number and pushing a hotfix once a year is in my perspective something I would accept. Think about vines who is reviewing that code? Nobody. Prosody has active development and a huge community.

> If we'd depend on a prosody installed via the systems package manager, the podmins just need to run pacman -Syu

This is something I cannot argue against. As I already said the benefit would be a fully automated installation process. And the possibility that we can configure prosody over the diaspora.yml

For me and you hitting the package manager and configuring a xmpp server isn't a big deal
but in my opinion everyone should be able to install diaspora as easy as possible.

Well it also bugs me that I started the chat implementation and now I cannot really provide a working backend.

Currently my plan is
* to improve the gem a little so that it includes test automatism for multiple OS's
* integrate diaspora configuration into the gem (or diaspora)

Well maybe it is a crappy plan. Please give me more concerns/objections (or approval ;)..

denschub:
> Also true, but increasing a version number and pushing a hotfix once a year is in my perspective something I would accept.

You did not get my point. You releasing the new version, us releasing the new diaspora version and the podmins updating could take several days. That's just too much time for a critical security update.

Idea: ship the configuration in the gem and make the path to the prosody binary configurable in your wrapper. That way, you still could manage the configs and runtimes of prosody in your gem, but you do not have to build prosody since that would get installed via the package manager.

Zauberstuhl:
> That way, you still could manage the configs and runtimes of prosody in your gem, but you do not have to build prosody since that would get installed via the package manager

I have slept on it for a night and I totally agree with you now. My further development will included the basic idea of this proposal but not the build thingy. Instead of checking prosody dependencies I will check for prosody itself.

Thanks @denschub for your thoughts :)

denschub:
Do you mind writing a migration guide to the wiki first? Otherwise, this looks fine and I'd merge it when I've got something in the wiki to link to in the changelog. :)

Zauberstuhl:
Yes totally! I will do that in the next few days. Cheers

Zauberstuhl:
https://wiki.diasporafoundation.org/Integration/Chat#Vines_to_Prosody

denschub:
That's merged, thanks. Although we do not support the the chat in our stable releases, I decided to merge this into `develop` only to avoid confusion.

PR #6782, opened 2016-04-03
Fix back to top not appearing on Webkit browsers + redesign

Flaburgan:
The `$("html, body").scrollTop()` was always returning 0 on Webkit browsers.

I changed the arrow based on #5729 while touching that, so this should supersedes #6713. I simply replaced the character by the CSS arrow designed by @SansPseudoFix

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -65,22 +65,25 @@ pre { word-wrap: break-word; }
 
 .back-to-top {
   background-color: $border-dark-grey;
-  border-radius: 10px;
+  border-radius: 4px;
   bottom: 20px;
   color: $white;
   display: block;
-  font-size: 2.9em;
-  line-height: 1.5;
+  font-size: 3.5em;
+  height: 50px;
+  line-height: 50px;
   opacity: 0;
-  padding: 0 12px;
   position: fixed;
   right: 54px;
+  transition: opacity ease 400ms;
+  width: 50px;
   z-index: 49;
 
   &:hover,
   &.visible:hover {
     color: $white;
     opacity: .85;
+    text-decoration: none;
   }
 
   &.visible { opacity: .5; }
--- a/app/assets/javascripts/app/views/back_to_top_view.js
+++ b/app/assets/javascripts/app/views/back_to_top_view.js
@@ -16,7 +16,7 @@ app.views.BackToTop = Backbone.View.extend({
   },
 
   toggleVisibility: function() {
-    if($("html, body").scrollTop() > 1000) {
+    if($(document).scrollTop() > 1000) {
       $("#back-to-top").addClass("visible");
     } else {
       $("#back-to-top").removeClass("visible");
--- a/app/views/people/contacts.haml
+++ b/app/views/people/contacts.haml
@@ -17,8 +17,7 @@
             = render partial: 'people/person', locals: hash
         = will_paginate @contacts_of_contact, renderer: WillPaginate::ActionView::BootstrapLinkRenderer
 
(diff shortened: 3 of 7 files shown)


app/assets/stylesheets/base.scss
```diff
@@ -65,16 +65,18 @@ pre { word-wrap: break-word; }
...
   display: block;
   font-size: 2.9em;
+  height: 50px;
   line-height: 1.5;
   opacity: 0;
-  padding: 0 12px;
   position: fixed;
   right: 54px;
+  transition: opacity ease 400ms;
+  width: 50px;
```
Flaburgan: @svbergerem pronto asked me to place the width there, feels weird to me, I use to place it next to height... Any reason it should be here?

svbergerem: https://github.com/diaspora/diaspora/blob/0f33a7e/.scss-lint.yml#L115 and https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#propertysortorder. Enforcing some order looked like a good idea and alphabetical order is the scss lint default. (also see the linked readme for more information) I don't have a strong opinion here but if you'd like to change this then let's discuss that somewhere else.

svbergerem:
A screenshot would be great. :)

Flaburgan:
capture du 2016-04-04 11-09-52
capture du 2016-04-04 11-11-09
On hover:
capture du 2016-04-04 11-11-24

https://github.com/diaspora/diaspora/pull/6782/commits/aad7cb7c2899993fa147b9901acfc9a7da5a2144 now includes only the webkit fix, so it can be backported to stable.

svbergerem:
Firefox 45:
back to top Firefox 45

We should find a solution that works in both browsers.

Flaburgan:
Hm that's weird, Firefox 45 is the browser I used to open this PR. I'll have a look thank you.

denschub:
Probably the same customizations on diaspora-fr that have caused such issues before...

Flaburgan:
@denschub I dev on the develop branch, and diaspora-fr is running develop code now that https://github.com/diaspora/diaspora/pull/6535 is merged. I don't have time to look at that right now but I will during the week-end.

Flaburgan:
I just tried again on another computer, with an empty Firefox 45 with a new profile and I'm not able to reproduce the misplaced arrow. Are you sure you run the last code?

svbergerem:
I also tried again and still have the same problem. It looks the same in the Firefox Developer version but everything is fine in Chromium. Would be great if someone else could try this.

Flaburgan:
I replaced margins by `position: absolute';` and left / top. Does that solve the problem for you @svbergerem ?

denschub:
I actually had a question in https://github.com/diaspora/diaspora/pull/6713#issuecomment-190188167, but nobody seemed to care really, so I ask again:

> Is there any reason not to use `entypo-chevron-up` or even `entypo-chevron-thin-up` ? Those lines in the top of that arrow are rather distracting.

Why adding CSS hacks to render an arrow (that, due to the lines in it) looks bad and is hard to position when we have an icon font that gets loaded anyway?

Flaburgan:
@denschub I just grabbed the code of @SansPseudoFix, tested it and rebased it on develop. If you want me to use an icon instead of CSS, I can do it right now.

denschub:
I preferred that in #6713 and I still do that. No messy lines, no even messier border hacks and no absolute positioning! Win-win-win!

Flaburgan:
@denschub done. We will also have a better browsers support with no call to `rotate()`.

svbergerem:
Thank you! :)

Flaburgan:
Thank *you*. You can probably backport the fix commit to the stable branch btw ;)

svbergerem:
Done.

PR #6783, opened 2016-04-05
Extend the generated fixtures set and refactor AccountDeletion

cmrd-senya:
This adds some more generation code to the fixtures builder. The AccountDeletion test is rewritten basing on the newly added fixtures. Some other tests  which depended on fixtures are fixed up to conform    the changes.

ref #6780

### DIFF
--- a/lib/account_deleter.rb
+++ b/lib/account_deleter.rb
@@ -18,7 +18,7 @@ class AccountDeleter
   attr_accessor :person, :user
 
   def initialize(diaspora_handle)
-    self.person = Person.where(:diaspora_handle => diaspora_handle).first
+    self.person = Person.by_account_identifier(diaspora_handle)
     self.user = self.person.owner
   end
 
@@ -45,7 +45,7 @@ class AccountDeleter
   #user deletions
   def normal_ar_user_associates_to_delete
     %i(tag_followings services aspects user_preferences
-       notifications blocks authorizations o_auth_applications pairwise_pseudonymous_identifiers)
+       notifications blocks authorizations o_auth_applications pairwise_pseudonymous_identifiers invitation_codes)
   end
 
   def special_ar_user_associations
@@ -101,7 +101,8 @@ class AccountDeleter
   end
 
   def ignored_or_special_ar_person_associations
-    %i(comments contacts notification_actors notifications owner profile conversation_visibilities pod)
+    %i(comments likes contacts notification_actors notifications owner profile conversation_visibilities pod
+       messages conversations poll_participations blocks)
   end
 
   def mark_account_deletion_complete
--- a/spec/controllers/notifications_controller_spec.rb
+++ b/spec/controllers/notifications_controller_spec.rb
@@ -167,20 +167,21 @@ describe NotificationsController, :type => :controller do
   describe "#read_all" do
     it 'marks all notifications as read' do
       request.env["HTTP_REFERER"] = "I wish I were spelled right"
-      FactoryGirl.create(:notification, :recipient => alice, :target => @post)
-      FactoryGirl.create(:notification, :recipient => alice, :target => @post)
+      FactoryGirl.create(:notification, recipient: alice, target: @post)
+      FactoryGirl.create(:notification, recipient: alice, target: @post)
 
-      expect(Notification.where(:unread => true).count).to eq(2)
+      expect(Notification.where(unread: true, recipient: alice).count).to eq(2)
       get :read_all
-      expect(Notification.where(:unread => true).count).to eq(0)
+      expect(Notification.where(unread: true, recipient: alice).count).to eq(0)
     end
     it 'marks all notifications in the current filter as read' do
       request.env["HTTP_REFERER"] = "I wish I were spelled right"
-      FactoryGirl.create(:notification, :recipient => alice, :target => @post)
-      FactoryGirl.create(:notification, :recipient => alice, :type => "Notifications::StartedSharing")
-      expect(Notification.where(:unread => true).count).to eq(2)
-      get :read_all, "type" => "started_sharing"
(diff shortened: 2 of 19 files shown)


cmrd-senya:
Well, the tests are sort of green...

Weird thing, looks like the dangling Lola broke the photo numbers feature.

lib/account_deleter.rb
```diff
@@ -17,8 +17,8 @@ class AccountDeleter
 
   attr_accessor :person, :user
 
-  def initialize(diaspora_handle)
-    self.person = Person.where(:diaspora_handle => diaspora_handle).first
+  def initialize(person)
```
denschub: Why? I'd rather not pass objects around workers... Why don't you use the guid here?

spec/controllers/notifications_controller_spec.rb
```diff
@@ -149,17 +149,17 @@
       FactoryGirl.create(:notification, :recipient => alice, :target => @post)
       FactoryGirl.create(:notification, :recipient => alice, :target => @post)
 
-      expect(Notification.where(:unread => true).count).to eq(2)
+      expect(Notification.where(unread: true).where(recipient: alice).count).to eq(2)
```
denschub: Everywhere:

```ruby
Notification.where(unread: true, recipient: alice).count
```

spec/controllers/notifications_controller_spec.rb
```diff
@@ -149,17 +149,17 @@
       FactoryGirl.create(:notification, :recipient => alice, :target => @post)
       FactoryGirl.create(:notification, :recipient => alice, :target => @post)
```
denschub: If you change to the new ruby syntax, please change the entire test.

spec/factories.rb
```diff
@@ -298,6 +298,21 @@ def r_str
     end
   end
 
+  factory(:report) do
```
denschub: Unnecessary parenthesis.

spec/factories.rb
```diff
@@ -298,6 +298,21 @@ def r_str
...
+    sequence(:text) {|n| "#{n} offensive content" }
+    association(:user)
+  end
+
+  factory(:report_on_post, parent: :report) do
+    association(:item, factory: :status_message)
+    item_type "Post"
+  end
+
+  factory(:role) do
```
denschub: Unnecessary parenthesis.

spec/factories.rb
denschub: Actually, most of the parenthesis here are unnecessary. Please see other factory definitions...

spec/factories.rb
```diff
@@ -312,32 +327,32 @@ def r_str
...
-    client_name "Diaspora Test Client"
+    sequence(:client_name) {|n| "Diaspora Test Client #{n}#{r_str}" }
     redirect_uris %w(http://localhost:3000/)
     ppid true
     sector_identifier_uri "https://example.com/uri"
   end
 
   factory :o_auth_application_with_multiple_redirects, class: Api::OpenidConnect::OAuthApplication do
-    client_name "Diaspora Test Client"
+    sequence(:client_name) {|n| "Diaspora Test Client #{n}#{r_str}" }
```
denschub: Why's that needed? If this is needed, you missed line 360.

spec/misc_spec.rb
```diff
@@ -14,6 +14,10 @@
   describe 'fixtures' do
     it 'loads fixtures' do
       expect(User.count).not_to eq(0)
+      expect(TagFollowing.count).to be > 0
```
denschub: Look at the line above. Use consistent code.

lib/account_deleter.rb
cmrd-senya: The change is not critical, I can revert it. But what's wrong with passing object here? It's not worker enqueueing here, so it's not passed via Redis, right?

I just thought, that in all the usages of the AccountDeleter (currently one, more in #6750) the person object is already available in some way and there is no point searching it via diaspora_handle.

For example, [here](https://github.com/diaspora/diaspora/pull/6783/files#diff-0754ce5ce1ca26e1c990fa947f873486R33) person is linked via belongs_to relationship, and perhaps it is found faster than a search via diaspora_handle. But that's a pure speculation.

spec/controllers/notifications_controller_spec.rb
cmrd-senya: Is it enough to change the corresponding "it" blocks? Or you mean to update the whole _spec.rb file?

denschub: I'd be fine with just changing the block. But two different syntaxes in one block look... weird :)

spec/factories.rb
cmrd-senya: If the name is the same "Diaspora Test Client" for all objects, then with my patch [this](https://github.com/diaspora/diaspora/blob/develop/spec/controllers/api/openid_connect/clients_controller_spec.rb#L127) test fails. It happens because [this](https://github.com/diaspora/diaspora/pull/6783/commits/41049694dc870bf827513cf5446c01f8bd1be532#diff-8c42b18d6c734a3f806c8df8c01016b4R103) object creation depends on ```:o_auth_application``` factory and therefore we have an existing object with client_name "Diaspora Test Client" in the DB, while the test I've mentioned above makes search by the client_name and finds an unexpected object.

It seems that it is enough to make the name sequenced only for ```:o_auth_application``` factory, so I can revert name to the plain string for all the others.

cmrd-senya:
Updates in this PR lead to random test suite failures due to the #6811. The issue is fixed via #6812 so I guess it should be merged and this PR then rebased to avoid the failures.

cmrd-senya:
I moved all the changes I found useful in this PR to #6726, so closing this.

PR #6784, opened 2016-04-06
Redirect to "/users/sign_in" after AccountDeletion

cmrd-senya:
There is no point to redirect to "/stream" after the account deletion
because the user is logged off.


cmrd-senya:
green

denschub:
Purple.

PR #6785, opened 2016-04-06
Add temporary user lock to admin interface

ghost:
With this patch the admin can lock an account for a certain amount of time (or permanently)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -550,6 +550,26 @@ class User < ActiveRecord::Base
     AccountDeletion.create(:person => self.person)
   end
 
+  def lock_access!(opts={})
+    if opts[:unlock_in]
+      self.lock_expires = true
+      self.locked_at = Time.now.utc + opts[:unlock_in]
+      save
+    else
+      self.lock_expires = false
+      super
+    end
+  end
+
+  def lock_expired?
+    locked_at && lock_expires && locked_at < Time.now.utc
+  end
+
+  def unlock_access!
+    self.lock_expires = false
+    super
+  end
+
   def closed_account?
     self.person.closed_account
   end
--- a/app/views/admins/_user_entry.haml
+++ b/app/views/admins/_user_entry.haml
@@ -26,7 +26,11 @@
 
         - unless user.closed_account?
           - unless user.access_locked?
-            = link_to t('admins.user_search.lock_account'), admin_lock_account_path(user), method: :post, data: { confirm: t('admins.user_search.are_you_sure_lock_account') }, class: 'btn btn-danger btn-block btn-xs'
+            = form_for "user", url: admin_lock_account_path(user), html: {method: :post, class: "form-group btn"} do |f|
+              = f.number_field_tag :duration, 0, min: 0, class: "form-control btn-block"
+              = f.submit t("admins.user_search.lock_account"),
+                data: {confirm: t("admins.user_search.are_you_sure_lock_account")},
+                class: "btn btn-danger btn-block btn-xs"
           - else
             = link_to t('admins.user_search.unlock_account'), admin_unlock_account_path(user), method: :post, data: { confirm: t('admins.user_search.are_you_sure_unlock_account') }, class: 'btn btn-danger btn-block btn-xs'
 
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -8,8 +8,12 @@ module Admin
     end
 
     def lock_account
+      opts = {}
       u = User.find(params[:id])
-      u.lock_access!
+      if (unlock_in = params[:duration].to_i) > 0
+        opts[:unlock_in] = unlock_in.minutes
+      end
+      u.lock_access!(opts)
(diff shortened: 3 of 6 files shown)


app/views/admins/_user_entry.haml
```diff
@@ -26,7 +26,10 @@
 
         - unless user.closed_account?
           - unless user.access_locked?
-            = link_to t('admins.user_search.lock_account'), admin_lock_account_path(user), method: :post, data: { confirm: t('admins.user_search.are_you_sure_lock_account') }, class: 'btn btn-danger btn-block btn-xs'
+            = form_for "user", url: admin_lock_account_path(user), html: {method: :post, class: "form-group btn"} do |f|
+              = f.number_field_tag :duration, 0, min: 0, class: "form-control btn-block"
+              = f.submit t("admins.user_search.lock_account"),
+                data: {confirm: t("admins.user_search.are_you_sure_lock_account")}, class: "btn btn-danger btn-block btn-xs"
```
diaspora-code-review: Line is too long. [124/120]

app/models/report.rb
```diff
@@ -51,6 +51,14 @@ def destroy_reported_item
     mark_as_reviewed
   end
 
+  def mark_as_nsfw
```
denschub: These changes will not federate, so the post will still be SFW everywhere else. I don't think that this is a good thing to implement then...

Flaburgan:
About the lock / unlock, there is [an unlock method](https://github.com/diaspora/diaspora/blob/b9972b54071d6da7348b5e896a9d65058d6c65ad/app/controllers/admin/users_controller.rb#L16), why don't simply expose it in the admin panel? (Tbh I thought it was already there...). Not sure adding a column etc is worth the effort.

app/models/user.rb
```diff
@@ -550,6 +550,26 @@ def close_account!
...
+      self.lock_expires = true
+      self.locked_at = Time.now.utc + opts[:unlock_in]
+      save
+    else
+      self.lock_expires = false
+      super
+    end
+  end
+
+  def lock_expired?
```
denschub: Have you checked that this actually gets called? The [devise docs](http://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Lockable) say this only gets called if the lock strategy is `time`, which is not the case for us.

app/models/report.rb
Zauberstuhl: true but it is always a local user which reports stuff and it happened to me so often.  
Currently the only possibility is to delete the whole post and this is in my opinion to extreme for a missing tag.

app/models/user.rb
Zauberstuhl: > Have you checked that this actually gets called?

yes it does. Its being called while authenticating.. I had no desire to read through the devise source code

app/models/report.rb
denschub: How about either ignoring whiny requests, talking to the user or setting the profile NSFW (which would federate)? I'm really not a fan of your "well, I'll just change it locally and don't care about the rest of the world" solution here.

Flaburgan: >setting the profile NSFW

This should only be done if the majority of the user's content is NSFW. I know people complaining about that for only one post of a user who definitely not deserves to be checked as NSFW.

IMO the easiest solution is to check manually: if the post really is porn related or whatever, then delete the post (the author is not compliant with the ToS), if it's not (just erotic or something like that), then ignore the request. It's not longer than to turn the post to nsfw anyway and I agree that inconsistency in the network is bad.

Zauberstuhl: > How about either ignoring whiny requests

I do that, but I decided for myself that NSFW is something which bugs me too.
And this is something you can decided for yourself I just implemented a option

> talking to the user or setting the profile NSFW

if you have 10 reports from the same user that is probably the better solution,
but if not you can spend either contacting ten different people or mark it quickly as nsfw

>   I'm really not a fan of your "well, I'll just change it locally and don't care about the rest of the world" solution here.

Never said that. Give me a federated-solution and I change it right away

Zauberstuhl:
@Flaburgan I overwrite https://github.com/diaspora/diaspora/blob/b9972b54071d6da7348b5e896a9d65058d6c65ad/app/controllers/admin/users_controller.rb#L12

this is necessary if you want to lock user non-permanently

app/models/report.rb
Flaburgan: >but if not you can spend either contacting ten different people or mark it quickly as nsfw

maybe it's time to set a community of volunteers for joindiaspora ? :) We now have the moderator role, just for that.

denschub: >  Give me a federated-solution and I change it right away

Mark the profile as NSFW or delete it. I'm not going to support adding inconsistencies by locally modifying posts, sorry. Maybe others have different opinions, though.

app/models/user.rb
Zauberstuhl: its called though the Authenticatable-module from devise

jhass:
I don't think hijacking `locked_at`s meaning like this is a good idea. How about adding a `locked_until` column instead of `lock_expires`? It being `nil` would mean locked forever.

Issue #6269, opened 2015-07-29
Use unicorn-worker-killer?


ldidry:
I saw a [blog post from Gitlab](https://about.gitlab.com/2015/06/05/how-gitlab-uses-unicorn-and-unicorn-worker-killer/) about their use of the [unicorn-worker-killer](https://github.com/kzk/unicorn-worker-killer) gem to stop memory leaks.

This is simple: every 16 requests, unicorn-worker-killer checks the memory used by the worker and kills it if a memory leak is detected.

They also have a [Sidekiq MemoryKiller](http://www.rubydoc.info/github/gitlabhq/gitlabhq/Gitlab/SidekiqMiddleware/MemoryKiller) (code [here](https://github.com/gitlabhq/gitlabhq/tree/master/lib/gitlab/sidekiq_middleware)). Maybe another hint to reduce Diaspora* memory usage ?

jhass:
https://github.com/diaspora/diaspora/blob/develop/lib/unicorn_killer.rb like this? ;)

ldidry:
Oops! Ok, for unicorn. For my defense, I only looked at the Gemfile.

There's still the question of sidekiq.

svbergerem:
@jhass Should we use the gem instead of the file you linked? (less code we have to maintain)

jhass:
Maybe, though that code so far was surprisingly maintenance free.

Flaburgan:
The new version 4 of Sidekiq really improves memory usage (see [this blogpost](http://www.mikeperham.com/2015/10/14/optimizing-sidekiq/)). We need to [drop sidetiq](https://github.com/diaspora/diaspora/issues/6538) first to be able to upgrade sidekiq, but looks like this is [coming soon](https://github.com/diaspora/diaspora/pull/6616).

Flaburgan:
What's the status of this? We now use sidekiq 4 since September. If it can still improve the situation and someone writes a patch, I can test it on diaspora-fr.

PR #6792, opened 2016-04-13
Move unicorn_killer to Gemfile

ghost:
I stumbled over that and though it is a good idea to use the gem file instead. There are a view changes since 2013.

There is also the possibility to kill workers by max_requests:

```
use Unicorn::WorkerKiller::MaxRequests, max_request_min, max_request_max
```

Do we need/want this? If so.. unfortunately I have no reference for some max/min values:
I found some examples with 500-600 request per worker.

related to #6269


config.ru
```diff
@@ -5,12 +5,15 @@
 # This file is used by Rack-based servers to start the application.
 
 require ::File.expand_path("../config/environment",  __FILE__)
-require ::File.expand_path("../lib/unicorn_killer",  __FILE__)
 require ::File.expand_path("../lib/rack/internet_explorer_version", __FILE__)
 
 # Kill unicorn workers really aggressively (at 300mb)
 if defined?(Unicorn)
-  use UnicornKiller::Oom, 300 * 1024
+  require 'unicorn/worker_killer'
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

config.ru
```diff
@@ -5,12 +5,15 @@
 # This file is used by Rack-based servers to start the application.
 
 require ::File.expand_path("../config/environment",  __FILE__)
-require ::File.expand_path("../lib/unicorn_killer",  __FILE__)
 require ::File.expand_path("../lib/rack/internet_explorer_version", __FILE__)
 
 # Kill unicorn workers really aggressively (at 300mb)
 if defined?(Unicorn)
-  use UnicornKiller::Oom, 300 * 1024
+  require "unicorn/worker_killer"
```
denschub: This `require` should be on top of the file.

jhass: Not necessarily, since the `defined?(Unicorn)` makes unicorn optional with no changes besides the Gemfile.

denschub: I was more worried about codestyle here - requires in the middle of the code look weird. But if you're fine with it, I can get over it. ;)

jhass: It's too weird and I mean there's another require just above it stil

denschub:
Merged. 

> Do we need/want this?

If we do, we can always add that later. :)

PR #6805, opened 2016-04-19
Fix up the meaning of the name parameter in mention

### LINKED ISSUE #6795: Mentions don't respect alternate text when rendered in mobile view
Our mention syntax is ```@{name; id}``` and name is supposed to be substituted as a link text for the mention. In usual view it is done so with the javascript code from [markdown-it-diaspora-mention](https://github.com/diaspora/markdown-it-diaspora-mention). In some cases, for example, for mobile view  of ```/stream``` posts are rendered with the MessageRenderer (e.g., [markdownify method](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/message_renderer.rb#L207)) which renders mentions with [Diaspora::Mentionable.format method](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/mentionable.rb#L31). So when mentions are rendered with this method, the ```name``` attribute of the mention is completely ignored and ```person.name``` value is put as the link text instead. However, ```name``` is rendered if corresponding person was not supplied to the method, so it works as a fallback, rather than the visible name as it works in [markdown-it-diaspora-mention](https://github.com/diaspora/markdown-it-diaspora-mention).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32995031-mentions-don-t-respect-alternate-text-when-rendered-in-mobile-view?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cmrd-senya:
The desktop frontend now treats the "name" parameter of mention as
a string to display unconditionally. But the Diaspora::Mentionable
renders mentions the different way: "name" is treated as a fallback
string which is rendered only if the person's name is unavailable.
This reflects on the mobile version ATM. This patch makes it behave
the same way as the current desktop version does.

fix #6795

### DIFF
--- a/lib/diaspora/mentionable.rb
+++ b/lib/diaspora/mentionable.rb
@@ -14,8 +14,8 @@ module Diaspora::Mentionable
     mention = mention_str.match(REGEX)[2]
     del_pos = mention.rindex(/;/)
 
-    name   = mention[0..(del_pos-1)].strip
-    handle = mention[(del_pos+1)..-1].strip
+    name = mention[0..(del_pos - 1)].strip
+    handle = mention[(del_pos + 1)..-1].strip
 
     [name, handle]
   end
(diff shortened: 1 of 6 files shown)


lib/diaspora/mentionable.rb
```diff
@@ -15,6 +15,7 @@ def self.mention_attrs(mention_str)
     del_pos = mention.rindex(/;/)
 
     name   = mention[0..(del_pos-1)].strip
+    name = nil if name.empty?
```
denschub: So, why is `nil` better than an empty string?

Also, nit, something is wrong with the indentation.

cmrd-senya: It was done for [this line](https://github.com/diaspora/diaspora/pull/6805/files#diff-515d5bec596509fb74bd38d45f5c3db2R100) to pass to `person.name` when empty. However, it's probably better to change the mentioned line then, instead of having this non-obvious expression.

cmrd-senya: updated with a more neat version

lib/diaspora/mentionable.rb
```diff
@@ -86,32 +86,32 @@ module MentionsInternal
     extend ::PeopleHelper
 
     # output a formatted mention link as defined by the given options,
-    # use the fallback name if the person is unavailable
+    # use the display name if the person is unavailable
```
jhass: You made the `display_name` take precedence over the `person.name` now, no?

cmrd-senya: fixed

lib/diaspora/mentionable.rb
```diff
@@ -86,32 +86,32 @@ module MentionsInternal
     extend ::PeopleHelper
 
     # output a formatted mention link as defined by the given options,
-    # use the fallback name if the person is unavailable
+    # use the person name the if the display name is unavailable
```
jhass: > if the display name is blank, fallback to the persons name.

lib/diaspora/mentionable.rb
```diff
@@ -86,32 +86,32 @@ module MentionsInternal
...
       else
-        person_link(person, class: PERSON_HREF_CLASS)
+        person_link(person, class: PERSON_HREF_CLASS, display_name: display_name)
       end
     end
 
-    # output a markdown formatted link to the given person or the given fallback
-    # string, in case the person is not present
+    # output a markdown formatted link to the given person or the given display
+    # name, in case the person is not present
```
jhass: Ditto

cmrd-senya:
updated

jhass:
Thank you!

PR #6812, opened 2016-04-25
Don't include wrong shareable types in scopes

### LINKED ISSUE #6811: Weird behavior of Photo.visible
I was hunting down the issue that lead to the wrong photo number in the [tests](https://travis-ci.org/diaspora/diaspora/jobs/124726853#L561) of #6783. Basically, it happens that [this test](https://github.com/diaspora/diaspora/blob/develop/spec/controllers/photos_controller_spec.rb#L107) generates 16 test photos, but when it checks the number with the ```Photo.visible``` method, the value is bigger, e.g. 19.

Finally I managed to get some database dump that shows weird result and seemingly  reproduces that very bug. I worked with mysql, but it happens on postgres as well ([example](https://travis-ci.org/diaspora/diaspora/jobs/124726857#L573)).

I [rebuilt](https://gist.github.com/cmrd-senya/0683ab0c8c024620392a9e525ed5f025) the db state similar to what we have in the test. If you load this DB dump and run the following commands within the rails console, you'll see the issue:
```
2.2.1 :001 > alice = User.find_by(username: "alice")
 => #<User id: 9, username: "alice", serialized_private_key: "-----BEGIN RSA PRIVATE KEY-----\nMIICWwIBAAKBgQDTH4...", getting_started: false, disable_mail: false, language: "en", email: "bob151f078@pivotallabs.com", encrypted_password: "$2a$04$acc98iRbowJhXqnH3I1SNuLI.MaVUAdj2/MyRC40GVA...", invitation_token: nil, invitation_sent_at: nil, reset_password_token: nil, remember_created_at: nil, sign_in_count: 0, current_sign_in_at: nil, last_sign_in_at: nil, current_sign_in_ip: nil, last_sign_in_ip: nil, created_at: "2016-04-22 17:28:17", updated_at: "2016-04-22 17:28:17", invitation_service: nil, invitation_identifier: nil, invitation_limit: nil, invited_by_id: nil, invited_by_type: nil, authentication_token: nil, unconfirmed_email: nil, confirm_email_token: nil, locked_at: nil, show_community_spotlight_in_stream: true, auto_follow_back: false, auto_follow_back_aspect_id: nil, hidden_shareables: {}, reset_password_sent_at: nil, last_seen: nil, remove_after: nil, export: "alice_diaspora_data_i-YlfKsk9vQLKIhtUvYpIA.json.gz", exported_at: nil, exporting: false, strip_exif: false, exported_photos_file: nil, exported_photos_at: nil, exporting_photos: false, color_theme: "original"> 
2.2.1 :002 > eve = User.find_by(username: "eve")
 => #<User id: 10, username: "eve", serialized_private_key: "-----BEGIN RSA PRIVATE KEY-----\nMIICWwIBAAKBgQDTH4...", getting_started: false, disable_mail: false, language: "en", email: "bob20d25d6@pivotallabs.com", encrypted_password: "$2a$04$GZReLqLhTFkGu0/gDzLs1eCzH2RQj4Q.gpbDlnVABtx...", invitation_token: nil, invitation_sent_at: nil, reset_password_token: nil, remember_created_at: nil, sign_in_count: 0, current_sign_in_at: nil, last_sign_in_at: nil, current_sign_in_ip: nil, last_sign_in_ip: nil, created_at: "2016-04-22 17:28:17", updated_at: "2016-04-22 17:28:17", invitation_service: nil, invitation_identifier: nil, invitation_limit: nil, invited_by_id: nil, invited_by_type: nil, authentication_token: nil, unconfirmed_email: nil, confirm_email_token: nil, locked_at: nil, show_community_spotlight_in_stream: true, auto_follow_back: false, auto_follow_back_aspect_id: nil, hidden_shareables: {}, reset_password_sent_at: nil, last_seen: nil, remove_after: nil, export: "eve_diaspora_data_JVcwGMM0IJA2zsxQwEobrg.json.gz", exported_at: nil, exporting: false, strip_exif: true, exported_photos_file: nil, exported_photos_at: nil, exporting_photos: false, color_theme: "original"> 
2.2.1 :003 > Photo.count
 => 8 
2.2.1 :004 > Photo.visible(alice, eve.person).count(:all)
 => 9 
```

So, basically, the method returns the visible photos count bigger than the total photos count, which is, apparently, a bug.

I can't say how does it reflect on the application operation, but I catch it sometimes in the testsuite of my PR #6783. I don't know why it happens only sometimes but not everytime. I guess the underlying SQL query of ```Photo.visible``` must be studied and I'm a bit tired to do that right now.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/33239642-weird-behavior-of-photo-visible?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cmrd-senya:
fix #6811

also moves diaspora_initialize to lib/diaspora/shareable.rb in sake of DRY

### DIFF
--- a/app/models/photo.rb
+++ b/app/models/photo.rb
@@ -82,11 +82,7 @@ class Photo < ActiveRecord::Base
   end
 
   def self.diaspora_initialize(params = {})
-    photo = self.new params.to_hash.slice(:text, :pending)
-    photo.author = params[:author]
-    photo.public = params[:public] if params[:public]
-    photo.pending = params[:pending] if params[:pending]
-
+    photo = shareable_initialize(params)
     photo.random_string = SecureRandom.hex(10)
 
     if photo.author.local?
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -133,9 +133,7 @@ class Post < ActiveRecord::Base
   #############
 
   def self.diaspora_initialize(params)
-    new(params.to_hash.stringify_keys.slice(*column_names)).tap do |new_post|
-      new_post.author = params[:author]
-    end
+    shareable_initialize(params)
   end
 
   # @return Returns true if this Post will accept updates (i.e. updates to the caption of a photo).
--- a/lib/diaspora/shareable.rb
+++ b/lib/diaspora/shareable.rb
@@ -21,11 +21,13 @@ module Diaspora
         scope :all_public, -> { where(public: true, pending: false) }
 
         scope :with_visibility, -> {
-          joins("LEFT OUTER JOIN share_visibilities ON share_visibilities.shareable_id = #{table_name}.id")
+          joins("LEFT OUTER JOIN share_visibilities ON share_visibilities.shareable_id = #{table_name}.id AND "\
+            "share_visibilities.shareable_type = '#{base_class}'")
         }
 
         scope :with_aspects, -> {
-          joins("LEFT OUTER JOIN aspect_visibilities ON aspect_visibilities.shareable_id = #{table_name}.id")
+          joins("LEFT OUTER JOIN aspect_visibilities ON aspect_visibilities.shareable_id = #{table_name}.id AND "\
+          " aspect_visibilities.shareable_type = '#{base_class}'")
         }
 
         def self.owned_or_visible_by_user(user)
@@ -57,7 +59,6 @@ module Diaspora
 
         def self.visible_by_user(user)
           ShareVisibility.arel_table[:user_id].eq(user.id)
-            .and(ShareVisibility.arel_table[:shareable_type].eq(base_class.to_s))
         end
         private_class_method :visible_by_user
       end
--- a/lib/diaspora/shareable.rb
+++ b/lib/diaspora/shareable.rb
@@ -57,6 +57,12 @@ module Diaspora
           user.person.send(table_name).where(pending: false)
         end
 
+        def self.shareable_initialize(params)
+          new(params.to_hash.stringify_keys.slice(*column_names)).tap do |new_shareable|
+            new_shareable.author = params[:author]
+          end
+        end
+
         def self.visible_by_user(user)
           ShareVisibility.arel_table[:user_id].eq(user.id)
         end
--- a/spec/lib/diaspora/shareable_spec.rb
+++ b/spec/lib/diaspora/shareable_spec.rb
@@ -17,5 +17,30 @@ describe Diaspora::Shareable do
         expect(Post.all_public.map(&:id)).to eq([])
       end
     end
+
+    context "having multiple objects with equal db IDs" do
+      before do
+        # Determine the next database key ID, free on both Photo and StatusMessage
+        id = [Photo, StatusMessage].map {|model| model.maximum(:id).try(:next).to_i }.push(1).max
+
+        alice.post(:status_message, id: id, text: "I'm #{alice.username}", to: alice.aspects.first.id, public: false)
+        alice.post(:photo, id: id, user_file: uploaded_photo, to: alice.aspects.first.id, public: false)
+        expect(StatusMessage.where(id: id)).to exist
+        expect(Photo.where(id: id)).to exist
+      end
+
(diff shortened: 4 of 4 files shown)


app/models/photo.rb
```diff
@@ -81,12 +81,12 @@ def ownership_of_status_message
...
   end
 
-  def self.diaspora_initialize(params = {})
-    photo = self.new params.to_hash.slice(:text, :pending)
-    photo.author = params[:author]
-    photo.public = params[:public] if params[:public]
-    photo.pending = params[:pending] if params[:pending]
+  class << self
+    alias_method :diaspora_initialize_shareable, :diaspora_initialize
+  end
```
SuperTux88: I don't like this, and I don't know if the DRY is worth this? You do this only to not write this twice?
```ruby
new(params.to_hash.stringify_keys.slice(*column_names)).tap do |photo|
  photo.author = params[:author]
```

cmrd-senya: Well, yes, there is now `diaspora_initialize` in both `Photo` and `Post` which do the same but implemented differently. I wanted to fix this.

But this aliasing is not too pretty.

What if I make method `shareable_initialize` in Diaspora::Shareable instead and call it from `Photo` and `Post` methods `diaspora_initialize`?

spec/lib/diaspora/shareable_spec.rb
```diff
@@ -17,5 +17,28 @@
         expect(Post.all_public.map(&:id)).to eq([])
       end
     end
+
+    context "having multiple objects with equal db IDs" do
+      before do
+        id = 60_000
```
cmrd-senya: I also feel now that hardcoding this constant is also not too good style. I know how to rewrite it without hardcoding. Do you think I should do it?

app/models/photo.rb
SuperTux88: `shareable_initialize` sounds better to me.

spec/lib/diaspora/shareable_spec.rb
SuperTux88: I also noticed that, and I think it's better if it's not hardcoded.

cmrd-senya:
Updated

app/models/post.rb
```diff
@@ -132,10 +132,8 @@ def like_for(user)
 
   #############
 
-  def self.diaspora_initialize(params)
-    new(params.to_hash.stringify_keys.slice(*column_names)).tap do |new_post|
-      new_post.author = params[:author]
-    end
+  class << self
+    alias diaspora_initialize shareable_initialize
```
SuperTux88: I would prefer
```ruby
  def self.diaspora_initialize(params)
    shareable_initialize(params)
  end
```
instead of the `class << self`. It's basically the same, but I think its better readable.

cmrd-senya: Changed

lib/diaspora/shareable.rb
```diff
@@ -55,9 +57,14 @@ def self.owned_by_user(user)
           user.person.send(table_name).where(pending: false)
         end
 
+        def self.shareable_initialize(params)
+          new(params.to_hash.stringify_keys.slice(*column_names)).tap do |new_post|
```
SuperTux88: `new_post` should be named `shareable` or so.

cmrd-senya: Maybe `new_shareable` then?

SuperTux88: I don't think we need the `new_` prefix here, but whatever...

cmrd-senya: I can still change it to `shareable` if you want :)

The greatest coding problem ever!

SuperTux88: I don't care :wink:

denschub:
Thanks for your patience, @cmrd-senya. Merged.

Thanks @SuperTux88 for your preliminary reviews!

PR #6814, opened 2016-04-27
Fix open_graph always showing default image

ghost:
have a look at `app/views/layouts/_open_graph.haml` it requires an instance var otherwise it will always fallback to the default image tag

### DIFF
--- a/spec/controllers/posts_controller_spec.rb
+++ b/spec/controllers/posts_controller_spec.rb
@@ -21,6 +21,11 @@ describe PostsController, type: :controller do
           expect(response).to be_success
         end
 
+        it "should instantiate post as global val for open_graph" do
+          get :show, id: post.id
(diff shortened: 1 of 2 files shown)


spec/controllers/posts_controller_spec.rb
```diff
@@ -21,6 +21,11 @@
           expect(response).to be_success
         end
 
+        it "should instantiate post as global val for open_graph" do
```
denschub: A test whether the actual URL is shown in the frontend would be better!

denschub:
Does it? OpenGraph works fine as far as I can tell. Can you provide an example where it does not?

svbergerem:
@Zauberstuhl #6998 refactored the open graph code a lot and at least now it works for me. (I didn't check it earlier)

PR #6816, opened 2016-04-28
A minor refactor of the Publisher

cmrd-senya:
There was a minor copy-paste issue with publisher_public and
publisher_explain methods. Fix it and do a little refactoring
of the publisher's code.


denschub:
Can you please elaborate what was wrong? I don't quite get that.

denschub:
Nevermind, I figured it out. Merged, thanks.

Issue #6796, opened 2016-04-16
Invite email would be a lot more useful if it told the Diaspora username


Mikaela:
I got invite to Diaspora from "Some Person". The invite email only named "Some Person" in the subject field.

I would otherwise join Diaspora, but I happened to already register to another pod before I was invited and now I am wondering who is "Some Person" on pod.example.com.

I think the invite email should say that their Diaspora ID is `some.person@pod.example.com`, so I could just immediately add them in the pod I am currently using.

goobertron:
That's a good idea.

svbergerem:
Fixed by #6817.

PR #6817, opened 2016-04-29
Add inviter diaspora-ID to invite email #6796

sakshi-keepworks:
The invite email should show the diaspora-ID of the person who has invited you to diaspora*. 

-- I am not sure how to add it for locales other than English.

### DIFF
--- a/app/views/notifier/invite.markerb
+++ b/app/views/notifier/invite.markerb
@@ -1,1 +1,1 @@
-<%= t('.message', :invite_url => invite_code_url(@invitation_code), :diasporafoundation_url => 'https://diasporafoundation.org/') %>
+<%= t('.message', :invite_url => invite_code_url(@invitation_code), :diasporafoundation_url => 'https://diasporafoundation.org/', :diaspora_id => @inviter.try(:diaspora_handle)) %>
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -851,11 +851,13 @@ en:
       message: |-
         Hello!
 
-        You have been invited to join diaspora*!
+        You have been invited to join diaspora* by %{diaspora_id}!
 
         Click this link to get started
 
         [%{invite_url}][1]
+
+        Or you can add %{diaspora_id} to your contacts if you already have an account.
 
 
         Love,
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -39,7 +39,7 @@ class InvitationsController < ApplicationController
       else
         params[:invitation_code]
       end
-
+    @inviter = user || InvitationCode.where(id: params[:invitation_code]).first.try(:user)
     if @invitation_code.present?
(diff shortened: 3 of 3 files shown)


SuperTux88:
I disagree with you, the email-address of the inviter is private, nobody can see it (not even contacts you are sharing with).

Also it is more or less useless for the usecase described in #6796. You should add the diaspora-ID of the inviter instead (as described in the linked issue). If the invited person already has an account on another pod, he can add the inviter with this ID (which is not possible with the email-address).

sjain1107:
Sounds right! Updated. 👍

SuperTux88:
Thanks, you should also change the commit-message.

And maybe it is better to add some text to the diaspora-ID, for people who don't know diaspora and don't know what to do with this "address", they could be confused because it looks like an email-address. ;)

app/views/notifier/invite.markerb
```diff
@@ -1 +1 @@
-<%= t('.message', :invite_url => invite_code_url(@invitation_code), :diasporafoundation_url => 'https://diasporafoundation.org/') %>
+<%= t('.message', :invite_url => invite_code_url(@invitation_code), :diasporafoundation_url => 'https://diasporafoundation.org/', :diaspora_id => @inviter.diaspora_handle) %>
```
SuperTux88: `@inviter` is not defined if you call this via `InvitationsController#email`, so you need to define it there too.

See broken spec: https://travis-ci.org/diaspora/diaspora/jobs/126729534#L591

config/locales/diaspora/en.yml
```diff
@@ -851,12 +851,15 @@ en:
...
         Hello!
 
-        You have been invited to join diaspora*!
+        You have been invited to join diaspora* by %{diaspora_id}!
 
         Click this link to get started
 
         [%{invite_url}][1]
 
+        You can add %{diaspora_id} to the pod you might be currently using.
```
sjain1107: Please let me know if this is a correct description. Also, I can't understand how to handle the case where `diaspora_id` is `nil`.

SuperTux88: I think "Or you can add %{diaspora_id} to your contacts if you already have an account." or something like this is clearer. Or maybe @goobertron has a better idea.

`diaspora_id` shouldn't be `nil`, because every user has a diaspora-ID. I don't know what happens if the user has closed his account, and if the invitation still works.

config/locales/diaspora/en.yml
```diff
@@ -851,12 +851,15 @@ en:
...
+
+
```
SuperTux88: Maybe you can remove one newline here?

sjain1107:
@SuperTux88 Thanks for review. I have updated the pull request.

denschub:
Merged as 4cd2f1d9c165807ce83dd79684380885baaf7756, thank you very much for your contribution!

Also, thanks @SuperTux88 for your preliminary reviews. :)

Issue #1851, opened 2011-08-30
User should be able to mention other users in comments


StepMiller:
You can't mention anyone in a comment.
It's really  curious: If you write for example

@`tester@joindiaspora.com`

it will be transformed to a link -> `http://tester@joindiaspora.com/` with an @ in front of it...
##

cmrd-senya:
If nobody minds, I'd like to work on this one.

jaywink:
That would be awesome @cmrd-senya , thanks!

goobertron:
I don't think anyone would mind that at all! Thank you.

Flaburgan:
You're very welcome to do it! There are couple of things to consider though, I discussed with @jhass about it a long time ago, I think the first step would be to write specifications because this is a non-trivial feature (choices have to be made).

jaywink:
@Flaburgan unless you mean pure technical details then the high level spec was already decided on [here](https://github.com/diaspora/diaspora/issues/1851#issuecomment-36557842) :dancer:

cmrd-senya:
I'm leaving for three weeks and I won't be able to work on it for a while. If the issue is not solved by the time I'm back by someone else I'll continue.

Flaburgan:
@jaywink I remember a discussion with @jhass a long time ago where he told me that it will be difficult (impossible?) to send notification to non-contact. The notification is, imo, the most important part of the mention (the link to the profile is not why I use notification). So I guess https://github.com/diaspora/diaspora/issues/5522 has to be solved first. There is also https://github.com/diaspora/diaspora/issues/4491 which is not a blocker, but which is going to be worse if not solved before this.

jaywink:
@Flaburgan 

> it will be difficult (impossible?) to send notification to non-contact.

Nothing is impossible when it comes to writing code ;) Bend the code to your will, not your will to the code!

It's possible something has to be done on the receiver end if the users are not connected or profile is not found on remote, but you can't say it is impossible - the relevant paths just need to be looked at and triggers fired to make sure a notification is created for the mentioned user.

pavi:
Its a nice feature to have. :+1:

cirosantilli:
+1

pravi:
@cmrd-senya has added this issue as a stretch goal to his account backup/restore crowd funding https://www.indiegogo.com/projects/account-backup-and-restore-for-diaspora--3#/ He will work on this if he reaches 4200 €, he'll work on this.

[...]

PR #6818, opened 2016-04-29
Mentions in comments backend changes

cmrd-senya:
This is an early preliminary PR with backend-changes to support mentions in comments (#1851).

It introduces changes to the DB schema making Mention model polymorphic and allowing a mention to belong to a comment. It also creates module Diaspora::MentionsContainer which incapsulates the logic of an object that supports keeping mentions. Now it includes logic I picked from the StatusMessage model.

ATM the mentions in comments are supported to the extent of placing user profile links instead of mention code in the mobile version (it is so because the mobile version doesn't require javascript code changes).

Changes summary:
- Make Mention model polymorphic;
- Introduce Diaspora::MentionsContainer which incapsulates the logic of an object that supports keeping mentions;
- We don't filter mentions anymore, instead we send notification/show in streams only that mentions, which are visible to the recipient;
- Create Notifications::Mentioned module which incapsulates common logic of mentioned notification;
- Add Notifications::MentionedInComment and move old Notifications::Mentioned to Notifications::MentionedInPost;
- Update the UI code to render new type of notification;
- Hack comments notifications to avoid having double notification when mentioned in a comment;

Drive-by refactorings:
- Move Notification.types out of the model to the controller;
- Move `StatusMessage#mentioned_people_names` out of the model. Create a `PersonPresenter.people_names` method instead;
- DRY UserController to use UserPreference::VALID_EMAIL_TYPES in the `user_params` method;

TODO:
- [x] Implement notifications for mentioning in comments
- [x] Implement restrictions of mentioning some kind of people (see [comment](https://github.com/diaspora/diaspora/issues/1851#issuecomment-36557842))
- [x] Ensure comment is federated to new discussion participants
- [x] Ensure mentions in comment are filtered on federation receive
- [x] ~~Ensure the filtered out mentions are replaced with links properly~~
- [x] Don't send redundant second notification when mentioned in comment.
- [x] Ensure email notifications work properly
- ~~Implement frontend javascript changes for comments publisher~~

### DIFF
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -2,40 +2,44 @@ module NotificationsHelper
   include PeopleHelper
   include PostsHelper
 
-  def object_link(note, actors)
+  def object_link(note, actors_html)
     target_type = note.popup_translation_key
-    actors_count = note.actors.size
+    opts = {actors: actors_html, count: note.actors.size}
 
-    if note.instance_of?(Notifications::Mentioned)
-      if post = note.linked_object
-        translation(target_type,
-                    actors:    actors,
-                    count:     actors_count,
-                    post_link: link_to(post_page_title(post), post_path(post)).html_safe)
-      else
-        t(note.deleted_translation_key, :actors => actors, :count => actors_count).html_safe
+    if note.respond_to?(:linked_object)
+      if note.linked_object.nil? && note.respond_to?(:deleted_translation_key)
+        target_type = note.deleted_translation_key
+      elsif note.is_a?(Notifications::Mentioned)
+        opts.merge!(opts_for_mentioned(note.linked_object))
+      elsif %w(Notifications::CommentOnPost Notifications::AlsoCommented Notifications::Reshared Notifications::Liked)
+            .include?(note.type)
+        opts.merge!(opts_for_post(note.linked_object))
       end
-    elsif note.instance_of?(Notifications::CommentOnPost) || note.instance_of?(Notifications::AlsoCommented) || note.instance_of?(Notifications::Reshared) || note.instance_of?(Notifications::Liked)
-      if post = note.linked_object
-        translation(target_type,
-                    actors:      actors,
-                    count:       actors_count,
-                    post_author: h(post.author_name),
-                    post_link:   link_to(post_page_title(post),
-                                         post_path(post),
-                                         data:  {ref: post.id},
-                                         class: "hard_object_link").html_safe)
-      else
-        t(note.deleted_translation_key, :actors => actors, :count => actors_count).html_safe
-      end
-    else #Notifications:StartedSharing, etc.
-      translation(target_type, :actors => actors, :count => actors_count)
     end
+    translation(target_type, opts)
   end
 
   def translation(target_type, opts = {})
-    {:post_author => nil}.merge!(opts)
     t("#{target_type}", opts).html_safe
+  end
+
+  def opts_for_post(post)
+    {
+      post_author: html_escape(post.author_name),
+      post_link:   link_to(post_page_title(post),
+                           post_path(post),
+                           data:  {ref: post.id},
+                           class: "hard_object_link").html_safe
+    }
+  end
+
+  def opts_for_mentioned(mentioned)
+    post = mentioned.instance_of?(Comment) ? mentioned.parent : mentioned
+    {
+      post_link: link_to(post_page_title(post), post_path(post)).html_safe
+    }.tap {|opts|
+      opts[:comment_path] = post_path(post, anchor: mentioned.guid).html_safe if mentioned.instance_of?(Comment)
+    }
   end
 
   def notification_people_link(note, people=nil)
--- a/app/mailers/notification_mailers/mentioned.rb
+++ b/app/mailers/notification_mailers/mentioned.rb
@@ -4,7 +4,7 @@ module NotificationMailers
     delegate :author_name, to: :post, prefix: true
 
     def set_headers(target_id)
-      @post = Mention.find_by_id(target_id).post
+      @post = Mention.find_by_id(target_id).mentions_container
 
       @headers[:subject] = I18n.t('notifier.mentioned.subject', :name => @sender.name)
       @headers[:in_reply_to] = @headers[:references] = "<#{@post.guid}@#{AppConfig.pod_uri.host}>"
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -11,6 +11,7 @@ class Comment < ActiveRecord::Base
 
   include Diaspora::Taggable
   include Diaspora::Likeable
+  include Diaspora::MentionsContainer
 
   acts_as_taggable_on :tags
   extract_tags_from :text
@@ -49,12 +50,20 @@ class Comment < ActiveRecord::Base
     participation.unparticipate! if participation.present?
   end
 
-  def message
-    @message ||= Diaspora::MessageRenderer.new text
-  end
-
   def text= text
      self[:text] = text.to_s.strip #to_s if for nil, for whatever reason
+  end
+
+  def people_allowed_to_be_mentioned
+    if parent.public?
+      :all
+    else
+      [*parent.comments.pluck(:author_id), *parent.likes.pluck(:author_id), parent.author_id].uniq
+    end
+  end
+
+  def add_mention_subscribers?
+    super && parent.author.local?
   end
 
   class Generator < Diaspora::Federated::Generator
@@ -68,7 +77,7 @@ class Comment < ActiveRecord::Base
     end
 
     def relayable_options
-      {:post => @target, :text => @text}
+      {post: @target, text: @text}
     end
   end
 end
--- a/app/models/mention.rb
+++ b/app/models/mention.rb
@@ -3,10 +3,14 @@
 #   the COPYRIGHT file.
 
 class Mention < ActiveRecord::Base
-  belongs_to :post
+  belongs_to :mentions_container, polymorphic: true
   belongs_to :person
-  validates :post, presence: true
+  validates :mentions_container, presence: true
   validates :person, presence: true
+
+  scope :local, -> {
+    joins(:person).where.not(people: {owner_id: nil})
+  }
 
   after_destroy :delete_notification
 
--- a/app/models/notifications/also_commented.rb
+++ b/app/models/notifications/also_commented.rb
@@ -1,5 +1,7 @@
 module Notifications
   class AlsoCommented < Notification
+    include Notifications::Commented
+
     def mail_job
       Workers::Mail::AlsoCommented
     end
@@ -8,17 +10,13 @@ module Notifications
       "notifications.also_commented"
     end
 
-    def deleted_translation_key
-      "notifications.also_commented_deleted"
-    end
-
     def self.notify(comment, _recipient_user_ids)
       actor = comment.author
       commentable = comment.commentable
       recipient_ids = commentable.participants.local.where.not(id: [commentable.author_id, actor.id]).pluck(:owner_id)
 
       User.where(id: recipient_ids).find_each do |recipient|
-        next if recipient.is_shareable_hidden?(commentable)
+        next if recipient.is_shareable_hidden?(commentable) || mention_notification_exists?(comment, recipient.person)
 
         concatenate_or_create(recipient, commentable, actor).try(:email_the_user, comment, actor)
       end
--- a/app/models/notifications/mentioned.rb
+++ b/app/models/notifications/mentioned.rb
@@ -1,30 +1,23 @@
 module Notifications
-  class Mentioned < Notification
-    def mail_job
-      Workers::Mail::Mentioned
-    end
-
-    def popup_translation_key
-      "notifications.mentioned"
-    end
-
-    def deleted_translation_key
-      "notifications.mentioned_deleted"
-    end
+  module Mentioned
+    extend ActiveSupport::Concern
 
     def linked_object
-      target.post
+      target.mentions_container
     end
 
-    def self.notify(mentionable, recipient_user_ids)
-      actor = mentionable.author
+    module ClassMethods
+      def notify(mentionable, recipient_user_ids)
+        actor = mentionable.author
+        relevant_mentions = filter_mentions(
+          mentionable.mentions.local.where.not(person: actor),
+          mentionable,
+          recipient_user_ids
+        )
 
-      mentionable.mentions.select {|mention| mention.person.local? }.each do |mention|
-        recipient = mention.person
-
-        next if recipient == actor || !(mentionable.public || recipient_user_ids.include?(recipient.owner_id))
-
-        create_notification(recipient.owner, mention, actor).try(:email_the_user, mention, actor)
+        relevant_mentions.each do |mention|
+          create_notification(mention.person.owner, mention, actor).try(:email_the_user, mention, actor)
+        end
       end
     end
   end
--- a/app/models/notifications/mentioned_in_comment.rb
+++ b/app/models/notifications/mentioned_in_comment.rb
@@ -1,0 +1,38 @@
+module Notifications
+  class MentionedInComment < Notification
+    include Notifications::Mentioned
+
+    def popup_translation_key
+      "notifications.mentioned_in_comment"
+    end
+
+    def deleted_translation_key
+      "notifications.mentioned_in_comment_deleted"
+    end
+
+    def self.filter_mentions(mentions, mentionable, _recipient_user_ids)
+      people = mentionable.people_allowed_to_be_mentioned
+      if people == :all
+        mentions
+      else
+        mentions.where(person_id: people)
+      end
+    end
+
+    def mail_job
+      if !recipient.user_preferences.exists?(email_type: "mentioned_in_comment")
+        Workers::Mail::MentionedInComment
+      elsif shareable.author.owner_id == recipient_id
+        Workers::Mail::CommentOnPost
+      elsif shareable.participants.local.where(owner_id: recipient_id)
+        Workers::Mail::AlsoCommented
+      end
+    end
+
+    private
+
+    def shareable
+      linked_object.parent
+    end
+  end
+end
--- a/app/models/notifications/mentioned_in_post.rb
+++ b/app/models/notifications/mentioned_in_post.rb
@@ -1,0 +1,22 @@
+module Notifications
+  class MentionedInPost < Notification
+    include Notifications::Mentioned
+
+    def mail_job
+      Workers::Mail::Mentioned
+    end
+
+    def popup_translation_key
+      "notifications.mentioned"
+    end
+
+    def deleted_translation_key
+      "notifications.mentioned_deleted"
+    end
+
+    def self.filter_mentions(mentions, mentionable, recipient_user_ids)
+      return mentions if mentionable.public
+      mentions.where(person: Person.where(owner_id: recipient_user_ids).ids)
+    end
+  end
+end
--- a/app/models/status_message.rb
+++ b/app/models/status_message.rb
@@ -23,13 +23,12 @@ class StatusMessage < Post
   attr_accessor :oembed_url
   attr_accessor :open_graph_url
 
-  after_create :create_mentions
   after_commit :queue_gather_oembed_data, :on => :create, :if => :contains_oembed_url_in_text?
   after_commit :queue_gather_open_graph_data, :on => :create, :if => :contains_open_graph_url_in_text?
 
   #scopes
   scope :where_person_is_mentioned, ->(person) {
-    joins(:mentions).where(:mentions => {:person_id => person.id})
+    owned_or_visible_by_user(person.owner).joins(:mentions).where(mentions: {person_id: person.id})
   }
 
   def self.guids_for_author(person)
@@ -50,36 +49,6 @@ class StatusMessage < Post
 
   def nsfw
     text.try(:match, /#nsfw/i) || super
-  end
-
-  def message
-    @message ||= Diaspora::MessageRenderer.new(text, mentioned_people: mentioned_people)
-  end
-
-  def mentioned_people
-    if self.persisted?
-      self.mentions.includes(:person => :profile).map{ |mention| mention.person }
-    else
-      Diaspora::Mentionable.people_from_string(text)
-    end
-  end
-
-  ## TODO ----
-  # don't put presentation logic in the model!
-  def mentioned_people_names
-    self.mentioned_people.map(&:name).join(', ')
-  end
-  ## ---- ----
-
-  def create_mentions
-    ppl = Diaspora::Mentionable.people_from_string(text)
-    ppl.each do |person|
-      self.mentions.find_or_create_by(person_id: person.id)
-    end
-  end
-
-  def mentions?(person)
-    mentioned_people.include? person
   end
 
   def comment_email_subject
@@ -124,6 +93,22 @@ class StatusMessage < Post
     super(recipient_user_ids)
 
     photos.each {|photo| photo.receive(recipient_user_ids) }
+  end
+
+  # Only includes those people, to whom we're going to send a federation entity
+  # (and doesn't define exhaustive list of people who can receive it)
+  def people_allowed_to_be_mentioned
+    @aspects_ppl ||=
+      if public?
+        :all
+      else
+        Contact.joins(:aspect_memberships).where(aspect_memberships: {aspect: aspects}).distinct.pluck(:person_id)
+      end
+  end
+
+  def filter_mentions
+    return if people_allowed_to_be_mentioned == :all
+    update(text: Diaspora::Mentionable.filter_people(text, people_allowed_to_be_mentioned))
   end
 
   private
--- a/app/services/post_service.rb
+++ b/app/services/post_service.rb
@@ -59,8 +59,20 @@ class PostService
   end
 
   def mark_mention_notifications_read(post_id)
-    mention_id = Mention.where(post_id: post_id, person_id: user.person_id).pluck(:id)
-    Notification.where(recipient_id: user.id, target_type: "Mention", target_id: mention_id, unread: true)
-      .update_all(unread: false) if mention_id
+    mention_ids = Mention.where(
+      mentions_container_id:   post_id,
+      mentions_container_type: "Post",
+      person_id:               user.person_id
+    ).ids
+    mention_ids.concat(mentions_in_comments_for_post(post_id).pluck(:id))
+
+    Notification.where(recipient_id: user.id, target_type: "Mention", target_id: mention_ids, unread: true)
+                .update_all(unread: false) if mention_ids.any?
+  end
+
+  def mentions_in_comments_for_post(post_id)
+    Mention
+      .joins("INNER JOIN comments ON mentions_container_id = comments.id AND mentions_container_type = 'Comment'")
+      .where(comments: {commentable_id: post_id, commentable_type: "Post"})
   end
 end
--- a/app/services/status_message_creation_service.rb
+++ b/app/services/status_message_creation_service.rb
@@ -19,18 +19,7 @@ class StatusMessageCreationService
 
   def build_status_message(params)
     public = params[:public] || false
-    filter_mentions params
     user.build_post(:status_message, params[:status_message].merge(public: public))
-  end
-
-  def filter_mentions(params)
-    unless params[:public]
-      params[:status_message][:text] = Diaspora::Mentionable.filter_for_aspects(
-        params[:status_message][:text],
-        user,
-        *params[:aspect_ids]
-      )
-    end
   end
 
   def add_attachments(status_message, params)
@@ -75,6 +64,7 @@ class StatusMessageCreationService
 
   def dispatch(status_message, services)
     receiving_services = services ? Service.titles(services) : []
+    status_message.filter_mentions # this is only required until changes from #6818 are deployed on every pod
     user.dispatch_post(status_message,
                        url:           short_post_url(status_message.guid, host: AppConfig.environment.url),
                        service_types: receiving_services)
--- a/app/views/notifier/mentioned.markerb
+++ b/app/views/notifier/mentioned.markerb
@@ -4,6 +4,6 @@
 <%= t('notifier.mentioned.limited_post') %>
 <% end %>
 
-[<%= t('notifier.comment_on_post.reply', :name => @notification.post_author_name) %>][1]
+[<%= t("notifier.comment_on_post.reply", name: @notification.post_author_name) %>][1]
 
 [1]: <%= post_url(@notification.post) %>
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -615,9 +615,11 @@ en:
       one: "%{actors} also commented on %{post_author}’s post %{post_link}."
       other: "%{actors} also commented on %{post_author}’s post %{post_link}."
     mentioned:
-      zero: "%{actors} have mentioned you in the post %{post_link}."
       one: "%{actors} has mentioned you in the post %{post_link}."
       other: "%{actors} have mentioned you in the post %{post_link}."
+    mentioned_in_comment:
+      one: "%{actors} has mentioned you in a <a href='%{comment_path}'>comment</a> to the post %{post_link}."
+      other: "%{actors} have mentioned you in a <a href='%{comment_path}'>comment</a> to the post %{post_link}."
     liked:
       zero: "%{actors} have liked your post %{post_link}."
       one: "%{actors} has liked your post %{post_link}."
@@ -640,9 +642,11 @@ en:
       one: "%{actors} reshared your deleted post."
       other: "%{actors} reshared your deleted post."
     mentioned_deleted:
-      zero: "%{actors} mentioned you in a deleted post."
       one: "%{actors} mentioned you in a deleted post."
       other: "%{actors} mentioned you in a deleted post."
+    mentioned_in_comment_deleted:
+      one: "%{actors} mentioned you in a deleted comment."
+      other: "%{actors} mentioned you in a deleted comment."
     index:
       notifications: "Notifications"
       mark_all_as_read: "Mark all as read"
@@ -655,7 +659,8 @@ en:
       also_commented: "Also commented"
       comment_on_post: "Comment on post"
       liked: "Liked"
-      mentioned: "Mentioned"
+      mentioned: "Mentioned in post"
+      mentioned_in_comment: "Mentioned in comment"
       reshared: "Reshared"
       started_sharing: "Started sharing"
       no_notifications: "You don't have any notifications yet."
@@ -689,6 +694,9 @@ en:
     mentioned:
         subject: "%{name} has mentioned you on diaspora*"
         limited_post: "You were mentioned in a limited post."
+    mentioned_in_comment:
+        limited_post: "You were mentioned in a comment to a limited post."
+        reply: "Reply to or view this conversation >"
     private_message:
         subject: "There’s a new private message for you"
         reply_to_or_view: "Reply to or view this conversation >"
@@ -1167,6 +1175,7 @@ en:
       started_sharing: "someone starts sharing with you"
       someone_reported: "someone sends a report"
       mentioned: "you are mentioned in a post"
+      mentioned_in_comment: "you are mentioned in a comment"
       liked: "someone likes your post"
       reshared: "someone reshares your post"
       comment_on_post: "someone comments on your post"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161024231443) do
+ActiveRecord::Schema.define(version: 20161107100840) do
 
   create_table "account_deletions", force: :cascade do |t|
     t.string   "diaspora_handle", limit: 255
@@ -204,13 +204,14 @@ ActiveRecord::Schema.define(version: 20161024231443) do
   end
 
   create_table "mentions", force: :cascade do |t|
-    t.integer "post_id",   limit: 4, null: false
-    t.integer "person_id", limit: 4, null: false
+    t.integer "mentions_container_id",   limit: 4,   null: false
+    t.integer "person_id",               limit: 4,   null: false
+    t.string  "mentions_container_type", limit: 255, null: false
   end
 
-  add_index "mentions", ["person_id", "post_id"], name: "index_mentions_on_person_id_and_post_id", unique: true, using: :btree
+  add_index "mentions", ["mentions_container_id", "mentions_container_type"], name: "index_mentions_on_mc_id_and_mc_type", length: {"mentions_container_id"=>nil, "mentions_container_type"=>191}, using: :btree
+  add_index "mentions", ["person_id", "mentions_container_id", "mentions_container_type"], name: "index_mentions_on_person_and_mc_id_and_mc_type", unique: true, length: {"person_id"=>nil, "mentions_container_id"=>nil, "mentions_container_type"=>191}, using: :btree
   add_index "mentions", ["person_id"], name: "index_mentions_on_person_id", using: :btree
-  add_index "mentions", ["post_id"], name: "index_mentions_on_post_id", using: :btree
 
   create_table "messages", force: :cascade do |t|
     t.integer  "conversation_id",  limit: 4,     null: false
--- a/features/desktop/notifications.feature
+++ b/features/desktop/notifications.feature
@@ -101,6 +101,26 @@ Feature: Notifications
     Then I should see "mentioned you in the post"
     And I should have 1 email delivery
 
+  Scenario: someone mentioned me in a comment
+    Given "alice@alice.alice" has a public post with text "check this out!"
+    And "bob@bob.bob" has commented mentioning "alice@alice.alice" on "check this out!"
+    When I sign in as "alice@alice.alice"
+    And I follow "Notifications" in the header
+    Then the notification dropdown should be visible
+    And I should see "mentioned you in a comment"
+    And I should have 1 email delivery
+
+  Scenario: I mark a notification as read
+    Given a user with email "bob@bob.bob" is connected with "alice@alice.alice"
+    And Alice has a post mentioning Bob
+    When I sign in as "bob@bob.bob"
+    And I follow "Notifications" in the header
+    Then the notification dropdown should be visible
+    And I wait for notifications to load
+    And I should see a ".unread .unread-toggle .entypo-eye"
+    When I click on selector ".unread .unread-toggle .entypo-eye"
+    Then I should see a ".read .unread-toggle"
+
   Scenario: filter notifications
     Given a user with email "bob@bob.bob" is connected with "alice@alice.alice"
     And Alice has a post mentioning Bob
--- a/lib/diaspora/mentions_container.rb
+++ b/lib/diaspora/mentions_container.rb
@@ -1,0 +1,38 @@
+module Diaspora
+  module MentionsContainer
+    extend ActiveSupport::Concern
+
+    included do
+      after_create :create_mentions
+      has_many :mentions, as: :mentions_container, dependent: :destroy
+    end
+
+    def mentioned_people
+      if persisted?
+        mentions.includes(person: :profile).map(&:person)
+      else
+        Diaspora::Mentionable.people_from_string(text)
+      end
+    end
+
+    def add_mention_subscribers?
+      public?
+    end
+
+    def subscribers
+      super.tap {|subscribers|
+        subscribers.concat(mentions.map(&:person).select(&:remote?)) if add_mention_subscribers?
+      }
+    end
+
+    def create_mentions
+      Diaspora::Mentionable.people_from_string(text).each do |person|
+        mentions.find_or_create_by(person_id: person.id)
+      end
+    end
+
+    def message
+      @message ||= Diaspora::MessageRenderer.new text, mentioned_people: mentioned_people
+    end
+  end
+end
--- a/spec/integration/mentioning_spec.rb
+++ b/spec/integration/mentioning_spec.rb
@@ -1,10 +1,39 @@
 module MentioningSpecHelpers
-  def default_aspect
-    @user1.aspects.where(name: "generic").first
+  def notifications_about_mentioning(user, object)
+    table = object.class.table_name
+
+    if object.is_a?(StatusMessage)
+      klass = Notifications::MentionedInPost
+    elsif object.is_a?(Comment)
+      klass = Notifications::MentionedInComment
+    end
+
+    klass
+      .where(recipient_id: user.id)
+      .joins("LEFT OUTER JOIN mentions ON notifications.target_id = mentions.id AND "\
+             "notifications.target_type = 'Mention'")
+      .joins("LEFT OUTER JOIN #{table} ON mentions_container_id = #{table}.id AND "\
+             "mentions_container_type = '#{object.class.base_class}'").where(table.to_sym => {id: object.id})
   end
 
-  def text_mentioning(user)
-    "this is a text mentioning @{#{user.name}; #{user.diaspora_handle}} ... have fun testing!"
+  def mention_container_path(object)
+    object.is_a?(Post) ? post_path(object) : post_path(object.parent, anchor: object.guid)
+  end
+
+  def mentioning_mail_notification(user, object)
+    ActionMailer::Base.deliveries.select {|delivery|
+      delivery.to.include?(user.email) &&
+        delivery.subject.include?(I18n.t("notifier.mentioned.subject", name: "")) &&
+        delivery.body.parts[0].body.include?(mention_container_path(object))
+    }
+  end
+
+  def also_commented_mail_notification(user, post)
+    ActionMailer::Base.deliveries.select {|delivery|
+      delivery.to.include?(user.email) &&
+        delivery.subject.include?(I18n.t("notifier.also_commented.limited_subject")) &&
+        delivery.body.parts[0].body.include?(post_path(post))
+    }
   end
 
   def stream_for(user)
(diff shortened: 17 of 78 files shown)


app/models/person.rb
```diff
@@ -172,6 +172,10 @@ def self.search_order
     }.call
   end
 
+  def self.people_names(people)
```
SuperTux88:
```
## TODO ----
# don't put presentation logic in the model!
```

Also, this is only used once, why don't you just move `.map(&:name).join(", ")` to where its used? Maybe the controller is also not the best place, but it's better than split this and add it to the model.

db/schema.rb
```diff
@@ -211,13 +211,14 @@
...
   create_table "mentions", force: :cascade do |t|
-    t.integer "post_id",   limit: 4, null: false
-    t.integer "person_id", limit: 4, null: false
+    t.integer "mentions_container_id",   limit: 4,                    null: false
+    t.integer "person_id",               limit: 4,                    null: false
+    t.string  "mentions_container_type", limit: 255, default: "Post", null: false
   end
 
-  add_index "mentions", ["person_id", "post_id"], name: "index_mentions_on_person_id_and_post_id", unique: true, using: :btree
+  add_index "mentions", ["mentions_container_id"], name: "index_mentions_on_mentions_container_id", using: :btree
```
SuperTux88: you should also add `mentions_container_type` to this index, because it is never used without the type.

app/models/mention.rb
```diff
@@ -5,16 +5,22 @@
 class Mention < ActiveRecord::Base
   REGEX = /@\{([^;]+); ([^\}]+)\}/
 
-  belongs_to :post
+  belongs_to :mentions_container, polymorphic: :true
+  belongs_to :post, ->(mention) { (mention.mentions_container_type == "Post") ? where("1 = 1") : none },
```
SuperTux88: WTF? what does this? and why?

cmrd-senya: This allow to write
```
Mention.where(mentions_container_id: @post.id, mentions_contrainer_type: "Post")
```
as
```
Mention.where(post: @post)
```

app/models/person.rb
cmrd-senya: Well, I like to have it as a separate method, if possible. I don't know of a better place for that though. Any ideas?

app/models/mention.rb
SuperTux88: ah, fancy stuff.

But when you do
```ruby
Mention.where(post: Post.find(post_id))
```
it needs two querys and I don't know why this should be better than
```ruby
Mention.where(mentions_container_id: post_id, mentions_contrainer_type: "Post")
```

cmrd-senya: I picked the idea from the second answer [here](http://stackoverflow.com/questions/21850995/association-for-polymorphic-belongs-to-of-a-particular-type). I don't like the self-reference approach from the first answer, it looks less-obvious for me.

cmrd-senya: Ok, I'll check it once again then. One of my purposes for that was to escape rewriting all the code which references to Mention through post.

cmrd-senya:
Ok, I have completed backend support for mentions in comments here. Front end changes are completely separate thing and I suppose they can be implemented&reviewed separetely. This PR can be tested using the markdown syntax of mention directly: `@{alice; alice@example.tld}`.

Can be reviewed now.

app/helpers/notifications_helper.rb
```diff
@@ -4,40 +4,46 @@ module NotificationsHelper
...
+    }
+  end
+
+  def opts_for_mentioned(note)
+    post = note.linked_object.instance_of?(Comment) ? note.linked_object.parent : note.linked_object
+    {
+      post_link: link_to(post_page_title(post), post_path(post)).html_safe
+    }.tap {|opts|
+      if note.linked_object.instance_of?(Comment)
+        opts.merge!(comment_path: "#{post_path(post)}##{note.linked_object.guid}".html_safe)
```
jhass: IIRC path helpers take an `anchor` option, `post_path(post, anchor: note.linked_object.guid)`.

lib/diaspora/mentions_container.rb
```diff
@@ -0,0 +1,37 @@
+module Diaspora
+  module MentionsContainer
+    def self.included(model)
+      model.instance_eval do
+        after_create :create_mentions
+        has_many :mentions, as: :mentions_container, dependent: :destroy
+      end
+    end
```
jhass: Well, we're inside Rails, so it might be more idiomatic to use `AS::Concern` here.

```rb
  extend ActiveSupport::Concern
  included do
    ....
  end
````

File: app/models/comment.rb [hunk 6]
```diff
@@ -46,14 +48,17 @@ class Comment < ActiveRecord::Base
...
-
   def text= text
      self[:text] = text.to_s.strip #to_s if for nil, for whatever reason
   end
 
+  def people_allowed_to_be_mentioned
+    return @people_allowed_to_be_mentioned unless @people_allowed_to_be_mentioned.nil?
+    contacts = author.owner.try(:contact_people) if parent.public?
+
+    @people_allowed_to_be_mentioned = parent.try(:participants).to_a.concat(contacts.to_a).push(parent.author).uniq
```
jhass: Can you try whether `AR::Relation is splattable? `[*parent.try(:participants), *contacts, parent.author].uniq` Though even with explicit `to_a`s that might be a little bit more readable :)

jhass:
Sorry I don't remember, since we allow to bring new people into the conversation by allowing the commenter to mention anyone of their contacts, did we decide how to federate the comment in case the mentioned person never got the post initially?

cmrd-senya:
I walked quickly through the discussions and as far as I can see there was no decision. We allow to bring new people into the conversation for the public posts only. In this case we can add a mentioned person to the list of the comments subscribers and the comment will be federated to a new pod. After that, the new receiver pod may fetch the post if it's yet unknown. AFAIK, with the new federation it is nothing that we can't do.

Also I'm a bit concerned with the way notifications work in this PR. In cases when we mention a participant, the participant received two notifications (e.g. AlsoCommented and MentionedInComment). That's not too elegant. I'll seek for a way to mute CommentOnPost and AlsoCommented in case when we send MentionedInComment to a person.

Since these two issues are unresolved in the PR I'll set the status to [WIP] again.

app/models/comment.rb
cmrd-senya: > I'm not quite sure I follow why we know parent.author always succeeds but parent.participants may not

We have a [spec](https://github.com/diaspora/diaspora/blob/develop/spec/models/photo_spec.rb#L209) for photos commenting, and photos don't have participants. This spec fails when I do `parent.participants` with no `try`. Am I right that we don't have separate photos comments allowed now?

Flaburgan: >Am I right that we don't have separate photos comments allowed now?

At the moment, the interface doesn't allow users to comment on a specific photo, they can only comment on the post which contains the photo. But I guess this can change in the future.

cmrd-senya: So I guess we should leave it as a resposibility of Photo to implement `participants` and stay with `try` or maybe stub this method with `[]` for now. Which is better?

jhass: I'm not sure, I think I would get rid of the spec for now. I don't see reintroducing photo specific comments coming any time soon.

cmrd-senya: > Can you try whether AR::Relation is splattable?
Seems so.

> I think I would get rid of the spec for now.
Ok, then I'll remove the spec if nobody minds.

app/models/notifications/mentioned_in_comment.rb
```diff
@@ -0,0 +1,27 @@
...
+    def self.notify(comment, recipient_user_ids)
+      # recipient_user_ids are used in Notifications::Mentioned.notify in order to filter notification
+      # recipients for private posts. It's not clear, whether this is required, since mentions filtering
+      # is performed on the post/comment creation stage and we could consider every mention object as a
+      # sufficient reason for sending a notification. The issue is that recipient_user_ids never
+      # contains parent owner id and therefore the post author doesn't receive the notification if
+      # mentioned in comment.
+      #
+      # I left intact the Notifications::Mentioned.notify and didn't investigate the message dispatcher
+      # but here I add parent_owner_id to the list so the filtering passes well.
```
SuperTux88: This doesn't work, because private posts are received multiple times (to create the ShareVisibilities), so the mention-notification is only sent, if the user is in the `recipient_ids` (the user who received the visibility). But comments are only received once (all others are ignored), because all visibilities are already there. See `AlsoCommented`.

cmrd-senya: Hmm, I just tried to remove this method and everything seems to work fine. I made a mistake, thanks for pointing!

SuperTux88: No, that will not work either. If you use the parent-method, it will work for public posts, but for private-posts it will only send a notification, if the mentioned person is the first who receives it.

I'm not even sure if we should extend from `Mentioned`, because the `notify`-method would be simpler (than the parent) if you completely write the logic, instead of trying to use some parent logic.

cmrd-senya: > No, that will not work either. If you use the parent-method, it will work for public posts, but for private-posts it will only send a notification, if the mentioned person is the first who receives it.

Ah, ok, that means my tests don't cover that :)

> I'm not even sure if we should extend from Mentioned, because the notify-method would be simpler (than the parent) if you completely write the logic, instead of trying to use some parent logic.

Yes, that what I probably should do.

cmrd-senya:
@SuperTux88, I ended up with keeping inheritance. Please check 048d6b0777ed6065a9c2df9cb821b4a5e689e4d9 and tell me if you have any objections on the approach.

app/models/notifications/mentioned.rb
```diff
@@ -13,19 +13,25 @@ def deleted_translation_key
...
+      get_mentions(mentionable, recipient_user_ids).each do |mention|
         recipient = mention.person
-
-        next if recipient == actor || !(mentionable.public || recipient_user_ids.include?(recipient.owner_id))
-
         create_notification(recipient.owner_id, mention, actor).email_the_user(mention, actor)
       end
     end
+
+    def self.get_mentions(mentionable, recipient_user_ids)
```
SuperTux88: I think we shouldn't use `get_`, rubocop has a cop for this, but only for instance methods.

lib/diaspora/mentions_container.rb
```diff
@@ -0,0 +1,41 @@
+module Diaspora
+  module MentionsContainer
+    extend ActiveSupport::Concern
+
+    included do
+      after_create :create_mentions
+      has_many :mentions, as: :mentions_container, dependent: :destroy
+    end
+
+    def local_mentions
```
SuperTux88: I would add a scope `local` to the `Mention` model, so you can use `mentionable.mentions.local`

app/models/notifications/mentioned.rb
```diff
@@ -13,19 +13,25 @@ def deleted_translation_key
...
+      relation = mentionable.local_mentions.where.not(person: mentionable.author)
```
SuperTux88: this is the same line as in the `MentionedInComment`

app/models/notifications/mentioned.rb
```diff
@@ -13,19 +13,25 @@ def deleted_translation_key
...
+      if mentionable.public
+        relation
+      else
+        relation.where(person_id: recipient_user_ids.map {|id| User.find(id).person_id })
```
SuperTux88: `Person.where(owner_id: recipient_user_ids).ids`

app/models/comment.rb
```diff
@@ -64,6 +75,12 @@ def initialize(person, target, text)
       super(person, target)
     end
 
+    def build(*args)
+      super.tap do |comment|
+        comment.filter_mentions(comment.people_allowed_to_be_mentioned)
```
SuperTux88: We only filter this on send? and not on receive? Or do I miss something? So what happens, if we receive a comment with a mention (maybe from another network that doesn't have this outgoing filter), and the mentioned user can not see the post? does he receive a notification?

app/models/status_message.rb
```diff
@@ -121,6 +89,13 @@ def post_location
     }
   end
 
+  def people_allowed_to_be_mentioned
```
SuperTux88: where is this method called?

app/models/comment.rb
cmrd-senya: It doesn't seem to be filtered properly at the receive via federation, but that probably also applies to mentions in StatusMessage since the #6902 merge (or earlier). Gonna write some tests for that.

app/models/status_message.rb
cmrd-senya: This method was replaced with [this line](https://github.com/diaspora/diaspora/pull/6818/files#diff-1b42e0e2c5d972f2a8ef8ea55b6b92dfR23) after rebase on 46cbc6e52a9d4a03dff8228ce11ac4876d247a15, so it's not used. Thanks for noticing.

app/models/comment.rb
SuperTux88: Yes, maybe status messages are also not properly filtered on receive. It was just an idea, that we may should also filter on receive, so we don't send notifications, if the user can't see the post. 😃

cmrd-senya: So [looks like](https://travis-ci.org/diaspora/diaspora/jobs/147856327#L653) mentioning in status message and post via federation may bypass some visibility restrictions.

app/models/notifications/mentioned.rb
```diff
@@ -13,19 +13,26 @@ def deleted_translation_key
...
+      relevant_mentions.each do |mention|
         recipient = mention.person
-
-        next if recipient == actor || !(mentionable.public || recipient_user_ids.include?(recipient.owner_id))
-
         create_notification(recipient.owner_id, mention, actor).email_the_user(mention, actor)
       end
     end
+
+    def self.filter_mentions(mentions, mentionable, recipient_user_ids)
```
SuperTux88: So the point where I don't like the inheritance here is, that the method on the parent class is the "special case" (filter on private posts), and the same method on the child class, should be the default case (just no filtering). But invert the inheritance (extend `Mentioned` from `MentionedInComment`) feels also wrong.

So maybe the solution is to add a `AbstractMentioned` with the default (no filtering) `filter_mentions`, `notify` and `linked_object` method, and then both of `Mentioned` and `MentionedInComment` extending from this, and add the translation_keys and override the `filter_mentions` in the `Mentioned` class.

Just some thoughts ...

spec/integration/mentioning_spec.rb
```diff
@@ -23,73 +31,254 @@ def mention_stream_for(user)
...
+    entity = FactoryGirl.build(
+      :status_message_entity,
+      author: remote_user_on_pod_b.diaspora_handle,
+      text:   text,
+      public: false
+    )
+
+    allow_callbacks(%i(queue_private_receive receive_entity fetch_related_entity))
+
+    xml = generate_xml(entity, remote_user_on_pod_b, recipient)
```
SuperTux88: You don't need this whole generate and parse xml here, just send the `entity` to the `:receive_entity` callback.

db/migrate/20160621151100_polymorphic_mentions.rb
```diff
@@ -0,0 +1,12 @@
+class PolymorphicMentions < ActiveRecord::Migration
+  def change
+    remove_index :mentions, %i(post_id)
+    remove_index :mentions, %i(person_id post_id)
+    rename_column :mentions, :post_id, :mentions_container_id
+    add_column :mentions, :mentions_container_type, :string, default: "Post", null: false
+    add_index :mentions, %i(mentions_container_id mentions_container_type), name: "index_mentions_on_mc_id_and_mc_type"
```
SuperTux88: the migration breaks on mysql:
```
Mysql2::Error: Specified key was too long; max key length is 767 bytes: CREATE  INDEX `index_mentions_on_mc_id_and_mc_type`  ON `mentions` (`mentions_container_id`, `mentions_container_type`)
```

app/models/notifications/mentioned.rb
```diff
@@ -13,19 +13,26 @@ def deleted_translation_key
...
 
     def linked_object
-      target.post
+      target.mentions_container
     end
 
     def self.notify(mentionable, recipient_user_ids)
       actor = mentionable.author
+      relevant_mentions = filter_mentions(
+        mentionable.find_or_create_mentions.local.where.not(person: mentionable.author),
```
SuperTux88: why did you change this to `find_or_create_mentions`? The `mentions` should always be there at this point, if you create them in the `after_create`. The problem with `find_or_create_mentions` is, that this always parses the whole text again, if the post has NO mentions, because `mentions` would ALWAYS be empty for this posts, so it tries to create it. And `after_create` does this once. (same goes for comments without mentions)

app/models/comment.rb
SuperTux88: I discovered that this incoming filter is a bit more complex than I thought first.

For **public** stuff it's easy: don't filter anything, the sender can mention anyone. You can't know the senders aspects, so you don't know if he is allowed to mention someone.

For **private comments** it's easy too: you can only mention other participants, every pod can validate and filter that.

For **private posts** it's a bit more complex: We can not know if the mentioned person can see the post, when we receive it (that's what I not considered before). So we can not filter anything here and need to trust the sender. But my notification logic already handles the notifications correctly, so the person only receives a notification, when he receives the visibility. So maybe we need to ensure that the person can not see the post on his stream, unless he also receives the visibility, instead of filter the mention?

So what you built now is, that you filter all mentions on private posts on receive, which makes mention impossible on private posts.

spec/integration/mentioning_spec.rb
```diff
@@ -1,13 +1,21 @@
 require "spec_helper"
 require "requests_helper"
+require "integration/federation/federation_helper"
```
SuperTux88: Can you try to write this tests without the federation helper? When you don't use `remote_user_on_pod_b` and `remote_user_on_pod_c` (which add stubs on the callbacks which you don't need), you also don't need to use  `allow_callbacks` for the tests (no stubs == everything allowed with call original, which is what you want here).

Have a look at `remote_raphael`.

app/models/notifications/mentioned.rb
jhass: Do we need classic inheritance at all? Or stub definitions in the shared code? I'd say extract a module for the common code and just the common code and include it into both.

app/models/comment.rb
cmrd-senya: > So what you built now is, that you filter all mentions on private posts on receive, which makes mention impossible on private posts.

Not really. Have a look on [this](https://github.com/diaspora/diaspora/pull/6818/files#diff-2bcec3a9d410c74617f05ba3d8408810R156) test -- it passes well, that means that mentions work on private receive. Actually, what I built now is no filtering at all (in the federation case). It rather creates mentions only for people who are allowed to be mentioned. However it doesn't work well right now in the case of multiple receive again.

The question is when do we create Mention objects for post/comment. Mention objects was originally created only when mention is allowed. That's why filtering was used - to avoid creation of Mention objects. If we change the logic of stream rendering so it hides posts when person is not allowed to see them irrespective of Mention object presence, then we could create Mention objects for all mentioned people immediately after receive. I guess this is what you suggest.

After some thoughts, I agree that this is more optimal way to implement visibility restriction for mentioning. This way we won't need any filtering at all and get more straightforward design.

app/models/notifications/mentioned.rb
cmrd-senya: That is actually the same thing that we do know when we request `mentioned_people` for StatusMessage (see https://github.com/diaspora/diaspora/blob/develop/app/models/status_message.rb#L61).

However, we won't need this, if we introduce changes we discussed above in https://github.com/diaspora/diaspora/pull/6818#discussion-diff-72363123

app/models/notifications/mentioned.rb
cmrd-senya: Actually, I like classic inheritance here more. @SuperTux88, what do you think?

If I extract a module, where do I put it and how do I call it? For example if I make `Notification::MentionedHelper` and put it somewhere in `lib/` it's less convenient than to have a base class along with children in the same folder.

jhass: If you can argue for the base class, which is never directly used directly to live in the same folder, so you can for the module. If you can argue for the base class to be called something with an artificial suffix, like `MentionedBase`, so you can for the module (Something like `Mentioned` and `PostMentioned`, `CommentMentioned` would deprive from doing that in either case btw.)

Ruby has no abstract base classes, it uses modules to share functionality.

jhass: Btw a pattern that emerges in the Rails community is to have a `concerns` subfolder and in there modules (that extend `ActiveSupport::Concern`). I'm not yet sure whether I like that, but it's another option.

cmrd-senya: Sounds convincing. So is everybody ok with putting the shared module with common code in `app/models/notification`? I think I can call it `MentionedShared` or `MentionedCommon`, or something like that.

I see that we already have some extension modules for `User` in our `app/models/user` folder which I can consider as an acceptable decision then.

jhass: I think renaming `Mentioned` to `MentionedInPost` in order to free the name for the shared module is fair.

app/models/comment.rb
SuperTux88: > Not really. Have a look on this test -- it passes well, that means that mentions work on private receive.

That test is not really useful:
1. you don't know who can see this post on receive, because you only receive the first visibility.
2. my local test (where I saw that it doesn't work with private) was: podA  has your code, user@podB mentions user@podC (which is also a receiver of this post), but the mention was filtered (because podA doesn't know about that).

You can filter local users and recreate filtered mentions on second receive, if the second user can see it now, but I doubt that this is practicable.

So that all is no real solution for the problem. I think the most reasonable solution is to filter the streams, so you can only view posts, that you are allowed to see. We should still filter on outgoing stuff (because old posts would otherwise show the posts to users who shouldn't see it), but incoming private posts are 
impossible to filter, because you don't know who can see it.

But my original intention with this topic (filter on receive) was, to ensure that a user don't receives a notification for a mention in a comment, if he can't see the post. If this is done on notification-level (like for notifications for mentions in posts, where this already works), I'm totally fine with it.

app/models/notifications/mentioned.rb
cmrd-senya: Does it mean that I have to write a migration in order to change the `type` of notification in the DB  or can I just override the convention like this?
```ruby
class Notification::MentionedInPost < Notification
...
  def self.sti_name
      "Notification::Mentioned"
  end
...
end
```

jhass: I think I would prefer a migration :)

db/migrate/20160621151100_polymorphic_mentions.rb
cmrd-senya: Works for me when I do
```
git checkout develop
RAILS_ENV=development FIXTURES_PATH=spec/fixtures bundle exec rake db:drop db:create db:schema:load db:fixtures:load
... posting post with mention ...
git checkout 1851-mentions-in-comments.new
bundle exec rake db:migrate
```
Does database content make difference here?
MySQL version is `mysqld  Ver 5.5.49-0ubuntu0.14.04.1 for debian-linux-gnu on x86_64 ((Ubuntu))`.

app/models/notifications/mentioned.rb
```diff
@@ -1,38 +1,29 @@
...
 
-    def self.notify(mentionable, recipient_user_ids)
-      actor = mentionable.author
-      relevant_mentions = filter_mentions(
-        mentionable.find_or_create_mentions.local.where.not(person: mentionable.author),
-        mentionable,
-        recipient_user_ids
-      )
+    included do
+      def self.notify(mentionable, recipient_user_ids)
```
jhass: When using `AS::Concern`, instead of this, define a nested module called `ClassMethods` which have the class methods defined as instance methods.

db/migrate/20160621151100_polymorphic_mentions.rb
SuperTux88: When I try the same, I get the error: MariaDB 10.0.23

But I can create the index manually in the DB.

Anyway, why don't you just add the `length` parameter as in other migrations where indexes are added? WIth the `length` it works.

db/migrate/20160621151100_polymorphic_mentions.rb
```diff
@@ -0,0 +1,12 @@
+class PolymorphicMentions < ActiveRecord::Migration
+  def change
+    remove_index :mentions, %i(post_id)
```
SuperTux88: This doesn't work for rollback: `remove_index is only reversible if given a :column option.`

app/services/status_message_creation_service.rb
```diff
@@ -19,20 +19,9 @@ def create(params)
 
   def build_status_message(params)
     public = params[:public] || false
-    filter_mentions params
```
SuperTux88: you should still filter on send, because otherwise old pods will display the post to people who shouldn't see it.

cmrd-senya: fixed in https://github.com/diaspora/diaspora/pull/6818/commits/56dd62ceb39ee0f372eb549a71ac0df66a4f51c9

cmrd-senya:
Ok, this feels completed to me. Setting [WIP] off. It is ready to, hopefully, final review.

But I guess it won't be merged before the 0.6.0.0 release anyway? (Since there is no UI yet for mentioning in comment publisher, therefore it is not a complete feature).

app/models/comment.rb
```diff
@@ -46,14 +48,23 @@ class Comment < ActiveRecord::Base
...
+  def people_allowed_to_be_mentioned
+    return @people_allowed_to_be_mentioned unless @people_allowed_to_be_mentioned.nil?
+    contacts = author.owner.try(:contact_people) if parent.public?
+
+    @people_allowed_to_be_mentioned = [*parent.participants, *contacts, parent.author].uniq
+  end
+
+  def subscribers
+    super.tap {|subscribers|
+      subscribers.concat(legitimate_mention_recipients) if parent.author.local?
```
SuperTux88: This doesn't work for relayed comments on public posts, where the comment-author is allowed to mention his contacts. The parent author would filter this persons and don't deliver the comment to them if they are not also in the contacts of the parent-author or have already participated.

Maybe we should just allow all people to be mentioned on public comments (like on public posts)? We can decide to send no notification for the mentioned person, if he doesn't have the information, that the author is sharing with him, if that is a problem, but we also don't do that for posts.

config/locales/diaspora/en.yml
```diff
@@ -634,9 +636,11 @@ en:
       one: "%{actors} reshared your deleted post."
       other: "%{actors} reshared your deleted post."
     mentioned_deleted:
-      zero: "%{actors} mentioned you in a deleted post."
       one: "%{actors} mentioned you in a deleted post."
       other: "%{actors} mentioned you in a deleted post."
+    mentioned_in_comment_deleted:
+      one: "%{actors} mentioned you in a comment to a deleted post."
```
SuperTux88: Maybe only the comment is deleted and not the whole post.

lib/diaspora/shareable.rb
```diff
@@ -57,7 +65,7 @@ def owned_or_visible_by_user(user)
         with_visibility.where(
           visible_by_user(user).or(arel_table[:public].eq(true)
                                      .or(arel_table[:author_id].eq(user.person_id)))
-        ).select("DISTINCT #{table_name}.*")
+        ).distinct_wrapper
```
SuperTux88: why? If `.distinct` doesn't work for postgres (I haven't tested that), just leave the `.select("DISTINCT #{table_name}.*")`, because the `.distinct` doesn't make something different for mysql.

cmrd-senya: `.count` called on the relation doesn't work if `.select("DISTINCT #{table_name}.*")` used with mysql. `.distinct` fixes the thing for mysql, but breaks for postgres. You can reproduce this failure by reverting the change and running tests for `spec/models/status_message_spec.rb`.

SuperTux88: So, this is only for tests? I don't like changing the code (and maybe making it less readable) only for tests.

Also, if you change your test to
```ruby
expect(StatusMessage.where_person_is_mentioned(@bob).ids).to match_array([post1.id, post2.id])
```
it works with both databases without changing anything on the code. Also this would actually test if you see the correct posts, and not two other (maybe wrong) posts.

cmrd-senya: TBH I haven't checked whether this is needed for anything in the production code. You are right, it doesn't worth changing anything for tests only. I'll verify that then.

spec/spec_helper.rb
```diff
@@ -128,6 +128,11 @@ def client_assertion_with_nonexistent_client_id_path
     ActionMailer::Base.deliveries.clear
   end
 
+  # Clear Warden environment
+  config.after(:each) do
```
SuperTux88: Do we need a `config.after(:each)` for each one?

Also, if this fixes random rspec fails, then you should create its own small PR for it, which then probably could be merged faster than this big one.

Personally I think that also the `pry-rescue`-change should have its own PR, don't add unrelated stuff to a PR that is already big enough. If you need the changes here, you can rebase this PR on top of the other PRs.

app/models/notifications/also_commented.rb
```diff
@@ -18,6 +18,7 @@ def self.notify(comment, _recipient_user_ids)
       recipient_ids = commentable.participants.local.where.not(id: [commentable.author_id, actor.id]).pluck(:owner_id)
 
       User.where(id: recipient_ids).find_each do |recipient|
+        next if Notifications::MentionedInComment.where(target: comment.mentions.where(person: recipient.person)).any?
```
SuperTux88: `.exists?` instead of `.any?`

SuperTux88:
Also, please cleanup your commits, squash some together (we don't need 6 "polishing" commits and fixups) and write meaningful messages for the rest.

lib/diaspora/mentions_container.rb
```diff
@@ -0,0 +1,41 @@
...
+    def legitimate_mention_recipients
+      visible_for_recepient_mentions.map(&:person)
+    end
+
+    def visible_for_recepient_mentions
+      ppl = people_allowed_to_be_mentioned
+      if ppl == "all"
+        mentions
+      else
+        mentions.where(person: ppl)
```
SuperTux88: I'm not sure about this, this generates a `WHERE IN (...)` query with potentially many people (when someone with many contacts shares a post with "all aspects").

cmrd-senya:
> (we don't need 6 "polishing" commits and fixups) and write meaningful messages for the rest.

There was no need pointing this, it's not my first pull request. It's strange to suggest I want it to merged in that state ;)

SuperTux88:
> t's strange to suggest I want it to merged in that state ;)

I was confused because of this: "Ok, this feels completed to me.", sorry about that.

cmrd-senya:
> I was confused because of this: "Ok, this feels completed to me.", sorry about that.

No worries. Well, I didn't squash because there will be some new fixes after review anyway %) (you already discovered some issues)  But probably I had too. Next time I'll be armed with --autosquash! :D

allschild:
Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: SenyaSent: Sunday 7 August 2016 01:06To: diaspora/diasporaReply To: diaspora/diasporaSubject: Re: [diaspora/diaspora] Mentions in comments backend changes (#6818)
(we don't need 6 "polishing" commits and fixups) and write meaningful messages for the rest.


There was no need pointing this, it's not my first pull request. It's strange to suggest I want it to merged in that state ;)

—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/diaspora/diaspora","title":"diaspora/diaspora","subtitle":"GitHub repository","main_image_url":"https://assets-cdn.github.com/images/modules/aws/aws-bg.jpg","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/diaspora/diaspora"}},"updates":{"snippets":[{"icon":"PERSON","message":"@cmrd-senya in #6818: \u003e (we don't need 6 \"polishing\" commits and fixups) and write meaningful messages for the rest.\r\n\r\nThere was no need pointing this, it's not my first pull request. It's strange to suggest I want it to merged in that state ;)"}],"action":{"name":"View Pull Request","url":"https://github.com/diaspora/diaspora/pull/6818#issuecomment-238044823"}}}

SuperTux88:
Just found something more: If someone disables "you are mentioned in a post/comment" in the email-settings, but has enabled the "someone comments on ...", then he will not receive an email for this comment (because no commented-notification is created, and the mentioned notification is not sent).

I don't know, who should have this strange combination of settings, and I also don't know how to solve this the best, or if we just should ignore it. But I wanted to mention that here.

db/migrate/20160621151100_polymorphic_mentions.rb
```diff
@@ -0,0 +1,29 @@
...
+              %i(person_id mentions_container_id mentions_container_type),
+              name:   "index_mentions_on_person_and_mc_id_and_mc_type",
+              length: {mentions_container_type: 191}
+
+    reversible(&method(:up_down))
+  end
+
+  def up_down(change)
+    change.up do
+      Mention.update_all(mentions_container_type: "Post")
```
SuperTux88: Can you add [migration models](http://blog.makandra.com/2010/03/how-to-use-models-in-your-migrations-without-killing-kittens/) for the migrations?

app/models/comment.rb
cmrd-senya: I don't see any other options either: we don't have means to restrict mentioned recipients if we want to include anyone except participants.

cmrd-senya:
> I don't know, who should have this strange combination of settings, and I also don't know how to solve this the best, or if we just should ignore it. But I wanted to mention that here.

I thought of a following possible solution: overload `email_the_user` for `MentionInComment` with a version that does something like

```ruby
class Notifications::MentionedInComment < Notification
  def email_the_user(target, actor)
    unless recipient.user_preferences.exists?(email_type: "mention"))
      recipient.mail(Workers::Mail::Mentioned, recipient_id, actor.id, target.id)
    elsif recipient.is_owner_of?(target)
      recipient.mail(Workers::Mail::CommentOnPost, recipient_id, actor.id, target.id)
    elsif recipient.participates_in?(target)
      recipient.mail(Workers::Mail::AlsoCommented, recipient_id, actor.id, target.id)
    end
  end
end
```

It's a bit hacky, since it repeats the logic of NotificationService and the other comment notifications. But it's the easiest thing I invented that doesn't require any refactoring of the notification code.

I also think it worth fixing, because I may imaging someone turns off "mention in comments" email notification, because he doesn't want to be spammed by people who want to include him into discussions. And after that someone mentions the person in a comment within a discussion he participated, and the person doesn't receive the email notification (but he expects it, because he is participating and doesn't care on whether someone mentions him or not). In this case the software will implicitly make other things than expected, which is not good.

Opinions?

cmrd-senya:
To simplify the review I've outlined the changes by this PR in the top message of the thread.

app/models/comment.rb
cmrd-senya: The changes allowing everyone to be mentioned in public comments have been applied.

app/helpers/notifications_helper.rb
```diff
@@ -4,40 +4,46 @@ module NotificationsHelper
...
   end
 
   def translation(target_type, opts = {})
-    {:post_author => nil}.merge!(opts)
     t("#{target_type}", opts).html_safe
   end
 
+  def opts_for_post(post)
+    {
+      post_author: h(post.author_name),
```
SuperTux88: Can you use `html_escape()` here? I needed my IDE to find what `h()` does.

app/helpers/notifications_helper.rb
```diff
@@ -4,40 +4,46 @@ module NotificationsHelper
...
-                    actors:    actors,
-                    count:     actors_count,
-                    post_link: link_to(post_page_title(post), post_path(post)).html_safe)
-      else
-        t(note.deleted_translation_key, :actors => actors, :count => actors_count).html_safe
+    if note.respond_to?(:linked_object)
+      if note.linked_object.nil? && note.respond_to?(:deleted_translation_key)
+        target_type = note.deleted_translation_key
+      elsif note.is_a?(Notifications::Mentioned)
+        opts.merge!(opts_for_mentioned(note))
```
SuperTux88: Maybe use `note.linked_object` here like for `opts_for_post`?

app/views/notifier/mentioned.markerb
```diff
@@ -4,6 +4,12 @@
 <%= t('notifier.mentioned.limited_post') %>
 <% end %>
 
-[<%= t('notifier.comment_on_post.reply', :name => @notification.post_author_name) %>][1]
+<% if @notification.post.is_a?(Post) %>
```
SuperTux88: I think we should use different templates here. I don't like this if/else here, and now if you are mentioned in a private comment you get a mail with "You were mentioned in a limited post.".

app/models/mention.rb
```diff
@@ -3,11 +3,16 @@
 #   the COPYRIGHT file.
 
 class Mention < ActiveRecord::Base
-  belongs_to :post
+  belongs_to :mentions_container, polymorphic: :true
+  alias_attribute :post, :mentions_container
```
SuperTux88: Do we need this alias here?

app/models/notifications/also_commented.rb
```diff
@@ -18,7 +18,11 @@ def self.notify(comment, _recipient_user_ids)
       recipient_ids = commentable.participants.local.where.not(id: [commentable.author_id, actor.id]).pluck(:owner_id)
 
       User.where(id: recipient_ids).find_each do |recipient|
-        next if recipient.is_shareable_hidden?(commentable)
+        if recipient.is_shareable_hidden?(commentable) ||
+          Notifications::MentionedInComment.where(target: comment.mentions.where(person: recipient.person)).exists?
```
SuperTux88: You can extract `Notifications::MentionedInComment.where(target: comment.mentions.where(person: recipient.person)).exists?` to a method `mention_notification_exists?(comment, recipient)`. That way you can make this a one liner again and I think that would increase readability here.

app/models/status_message.rb
```diff
@@ -120,6 +89,29 @@ def post_location
...
+    @aspects_ppl ||=
+      if public?
+        "all"
+      else
+        [
+          *AspectMembership.where(aspect_id: aspects).map(&:person),
```
SuperTux88: This line creates a DB-query for each person in this aspects, but this can be done with an all-in-one query:
```ruby
Contact.joins(:aspect_memberships).where(aspect_memberships: {aspect: aspects}).distinct.pluck(:person_id)
```

app/models/status_message.rb
```diff
@@ -120,6 +89,29 @@ def post_location
...
+          *share_visibilities.map {|sv| sv.user.person }
```
SuperTux88: This line is not needed, because `people_allowed_to_be_mentioned` is only used with the author of the post, and the people from the aspects (see above) are all you need here.

SuperTux88: Oh, I missed the use in the `Notifications::Mentioned` ...

But I think it is better to create its own filter there, and only test if the person can see the post there (and for comments if the person can see the parent post)?

app/services/post_service.rb
```diff
@@ -59,7 +59,11 @@ def mark_comment_reshare_like_notifications_read(post_id)
   end
 
   def mark_mention_notifications_read(post_id)
```
SuperTux88: I think we should also mark comment-mention-notifications as read here.

spec/integration/mentioning_spec.rb
```diff
@@ -22,74 +35,275 @@ def mention_stream_for(user)
...
   before do
-    @user1 = FactoryGirl.create :user_with_aspect
-    @user2 = FactoryGirl.create :user
-    @user3 = FactoryGirl.create :user
+    @user2 = bob
+    @user3 = eve
 
-    @user1.share_with(@user2.person, default_aspect)
-    sign_in @user1
+    expect(users_connected?(alice, bob)).to be true
```
SuperTux88: I think you should simply create what you need here (like before) instead of checking if nobody changes alice, bob and eve.

spec/integration/mentioning_spec.rb
```diff
@@ -22,74 +35,275 @@ def mention_stream_for(user)
...
+    expect(@user2).to be_mentioned_in(status_msg)
+    expect(status_msg).to be_in_streams_of(@user2)
+  end
+
+  context "in comments" do
+    context "with public post" do
+      let(:status_msg) { FactoryGirl.create(:status_message, author: alice.person, public: true) }
+
+      context "a commenter may mention a person" do
+        [[bob, eve], [eve, bob], [alice, bob], [bob, alice], [local_luke, local_leia], [alice, eve]]
```
SuperTux88: Do we really need all of these here?

cmrd-senya: Yes, these all are different type of relationships between users. I'll
check once again, but we need all of them because receive works
different depending on the relationship type. So it's easy to screw it up.

spec/integration/mentioning_spec.rb
cmrd-senya: What do you think of removing these checks at all and relying on the
fixture builder generated data?

SuperTux88: No, don't use alice, bob and eve from fixture builder and create user1, 2 and 3 like before and connect them like you need it.

spec/integration/mentioning_spec.rb
SuperTux88: Then you should probably create tests for the different combinations, so you can see which combination broke if one of these fails.

cmrd-senya: What do you mean by "create tests for the different combinations"? I can
make context names express it more explicit, but the test body is the
same every time. Is it fine?

SuperTux88: Yes, same test, but write what "bob+eve" makes them special compared to "bob+alice".

spec/services/notification_service_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        )
+      }
+
+      [alice, eve].each do |user|
+        context "with recipient #{user.username}" do
+          it "sends only mention notification" do
+            expect(Workers::Mail::Mentioned).to receive(:perform_async).with(
+              alice.id,
+              bob.person.id,
+              *comment.mentions.where(person: alice.person).ids
```
SuperTux88: You can use `user` here to, so you don't test twice if alice receives a mail.

spec/services/notification_service_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+              *comment.mentions.where(person: eve.person).ids
+            )
+
+            expect {
+              NotificationService.new.notify(comment, [user.id])
+            }.to change { Notification.where(recipient_id: user).count }.by(1)
+
+            expect(
+              Notifications::MentionedInComment.where(target: comment.mentions, recipient_id: user.id)
+            ).not_to be_empty
```
SuperTux88: I think `exist` is better than `be_empty` here.

SuperTux88:
> I thought of a following possible solution: overload `email_the_user` for `MentionInComment`

Hmm, I don't know. It fixes the problem and it is relatively simple. And I don't have a better idea, but we can always change that if someone has a better solution.

app/models/notifications/also_commented.rb
cmrd-senya: `mention_notification_exists?` is going to be used both in this and `CommentOnPost` classes. Is it fine to place this method to the `Notification` model? I'd place it to `Notifications::MentionedInComment` as class-method, however it wouldn't be one-liner then (I checked, it doesn't fit).

SuperTux88: Maybe you can also create a `Commented` module like the `Mentioned` with common logic, because `deleted_translation_key` is also the same for both. Otherwise I have no problem if it isn't a one-liner, I still think the method-name `mention_notification_exists?` is easier to read/understand than the AR-query and that was my main goal here. But if it is only used for commented-notifications it doesn't belong to `Notification`.

app/mailers/notification_mailers/mentioned.rb
```diff
@@ -4,9 +4,10 @@ class Mentioned < NotificationMailers::Base
     delegate :author_name, to: :post, prefix: true
 
     def set_headers(target_id)
-      @post = Mention.find_by_id(target_id).post
+      @post = Mention.find_by_id(target_id).mentions_container
 
       @headers[:subject] = I18n.t('notifier.mentioned.subject', :name => @sender.name)
+      @headers[:template_name] = "mentioned_in_comment" if post.is_a?(Comment)
```
SuperTux88: I see what you are trying to do here ... and I'm unsure about it.

On one side you try to minimize code duplication, on the other side you need a workaround to make this work. Also you have `@post` here, which leads to stuff like `.post.parent` and `.post.guid` in the `mentioned_in_comment.markerb`.

If you create a separate class for `MentionedInComment` you would have a small code duplication, but you wouldn't need the workaround and you can name the variables properly (and you don't even need the `post_author_name` in the `mentioned_in_comment` template, so less duplication) and you don't need a `if` here.

And creating a separate class would also mean that you have separate checkboxes in the settings to enable mails for mentions on posts and comments.

At the end I'm more on the "split this completely"-side, I think that's the cleaner solution. What do you think?

cmrd-senya: I'm fine with both approaches. If you like "split" approach more, I'm fine with implementing it.

However, speaking about deduplication, I see that the workers for Notifications::*  in `app/workers/mail` are almost identical. So are respective methods in `Notifier`. I can refactor them to use some common code instead of deepen the duplication by introducing `Workers::Mail::MentionedInComment`. What do you think?

SuperTux88: Creating a `Workers::Mail::Base` sounds like a good idea.

cmrd-senya: However I'm not sure about `Workers::Mail::Base` as a good name. First it'll make confusion with `Workers::Base`. Second not every `Workers::Mail::*` will be child of this `Workers::Mail::Base`. There are `ReportWorker` or `InviteEmail` who have different logic. All `Workers::Mail::*` that we can commonify use `Notifier` class. So we can call the base class for it `Workers::Mail::NotifierBase` or something. ?

SuperTux88: Yes, `Workers::Mail::NotifierBase` is ok.

app/models/notifications/mentioned_in_post.rb
```diff
@@ -0,0 +1,28 @@
...
+      "notifications.mentioned_deleted"
+    end
+
+    def self.filter_mentions(mentions, mentionable, recipient_user_ids)
+      return mentions if mentionable.public
+      mentions.where(person: Person.where(owner_id: recipient_user_ids).ids)
+    end
+
+    def self.people_allowed_to_be_notified(mentions_container)
+      super.tap do |ppl|
```
SuperTux88: hmm, after thinking about this again: you don't even need this here, because `filter_mentions` does already the correct thing for mentions on posts. And for mentions on comments it is probably better to filter allowed mentions in `filter_mentions` too, so we don't have to filter twice.

app/models/comment.rb
```diff
@@ -49,14 +51,24 @@ class Comment < ActiveRecord::Base
...
-  def message
-    @message ||= Diaspora::MessageRenderer.new text
-  end
-
   def text= text
      self[:text] = text.to_s.strip #to_s if for nil, for whatever reason
   end
 
+  def people_allowed_to_be_mentioned
+    @people_allowed_to_be_mentioned ||= if parent.public?
```
SuperTux88: If I see it correctly, you only call this method once, so no need to cache the result in `@people_allowed_to_be_mentioned`?

cmrd-senya: It is called twice on each comment generation, but in different workers, so caching doesn't help.

app/models/notifications/mentioned_in_comment.rb
```diff
@@ -10,8 +10,15 @@ def deleted_translation_key
...
     end
 
-    def self.filter_mentions(mentions, _mentionable, _recipient_user_ids)
-      mentions
+    def self.filter_mentions(mentions, mentionable, _recipient_user_ids)
+      ppl = mentionable.people_allowed_to_be_mentioned
+      if ppl == "all"
+        mentions
+      else
+        mentions.to_ary.delete_if {|mention|
```
SuperTux88: You can write this as `mentions.where(person_id: ppl)`.

app/models/comment.rb
```diff
@@ -56,11 +56,11 @@ def text= text
...
   def people_allowed_to_be_mentioned
-    @people_allowed_to_be_mentioned ||= if parent.public?
-                                          "all"
-                                        else
-                                          [*parent.participants, parent.author].map(&:id).uniq
-                                        end
+    if parent.public?
+      "all"
+    else
+      [*parent.participants, parent.author].map(&:id).uniq
```
SuperTux88: you can do this with `[*parent.participations.pluck(:author_id), parent.author_id].uniq`, without joining the people table.

app/models/notifications/mentioned_in_comment.rb
cmrd-senya: Right, but won't that be an inefficient `WHERE people.id IN (....)` expression? I remember before I was adviced to avoid this idiom in case when there are many ids in the list. Often there will be few mentions and a number of people ids (10, 100, even more).

SuperTux88: Yes, I was also unsure about this, but I talked with @CSammy about that, and he meant that the DB should manage it, because there are other `WHERE` conditions to minimize the result set. 

I did a few tests locally now, and yes, it gets slower if you add more IDs to the `IN`, but 10000 are still ~10ms, so it should be "ok" (not really slow). Also I don't expect it to be more than 100, because I only know public posts with really many participants, private posts are mostly less active (at least the private posts I know about).

But I'm still unsure ... I thought the `where` would be more readable, but if you prefer the `delete_if` I'm also fine with it. But maybe `keep_if` is more intuitive then?

cmrd-senya: Well, the default scope is "All aspects" which means all contacts, but if it is fine with 10000, then I think it can hardly cause any issues.

I'd vote for readability and `where` clause then. If it cause any performance problems in future, it can be solved as a separate issue. I guess it comforms the Ruby/Rails way "features first, performance second".

SuperTux88: > "features first, performance second"

That shouldn't be an excuse to ignore performance ;)

But I think in this case the `where` is ok for the beginning, it shouldn't cause any issues, and if it does, we can change this in the future. This is also only in background processing (sidekiq), so no slow frontend actions because of this. But with 10000 it isn't fast anymore, compared to 0.x ms with 10, but I think we can live with it, it doesn't kill the DB.

> Well, the default scope is "All aspects"

That still doesn't mean that all contacts are participating.

app/models/comment.rb
SuperTux88: While thinking about the filter in the notifications, I remembered, that this doesn't work as expected. With the new federation, `participations` only contain participations of all participants, if the owner of the post is local, otherwise it only contains the local participations.

While this is perfect for the notification filter (means less person_ids in the `IN` query, because it only contains local IDs), this doesn't work for the send filter. Do we really want all participants to be mentionable (comment, like or poll_participation?), or only people who commented?

I suggest the latter, so it would be: `[*parent.comments.pluck(:author_id), parent.author_id].uniq`

cmrd-senya: Hmm, participations and poll_participations are generally non-visible. Comments and likes are visible. So there are few sense to include the former: strictly saying participation means "following" the discussion rather than participating in it and nobody sees your participation. On contrary comments are visible participation in the discussion. Likes are similar to comments in this respect. E.g. a user notices someone has liked the post and wants to say `@dude why do you like it?`

So, I guess, if changing that, I'd prefer 
`[*parent.comments.pluck(:author_id), *parent.likes.pluck(:author_id), parent.author_id].uniq`

Right?

SuperTux88: sounds good. 👍

cmrd-senya:
Ok, looks like I made all changes that was requested by @SuperTux88 so far. I'd love to test all of the changes for some more time, but I'm unsure I'll have the possibility to do so in the next days. It _should_ be fine, but I can't claim that I've conducted comprehensive examination.

cmrd-senya:
Okay, I've done some manual testing. I found a couple of minor issues and fixed them. If you think of some edge cases that I could have missed, please tell me, I can test them. Otherwise, everything works as expected for me and I think can be merged (after squash).

app/models/comment.rb
SuperTux88: When the (mentioned) post author is local and the (mentioning) comment author is remote, it sends two notification mails. That is because you add the local mentioned people (which already received it) again to the subscribers for relaying, so they receive it again.

SuperTux88 (changes requested):
I found some things, but it looks good so far.

But I have not done much manual testing yet.

app/helpers/notifications_helper.rb
```diff
@@ -4,40 +4,44 @@ module NotificationsHelper
 
   def object_link(note, actors)
     target_type = note.popup_translation_key
-    actors_count = note.actors.size
+    opts = {actors: actors, count: note.actors.size}
```
svbergerem: I guess @SuperTux88 was talking about the value instead of the key.

cmrd-senya: That is what I wanted to clarify.

SuperTux88: Oh yeah, changing the locales is not what I intended. Your current change is OK, thanks 👍

spec/integration/mentioning_spec.rb
```diff
@@ -19,77 +28,337 @@ def mention_stream_for(user)
...
+  context "in private post via federation with multiple recipients" do
+    shared_context "in the first possible recipient order" do
+      let(:recipients) { [user3, user2] }
+    end
+
+    shared_context "in the second possible recipient order" do
+      let(:recipients) { [user2, user3] }
+    end
+
+    ["in the first possible recipient order", "in the second possible recipient order"].each do |in_context|
```
SuperTux88: I don't know if we really need to test different orders here, but why not ... But I think we should probably use more meaningful names like "with user2 receiving before user3" or something like this.

cmrd-senya:
It was my original intention to test different orders with this
examples, but I lost it somewhere on the way. When you commented on this
part, I recalled it. Why I wanted this is because we receive a
federation entity one by one and Notifications::Mentioned.notify is
called with different recipient_user_ids. It is important to test that
we don't fail if Notifications::Mentioned.notify was already called
before with a different user id. For instance, a possible wrong flow is
to receive entity for user2, filter out mention for user3 because there
is no user3 among recipient_user_ids and then fail to notify user3 when
we receive the same entity for the second time. That doesn't happen now,
but who knows, how it can be refactored in future.

spec/integration/mentioning_spec.rb
cmrd-senya: It was my original intention to test different orders with this examples, but I lost it somewhere on the way. When you commented on this part, I recalled it. Why I wanted this is because we receive a federation entity one by one and `Notifications::Mentioned.notify` is called with different recipient_user_ids. It is important to test that we don't fail if `Notifications::Mentioned.notify` was already called before with a different user id. For instance, a possible wrong flow is to receive entity for user2, filter out mention for user3 because there is no user3 among recipient_user_ids and then fail to notify user3 when we receive the same entity for the second time. That doesn't happen now, but who knows, how it can be refactored in future.

(sorry for double posting, github sent my message to the general PR discussion instead of this subthread)

SuperTux88: But does the order matter then?

Maybe you can also add some more unit-tests for `Notifications::Mentioned.notify`?

cmrd-senya: > But does the order matter then?

On the second thought I think we don't really need this ordering test. I'll change it back to the fixed order.

> Maybe you can also add some more unit-tests for Notifications::Mentioned.notify?

Sure, why not.

cmrd-senya:
I've addressed all the notes of @SuperTux88 so far.

cmrd-senya:
Ok, I made a self-review and fixed some issues. Please move the PR to "Needs review". Hopefully it'll be a final one. After that I'll squash it. Dunno if we need any more deep testing. Nobody replied on my call for testers at diaspora, however @dadall and @AugierLe42e discussed a possibility to merge it on diaspote. If we need some more testing before merge it is an opportunity.

AugierLe42e:
Yep, I'll merge this during this week-end. Is it already usable? Is the front-end present here?

Flaburgan:
No there is no front-end at the moment. I'm also open to merge this on d-fr if needed, though I agree with @SuperTux88 that it doesn't bring anything more than the local testing as it brings nothing new in the UI>

AugierLe42e:
I think we can write a quick UI then. The code is already available. We can reuse the mention system used in the publisher. I won't be a big deal.

cmrd-senya:
> I think we can write a quick UI then.

It's not necessary. We can test mentions writing mention markdown clause in comments directly: `@{name; id@pod.tld}`.  If we have a couple of pods running this PR, we can test any imaginable usecase live. Then, when this is merged, we can make the UI without rush.

AugierLe42e:
I'm not sure our users will use it with the `@{name;handle}` syntax. I will merge this in diaspote and add the UI. Does this PR have a path to query the users that can be mentionned? Like [there is for the publisher](https://github.com/diaspora/diaspora/blob/develop/app/controllers/contacts_controller.rb#L18)?

cmrd-senya:
> Does this PR have a path to query the users that can be mentionned?

No, there isn't since I haven't ever started the UI part.

> I'm not sure our users will use it with the @{name;handle} syntax.

The syntax is `@{name; handle}`, the space after `;` is important (that is a hardcoded regexp we have there). And I don't think it is necessary that every user of your pod will test this feature. It is fine to test it among a little group of people, even only ourselves. Again, testing it on "usual users" massively is not how I think we should do. It's better if manual testing is conducted by a person who have some technical skills and understanding of technologies. Most probably, that if a user fails to use mention markdown syntax correctly, then he would also fail to give a decent feedback. Then, if we rush with UI, it could make things more complicated, because you'll have to introduce even more changes, some of them to controllers. If there are any issues with the quickly made UI, it'll get harder to find what is broken.

It's enough even if one person (except of me) conducts testing if it is done well. If it is fine to test changes on the end-users right away, then why don't we just merge the PR to the develop branch right away without any more manual tests?

SuperTux88:
> It's enough even if one person (except of me) conducts testing if it is done well.

Yes. I don't think that we have much benefit if we deploy this to a prod-pod and ask the users to test it. It is better when one person does a careful local testing with different users and as many edge cases the person can think of. This person can then also check the logs and check the expected behavior on every involved user. If you let multiple users test this in production, then they probably test it with their best friends (and always have a mutual connection), and that doesn't help much. But I can't stop somebody if he wants to test it in production, but I think local testing is easier and better.

AugierLe42e:
>  And I don't think it is necessary that every user of your pod will test this feature. It is fine to test it among a little group of people, even only ourselves. 

The users of our pod is already a small group. We are talking about less than 50 persons here ;)

> Then, if we rush with UI, it could make things more complicated, because you'll have to introduce even more changes, some of them to controllers.

Really, the UI is a one-liner. [mention_view.js](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/publisher/mention_view.js) can be used as is, as long as there's a path to query mentionnable people. See, I'm not kidding, [it's really a one-liner](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/publisher_view.js#L99):

```javascript
this.mention = new app.views.PublisherMention({ el: this.$("#publisher_textarea_wrapper") });
```

That's it.

app/models/comment.rb
```diff
@@ -49,14 +50,26 @@ class Comment < ActiveRecord::Base
...
+      :all
+    else
+      [*parent.comments.pluck(:author_id), *parent.likes.pluck(:author_id), parent.author_id].uniq
+    end
+  end
+
+  def subscribers
+    super.tap {|subscribers|
+      if parent.public? && parent.author.local?
+        subscribers.concat(legitimate_mention_recipients.select(&:remote?))
```
SuperTux88: This is now somehow redundant. You check `public?` here and also check public in `legitimate_mention_recipients` (which calls people_allowed_to_be_mentioned). Also we never need to add subscribers for private posts (status messages and comments)

I would rename `legitimate_mention_recipients` to `add_mention_recipients`, then you can do something like this:
```ruby
    def add_mention_recipients(subscribers)
      subscribers.concat(mentions.map(&:person).select(&:remote?)) if public?
    end
```

And you can use this for comments and status messages.

Just some quick thoughts, maybe you have a better idea how to simplify this? :)

spec/models/notifications/mentioned_spec.rb
```diff
@@ -1,72 +1,58 @@
...
+  describe ".notify" do
+    let(:status_message) {
+      FactoryGirl.create(:status_message, text: text_mentioning(remote_raphael, alice, bob, eve), author: eve.person)
+    }
+
+    it "calls filter_mentions on self" do
+      expect(TestNotification).to receive(:filter_mentions).with(
+        Mention.where(mentions_container: status_message, person: [alice, bob].map(&:person)),
+        status_message,
+        :passthrough_argument
```
SuperTux88: This confused me, what about `[alice.id, bob.id]`?

spec/models/notifications/mentioned_spec.rb
```diff
@@ -1,72 +1,58 @@
...
+      )
 
-      it "does not call create_notification if the mentioned person is not a recipient of the post" do
-        expect(Notifications::Mentioned).not_to receive(:create_notification)
+      TestNotification.notify(status_message, nil)
+    end
 
-        Notifications::Mentioned.notify(private_sm, [alice.id])
-      end
+    it "doesnt't create notification if it was filtered out by filter_mentions" do
```
SuperTux88: "doesn't ..."

spec/models/notifications/mentioned_spec.rb
```diff
@@ -1,72 +1,58 @@
...
+    def self.filter_mentions(mentions, *)
+      mentions
     end
 
-    it "sends an email to the mentioned person" do
-      allow(Notifications::Mentioned).to receive(:create_notification).and_return(mentioned_notification)
-      expect(bob).to receive(:mail).with(Workers::Mail::Mentioned, bob.id, sm.author.id, sm.mentions.first.id)
-
-      Notifications::Mentioned.notify(sm, [])
+    def email_the_user(*)
```
SuperTux88: You can remove this.

SuperTux88 (changes requested):
Temporary review (but I wanted to give some quick feedback, so you can already fix some things, so we can merge this faster :) ).

I'll do a more careful review again on the weekend and also some testing.

app/models/comment.rb
cmrd-senya: see https://github.com/diaspora/diaspora/pull/6818/commits/39cd3caf5f33fc1f7b98f73171675bc465afc19a

app/models/status_message.rb
```diff
@@ -111,10 +111,8 @@ def filter_mentions
     update(text: Diaspora::Mentionable.filter_people(text, people_allowed_to_be_mentioned))
   end
 
-  def subscribers
-    super.tap {|subscribers|
-      subscribers.concat(legitimate_mention_recipients)
-    }
+  def add_mention_subscribers?
+    super && author.local?
```
SuperTux88: Do we really need the `author.local?` here? When sending a StatusMessage the author is always local, and the subscribers of the StatusMessage are only added to the relayable-subscribers if the author is local.

features/desktop/notifications.feature
```diff
@@ -101,6 +101,26 @@ Feature: Notifications
...
+    Given a user with email "bob@bob.bob" is connected with "alice@alice.alice"
+    And "alice@alice.alice" has a public post with text "check this out!"
+    And "bob@bob.bob" has commented mentioning "alice@alice.alice" on "check this out!"
+    When I sign in as "alice@alice.alice"
+    And I follow "Notifications" in the header
+    Then the notification dropdown should be visible
+    Then I should see "mentioned you in the comment"
+    And I should have 1 email delivery
+
+  Scenario: I mark a mention in comment notification as read
```
SuperTux88: > Why name it generic "mark a notification as read" if it tests mark as read for "mention in comment" notification?

Because it doesn't matter. The "mark as read/unread" feature doesn't depend on the notification type. So it is OK to have a scenario that tests it in the frontend, but I don't want cucumber tests for every notification type to test if you can mark it as read. (that is something you can do in unit-test, which are faster).

And if it doesn't depend on the notification type, I think we should use a generic description. And because you don't need "mention in comment" you can also test it with a normal "mention" and use the existing "And Alice has a post mentioning Bob" step for it.

AugierLe42e:
Keep up the good work @cmrd-senya!

img-img-1

features/desktop/notifications.feature
cmrd-senya: > Because it doesn't matter. The "mark as read/unread" feature doesn't depend on the notification type. So it is OK to have a scenario that tests it in the frontend, but I don't want cucumber tests for every notification type to test if you can mark it as read. (that is something you can do in unit-test, which are faster).

I disagree. IMO, ideally, we should have an integration tests covering each client behaviour scenario. We can't rely on unit tests here. Unit tests are not for prooving that our application works as expected in every usage scenario. Ideally, we should have an integration test for every possible situation that may happen from the user point of view. The execution flow for "mark as read" for mentions in comments and posts is the same, but that is an implementation detail. We can't use that knowledge to claim that 
> "mark as read/unread" feature doesn't depend on the notification type.

In the end, I'll implement what you say, but I don't feel it right to have only one test scenario for "mark as read" for mentioning.

SuperTux88 (changes requested):
Looks really good so far, I think we can merge this soon :)

I also did some local testing now with different scenarios, and everything works as expected 👍 

Can you please fix these last remarks and then squash your commits? :)

SuperTux88:
I have just merged #7223 and that causes some conflicts with this PR (sorry about that, but they should be easy to resolve :) ). When rebasing, can you please also remove `require "spec_helper"` from any new specs you added (where you don't have a conflict).

features/desktop/notifications.feature
svbergerem: I agree with @SuperTux88 here. We use cukes to test the combination of front- and backend. And when doing that at least I also have the implementation in mind to know, which cases I have to go through.

You could use different “mark read/unread” cukes for different notification types. You could also use different “mark read/unread” cukes for contact / non-contact actors. Or users on the same pod vs. users on different pods. Or for posts with photos and posts without photos. Or... (I hope you get my point ;))

In this case there is exactly one thing we have to test: Do we send the correct request to the correct route when clicking the button. And the requests all look the same for the different notification types. And the buttons also look the same.

I still remember the times when our cucumber tests exceeded the timeout of 50 minutes. Now they need about 15 minutes but we have a few problems with randomly failing tests. Thus I'd still like to keep the number of cukes rather small. Let's keep using cukes to test the different combinations of front- and backend. And let's keep using rspec/jasmine tests for everything else. In this case I don't think that we need more than one cuke for the “mark read/unread” feature.

cmrd-senya:
Remarks corrected, squashed, rebased, signed. Tests are passing, except of random cukes.

SuperTux88 (approved):

SuperTux88:
Merged to develop branch, thanks @cmrd-senya :)

cmrd-senya:
Great! Thanks for your work either. I know it required many much efforts to review it multiple times. I'll do my best to produce a better code in my next contributions :)

We also need UI to get #1851 fixed. I'm going to work on it soon. I don't really think it is a one-liner, as @AugierLe42e says, when I looked at it last time I didn't quite get it. But that's on my list anyway.

goobertron:
Whoo whoo whoo! Thanks so much, @cmrd-senya, and thanks to all those who helped and reviewed.

Flaburgan:
A massive thank you for this very needed work, this was definitely a missing key feature on diaspora.

AugierLe42e:
Same as @Flaburgan : huge thanks to @cmrd-senya and @SuperTux88.

> when I looked at it last time I didn't quite get it. But that's on my list anyway.

As explained, as soon as there's a route that answers mentionnable people for a particular post, it's easy. Maybe not a one-liner, but 3 or 4 lines at most. As I refactored the mention system, I can help you or even write it myself. Just ping me on IRC if you need it.

Issue #3653, opened 2012-10-11
Photo upload broken


balasankarc:
The photo upload facility is completely broken today..  The icon (rotating one) showing "uploading" rotates till infinity... It just won't stop... And still photo is not being uploaded... Please fix it...

Balu

svbergerem:
We would have to apply the solution @taratatach [mentioned here](https://github.com/diaspora/diaspora/issues/3653#issuecomment-24876887) to the [apache conf](https://gist.github.com/jhass/719014).

jaywink:
I don't see how setting a larger body size in the web server will help in the case that the connection is just slow? The server side doesn't support chunking uploads and until it does it makes little sense to bump the 4mb limit or even suggest bumping it. Some users will just get more timeouts the larger the limit is increased. Photo uploads really should fall back or be able to use smaller chunks so that timeouts don't happen, which is mentioned in #4997.

jaywink:
Any objections to closing this?

jhass:
Nope.

PR #6819, opened 2016-04-30
[WIP] Make max image upload size configurable

Flaburgan:
4194304 bytes is actually hardcoded everywhere. It's becoming too small in 2016, I wanted to make that customizable.

This is not working actually, I've got "gon is undefined", is there something special I should do to make gon available?

### DIFF
--- a/app/assets/javascripts/app/views/publisher/uploader_view.js
+++ b/app/assets/javascripts/app/views/publisher/uploader_view.js
@@ -7,7 +7,7 @@
 app.views.PublisherUploader = Backbone.View.extend({
 
   allowedExtensions: ["jpg", "jpeg", "png", "gif", "tif", "tiff"],
-  sizeLimit: 4194304,  // bytes
+  sizeLimit: gon.appConfig.settings.max_upload_size,
 
   initialize: function(opts) {
     this.publisher = opts.publisher;
--- a/app/assets/javascripts/mobile/mobile_file_uploader.js
+++ b/app/assets/javascripts/mobile/mobile_file_uploader.js
@@ -11,7 +11,7 @@ function createUploader(){
        action: "/photos",
        debug: false,
        button: document.getElementById('file-upload-publisher'),
-       sizeLimit: 4194304,
+       sizeLimit: gon.appConfig.settings.max_upload_size,
 
        onProgress: function(id, fileName, loaded, total){
         var progress = Math.round(loaded / total * 100 );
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/app/views/publisher/uploader_view.js
```diff
@@ -7,7 +7,7 @@
 app.views.PublisherUploader = Backbone.View.extend({
 
   allowedExtensions: ["jpg", "jpeg", "png", "gif", "tif", "tiff"],
-  sizeLimit: 4194304,  // bytes
+  sizeLimit: gon.appConfig.settings.max_upload_size,
```
diaspora-code-review: Identifier 'max_upload_size' is not in camel case.

app/assets/javascripts/mobile/mobile_file_uploader.js
```diff
@@ -11,7 +11,7 @@ function createUploader(){
        action: "/photos",
        debug: false,
        button: document.getElementById('file-upload-publisher'),
-       sizeLimit: 4194304,
+       sizeLimit: gon.appConfig.settings.max_upload_size,
```
diaspora-code-review: Identifier 'max_upload_size' is not in camel case.

app/controllers/application_controller.rb
```diff
@@ -150,7 +150,10 @@ def current_user_redirect_path
   def gon_set_appconfig
     gon.push(appConfig: {
                chat:     {enabled: AppConfig.chat.enabled?},
-               settings: {podname: AppConfig.settings.pod_name},
+               settings: {
+                 podname: AppConfig.settings.pod_name,
```
diaspora-code-review: Align the elements of a hash literal if they span more than one line.

app/assets/javascripts/mobile/mobile_file_uploader.js
Flaburgan: I kept the same name in JS than in the diaspora.yml file here, should I change that?

denschub:
Sorry, but no.

We've discussed this in #3653 before. We are not able to simply increase the upload limit and we should absolutely not expose a config option for that. Increasing the max image size would require the podmins to change their nginx/apache configs to increase the maximal body size. That, in turn, will lead to people increasing their max body sizes to a gigabyte or something and totally blowing up their ram.

Instead of randomly increasing the body limit and exposing configs that may break stuff if the podmin does not change their webserver configs we should implemented chunked uploading, which is tracked in #4997.

Flaburgan:
Up to you, maybe if we don't expose it we can at least use a constant instead of a hard value everywhere, and increase the default value.

Flaburgan:
So what do you think? Can we use a unique variable instead of hardcoded values, and where should I put this variable if we don't want to expose it?

PR #6821, opened 2016-05-05
Fix time ago when making a notification read/unread.

### LINKED ISSUE #6798: Timeago in notifications
Marking a notification as unread resets the timeago stamp causing the times to look wrong. 
It can be reproduced by marking an old notification as unread. 

(0.5.99 Development Branch)

bildschirmfoto 2016-04-16 um 08 38 06

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/33011822-timeago-in-notifications?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ralinc:
Marking a notification as unread resets the timeago stamp causing
the times to look wrong. It can be reproduced by marking an old
notification as unread. Using the update_colum instead of
update_attribute will not touch the updated_at attribute, and
thus will not affect the updated time ago in the view.

Fixes #6798.

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -9,7 +9,7 @@ class Notification < ActiveRecord::Base
   belongs_to :target, :polymorphic => true
 
   attr_accessor :note_html
- 
+
   def self.for(recipient, opts={})
     self.where(opts.merge!(:recipient_id => recipient.id)).order('updated_at desc')
   end
@@ -29,7 +29,7 @@ class Notification < ActiveRecord::Base
                                       actor, note_type)
     end
     return_note.email_the_user(target, actor) if return_note
-    return_note 
+    return_note
   end
 
   def as_json(opts={})
@@ -41,7 +41,7 @@ class Notification < ActiveRecord::Base
   end
 
   def set_read_state( read_state )
-    self.update_attributes( :unread => !read_state )
+    update_column(:unread, !read_state)
   end
 
   def mail_job
--- a/spec/controllers/notifications_controller_spec.rb
+++ b/spec/controllers/notifications_controller_spec.rb
@@ -31,6 +31,19 @@ describe NotificationsController, :type => :controller do
       get :update, "id" => note.id, :set_unread => "true", :format => :json
     end
 
+    it "marks a notification as unread without timestamping" do
+      note = Timecop.travel(1.hour.ago) do
+        FactoryGirl.create(:notification, recipient: alice, unread: false)
+      end
+
+      get :update, "id" => note.id, :set_unread => "true", :format => :json
+      expect(response).to be_success
+
+      updated_note = Notification.find(note.id)
+      expect(updated_note.unread).to eq(true)
+      expect(updated_note.updated_at.iso8601).to eq(note.updated_at.iso8601)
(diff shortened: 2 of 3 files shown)


app/models/notification.rb
```diff
@@ -41,7 +41,11 @@ def email_the_user(target, actor)
   end
 
   def set_read_state( read_state )
-    self.update_attributes( :unread => !read_state )
+    if self.new_record?
```
SuperTux88: This method is only called on existing notifications, so this `if` shouldn't be needed.

spec/controllers/notifications_controller_spec.rb
```diff
@@ -31,6 +31,14 @@
       get :update, "id" => note.id, :set_unread => "true", :format => :json
     end
 
+    it 'marks a notification as unread without timestamping' do
```
SuperTux88: You should use `TimeCop` here to change the time. Because this spec runs in the same second as the Notification was created, so the time would be the same anyway.

Also this spec is red in travis, because after save it to the database some milliseconds were truncated. So you need to compare only date, hours, minutes and seconds (or only the one you changed with `TimeCop`).

app/models/notification.rb
```diff
@@ -41,7 +41,11 @@ def email_the_user(target, actor)
...
+      self.update_attribute(:unread, !read_state)
```
diaspora-code-review: Redundant `self` detected.

app/models/notification.rb
```diff
@@ -41,7 +41,11 @@ def email_the_user(target, actor)
...
+    else
+      self.update_column(:unread, !read_state)
```
diaspora-code-review: Redundant `self` detected.

spec/controllers/notifications_controller_spec.rb
```diff
@@ -31,6 +31,19 @@
...
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/notification.rb
ralinchimev: The if-statement is needed because of some failing model unit tests. Or we have to change the tests somehow.

spec/controllers/notifications_controller_spec.rb
ralinchimev: Done. Thanks for the feedback.

app/models/notification.rb
SuperTux88: We shouldn't change the code only to make tests running. Please change the tests that they use saved notifications, so the tests correspond to the real conditions.

ralinchimev: Of course. I agree.

denschub:
Merged as 1773e3e35d3cf706a06088d4c36155a5ff4ab916, thank you for your contribution!

Flaburgan:
Thank you!

Issue #6823, opened 2016-05-06
Paragraph in *conversation* is wider but its container if containing big images


DeusFigendi:
Okay first of all I'm talking about Code 0.5.99.0-pabca848b running on Geraspora at the moment.

It seems when a conversation message contains an image and that image has a higher width but... 650px (or the like) it does not shrink to fit into the *stream_element message* it just overflows hidden.

Bad enough it also affects paragraphs to do the same.

[ ![Demo](https://pod.geraspora.de/uploads/images/scaled_full_f549b88cfa880f51a8c8.png) ](https://pod.geraspora.de/uploads/images/f549b88cfa880f51a8c8.png) (click image to open full size)

and this is the markdown I used:

````
1 Officia reiciendis quisquam consequatur sed nulla eos culpa asperiores. 2 Aut sit veniam adipisci. 3 Voluptatem et ut ea. 4 Nihil facere ut porro. 5 Cupiditate at aut non molestiae. 6 Magnam pariatur voluptatum sed. 7 Dolor dolores porro voluptate dignissimos. 8 Tempora est repellendus rerum. 9 Earum facere iste inventore temporibus. 10 Ad eius expedita odit. 11 Tenetur voluptatem inventore repellat. 12 Repellendus reprehenderit quo est. 13 Suscipit dolorum itaque magni omnis qui error rerum libero. 14 Quae quis aut amet. 15 Aut ab dolor aut ut alias. 16 Blanditiis id omnis animi numquam a dolor id. 17 Odio consequuntur pariatur a. 18 Aliquid ullam sit vero asperiores et cupiditate ut voluptates. 19 Aut qui ipsum a culpa quia quibusdam sit harum. 20 Dolor quidem provident sit exercitationem sunt voluptatibus aut.
````

and

````
1 Officia reiciendis quisquam consequatur sed nulla eos culpa asperiores. 2 Aut sit veniam adipisci. 3 Voluptatem et ut ea. 4 Nihil facere ut porro. 5 Cupiditate at aut non molestiae. 6 Magnam pariatur voluptatum sed. 7 Dolor dolores porro voluptate dignissimos. 8 Tempora est repellendus rerum. 9 Earum facere iste inventore temporibus. 10 Ad eius expedita odit.

![1000 x 1000](https://upload.wikimedia.org/wikipedia/commons/5/53/Lorem_Ipsum_Gill_Sans.png)

[...]

PR #6831, opened 2016-05-12
Fix message content overflow in conversations

svbergerem:
Fixes everything mentioned in the first comment of #6823. The codeblocks issue mentioned by @denschub in the comments over there looks to me like a Bootstrap upstream issue related to https://github.com/twbs/bootstrap/issues/15958. I'm pretty sure I already linked a related Bootstrap issue somewhere else but I'm currently unable to find it.

### DIFF
--- a/app/assets/stylesheets/post-content.scss
+++ b/app/assets/stylesheets/post-content.scss
@@ -1,3 +1,4 @@
+.message-content,
 .post-content {
   img { max-width: 100%; }
   .photo_attachments {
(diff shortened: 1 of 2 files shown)


denschub:
Thank you for the fix!

> I'm pretty sure I already linked a related Bootstrap issue somewhere else but I'm currently unable to find it.

Actually, I do remember that now. I wasn't thinking about that when I wrote my comment, though. Thanks for the reminder!

PR #6836, opened 2016-05-17
Set ruby version to 2.2.3

cmrd-senya:
The latest version bumps make `bundle install` fail with ruby versions less than 2.2.3.

I get the error message

```
Installing listen 3.1.3

Gem::InstallError: listen requires Ruby version >= 2.2.3, ~> 2.2
```


jhass:
`listen` is a development dependency (through `guard`). Production installs should be unaffected as you can see by Travis still using Ruby 2.2.0.

SuperTux88:
the latest version of rvm uses 2.2.4 if you only write `2.2`, so maybe you simply need to run `rvm get latest`

cmrd-senya:
I see. So it's rather my local installation issue then. I guess it could be closed then. Sorry for making noise.

Flaburgan:
If you're using `rbenv` like me, you need to add the hotfix number manually in `.ruby-version`. But as we recommend using `rvm` at the moment, we should not modify it in diaspora code.

Flaburgan:
Hm @cmrd-senya you raise a valid point here though, we say in the changelog that 2.1.x is still supported for diaspora* 0.6. So we can't bump listen this way. I just tested with ruby 2.1.5 and the error message is preatty clear: `Gem::InstallError: listen requires Ruby version >= 2.2.3, ~> 2.2.`

SuperTux88:
2.1 is supported for prod-env, but not for development.

Flaburgan:
Oh my bad I didn't read carefully enough @jhass's message.

PR #6844, opened 2016-05-27
Fix possible duplication of AspectVisibility

cmrd-senya:
No uniqueness control on AspectVisibility resulted in possible having multiple AspectVisibility objects in the DB for the same aspect and shareable which doesn't make sense. Introduce uniqueness validation and fix up tests where duplication happened.

### DIFF
--- a/db/migrate/20160531170531_remove_duplicate_aspect_visibilities.rb
+++ b/db/migrate/20160531170531_remove_duplicate_aspect_visibilities.rb
@@ -1,0 +1,11 @@
+class RemoveDuplicateAspectVisibilities < ActiveRecord::Migration
+  def up
+    where = "WHERE a1.aspect_id = a2.aspect_id AND a1.shareable_id = a2.shareable_id AND "\
+      "a1.shareable_type = a2.shareable_type AND a1.id > a2.id"
+    if AppConfig.postgres?
+      execute("DELETE FROM aspect_visibilities AS a1 USING aspect_visibilities AS a2 #{where}")
+    else
+      execute("DELETE a1 FROM aspect_visibilities a1, aspect_visibilities a2 #{where}")
+    end
+  end
+end
--- a/app/models/aspect_visibility.rb
+++ b/app/models/aspect_visibility.rb
@@ -10,4 +10,5 @@ class AspectVisibility < ActiveRecord::Base
   belongs_to :shareable, :polymorphic => true
   validates :shareable, :presence => true
 
+  validates :aspect, uniqueness: {scope: %i(shareable_id shareable_type)}
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160327103605) do
+ActiveRecord::Schema.define(version: 20160531170531) do
 
   create_table "account_deletions", force: :cascade do |t|
     t.string   "diaspora_handle", limit: 255
--- a/spec/integration/mentioning_spec.rb
+++ b/spec/integration/mentioning_spec.rb
@@ -43,9 +43,9 @@ describe 'mentioning', :type => :request do
     expect(users_connected?(@user1, @user3)).to be false
 
     status_msg = nil
-    expect do
+    expect {
       status_msg = @user1.post(:status_message, {text: text_mentioning(@user3), to: default_aspect})
-    end.to change(Post, :count).by(1)
+    }.to change(Post, :count).by(1).and change(AspectVisibility, :count).by(1)
 
     expect(status_msg).not_to be_nil
     expect(status_msg.public?).to be false
--- a/spec/models/aspect_visibility_spec.rb
+++ b/spec/models/aspect_visibility_spec.rb
@@ -1,0 +1,32 @@
+require "spec_helper"
+
+describe AspectVisibility, type: :model do
+  let(:status_message) { FactoryGirl.create(:status_message) }
+  let(:aspect) { FactoryGirl.create(:aspect) }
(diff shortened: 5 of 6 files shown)


SuperTux88:
Did you check if there are already duplicates in a production database which need a cleanup migration? Or did this only happen in specs?

cmrd-senya:
Well, I don't have an access to a proper production DB, so no, I didn't. Maybe you could check your pods?

And I think it doesn't happen with the present code base, but who knows what was there before!

SuperTux88:
I have one shareable duplicate (in 5 aspects, so 5 duplicates) in my DB. But this will be cleaned up by `20160302025129_cleanup_aspect_visibility.rb` because the shareable itself doesn't exist anymore. So I think it is fine without cleanup migration.

Flaburgan:
@SuperTux88 can you please give me the SQL query to run so I can check on d-fr database?

SuperTux88:
@Flaburgan 
```
SELECT a.id, a.aspect_id, a.shareable_id, a.shareable_type
FROM aspect_visibilities a
INNER JOIN aspect_visibilities b ON a.shareable_id = b.shareable_id AND a.shareable_type = b.shareable_type AND a.aspect_id = b.aspect_id
WHERE a.id <> b.id
```

Flaburgan:
The same query produces 1150 lines on framasphere (running last master).

SuperTux88:
Then we probably need a cleanup-migration.

cmrd-senya:
Ok!

cmrd-senya:
A clean up migration is here. I kinda tested it manually, but I would feel safer with some reproducible test, but I don't see any tests for migrations used in diaspora.

db/migrate/20160531170531_remove_duplicate_aspect_visibilities.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveDuplicateAspectVisibilities < ActiveRecord::Migration
+  def up
+    min_ids = AspectVisibility.group(:aspect_id, :shareable_id, :shareable_type).pluck(:id)
+    AspectVisibility.where.not(id: min_ids).destroy_all
```
SuperTux88: my production-db contains 672921 entries (with 5 duplicates), so you want so use a query with `WHERE (aspect_visibilities.id NOT IN (672916 IDs))`?

This raises a `ActiveRecord::StatementInvalid: Mysql2::Error: MySQL server has gone away` on my pod :scream_cat: I apologized and promised that I never ever do this again, when he comes back, so everything is fine :smiley_cat: but we need another query for this, where my mysql server doesn't walk away. :wink:

cmrd-senya: Oops, sorry for that! I'll reconsider that!

cmrd-senya:
Migration is updated, hopefully, your mysql won't be so upset by this one.

SuperTux88:
First, my brain doesn't parse this query :wink: But mysql seems to like it \o/ But now postgres doesn't like it: `ActiveRecord::StatementInvalid: PG::GroupingError: ERROR:  column "aspect_visibilities.id" must appear in the GROUP BY clause or be used in an aggregate function`

I don't know why you need to select everything first, because you can delete it directly on the database (delete vs destroy), because `AspectVisibility` doesn't do anything on destroy. Also we don't need a subselect here :smiley:

You can use this where-clause:
`WHERE a1.aspect_id = a2.aspect_id AND a1.shareable_id = a2.shareable_id AND a1.shareable_type = a2.shareable_type AND a1.id > a2.id`

But MySQL and Postgres have a different syntax for the `DELETE` with "multiple" tables:
Postgres: `DELETE FROM aspect_visibilities AS a1 USING aspect_visibilities AS a2 #{where}`
MySQL: `DELETE a1 FROM aspect_visibilities a1, aspect_visibilities a2 #{where}`

cmrd-senya:
Okay, I made everything pretty much as you said with an `if` condition on the database type.

I wonder if it is possible to overcome the syntax difference by using ARel, but I'm sure  the complication doesn't worth the case.

db/migrate/20160531170531_remove_duplicate_aspect_visibilities.rb
```diff
@@ -0,0 +1,13 @@
+class RemoveDuplicateAspectVisibilities < ActiveRecord::Migration
+  def up
+    where = "WHERE a1.aspect_id = a2.aspect_id AND a1.shareable_id = a2.shareable_id AND "\
+      "a1.shareable_type = a2.shareable_type AND a1.id > a2.id"
+    if AppConfig.postgres?
+      AspectVisibility.connection.execute(
```
SuperTux88: you can use `execute` directly without `AspectVisibility.connection.` in migrations :smiley:

SuperTux88:
> I wonder if it is possible to overcome the syntax difference by using ARel, but I'm sure the complication doesn't worth the case.

No, you can't use the same table twice with AR

cmrd-senya:
I'm talking about [Arel](https://github.com/nkallen/arel)

db/migrate/20160531170531_remove_duplicate_aspect_visibilities.rb
cmrd-senya: fixed

cmrd-senya:
I meant this one of course: https://github.com/rails/arel

SuperTux88:
hmm, I don't know if it's possible with Arel then, but I think its good as it is now, because it is clearly readable what it does.

denschub:
> I wonder if it is possible to overcome the syntax difference by using ARel,

I'm pretty sure it is, but I'm also fairly confident that an Arel solution would be less readable in this case. Anyway, this PR looks fine to me, merged!

PR #6848, opened 2016-06-03
Fix like and reshare from the profile page

### LINKED ISSUE #6841: mobile web ui: like / reshare of posts from a user profile not possible.
Seen on Android device, using Chrome or Firefox: Clicked on a user pic in my stream to get to the user profile. Scrolled down. Clicked on "heart" or "reshare" to interact with some post on the profile page. Outcome: The browser will scroll back to the top of the profile page without performing the action I actually tried to do. Known issue?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/34490088-mobile-web-ui-like-reshare-of-posts-from-a-user-profile-not-possible?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cmrd-senya:
Fix class assignment in the mobile view of the profile page. Add a cuke testing the "liking from the profile view" feature.

fix #6841

### DIFF
--- a/app/views/people/show.mobile.haml
+++ b/app/views/people/show.mobile.haml
@@ -19,7 +19,7 @@
 - if @post_type == :photos
   = render "photos/index", photos: @posts
 - else
-  #main_stream
+  .stream#main_stream
     - if @stream.stream_posts.length > 0
       = render "shared/stream", posts: @stream.stream_posts
       = render "shared/stream_more_button"
--- a/features/mobile/reactions.feature
+++ b/features/mobile/reactions.feature
@@ -15,7 +15,16 @@ Feature: reactions mobile post
 
   Scenario: like on a mobile post
     When I should see "No reactions" within ".show-comments"
-    And I click on selector "span.show-comments"
+    And I click on selector "a.like-action.inactive"
+    Then I should see a "a.like-action.active"
+    When I go to the stream page
+    And I should see "1 reaction" within ".show-comments"
+    And I click on selector "a.show-comments"
+    Then I should see "1" within ".like-count"
+
+  Scenario: liking from the profile view
+    When I am on "alice@alice.alice"'s page
+    Then I should see "No reactions" within ".show-comments"
     And I click on selector "a.like-action.inactive"
     Then I should see a "a.like-action.active"
     When I go to the stream page


app/views/people/show.mobile.haml
```diff
@@ -19,7 +19,7 @@
 - if @post_type == :photos
   = render "photos/index", photos: @posts
 - else
-  #main_stream
+  #main_stream.stream
```
diaspora-code-review: Classes should be listed before IDs (.stream should precede #main_stream)

features/support/application_cuke_helpers.rb
```diff
@@ -32,6 +32,9 @@ def check_fields_validation_error(field_list)
     end
   end
 
+  def mobile_version?
```
Flaburgan: There is probably a cleaner way to know in the cukes if we are using the mobile version,

Flaburgan:
Thanks for the fix! And my bad for the regression :(

Flaburgan:
I noticed that we have a [reaction tests file](https://github.com/diaspora/diaspora/blob/develop/features/mobile/reactions.feature), maybe you want to put your test there?

features/mobile/likes.feature
```diff
@@ -0,0 +1,22 @@
...
+  I want to like their posts
+
+  Background:
+    Given following users exist:
+      | username    | email             |
+      | Bob Jones   | bob@bob.bob       |
+      | Alice Smith | alice@alice.alice |
+    And a user with email "bob@bob.bob" is connected with "alice@alice.alice"
+    And "bob@bob.bob" has a public post with text "I like unicorns"
+    And I sign in as "alice@alice.alice"
```
Flaburgan: The other tests have `And I sign in as "bob@bob.bob" **on the mobile website**`. I don't know if that change anything, I ran the test locally and it was correctly using the mobile version.  I wanted to point that just in case.

features/support/application_cuke_helpers.rb
svbergerem: I don't know a better solution.

svbergerem:
> I noticed that we have a reaction tests file, maybe you want to put your test there?

I agree with @Flaburgan that we shouldn't have tests for liking posts in two different files. Either move your new test to the reactions tests or move the other liking tests from the reactions file to your new file.

While doing that it would be great if you could unify the tests for liking posts. The only difference should be the page you are using for the tests.

features/support/application_cuke_helpers.rb
cmrd-senya: I was searching for something better, but I don't think that mode -- mobile or desktop is stored somewhere in that context. It is stored in the user session data, but it's accessible from controllers, but not from the Capybara context. At the same time we have the `@mobile` tag in cukes, but as it seemed to me that it's rather switches the browser emulation mode and that is why mobile version is picked automatically. So I don't think that there is anything more reliable than just getting it from the metadata in the page content, if we need to do that at all.

cmrd-senya:
Updated. The test moved to the `reactions.feature`. Unification makes other modifications unnecessary.

features/mobile/reactions.feature
```diff
@@ -23,6 +23,17 @@ Feature: reactions mobile post
     And I click on selector "a.show-comments"
     Then I should see "1" within ".like-count"
 
+  Scenario: liking from the profile view
+    When I am on "alice@alice.alice"'s page
+    Then I should see "No reactions" within ".show-comments"
+    And I click on selector "span.show-comments"
```
Flaburgan: There is no event bind when there is no reactions. Even when the click is bind because there are comments, the `.like-action` is shown without having to open the panel. Can you please check if the test succeed without this line? If so, can you please remove it from this test and the one in the main stream, for both "No reaction" and "1 reaction" cases?

cmrd-senya: Yes, it works perfectly. So I remove two lines: this and number 18?

Flaburgan: Yes please ;)

cmrd-senya:
Updated according to [the note](https://github.com/diaspora/diaspora/pull/6848#discussion_r65908990).

Flaburgan:
I tested this locally, I can confirm that the test succeed with the patch and failed without it.
I also tested it manually with the same result for both like and reshare actions. This seems ready to be merged to me. Or maybe also add a test for the reshare action?

svbergerem:
Thanks!

PR #6854, opened 2016-06-07
6853 fix usernames overflow

### LINKED ISSUE #6853: really long usernames "break out" of their ui/css boxes (on /tags pages)
see example

https://therealtalk.org/tags/love

love 2016-06-07 10-22-01

it's really a minor thing. but it's even doing it when "maxwith" is hit.
i think soloution should be something like they're get cut and dottet out (while link stays valid of course)

affects 0.5.99

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/34931979-really-long-usernames-break-out-of-their-ui-css-boxes-on-tags-pages?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SansPseudoFix:
As far as I can see: 

```
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```

is no more needed, but tell me if I'm wrong.

New behavior:

6853-overflow

I'm just concerned about tags background with small width:

6853-overflow-min

Maybe a little bit ugly, no?

_Super Overflow_ strikes again! (should fix #6853 )

### DIFF
--- a/app/assets/stylesheets/tag.scss
+++ b/app/assets/stylesheets/tag.scss
@@ -33,9 +33,7 @@ h1.tag {
       .side_stream #people_stream {
         .name { display: block; }
         .name, .diaspora_handle {
-          overflow: hidden;
-          text-overflow: ellipsis;
-          white-space: nowrap;
+          word-break: break-all;
         }
       }
     }


Flaburgan:
What happens for someone with a long name without spaces, and for a even longer diaspora* ID?

SansPseudoFix:
@Flaburgan how dare you!

6853-overflow-verylong

:)

svbergerem:
Thank you!

PR #6861, opened 2016-06-12
Improve status message creation service spec

svbergerem:
This failed in https://travis-ci.org/diaspora/diaspora/jobs/137026557


jhass:
Thanks!

PR #6863, opened 2016-06-14
remove REDISTOGO_URL deprecation

SuperTux88:
I think we can remove this deprecation-handling after more than 3,5 years.


denschub:
[Aaaaand it's gone](https://www.youtube.com/watch?v=-DT7bX-B1Mg).

PR #6868, opened 2016-06-18
Eslint drop max nested callbacks and add "=" marker to spaced comment

svbergerem:
- We need more nested callbacks in spec files
- We need a `=` at the beginning of a comment (`//= require leaflet`)


denschub:
:shipit:

PR #6870, opened 2016-06-18
Update status message text after prefilling mentions

svbergerem:
Before the hidden field `#status_message_text` hasn't been updated after prefilling mentions. This only happend after the user added some more text.


denschub:
Merged, thanks. :)

Issue #6871, opened 2016-06-19
External sites opened from Diaspora can access window.opener


gandaro:
### How to reproduce the bug?
In the Diaspora stream or in a private message on Diaspora, click on a link to a website that contains JavaScript of the following kind (e.g. [this one](https://mathiasbynens.github.io/rel-noopener/malicious.html)):
```js
if (window.opener) {
    opener.location = 'https://example.com/';
}
```
### What’s happening?
The website containing the JavaScript will open in a new tab (because of the attribute `target="_blank"`, which is set on all external links in Diaspora).

The JavaScript will be executed, and the tab containing the Diaspora stream will relocate to the URL given in the JavaScript (e.g. https://example.com/).

You can see the behavior in this animation:
![Illustration](https://cloud.githubusercontent.com/assets/811907/16174569/4113e3c4-35c7-11e6-8efd-18bab59d423d.gif)
### Why is that a problem?
No one expects their tabs to switch to completely different locations because they open a link in a new tab. This might be abused, for example, by posting links to malicious sites that send the Diaspora tab to a site mimicking the user’s Diaspora pod, showing a sign in page with a “your session timed out” message, potentially phishing credentials.
### How to prevent this?
Set `rel="noopener noreferrer"` on all (or at least user-provided) external links. See https://mathiasbynens.github.io/rel-noopener/#recommendations

denschub:
I'm a bit concerned about the way this issue is "promoted" here and in the demo page you've linked, to be honest. I do not like people who try to make a big deal about issues that are a bit more complex. For everyone really interested in this, I suggest to read the discussion in Chrome bug tracker and all related issues.

The actual issue is not a "missing" rel attribute, but rather an implementation offset included in all browsers. The [spec is already changed](https://html.spec.whatwg.org/multipage/browsers.html#allowed-to-navigate) and the browser manufactors are working on implementing the changes:

> A browsing context A is allowed to navigate a second browsing context B if the following algorithm terminates positively:
>
> 1. If A is not the same browsing context as B, and A is not one of the ancestor browsing contexts of B, and B is not a top-level browsing context, and A's active document's active sandboxing flag set has its sandboxed navigation browsing context flag set, then abort these steps negatively.
> 2. Otherwise, if B is a top-level browsing context, and is one of the ancestor browsing contexts of A, and A's active document's active sandboxing flag set has its sandboxed top-level navigation browsing context flag set, then abort these steps negatively.
[...]

PR #6872, opened 2016-06-19
Add rel="noopener noreferrer" to autolinks

skyschub:
As per request in #6871. This is not a real solution, but rather a hot glue repair job that will not fix the issue itself 100%. I'll add some more explanations in the originating issue report. Since this is a "small enough" fix and it seems to make people happy, it's a good compromise, I guess.

### DIFF
--- a/spec/lib/diaspora/message_renderer_spec.rb
+++ b/spec/lib/diaspora/message_renderer_spec.rb
@@ -169,7 +169,7 @@ describe Diaspora::MessageRenderer do
       it 'should process text with both a hashtag and a link' do
         expect(
           message("Test #tag?\nhttps://joindiaspora.com\n").markdownified
-        ).to eq %{<p>Test <a class="tag" href="/tags/tag">#tag</a>?<br>\n<a href="https://joindiaspora.com" rel="nofollow" target="_blank">https://joindiaspora.com</a></p>\n}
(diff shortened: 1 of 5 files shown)


spec/lib/diaspora/message_renderer_spec.rb
```diff
@@ -169,7 +169,7 @@ def message(text, opts={})
       it 'should process text with both a hashtag and a link' do
         expect(
           message("Test #tag?\nhttps://joindiaspora.com\n").markdownified
-        ).to eq %{<p>Test <a class="tag" href="/tags/tag">#tag</a>?<br>\n<a href="https://joindiaspora.com" rel="nofollow" target="_blank">https://joindiaspora.com</a></p>\n}
+        ).to eq %{<p>Test <a class="tag" href="/tags/tag">#tag</a>?<br>\n<a href="https://joindiaspora.com" rel="nofollow noopener noreferrer" target="_blank">https://joindiaspora.com</a></p>\n}
```
diaspora-code-review: `%`-literals should be delimited by `(` and `)`.

diaspora-code-review: Line is too long. [194/120]

jhass:
Thanks!

Issue #6579, opened 2015-12-07
duplicate "also commented" note from local users 


tchncs:
1. like something on a post from an user on my pod
2. someone also commented something on this post
3. i receive two notifications

This pod is on the master branch... .
Other users dont have this problem.
I guess it only happened, if the "also commented" user is from an external pod... .
The last time, it happened from nerdpol.ch - to reproduce it with my old Geraspora acc has failed.

tchncs:
So...its labeled now... .
I will love to give more information. 
I cant find anything interesting in the logs. Sadly.
Also it doesnt happens on *every* post.
Also again, it doesnt only happens on my "own" pod. 
I maintaining an second one, which is not public yet. 
The first guys from the community where the pod is for, joined a few hours ago. 
I received two notifications for just one comment under a #newhere post i just commented.

Its Debian sid with stable-updates and Ruby 2.1.7.
My Pod is using Ruby 2.1.5 on Debian 8.2.

denschub:
@SuperTux88 didn't you some time in the past talk about this?

denschub:
Closed by #6873.

Issue #4727, opened 2014-01-22
Participations on threads for unconnected users are broken.


rsofaer:
Diaspora rejects federated messages from senders to recipients that the sender is not connected to, unless the incoming message is a request.  

I propose we should make accepting participations from strangers consistent with accepting comments from strangers, and if we reject them, we should stop sending them.

2014-01-22T00:54:19.820498+00:00 app[salmon.1]: event=receive status=abort reason='sender not connected to recipient' recipient=*@joindiaspora.com sender=*@*
2014-01-22T00:54:19.820948+00:00 app[salmon.1]: Failure to receive Participation from sender:* for user:*: Diaspora::ContactRequiredUnlessRequest
2014-01-22T00:54:19.820948+00:00 app[salmon.1]: #<Participation id: nil, guid: "*", target_id: *, target_type: "Post", author_id: *, author_signature: "*", parent_author_signature: "*", created_at: nil, updated_at: nil>
2014-01-22T00:54:19.821049+00:00 app[salmon.1]: error on receive: Diaspora::ContactRequiredUnlessRequest

denschub:
Closed by #6873.

Issue #4697, opened 2014-01-16
Photo displayed in post multiple times


svbergerem:
Example: https://pod.geraspora.de/posts/1836156
The photo is displayed twice. When looking at the [json](https://pod.geraspora.de/posts/1836156.json) you see that the two photos have the same guid but a different id.

aykevl93:
I still see the bug occurring on diasp.eu and pod.orkz.net. It looks like it only occurs while federating (people on the same pod don't see duplicated images) and that a photo is shown up to six times. So maybe it's related to federation and retrying to push images?
I'm using diasp.eu (version 0.4.1.2-p3c87df88) with a lightbox but that doesn't change anything. Multiple images are shown in the stream and even the notifications say there are six images while there was only one posted.

Flaburgan:
>the notifications say there are six images

Where do you see that?

aykevl93:
It says:

> &lt;person A> and &lt;person B> also commented on &lt;person B>'s post 6 photos by &lt;person B>.

Flaburgan:
I think you have an addon or a script like Greasemonkey, I never saw that myself in diaspora*.

aykevl93:
Well, no. It is exactly the same on Firefox for Android with no add-ons installed.
Note that it is a post without text, just one photo.

denschub:
Closed by #6873.

PR #6873, opened 2016-06-19
Federation rewrite with diaspora_federation gem

SuperTux88:
This is the last step of #5114. I rewrote the whole incoming and outgoing federation code to use the new [federation-gem](https://github.com/SuperTux88/diaspora_federation).

This PR currently uses the latest commit on the develop-branch of the federation gem. I would release the 0.1.0 version  and update the PR after this PR was reviewed and tested successfully. Also I suggest to move the gem to the diaspora-organization after this PR is merged.

This PR also includes a complete rewrite of the notification-system, to fix the "sometimes no, sometimes one and sometimes two notification mails"-problem.

The following issues are fixed with this PR: #4727, #4697, #3666, #6579 (and many more bugs, where I didn't find an issue)

I think that #3519 also can be closed after this is merged, because this code doesn't exist anymore.

Note: The migration to cleanup participation (04c8329f7c895a3c132134f2e23a381439b9411c) can cause problems on big mysql databases if they have a `innodb_data_file_path` with `max:`. But this is not enabled by default on mysql (but somehow was on my local test-database). Should we add a note to the Changelog/Update-Notes for that? (see: http://stackoverflow.com/questions/21499169/innodb-delete-table-is-full)

Maybe someone can test the migrations with a bigger postgresql database? I only have a small test-DB (so the syntax should be OK), but I don't know if they cause trouble on bigger DBs (edge cases on old installations). I tested the migrations on a backup of my production mysql-database, so mysql should be OK.

I didn't write migrations to cleanup the `author_signature` of `Like`, `Comment` and `PollParticipation`, because @CSammy wants to make more database-optimizations and said that he can cleanup this then. Also the `author_signature` column of `Message` can be dropped completely after https://github.com/SuperTux88/diaspora_federation/issues/36 is done, so I didn't touch it now.

@cmrd-senya: maybe you can test this PR with your federation-tests?

---

(This part is not part of the initial PR text and gets abused as a ToDo list.)

ToDos before merging:
- [x] Move `SuperTux88/diaspora_federation` to `diaspora/diaspora_federation`
- [x] Verify migration compatibility on PostgreSQL
- [x] Manual testing of all the things
- [x] Second round of `diaspora_federation` gem reviewing
- [x] Release v0.1.0 of `diaspora_federation` gem

ToDos after merging:
- [ ] Work through all federation related issues, close them or move them into the federation repository if applicable
- [ ] Work through all notification related issues and see if those are fixed
- [ ] Create metabug for tracking database optimizations (actually, a milestone and single issues would work even better)
- [ ] Add a note regarding "full" MySQL tables into the upgrade notes.


app/views/notifications/_notification.haml
```diff
@@ -1,9 +1,9 @@
 .media.stream_element{:data=>{:guid => note.id, :type => (Notification.types.key(note.type) || '') }, :class => (note.unread ? 'unread' : 'read')}
   .unread-toggle.pull-right
     %i.entypo-eye{title: (note.unread ? t("notifications.index.mark_read") : t("notifications.index.mark_unread"))}
-  - if note.type == "Notifications::StartedSharing" && contact = current_user.contact_for(note.effective_target)
+  - if note.type == "Notifications::StartedSharing" && contact = current_user.contact_for(note.target)
```
diaspora-code-review: Assignment in condition - you probably meant to use `==`.

app/workers/send_base.rb
```diff
@@ -0,0 +1,28 @@
+module Workers
+  class SendBase < Base
+    sidekiq_options queue: :http, retry: 0
+
+    MAX_RETRIES = (rt = AppConfig.environment.sidekiq.retry.get) ? rt.to_i : 0
```
diaspora-code-review: Assignment in condition - you probably meant to use `==`.

lib/diaspora/federation/entities.rb
```diff
@@ -0,0 +1,258 @@
+module Diaspora
+  module Federation
+    module Entities
+      def self.build(entity)
```
diaspora-code-review: Cyclomatic complexity for build is too high. [14/6]

Gemfile
```diff
@@ -136,7 +136,6 @@ gem "leaflet-rails",       "0.7.7"
 gem "nokogiri",          "1.6.8"
 gem "redcarpet",         "3.3.4"
 gem "twitter-text",      "1.13.4"
-gem "roxml",             "3.1.6"
```
denschub: 🎉 🎈 🍰 ❤️

app/models/contact.rb
```diff
@@ -93,26 +80,45 @@ def self.contact_contacts_for(user, person)
...
+  def object_to_receive
+    Contact.create_or_update_sharing_contact(person.owner, user.person)
+  end
+
+  # @return [Array<Person>] The recipient of the contact
+  def subscribers
+    [person]
+  end
+
+  # creates or updates a contact with active sharing flag. Returns nil if allready sharing.
```
denschub: Nit: already.

app/models/contact.rb
```diff
@@ -93,26 +80,45 @@ def self.contact_contacts_for(user, person)
...
+  def self.create_or_update_sharing_contact(recipient, sender)
+    contact = recipient.contacts.find_or_initialize_by(person_id: sender.id)
+
+    return if contact.sharing
+
+    contact.tap do |contact|
```
denschub: You like `.tap`, don't you? :) However, I am not too sure if it adds anything here besides looking fancy. Not using tap and returning the object without tap actually is a line shorter and looks less complicated:

```ruby
contact.sharing = true
contact.save!
contact
```

Also, if you just update a single attribute,

```ruby
contact.update(sharing: true)
contact
```

Would work, too.

SuperTux88: yes, I think the solution with `update` is better, I'll change that.

app/views/notifications/_notification.haml
denschub: Ignore.

app/workers/send_base.rb
denschub: `to_i` [returns 0 if an invalid string/nil is provided](http://ruby-doc.org/core-2.2.0/String.html#method-i-to_i):

```ruby
MAX_RETRIES = AppConfig.environment.sidekiq.retry.get.to_i
```

config/defaults.yml
```diff
@@ -187,7 +187,7 @@ defaults:
   relay:
     outbound:
       send: false
-      handle: 'relay@relay.iliketoast.net'
+      url: 'https://relay.iliketoast.net/receive/public'
```
denschub: 👍

config/initializers/diaspora_federation.rb
```diff
@@ -63,64 +63,95 @@
...
-    on :receive_entity do
-      # TODO
+    on :receive_entity do |entity, recipient_id|
+      case entity
+      when DiasporaFederation::Entities::AccountDeletion
+        Diaspora::Federation::Receive.account_deletion(entity)
+      when DiasporaFederation::Entities::Retraction
+        Diaspora::Federation::Receive.retraction(entity, recipient_id)
+      else
+        persisted = case entity
```
denschub: Note to myself: think about if/how we can make this prettier.

config/initializers/diaspora_federation.rb
```diff
@@ -63,64 +63,95 @@
...
+      pod = Pod.find_or_create_by(url: url)
+
+      if status.is_a? Symbol
+        pod.status = Pod::CURL_ERROR_MAP.fetch(status, :unknown_error)
+        pod.error = "FederationError: #{status}"
+      elsif status >= 200 && status < 300
+        pod.status = :no_errors unless Pod.statuses[pod.status] == Pod.statuses[:version_failed]
+      else
+        pod.status = :http_failed
+        pod.error = "FederationError: HTTP-Status-Code was: #{status}"
```
denschub: Nit: english does not use hyphens to combine words
```diff
- HTTP-Status-Code
+ HTTP status code
```

features/support/poor_mans_webmock.rb
```diff
@@ -9,8 +9,14 @@ def perform(_sender_atom_url)
     end
   end
 
-  class HttpMulti < Base
-    def perform(_user_id, _enc_object_xml, _person_ids, _retry_count=0)
+  class SendPrivate < Base
+    def perform(_sender_id, _obj_str, _targets, _retry_count=0)
```
denschub: What do you think about `def perform(*_args)`? This would reduce the amount of changes we need to do if we ever change the signature of this function and we have other specs to check if the functions are called correctly.

lib/diaspora/federation/entities.rb
denschub: Yes, @diaspora-code-review, you are perfectly right, the cyclomatic complexity is way too high for a normal function. But dude, take a look at the source! It's simply calling some function depending on the incoming objects, mate. That's not an issue! I am sure you can be happy with that, can't you?

Oh, wait, I am talking to a bot. It's going down.

lib/diaspora/federation/receive.rb
```diff
@@ -0,0 +1,322 @@
...
+            created_at:  entity.created_at,
+            text:        entity.text,
+            commentable: Post.find_by(guid: entity.parent_guid)
+          )
+        end
+      end
+
+      def self.contact(entity)
+        recipient = Person.find_by(diaspora_handle: entity.recipient).owner
+        if entity.sharing.to_s == "true"
```
denschub: Note to myself: No, @SuperTux88 is not crazy. The federation gem is only capable of passing strings at the moment and the conversion to a Boolean happens by AR. This is the only place where a check of that kind is needed, so the adjustment to the federation gem has a low priority, but will be done eventually.

lib/diaspora/taggable.rb
```diff
@@ -1,3 +1,4 @@
+
```
denschub: Oh hi there little newline! You look pretty out of place here, but I will help you finding your original destination!

spec/workers/send_base_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "spec_helper"
+
+describe Workers::SendBase do
+  it "retries first time after at least 256 seconds" do
+    retry_delay = Workers::SendBase.new.send(:seconds_to_delay, 1)
+    expect(retry_delay).to be >= 256
+    expect(retry_delay).to be < 316
+  end
+
+  it "increases the retry time for every retry exponentially " do
```
denschub: Nit: this is not an exponential increase.

denschub:
So, before adding any more comments: **Thank you for your work**. I know that you spent over a year on the federation gem and on getting this first PR out. This is awesome and I guess it is fair to say that this is the biggest change for diaspora so far. It is very sad that, if everything goes well, no user will ever notice what was achieved here, but that's the fun of backend work, isn't it?

I finished a first round of review, mainly looking for obvious mistakes (which I did not find any) and style improvements. The migrations look fine to me, but I will run them on a copy of Geraspora's database (which, coincidentally, is a Postgres database) to make sure. Also, I do expect some more changes to come.

But... great work so far! ❤️

*Finally, a note to everyone else: I know everyone is excited, but please keep the comments of this PR rather clean. Please add your heart and party emoticons to the PR, but please do not post any comments unless needed. If you have questions about a change or you want to suggest something but you are not sure if this is something that should be done, please contact a core team member via IRC. I do expect some technical discussions about some implementations in this PR and everything will be easier if we reduce the amount of off-topic comments. Thanks! <3*

denschub:
As for answering the questions in the initial PR text as well as keeping track of open tasks, I've added a list to the PR description.

cmrd-senya:
> @cmrd-senya: maybe you can test this PR with your federation-tests?

Yes, I'll test this during the next week.

app/controllers/blocks_controller.rb
```diff
@@ -35,9 +35,8 @@ def destroy
   private
 
   def disconnect_if_contact(person)
-    if contact = current_user.contact_for(person)
-      current_user.disconnect(contact, :force => true)
-    end
+    contact = current_user.contact_for(person)
+    current_user.disconnect(contact) if contact
```
jhass: Style question: Should start to make use of `try` for these kind of cases?

```rb
current_user.contact_for(person).try {|contact| current_user.disconnect(contact) }
```

app/models/conversation.rb
```diff
@@ -59,15 +44,12 @@ def set_read(user)
     end
   end
 
-  def public?
-    false
-  end
-
   def participant_handles
-    self.participants.map{|p| p.diaspora_handle}.join(";")
+    participants.map(&:diaspora_handle).join(";")
```
jhass: How likely is it that we'll need the whole objects anyway if this is called? In other words could it be beneficial to `pluck` instead or are we going to load the whole objects anyway?

app/models/message.rb
```diff
@@ -1,58 +1,22 @@
...
-    self.conversation
-  end
+  validates :text, presence: true
+  validate :participant_of_parent_conversation
 
-  def parent= parent
-    self.conversation = parent
+  def conversation_guid=(guid)
+    cnv = Conversation.find_by(guid: guid)
+    self.conversation = cnv if cnv
```
jhass: Similar micro optimization consideration, how likely is it that we'll need the whole `Conversation` object later if this was called? Else we could save the allocation with `super if Conversation.exists?`.

Silently ignoring the assignment feels worrying too, but probably out of scope here to fix. Might be worth a TODO comment nonetheless.

app/models/message.rb
```diff
@@ -62,17 +26,23 @@ def increase_unread(user)
...
   def message
     @message ||= Diaspora::MessageRenderer.new text
   end
 
+  # @return [Array<Person>]
+  def subscribers
+    if author.local?
+      conversation.participants
+    else # for relaying, TODO: can be removed when messages are not relayed anymore
+      conversation.participants.select(&:remote?)
```
jhass: Same again, is the whole collection loaded anyway? If not can we turn this into a DB query?

app/models/notification.rb
```diff
@@ -3,93 +3,56 @@
...
-      s_target = target.is_a?(Reshare) ? target.root : target.parent
-      note_type.concatenate_or_create(recipient, s_target,
-                                          actor, note_type)
-    else
-      note_type.make_notification(recipient, target,
-                                      actor, note_type)
-    end
-    return_note.email_the_user(target, actor) if return_note
-    return_note
+    where(opts.merge!(recipient_id: recipient.id)).order("updated_at desc")
```
jhass: Nit: capitalize `DESC`.

app/models/notification.rb
```diff
@@ -3,93 +3,56 @@
...
-      rescue ActiveRecord::RecordNotUnique
-        nil
-      end
-      n
-    else
-      make_notification(recipient, target, actor, notification_type)
+    find_or_initialize_by(recipient: recipient, target: target, unread: true).tap do |notification|
+      notification.actors |= [actor]
+      # Explicitly touch the notification to update updated_at whenever new actor is inserted in notification.
+      notification.touch unless notification.new_record?
```
jhass: Micro: We shouldn't need to `save!` if the record has no changes, just `touch` and we shouldn't need to `touch` if the record has changes, `save!` will do it:
```rb
if notification.new_record? || notification.changed?
  notification.save!
else
  notification.touch
end

File: app/models/notifications/also_commented.rb [hunk 20]
```diff
@@ -1,17 +1,26 @@
...
+    end
+
+    def deleted_translation_key
+      "notifications.also_commented_deleted"
+    end
+
+    def self.notify(comment, _recipient_user_ids)
+      actor = comment.author
+      commentable = comment.commentable
+      recipients = commentable.participations.map(&:author).select(&:local?) - [commentable.author, actor]
```
jhass: Same questions as before, do we need to force the collections here or could this be a `joins().where().not()`?

app/models/notifications/also_commented.rb
```diff
@@ -1,17 +1,26 @@
...
 
-  def linked_object
-    Post.where(:id => self.target_id).first
+      recipients.each do |recipient|
+        notification = concatenate_or_create(recipient.owner, commentable, actor)
+        notification.email_the_user(comment, actor) if notification
```
jhass: Another `.try { }` candidate.

app/models/notifications/private_message.rb
```diff
@@ -1,15 +1,29 @@
...
+        notify_message(object)
+      end
+    end
+
+    def self.notify_message(message)
+      recipients = message.conversation.participants.select(&:local?) - [message.author]
+      recipients.map(&:owner).each do |recipient|
+        message.increase_unread(recipient)
+        new(recipient: recipient).email_the_user(message, message.author)
+      end
```
jhass: Same story about loading the entire collections, I'll skip any further ones for now.

app/models/photo.rb
```diff
@@ -81,13 +72,11 @@ def ownership_of_status_message
     end
   end
 
-  def self.diaspora_initialize(params = {})
-    photo = shareable_initialize(params)
+  def self.diaspora_initialize(params={})
+    photo = new(params.to_hash.stringify_keys.slice(*column_names, "author"))
```
jhass: `to_h`

app/models/poll_participation.rb
```diff
@@ -1,43 +1,18 @@
...
-  def diaspora_handle
-    self.author.diaspora_handle
-  end
+  validate :not_already_participated
 
-  def diaspora_handle= nh
-    self.author = Person.find_or_fetch_by_identifier(nh)
+  def poll_answer_guid=(new_poll_answer_guid)
+    self.poll_answer = PollAnswer.find_by(guid: new_poll_answer_guid)
   end
```
jhass: Same question as for `conversation_guid=`.

app/models/post.rb
```diff
@@ -132,13 +129,13 @@ def like_for(user)
...
+  # @return [Integer]
+  def update_reshares_counter
+    self.class.where(id: id).update_all(reshares_count: reshares.count)
   end
 
-  # @return Returns true if this Post will accept updates (i.e. updates to the caption of a photo).
-  def mutable?
-    false
+  def self.diaspora_initialize(params)
+    new(params.to_hash.stringify_keys.slice(*column_names, "author"))
```
jhass: `to_h` vs `to_hash`.

app/models/post.rb
```diff
@@ -152,4 +149,10 @@ def comment_email_subject
   def nsfw
     self.author.profile.nsfw?
   end
+
+  def subscribers
+    subs = super
+    subs += resharers + participants if public?
```
jhass: Do we own the object we get from the parent at this point? `Array.concat` could save a few array allocations here.

app/models/user/connecting.rb
```diff
@@ -2,56 +2,59 @@
...
+      contact.aspect_memberships.delete_all
 
-  def remove_contact(contact, opts={:force => false, :retracted => false})
-    if !contact.mutual? || opts[:force]
-      contact.destroy
-    elsif opts[:retracted]
-      contact.update_attributes(:sharing => false)
-    else
-      contact.update_attributes(:receiving => false)
+      disconnect_contact(contact, :receiving, !contact.sharing)
```
jhass: This call could benefit from keyword arguments:
```rb
disconnect_contact(contact, direction: :receiving, destroy: !contact.sharing)
``

File: app/services/notification_service.rb [hunk 28]
```diff
@@ -0,0 +1,21 @@
+class NotificationService
+  NOTIFICATION_TYPES = {
+    Comment       => [Notifications::CommentOnPost, Notifications::AlsoCommented],
+    Like          => Notifications::Liked,
+    StatusMessage => Notifications::Mentioned,
+    Conversation  => Notifications::PrivateMessage,
+    Message       => Notifications::PrivateMessage,
+    Reshare       => Notifications::Reshared,
+    Contact       => Notifications::StartedSharing
+  }.freeze
```
jhass: Always wrapping the values into an array and getting rid of the `[*...]` below is probably a bit cleaner here.

app/services/status_message_creation_service.rb
```diff
@@ -42,19 +42,24 @@ def add_poll(status_message, params)
...
 
   def add_photos(status_message, photos)
-    status_message.attach_photos_by_ids(photos)
-    status_message.photos.each {|photo| photo.public = status_message.public }
+    if photos.present?
+      status_message.photos << Photo.where(id: photos, author_id: status_message.author_id)
+      status_message.photos.each do |photo|
+        photo.public = status_message.public
+        photo.pending = false
+      end
```
jhass: Can this be an `update_all` or do we rely on the validations here?

app/workers/deferred_retraction.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module Workers
+  class DeferredRetraction < Base
+    sidekiq_options queue: :dispatch
+
+    def perform(user_id, retraction_data, recipient_ids, opts)
+      user = User.find(user_id)
+      subscribers = Person.where(id: recipient_ids)
+      object = Retraction.new(retraction_data.deep_symbolize_keys, subscribers)
+      opts = HashWithIndifferentAccess.new(opts)
```
jhass: Okay this probably a bit too fundamental at this point, but I would have loved for the lower level to not rely on HWIAs and expect normalized to string keys option hashes. Or given Ruby 2.3's Symbol GC we could even consider normalizing to symbols.

config/initializers/diaspora_federation.rb
```diff
@@ -6,8 +6,8 @@
   config.certificate_authorities = AppConfig.environment.certificate_authorities.get
 
   config.define_callbacks do
-    on :fetch_person_for_webfinger do |handle|
-      person = Person.find_local_by_diaspora_handle(handle)
+    on :fetch_person_for_webfinger do |diaspora_id|
+      person = Person.where(diaspora_handle: diaspora_id, closed_account: false).where.not(owner: nil).take
```
jhass: Why `take` over `first`? It's unlikely for this to ever match more than one record, but if for some reason it does, `first` would at least give a deterministic result, due ordering by PK.

config/initializers/diaspora_federation.rb
jhass: A hash, moving it out of sight of the glue code and `public_send` is probably acceptable here.

```rb
Diaspora::Federation::Receive.perform(entity)

def self.perform(entity)
  receiver = ENTITY_RECEIVERS.fetch(entity.class) { raise ... }
  public_send(receiver, entity)
end

File: lib/diaspora/federation/dispatcher.rb [hunk 32]
```diff
@@ -0,0 +1,84 @@
...
+      def self.build(sender, object, opts={})
+        if object_should_be_processed_as_public?(object)
+          Public.new(sender, object, opts)
+        else
+          Private.new(sender, object, opts)
+        end
+      end
+
+      def self.defer_dispatch(sender, object, opts={})
+        object.save if object.new_record?
```
jhass: Why the explicit check, is there a reason why it shouldn't be saved if there are just changed attributes? Otherwise `save` is a noop if the object is `persisted? && !changed?`

lib/diaspora/federation/dispatcher.rb
```diff
@@ -0,0 +1,84 @@
...
+        deliver_to_services
+        deliver_to_subscribers
+      end
+
+      private
+
+      attr_reader :sender, :object, :opts
+
+      def self.object_should_be_processed_as_public?(object)
+        object.respond_to?(:public?) && object.public?
```
jhass: Style alternative: `object.try(:public?) || false`

lib/diaspora/federation/dispatcher.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+      private_class_method :object_should_be_processed_as_public?
```
jhass: `def` returns the methods name as symbol now, so in theory you could write `private_class_method def self.`, not sure we want this though.

lib/diaspora/federation/dispatcher.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      def deliver_to_subscribers
+        local_people, remote_people = subscribers.partition(&:local?)
+
+        deliver_to_local(local_people) unless local_people.empty?
+        deliver_to_remote(remote_people)
+      end
+
+      def deliver_to_local(people)
+        obj = object.respond_to?(:object_to_receive) ? object.object_to_receive : object
```
jhass: `object.try(:object_to_receive) || object`

lib/diaspora/federation/dispatcher.rb
```diff
@@ -0,0 +1,84 @@
...
+        Workers::ReceiveLocal.perform_async(obj.class.to_s, obj.id, people.map(&:owner_id)) if obj
```
jhass: The guard if feels a bit hidden here. To please rubocop about it I'd use a guard clause instead `return unless obj`

lib/diaspora/federation/dispatcher.rb
```diff
@@ -0,0 +1,84 @@
...
+      def deliver_to_user_services
+        case object
+        when StatusMessage
+          for_services {|service| Workers::PostToService.perform_async(service.id, object.id, opts[:url]) }
+        when Retraction
+          for_services {|service| Workers::DeletePostFromService.perform_async(service.id, opts) }
+        end
+      end
+
+      def for_services
```
jhass: Just `each_service`?

lib/diaspora/federation/dispatcher/public.rb
```diff
@@ -0,0 +1,46 @@
...
+        end
+
+        def target_urls(people)
+          Pod.where(id: people.map(&:pod_id).uniq).map {|pod| pod.url_to("/receive/public") }
+        end
+
+        def additional_target_urls
+          [].tap do |urls|
+            urls << AppConfig.relay.outbound.url if AppConfig.relay.outbound.send? && object.instance_of?(StatusMessage)
+          end
```
jhass: Too smart IMO, just
```rb
return unless AppConfig.relay.outbound.send? || !object.instance_of(StatusMessage)
[AppConfig.relay.outbound.url]
```

lib/diaspora/federation/entities.rb
```diff
@@ -0,0 +1,258 @@
...
+        when Conversation then conversation(entity)
+        when Like then like(entity)
+        when Message then message(entity)
+        when Participation then participation(entity)
+        when Photo then photo(entity)
+        when PollParticipation then poll_participation(entity)
+        when Profile then profile(entity)
+        when Reshare then reshare(entity)
+        when Retraction then build_retraction(entity)
+        when StatusMessage then status_message(entity)
```
jhass: Mh, so this is essentially the same list and mapping we have in the glue code earlier?

lib/diaspora/federation/entities.rb
```diff
@@ -0,0 +1,258 @@
...
+        when "Post"
+          DiasporaFederation::Entities::SignedRetraction.new(retraction.data)
+        else
+          DiasporaFederation::Entities::Retraction.new(retraction.data)
+        end
+      end
+
+      def self.post(post)
+        case post
+        when StatusMessage then status_message(post)
```
jhass: Style nit: I kinda dislike `then` personally and prefer newlines

lib/diaspora/federation/receive.rb
```diff
@@ -0,0 +1,322 @@
...
+        ignore_existing_guid(Participation, entity.guid, author) do
+          parent = entity.parent_type.constantize.find_by(guid: entity.parent_guid)
+
+          Participation.create!(author: author, guid: entity.guid, target: parent) if parent.author.local?
+        end
+      end
+
+      def self.photo(entity)
+        author = author_of(entity)
+        persisted_photo = load_from_db(Photo, entity.guid, author)
```
jhass: Let's avoid acronyms as much as possible, `db` -> `database`.

lib/diaspora/federation/receive.rb
```diff
@@ -0,0 +1,322 @@
...
+            created_at:            entity.created_at,
+            provider_display_name: entity.provider_display_name,
+            public:                entity.public,
+            root_guid:             entity.root_guid
+          )
+        end
+      end
+
+      def self.retraction(entity, recipient_id)
+        object = entity.target_type.constantize.where(guid: entity.target_guid).take!
```
jhass: Does the layer guarantee that the `constantize` can't fail here?

lib/diaspora/shareable.rb
```diff
@@ -29,50 +30,61 @@ def self.included(model)
...
+      ShareVisibility.batch_import(recipient_user_ids, self)
+    end
+
+    # The list of people that should receive this Shareable.
+    #
+    # @return [Array<Person>] The list of subscribers to this shareable
+    def subscribers
+      user = author.owner
+      if public?
+        user.contact_people + [author]
```
jhass: `[*user.contact_people, author]` should avoid one array allocation here.

lib/diaspora/shareable.rb
```diff
@@ -29,50 +30,61 @@ def self.included(model)
...
+      def for_visible_shareable_sql(max_time, order, limit=15, types=Stream::Base::TYPES_OF_POST_IN_STREAM)
+        by_max_time(max_time, order).order(table_name + ".id DESC").where(type: types).limit(limit)
+      end
+
+      def by_max_time(max_time, order="created_at")
+        where("#{table_name}.#{order} < ?", max_time).order("#{table_name}.#{order} DESC")
+      end
+
+      def owned_by_user(user)
+        user.person.send(table_name).where(pending: false)
```
jhass: `public_send`

jhass:
So far I couldn't spot a merge blocker, good work!

app/models/conversation.rb
SuperTux88: this method is called while dispatching, and the participants are probably already loaded, because he calls `subscribers` before he calls `participant_handles`.

app/models/post.rb
SuperTux88: > Do we own the object we get from the parent at this point?

Yes, the parent is only called if we own it.

jhass: I mean does the parent hold a reference to it otherwise?

SuperTux88: Sorry, I misunderstood you, no the parent doesn't hold a reference to it.

app/workers/deferred_retraction.rb
SuperTux88: I think we should normalize to symbols once diaspora officially supports ruby 2.3.

lib/diaspora/federation/dispatcher.rb
SuperTux88: this doesn't work, because `object_to_receive` can return nil, and then it shouldn't use `object`.

jhass: The cleaner variant would be for all possible objects to implement `object_to_receive` and return `self`, but I have no idea if that's feasible here then.

SuperTux88: I can add a `object_to_receive` to `Diaspora::Federated::Base` then?

lib/diaspora/federation/entities.rb
SuperTux88: same mapping, but other direction. This is outgoing (rails-model -> federation-entity), the other is incoming (federation-entity -> rails-model)

lib/diaspora/federation/receive.rb
SuperTux88: Yes, the federation-gem already checks the target on receive, and checks if the sender of the retraction is allowed to delete this entity, otherwise it would raise a `DiasporaFederation::Entities::Retraction::TargetNotFound` or `DiasporaFederation::Federation::Receiver::InvalidSender`

jhass: But `target_type` comes from the XML? Didn't we want to decouple the XML/entity names from the class names?

lib/diaspora/federation/dispatcher.rb
jhass: Fine with me.

lib/diaspora/federation/receive.rb
SuperTux88: But currently they are the same. Should I create a mapping for String <-> Class instead of `.to_s` and `constantize`?

jhass: I would say so. This is no different to using `const_get` or similar where your big case/when's are currently.

SuperTux88:
@denschub and @jhass thx for the review so far, I updated the PR with the changes you mentioned.

lib/diaspora/federation/entities.rb
SuperTux88: fixed :smiley_cat:

cmrd-senya:
I set up the test machine to test `origin/develop` against your branch. The testsuite is very limited ATM, and the little set I have now [passes](http://190.15.209.124:8010/builders/regression_testing/builds/21/steps/shell_5/logs/stdio) well.

I can dedicate this week to extending the testsuite. Then I can set the suite to test your branch against `origin/stable` and itself also (or any other version actually).

@SuperTux88, if you have any suggestions what features to test first, for example if you're unsure with something or some part is more complex and therefore requires more attention, tell me.

SuperTux88:
Thanks. I think that we don't need many tests in this federation-tests, only the basics to test that federation itself work, and we can test the other cases better with normal unit-/integration-tests.

So if you have start/stop sharing, write/delete public/private post and write/delete comments (best with 3 pods, to test relayables), it would cover most of the federation.

I tested locally with 3 pods (one with this branch and two develop), and everything that didn't work properly was because of stuff still broken in develop :wink:

app/workers/receive_base.rb
```diff
@@ -0,0 +1,33 @@
+module Workers
+  class ReceiveBase < Base
+    sidekiq_options queue: :receive
+
+    include Diaspora::Logging
+
+    # don't retry for errors that will fail again
+    def filter_errors_for_retry
```
diaspora-code-review: Method has too many lines. [21/20]

SuperTux88: yes, I know, but after removing the deprecated errors, it will be under 20 lines again. I hope that is OK for you, holy @diaspora-code-review :innocent:

SuperTux88:
I did a few more improvements on this PR and on the gem.

SuperTux88:
aaaaaand I broke something :scream_cat:

I finally found why a user participates his own posts ... so that the posts are shown in his "my activity"-stream. Funny is, that the tests for this are broken anyway, because there is [a test](https://github.com/diaspora/diaspora/blob/develop/features/mobile/activity_stream.feature#L12-L15) that a users own posts doesn't appear in the activity-stream (which they do in the current stable and develop version). Because of this broken test, I didn't notice this before I opened this PR ... only the second test fails now. So I need to fix the tests too.

But I need to sleep now, I'll fix this later :sleeping:

cmrd-senya:
Intermediate test results.

I [extended](https://github.com/cmrd-senya/diaspora_federation_behavior/commit/d077a5f41f3a21334853b96187b7b2e3def92ff5) the testsuite with a test for conversation (post and comments and retractions) for two people on different pods. The test runs [properly](http://190.15.209.124:8010/builders/regression_testing/builds/26/steps/shell_5/logs/stdio) with this branch against `origin/develop`.

Now, I'll implement and conduct similar tests for 3 pods case to test relayables properly.

cmrd-senya:
Ok, I [implemeted](https://github.com/cmrd-senya/diaspora_federation_behavior/commit/50ae0d349ee07490d4a95cf846a1efecc5c4a3ce) [a test](https://github.com/cmrd-senya/diaspora_federation_behavior/commit/50ae0d349ee07490d4a95cf846a1efecc5c4a3ce#diff-16577f811afc79e70a514367c7880c9fR81) for 3 pods case conversation.

I set the test suite to run the following sets of branches:

pod1 | pod2 | pod3
------------ | ------------- | -------------
origin/develop | origin/stable | origin/stable
origin/develop | origin/stable | origin/develop
origin/develop | SuperTux88/federation-gem-salmon | origin/stable
origin/develop | SuperTux88/federation-gem-salmon | origin/develop

The following steps are taken in the test:
0) user@pod2 shares with user@pod1 and user@pod3
1) Post by user@pod2
2) Comment by user@pod1 and user@pod3
3) Comment retractions by user@pod2
4) Post retraction by user@pod2

Each step is verified to be reflected on every pod. The test is run using both public and private post method.

All the tests [have passed](http://190.15.209.124:8010/builders/regression_testing_3pods/builds/4/steps/shell_5/logs/stdio) with no failures.

The buildbot installation is set to track changes in origin/develop and SuperTux88/federation-gem-salmon branches. Once it notice some changes, it'll run tests automatically, and new builds will appear [here](http://190.15.209.124:8010/builders/regression_testing) for 2-pods tests and [here](http://190.15.209.124:8010/builders/regression_testing_3pods) for 3-pods tests. But I haven't tested that too well and I also can't promise that the test framework will work stable without my supervision.

Note that all the test I perform work in a single process mode of diaspora run (like in develop). If anyone thinks it's important to test it with sidekiq, I could change the setting and rerun the tests.

SuperTux88:
Thanks!

Single process is ok, but you need at least two unicorn worker for working federation (one to send out the entity, and one to handle the incoming webfinger request if the other pod doesn't know the sender already). I don't know if that is the default, because I had deadlock-problems once because of this. If the pods webfinger the persons before they receive the first entity, it works also with one worker.

cmrd-senya:
I have `unicorn_worker` set to 4 on all pods. By single process mode I meant the `single_process_mode`  setting in `diaspora.yml`, which, I guess, only embeds jobs processing to unicorn, but still keeps workers count > 1.

denschub:
Migrations are working on PostgreSQL and a copy of Geraspora's production database. To my surprise, the migrated database is 29,55% smaller than the old one. Quite impressive given the fact that even more database optimizations will follow!

denschub:
Yihhhaaaa! 🎉 

I've merged this so I'm able to test this on Geraspora. I hope this is stable enough for a dev-merge, but we can always file follow-ups.

Thanks again for your work here! ❤️

Flaburgan:
>the migrated database is 29,55% smaller than the old one

Was that migration done from a master database or from a develop one with #6586 already applied?

Awesome news! Big up for the new federation!

denschub:
Yes, #6586 was already applied, so the size reduction for 0.5.0.0 -> 0.6.0.0 will be even more impressive.

Flaburgan:
I updated d-fr which now runs the last develop commits.
Before the pg_dump was 7972654124B
After it is 4846210865B
Here is the result of the `db:migrate` if you're interested about the time: https://pastebin.mozilla.org/8880142

Thank you very much!

PR #6860, opened 2016-06-12
Use eslint instead of jshint

svbergerem:

PR #6874, opened 2016-06-19
Fix guard setup

skyschub:
I just pushed a small fixup (7fb6ffbb4dca257acfb617ec85c1124c9c006e73) that prevents guard-rspec from requiring rspec all the time in development since I discovered a warning in my Rails console. During that, I noticed that our guard setup is... somewhat broken.

With this PR, I
- removed pry-debundle since its very unmaintained and [broken with bundler 1.12](https://github.com/ConradIrwin/pry-debundle/issues/12). If everyone actually used it, there is a way to install it locally.
- removed the jshint guard config which was missed in #6860.
- fixed the guard-cocumber configuration (there were some deprecated values in there)
- told bundler to not require any of the guard-\* gems as a follow-up for 7fb6ffbb4dca257acfb617ec85c1124c9c006e73. guard will require those gems itself, so there is no need for us to do so (especially not if we're in a rails console or the development runserver...)

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -37,9 +37,3 @@ guard :rubocop, all_on_start: false, keep_failed: false do
   watch(/(?:app|config|db|lib|features|spec)\/.+\.rb$/)
   watch(/(config.ru|Gemfile|Guardfile|Rakefile)$/)
 end
-
-guard :jshintrb do
-  watch(/^app\/assets\/javascripts\/.+\.js$/)
-  watch(/^lib\/assets\/javascripts\/.+\.js$/)
-  watch(/^spec\/javascripts\/.+\.js$/)
-end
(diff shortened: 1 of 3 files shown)


Guardfile
```diff
@@ -22,8 +22,7 @@ guard :rspec, cmd: "bin/spring rspec", all_on_start: false, all_after_pass: fals
 end
 
 guard(:cucumber,
-      command_prefix: "bin/spring",
-      bundler:        false,
+      cmd: "bin/spring cucumber",
```
diaspora-code-review: Align the elements of a hash literal if they span more than one line.

jhass:
Merged as 76369f33c5e96fc499b10356ccbb68f071b58eee 2b1942a9b31037d596bba97780bf7d4f2c9091aa 3ae3bfe5203caca5a804ca866b7e4dfdbd0486f4

Thanks!

joallard:
If someone wants to use Pry Debundle outside the bundle, I maintain [a fork](https://github.com/joallard/pry-debundle/tree/fork) which includes a fix for Bundler 1.12.

PR #6877, opened 2016-06-19
Throw exception if no templateName gets provided

skyschub:
but allow setting templateName to false explicitly...

During cucumber debugging, we noticed some strange console loggings. Turns out the profile page had no `templateName` specified, causing the messages. I've refactored the base view to throw if no `templateName` is specified. In the rare case no template is needed, setting it to `false` will mute the exception.


jhass:
Merged as 898c3d404f5e986b8bca2078b722ac0b72045721

Thanks!

PR #6268, opened 2015-07-29
Redesign default.haml

svbergerem:
default.haml

Please comment if you have any idea what we should put on that page.

PR #6881, opened 2016-06-20
Change 'Sign up' to 'Create account'

svbergerem:
See https://github.com/diaspora/diaspora/pull/6268#issuecomment-227121650 and https://github.com/diaspora/diaspora/pull/6268#issuecomment-227236316.


denschub:
sure.

PR #6883, opened 2016-06-22
Remove unused strings from translation files

gandaro:


config/locales/diaspora/en.yml
```diff
@@ -1315,7 +1311,6 @@ en:
       hide_and_mute: "Hide and mute post"
       like: "Like"
       unlike: "Unlike"
-      dislike: "Dislike"
```
gandaro: I wonder if this string is really not needed anymore… I couldn’t find it used anywhere, but there are references to dislikes in *lib/diaspora/likeable.rb* and *spec/controllers/likes_controller_spec.rb*.

gandaro:
Please check my comment on the diff before merging.

denschub:
Have you checked if there are any other unused keys?

denschub:
#6901.

PR #6886, opened 2016-06-26
use the parent author as salmon sender, if the parent author is local

SuperTux88:
this is a problem with old-federation-pods

### DIFF
--- a/spec/lib/diaspora/federation/dispatcher_spec.rb
+++ b/spec/lib/diaspora/federation/dispatcher_spec.rb
@@ -32,6 +32,27 @@ describe Diaspora::Federation::Dispatcher do
 
       expect(dispatcher).to be_instance_of(Diaspora::Federation::Dispatcher::Private)
     end
+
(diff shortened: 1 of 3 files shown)


spec/lib/diaspora/federation/dispatcher_spec.rb
```diff
@@ -32,6 +32,17 @@
...
+
+    it "uses the parent author as sender for a comment if the parent is local" do
+      comment = FactoryGirl.create(:comment, author: bob.person, post: post)
+
+      expect(Diaspora::Federation::Dispatcher::Public).to receive(:new).with(alice, comment, {}).and_call_original
+
+      dispatcher = described_class.build(alice, comment)
+
+      expect(dispatcher).to be_instance_of(Diaspora::Federation::Dispatcher::Public)
+
```
diaspora-code-review: Extra empty line detected at block body end.

denschub:
(green specs, ignoring other tests.)

thanks for the quick fix.

PR #6887, opened 2016-06-26
Fix relayable retraction sender

SuperTux88:
this is a problem with old-federation-pods


denschub:
Spec green, PR merged.

PR #6888, opened 2016-06-26
fix sender for messages for old pods

SuperTux88:
same problem as #6886

Send a second message to remote recipients with conversation-author
as salmon author if the conversation-author is local.

The first dispatch will fail on old pods. New pods will ignore the
second message, if the guid already exist.


jhass:
Thanks1

PR #6889, opened 2016-06-26
bump diaspora_federation to 0.1.1

SuperTux88:
This fetches reshare roots if they are not available locally already.


jhass:
Thanks!

PR #6891, opened 2016-06-27
start cucumber jobs before other jobs

SuperTux88:
the whole build is faster, if the workers start first with the longer
running jobs.


jhass:
Thanks!

PR #6895, opened 2016-06-27
Fix up nodeinfo tests

cmrd-senya:
Tests for nodeinfo failed if services config are non-default. Fix it to pick the data for comparation from application settings.


jhass:
Thanks!

Issue #6632, opened 2016-01-10
notification for hidden and muted private conversation


schubeda:
I got an email notification for a private conversation that was already hidden and muted. Obviously I couldn't see the message when I clicked the link. The other person in the conversation actually sent a message, but I guess I shouldn't get a notification if it's muted.

jaywink:
Could you list your diaspora pod version please? It is in the right sidebar, or go to `/statistics` on the pod you are on, thanks!

schubeda:
The version is 0.5.5.99-p0a02e81b

svbergerem:
Fixed by #6873, test added by #6899.

PR #6899, opened 2016-06-29
Add cuke for deleting conversations

svbergerem:
This adds a test for #6632 which fails when running on the old federation code. So it looks like #6632 also has been fixed by #6873.

While testing this I found out that the color of the flash message for new conversations is wrong, so I fixed that in another commit.


denschub:
Merged, thank you!

PR #6901, opened 2016-07-03
Remove unused translations

svbergerem:
Supersedes #6883. I wasn't able to find any use for these translations and I guess we should see some failing tests if some of those were still in use.

I went through all translations in `config/locales/diaspora/en.yml` and `config/locales/javascript/javascript.en.yml` but there might still be some unused translations that I missed.


jhass:
Thank you!

PR #6902, opened 2016-07-04
Filter mentions on status message creation

svbergerem:
Aspects are added to the status message after creation, so `filter_mentions` didn't do anything since `self.aspects` was always empty. This PR moves the logic to the `StatusMessageCreationService`.

### DIFF
--- a/app/services/status_message_creation_service.rb
+++ b/app/services/status_message_creation_service.rb
@@ -19,7 +19,18 @@ class StatusMessageCreationService
 
   def build_status_message(params)
     public = params[:public] || false
+    filter_mentions params
     user.build_post(:status_message, params[:status_message].merge(public: public))
+  end
+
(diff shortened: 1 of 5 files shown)


app/services/status_message_creation_service.rb
```diff
@@ -19,9 +19,18 @@ def create(params)
 
   def build_status_message(params)
     public = params[:public] || false
+    params[:status_message][:text] = filter_mentions params
```
jhass: Since this is a rather internal method and it gets the entire hash, I think it's okay for it to modify it too, the current line looks a bit weird to me, "filter a hash and get a string back". So just `filter_mentions params` and change `:text` inside.

svbergerem: done.

jhass:
Thank you!

Issue #6852, opened 2016-06-07
on /tags pages when there is no asociated post found, there sould be a message stating that.


paskalito:
With 0.6 now there is a great indication that the site is searching for posts if you search for a tag.
but if it does not find any post with your searched tag the nice loading bar just disappears and nothing further happens.

there sould be a message like. "No posts found" ( for you to see on this pod) ^^

so people don't have to assume that but can be sure..

https://therealtalk.org/tags/lepetitprince
(example of empty post page)

I tested against realtalk beacause of 0.5.99

Issue #4503, opened 2013-09-07
followed tags don't work anymore?


faustdotmef:
i had account at diasp.org and dropped it because it didn't show posts with tags i followed. i thought it's problem of diasp.org but now i see that my new account don't show posts with followed tags anymore as well. and i saw few posts with #bug who have this trouble.
##

svbergerem:
Is anyone still able to reproduce this?

svbergerem:
There was no answer for more than 3 months, so I'll close this issue.

SuperTux88:
I'm reopening this, because I now had this once too. I had a reshare of a post in my stream, but since I follow the #diaspora tag (with which the original post was tagged) the original post should be in my stream too, but wasn't (should be right below the reshare). The reshare was the eleventh post from the top in my stream, so not at the break between too batches. The next post was more than 20 minutes older than the original post, while the reshare was 15 minutes newer then the original, so it should definitely be between these two posts.

I try to debug and reproduce this later.

SuperTux88:
OK, now as soon as the reshare is the fifteenth post (the last of the batch), the missing post is in the next batch (there were even more posts missing I didn't notice).

PR #6903, opened 2016-07-04
Improve /tag page performance.

### LINKED ISSUE #6152: /tags pages are slow
Loading of tag pages takes up to 45 secounds. This seems to be an issue since 0.5.0.0.
Maybe a mysql only issue.

Post regarding this isssue: https://social.mrzyx.de/posts/79584d30f4ea0132c47000259069449e

Running version: 0.5.1.1 
DB: mysql Ver 5.5.43-0
OS: Debian 8.1

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23524853-tags-pages-are-slow?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

CSammy:
Fixes #6152.

In addition, this concerns #6852. Currently, there is a hack in place to make the page work even if the requested tag does not exist. More discussion is necessary in #6852 to establish the desired behaviour on what to do if the result set is empty.

I also noticed there is no test to check if additional posts are loaded correctly when scrolling down. This should be fixed as well.

Work in progress:
- [x] Simple speed fix (#6152)
- [x] Correct behaviour on empty list (#6852)
- [x] Tests for load on scroll-down

### DIFF
--- a/lib/stream/tag.rb
+++ b/lib/stream/tag.rb
@@ -29,7 +29,16 @@ class Stream::Tag < Stream::Base
   end
 
   def posts
-    @posts ||= construct_post_query
+    @posts ||= if user
+                 StatusMessage.user_tag_stream(user, tag.id)
+               else
+                 StatusMessage.public_tag_stream(tag.id)
+               end
+  end
+
+  def stream_posts
+    return [] unless tag
+    super
   end
 
   def tag_name=(tag_name)
@@ -41,14 +50,5 @@ class Stream::Tag < Stream::Base
   # @return [Hash]
   def publisher_opts
     {:open => true}
-  end
-
-  def construct_post_query
-    posts = if user.present?
-              StatusMessage.owned_or_visible_by_user(user)
-            else
-              StatusMessage.all_public
-            end
-    posts.tagged_with(tag_name, :any => true)
   end
 end
--- a/spec/controllers/tags_controller_spec.rb
+++ b/spec/controllers/tags_controller_spec.rb
@@ -107,6 +107,18 @@ describe TagsController, :type => :controller do
           get :show, :name => 'foo', :format => :mobile
           expect(response).to be_success
         end
+
+        it "returns the post with the correct age" do
+          post2 = eve.post(
+            :status_message,
+            text:       "#what #yes #hellyes #foo tagged second post",
+            public:     true,
+            created_at: @post.created_at - 1.day
+          )
+          get :show, name: "what", max_time: @post.created_at, format: :json
+          expect(JSON.parse(response.body).size).to be(1)
+          expect(JSON.parse(response.body).first["guid"]).to eq(post2.guid)
+        end
       end
     end
   end
--- a/spec/lib/stream/tag_spec.rb
+++ b/spec/lib/stream/tag_spec.rb
@@ -85,6 +85,25 @@ describe Stream::Tag do
     it_should_behave_like 'it is a stream'
   end
 
+  describe '#stream_posts' do
+    it "returns an empty array if the tag does not exist" do
+      stream = Stream::Tag.new(FactoryGirl.create(:user), "test")
+      expect(stream.stream_posts).to eq([])
+    end
+
+    it "returns an empty array if there are no visible posts for the tag" do
+      alice.post(:status_message, text: "#what", public: false, to: "all")
+      stream = Stream::Tag.new(nil, "what")
+      expect(stream.stream_posts).to eq([])
+    end
+
+    it "returns the post containing the tag" do
+      post = alice.post(:status_message, text: "#what", public: true)
+      stream = Stream::Tag.new(FactoryGirl.create(:user), "what")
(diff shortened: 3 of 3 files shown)


Flaburgan:
On which dataset did you test the performance improvement? If you generated one, mind sharing the command to do so? Did you test this on both postgreSQL and MariaDB?

lib/stream/tag.rb
```diff
@@ -29,7 +29,12 @@ def tagged_people_count
   end
 
   def posts
-    @posts ||= construct_post_query
+    tag_id = [tag.try(:id)]
```
jhass: so the `try` here is the hack?

CSammy: Yes, that's the hack. Before, it worked with a non-existent tag name and did simply not return results. Now I need a tag ID and retrieving that may fail.

CSammy:
@Flaburgan To be honest, I have no test data set. However, the query I use now is used in the Tag Stream, and the public query is used in the Multi Stream, both of which have not caused as much negative feedback regarding performance. From that I guessed it should be pretty safe.

In addition, this way there is only one query to maintain for a stream of tagged posts.

Flaburgan:
OK so to test the queries with real production data, I removed all my followed tags excepted #diaspora and checked the response time of the XHR request on `/tags/diaspora` and `/followed_tags`.

- diaspora-fr (last develop code, PG, powerful server, not really used), #diaspora: 1200ms, followed tags: 1000ms.
- framasphere (0.5.9, PG, loaded server) , #diaspora: 2400ms, followed tags: 2000ms
- diasp.org (0.5.9, I don't know if already migrated to PG or still on mysql), #diaspora: 3200ms, followed_tags: 2300ms
- joindiaspora (0.5.10.99) #diaspora the request timeout, followed tags: 1500ms

So there is clearly an improvement :)
However I discovered a really strange bug doing those tests, maybe it is #4503:
On diasp.org, the most recent post on the followed_tags page is 4 months old. All following posts look close to it. 4 months is maybe the last time I logged in diasp.org and loaded `/stream`. On the /tags/diaspora page though I have the recent #diaspora post.

I tried to unfollow / follow back #diaspora on the /followed_tags, and did the same on the `/tags/diaspora` page, to visit the profile of contacts who shared posts with #diaspora and to scroll my /stream until I see some posts with #diaspora, but nothing change, the most recent post is still 4 months old on the followed_tags page. 

So maybe we should inspect the followed_tags query first before to merge this?

Flaburgan:
Hm, note that a lot of users complaining on #4503 are from diasp.org, so maybe this bug is pod specific? We would need @dmorley here.

CSammy:
Thanks to @svbergerem and his PR #6974, there is a proper message now to inform a user that no posts have been found. I have tested the behaviour on my test setup after adjusting the code. The only thing missing now is the test for posts being fetched on scroll-down.

app/controllers/tags_controller.rb
```diff
@@ -39,7 +39,14 @@ def show
     end
     @stream = Stream::Tag.new(current_user, params[:name], :max_time => max_time, :page => params[:page])
     respond_with do |format|
-      format.json { render :json => @stream.stream_posts.map { |p| LastThreeCommentsDecorator.new(PostPresenter.new(p, current_user)) }}
+      format.json do
+        json = if @stream.tag
+                 @stream.stream_posts.map {|p| LastThreeCommentsDecorator.new(PostPresenter.new(p, current_user)) }
+               else
+                 []
+               end
```
jhass: I think I would prefer

```rb
def stream_posts
  return [] unless tag
  super
end
```

in `Stream::Tag`.

CSammy:
This PR is now complete and I ask for a merge. The issue with missing posts on certain pods will be followed up in the respective issue.

spec/controllers/tags_controller_spec.rb
```diff
@@ -107,6 +107,17 @@
...
+
+        it "returns the post with the correct age" do
+          post2 = eve.post(
+            :status_message,
+            text:       "#what #yes #hellyes #foo tagged second post",
+            public:     true,
+            created_at: @post.created_at - 1.day
+          )
+          get :show, name: "what", max_time: @post.created_at, format: :json
+          expect(JSON.parse(response.body).first["guid"]).to eq(post2.guid)
```
svbergerem: I'd maybe add
```ruby
expect(JSON.parse(response.body).length).to be(1)
```
(or `expect(JSON.parse(response.body).size).to be(1)` @jhass ?)

jhass: The latter ;)

CSammy: Thank you for the hint, indeed a good addition.

jhass:
Thank you!

PR #6904, opened 2016-07-04
Update .ruby-version

padi:
To latest patch version. In order to prevent `rbenv` users from getting warning: `rbenv: version`2.2' is not installed`


Flaburgan:
I have this problem too as a user of `rbenv` but to not precise the patch version allows rvm's users to always use the last one and as we currently recommend `rvm` during the installation process, we should focus on these users first.

jhass:
rbenv is the clear black sheep here, both RVM and chruby allow fuzzy matching to the latest available patch release, and that's the vast majority of users, not only within the user base of diaspora but in general.

I recommend to use chruby or a rbenv plugin such as https://github.com/tpope/rbenv-aliases to create an alias for the latest patch release.

padi:
I was trying to find the recommended installation process starting from https://wiki.diasporafoundation.org/Getting_started_with_contributing and it didn't occur to me to search for rvm in the search bar. Thanks for clearing that up @jhass @Flaburgan.

Issue #6847, opened 2016-06-02
Change email address for one already used cause error


SansPseudoFix:
Steps to reproduce:

* 2 accounts with different email address
* change one with the same email address than the other

You receved email notification on the changed one, but when you click on the link, you have error: **Email could not be activated. Wrong link?**

Expected: an alert to inform that e-mail addres is already used.

Tested on framasphere.org - version:  0.5.9.0-p26439976

PR #6905, opened 2016-07-05
Fix issue #6847

aoh0x7DE:


jhass:
Thank you! Could you add a test or two?

aoh0x7DE:
Sure. I'm new to rspec, so you might want to double-check me on those.

app/models/user.rb
```diff
@@ -492,6 +503,16 @@ def guard_unconfirmed_email
...
 
+  # Whenever email is set, clear all unconfirmed emails which match
+  def remove_invalid_unconfirmed_emails
+    if email_changed?
+      User.where("unconfirmed_email = ?", email).find_each do |problem_user|
+        problem_user.unconfirmed_email = nil
+        problem_user.save
+      end
+    end
+  end
```
jhass: I don't think setting it to `nil` can ever make a user fail validation, so `User.where(unconfirmed_email: email).update_all(unconfirmed_email: nil)` should be much, much faster.

spec/models/user_spec.rb
```diff
@@ -305,6 +305,13 @@
         alice.email = "somebody@anywhere"
         expect(alice).not_to be_valid
       end
+
+      it "resets a matching unconfirmed_email on save" do
+        eve.update_attribute :unconfirmed_email, "new@email.com"
+        alice.update_attribute :email, "new@email.com"
```
jhass: Let's use `example.com` rather than a potentially valid address.

app/models/user.rb
```diff
@@ -484,6 +488,13 @@ def mine?(target)
   end
 
 
+  # Ensure that the unconfirmed email isn't already someone's email
+  def unconfirmed_email_quasiuniqueness
+    if User.exists?(["id != ? AND email = ?", id, unconfirmed_email])
+      errors.add(:unconfirmed_email, "is already in use")
```
jhass: This is a user facing error, right? It should use a translation then :)

aoh0x7DE:
Okay, I'll make those changes. Should I change the other tests which use new@email.com while I'm at it?

Forgive my noobishness, but is there an example of an error using a translation somewhere in the codebase I could look at? Not quite sure how to go about it as-is.

Also, just changed the controller code a little. Pronto is now complaining about the line after the one I changed having single quotes instead of double; I'm tentatively ignoring this (?) because it matches all the others.

jhass:
Yeah, changing the other tests seems fine.

http://guides.rubyonrails.org/i18n.html#translations-for-active-record-models goes a bit into validation error translations, I actually never tried myself, but if the interpolation variant works for custom errors, it would be highly preferable. Would be awesome if you could find out. I think we also have some stuff to clean up if so ;) If you need to add any translation strings, only edit the English locale.

You always should follow pronto's messages. We introduced a fixed styleguide much much later and are porting the codebase to it as we touch things. Porting everything to a consistent style at once would just be too many changes at once.

aoh0x7DE:
Got everything sorted. As it turns out, there was a message (errors.messages.taken) that matched what I was after, so I didn't end up learning about custom messages.

However, the message might not be relevant at the moment. Currently, the users controller flashes generic messages on a failed save (e.g., "Email change failed") rather than making any reference to the errors in question. Is this behavior that should be modified?

app/controllers/users_controller.rb
```diff
@@ -111,7 +112,7 @@ def public
       respond_to do |format|
         format.atom do
           @posts = Post.where(author_id: @user.person_id, public: true)
-                    .order('created_at DESC')
+                    .order("created_at DESC")
```
diaspora-code-review: Align `.order` with `.where` on line 114.

app/controllers/users_controller.rb
```diff
@@ -120,7 +121,7 @@ def public
         format.any { redirect_to person_path(@user.person) }
       end
     else
-      redirect_to stream_path, :error => I18n.t('users.public.does_not_exist', :username => params[:username])
+      redirect_to stream_path, :error => I18n.t("users.public.does_not_exist", :username => params[:username])
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

app/controllers/users_controller.rb
```diff
@@ -171,9 +172,9 @@ def user_photo
 
   def confirm_email
     if current_user.confirm_email(params[:token])
-      flash[:notice] = I18n.t('users.confirm_email.email_confirmed', :email => current_user.email)
+      flash[:notice] = I18n.t("users.confirm_email.email_confirmed", :email => current_user.email)
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

jhass:
> Is this behavior that should be modified?

Perhaps, but not in this PR ;)

jhass:
Thanks!

PR #6906, opened 2016-07-05
Use the full bind path instead of the url

ghost:
In case your domain is different to the configured one
you will run into cross domain issues.

e.g. if you have an onion address you'll have problems with cross-domain
and more importantly sending traffic to the clear web.

### DIFF
--- a/app/helpers/jsxc_helper.rb
+++ b/app/helpers/jsxc_helper.rb
@@ -7,6 +7,6 @@ module JsxcHelper
     unless AppConfig.chat.server.bosh.proxy?
       return "http://#{host}:#{port}#{bind}"
     end
-    AppConfig.url_to bind
+    bind
   end
 end


jhass:
Perhaps we can simplify the configuration story there a bit? https://github.com/diaspora/diaspora/blob/9aaf58de120e98d1a6ada24718fc3ffc411a5f0b/config/diaspora.yml.example#L235 seems outdated anyway. Can we drop the `proxy` and `port` setting and just do something like

```rb
bind = AppConfig.url_to bind if  bind.start_with? "/"
bind
```

?

Zauberstuhl:
> Can we drop the proxy and port setting and just do something like

Forcing the podmin to configure his webserver? why not
The only downside would be that a podmin has to manually change the prosody-config if he wants to set a different port (in respect of the prosody wrapper). But I can live with that

also you might have some more work if you want to test jsxc in development mode.

jhass:
Well with this `https://xmpp.example.org:5280/bind` would be a valid value for `bind` anyway, no?

Zauberstuhl:
mh you're right.. classic .. too much normalization

Flaburgan:
So what do we want to do with this?

SuperTux88:

PR #6908, opened 2016-07-06
Add 'Be excellent to each other!' to the sidebar

svbergerem:
Be excellent to each other

### DIFF
--- a/app/assets/javascripts/app/views/sidebar_view.js
+++ b/app/assets/javascripts/app/views/sidebar_view.js
@@ -4,16 +4,11 @@ app.views.Sidebar = app.views.Base.extend({
   el: ".info-bar",
 
   events: {
-    "click input#invite_code": "selectInputText",
-    "click .section .title": "toggleSection"
+    "click input#invite_code": "selectInputText"
   },
 
   selectInputText: function(event) {
     event.target.select();
-  },
-
-  toggleSection: function(e) {
-    $(e.target).closest(".section").toggleClass("collapsed");
   }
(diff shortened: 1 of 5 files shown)


jhass:
Perhaps we should undo the collapsed items and always display them again?

svbergerem:
Seems legit, done.

no collapsed sections

Flaburgan:
Hm, I really liked them collapsed, it makes the ui clean and sober. What made you change your mind?

jhass:
Well, the point of this is having something visible all the time. Having to uncollapse it first defeats it.

Flaburgan:
So maybe it can be placed in the footer instead?

AugierLe42e:
I really like them collapsed too. If you want to keep them uncollapsed, then there should definitely a option to hide that sidebar.

jhass:
Being able to hide it, too, defeats  the point of having something visible all the time.

SuperTux88:
Must the whole sidebar be always uncollapsed? Somehow I don't like the uncollapsed sidebar under the other sidebar (and it is really big on followed tags page), maybe move it back to the left side again?

AugierLe42e:
> Being able to hide it, too, defeats the point of having something visible all the time.

There's no point in having something visible all the time for experienced users. I, for instance, have nerver had any use of this sidebar. It's just visual pollution for me.

jhass:
This PR is not primarily about uncollapsing the sidebar, that was just done as a solution to the primary purpose.

Issue #6850, opened 2016-06-05
Don't include subjects of private messages in email notifications


svbergerem:
As reported in this comment:
https://github.com/diaspora/diaspora/issues/4266#issuecomment-222394415

Changing [these two lines](https://github.com/diaspora/diaspora/blob/348a7b/app/mailers/notification_mailers/private_message.rb#L11-12) should fix this issue.

SaiVardhan:
@svbergerem  Can you give me a road map for this to kick start as I am a newcomer.

Right now, I can see the private message(Lets say "Hi" is the subject) is getting displayed as "**Hi** *There’s a new private message in diaspora* for you to check out.* "

So how it should be changed

svbergerem:
@SaiVardhan [These are the two lines](https://github.com/diaspora/diaspora/blob/348a7b/app/mailers/notification_mailers/private_message.rb#L11-12) that set the subject of a private message. For comments on private posts [we already don't add any private information to the subject](https://github.com/diaspora/diaspora/blob/aa6b590/app/mailers/notification_mailers/comment_on_post.rb#L12) so that might be something to compare this to. Right now we add a “Re:” in front of the subject for private messages if this isn't the first message in the conversation. If we remove the private information I think the “Re: ” will be useless so we could also remove that extra case. The translations for the private posts are [here](https://github.com/diaspora/diaspora/blob/bef5241/config/locales/diaspora/en.yml#L678) and there are [already some for private messages](https://github.com/diaspora/diaspora/blob/bef5241/config/locales/diaspora/en.yml#L685) so you might want to add the translation for the subject of private message notifications there. Afterwards you will have to add or modify tests in [spec/mailers/notifier_spec.rb](https://github.com/diaspora/diaspora/blob/0f04414/spec/mailers/notifier_spec.rb#L190-L233).

So here is your roadmap:

1. Replace the current subject of private message notification mails with a translated subject that doesn't include any private information
2. Drop the extra “Re: ” case
3. Add/Modify tests

Feel free to ask if you have any further questions. If you get stuck while working on this then please open an pull request with the work you did so far. That's the easiest way for us to help you with your code.

SaiVardhan:
can you suggest subject I am thinking to have it as "Diaspora: Private Message"

PR #6910, opened 2016-07-08
 [Fix] #6850 Don't include subjects of private message

SaiVardhan:
#6850  Don't include subjects of private messages in email notifications

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -683,6 +683,7 @@ en:
         subject: "%{name} has mentioned you on diaspora*"
         limited_post: "You were mentioned in a limited post."
     private_message:
+        subject: "There’s a new private message for you"
         reply_to_or_view: "Reply to or view this conversation >"
     liked:
         liked: "%{name} liked your post"
--- a/spec/mailers/notifier_spec.rb
+++ b/spec/mailers/notifier_spec.rb
@@ -212,15 +212,12 @@ describe Notifier, type: :mailer do
       expect(@mail["From"].to_s).to eq("\"#{@cnv.author.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>")
     end
 
-    it "SUBJECT: has a snippet of the post contents" do
-      expect(@mail.subject).to eq(@cnv.subject)
+    it "should use a generic subject" do
+      expect(@mail.subject).to eq(I18n.translate("notifier.private_message.subject"))
     end
 
-    it "SUBJECT: has 'Re:' if not the first message in a conversation" do
-      @cnv.messages << Message.new(text: "yo", author: eve.person)
-      @mail = Notifier.private_message(bob.id, @cnv.author.id, @cnv.messages.last.id)
-
-      expect(@mail.subject).to eq("Re: #{@cnv.subject}")
+    it "SUBJECT: should not has a snippet of the private message contents" do
+      expect(@mail.subject).not_to eq(@cnv.subject)
     end
 
     it "BODY: does not contain the message text" do
--- a/app/mailers/notification_mailers/private_message.rb
+++ b/app/mailers/notification_mailers/private_message.rb
@@ -8,8 +8,7 @@ module NotificationMailers
       @participants = @conversation.participants
 
       @headers[:from] = "\"#{@message.author_name} (diaspora*)\" <#{AppConfig.mail.sender_address}>"
-      @headers[:subject] = @conversation.subject.strip
-      @headers[:subject] = "Re: #{@headers[:subject]}" if @conversation.messages.size > 1
+      @headers[:subject] = I18n.t("notifier.private_message.subject")
     end
   end
 end


spec/mailers/notifier_spec.rb
```diff
@@ -213,14 +213,14 @@
     end
 
     it "SUBJECT: has a snippet of the post contents" do
-      expect(@mail.subject).to eq(@cnv.subject)
+      expect(@mail.subject).to eq(I18n.translate 'notifier.private_message.subject')
```
diaspora-code-review: Add parentheses to nested method call `I18n.translate 'notifier.private_message.subject'`.

diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/mailers/notifier_spec.rb
```diff
@@ -213,14 +213,14 @@
...
     end
 
     it "SUBJECT: has 'Re:' if not the first message in a conversation" do
       @cnv.messages << Message.new(text: "yo", author: eve.person)
       @mail = Notifier.private_message(bob.id, @cnv.author.id, @cnv.messages.last.id)
 
-      expect(@mail.subject).to eq("Re: #{@cnv.subject}")
+      expect(@mail.subject).to eq(I18n.translate 'notifier.private_message.subject')
```
diaspora-code-review: Add parentheses to nested method call `I18n.translate 'notifier.private_message.subject'`.

diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

SaiVardhan: @diaspora-code-review 
Thanks for the review. Can I know why you  perfer double quotes for string

Flaburgan: It's a bot set on our coding style ;)

config/locales/diaspora/en.yml
```diff
@@ -683,6 +683,7 @@ en:
         subject: "%{name} has mentioned you on diaspora*"
         limited_post: "You were mentioned in a limited post."
     private_message:
+        subject: "Diaspora: Private Message"
```
Flaburgan: The sender of the e-mail is diaspora*, I'm not sure it is necessary to add it there.

SaiVardhan: So what can be a suggested subject? Is "New Private Message" fine?

svbergerem:
@SaiVardhan We have some [styleguides](https://wiki.diasporafoundation.org/Styleguide) and once you commited your changes you can check your changes before pushing them with
```
bin/pronto run -c upstream/develop
```

config/locales/diaspora/en.yml
svbergerem: The subject for `also_commented` notifications for private posts is
> There's a new comment on a post you commented

so maybe 
> There’s a new private message for you

spec/mailers/notifier_spec.rb
```diff
@@ -213,14 +213,14 @@
     end
 
     it "SUBJECT: has a snippet of the post contents" do
```
svbergerem: Please change the description since it isn't correct anymore.

spec/mailers/notifier_spec.rb
```diff
@@ -213,14 +213,14 @@
...
+      expect(@mail.subject).to eq(I18n.translate("notifier.private_message.subject"))
     end
 
     it "SUBJECT: has 'Re:' if not the first message in a conversation" do
       @cnv.messages << Message.new(text: "yo", author: eve.person)
       @mail = Notifier.private_message(bob.id, @cnv.author.id, @cnv.messages.last.id)
 
-      expect(@mail.subject).to eq("Re: #{@cnv.subject}")
+      expect(@mail.subject).to eq(I18n.translate("notifier.private_message.subject"))
     end
```
svbergerem: We dropped the functionality so we can also drop this test.

config/locales/diaspora/en.yml
SaiVardhan: @svbergerem 
I too thought of the same but if we check the mail message we are getting it is

"**Subject**- _There’s a new private message in diaspora* for you to check out_" 
 which shows the body in italic and if we add the content it would be confusing for User

PFA 

test

SaiVardhan:
@svbergerem 
 When I run
 `bin/pronto run -c origin/develop` 
I am seeing these warnings 
`warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.5-compliant syntax, but you are running 2.2.4.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri`

I am using Ubuntu 14.04 and rvm but I see parser is compatible with Macruby

Please help in this regard

svbergerem:
@SaiVardhan This is just a warning and I guess it shouldn't be a problem. @jhass Anything you'd like to add?

config/locales/diaspora/en.yml
svbergerem: That depends on your email client but I get your point. Does anyone have a better idea? The subjects of emails we send tend to be more verbose than “New private message” so to me it feels like that doesn't really fit.

spec/mailers/notifier_spec.rb
```diff
@@ -212,15 +212,8 @@
...
-    it "SUBJECT: has a snippet of the post contents" do
-      expect(@mail.subject).to eq(@cnv.subject)
-    end
-
-    it "SUBJECT: has 'Re:' if not the first message in a conversation" do
-      @cnv.messages << Message.new(text: "yo", author: eve.person)
-      @mail = Notifier.private_message(bob.id, @cnv.author.id, @cnv.messages.last.id)
-
-      expect(@mail.subject).to eq("Re: #{@cnv.subject}")
+    it "SUBJECT: should include translation fallback" do
```
jhass: It's not a fallback :)

Also a second spec that the subject is actually *not* included would be awesome.

jhass:
It's indeed just a warning you can safely ignore, if it annoys you too much you can do `rvm get head; rvm install 2.2.5`.

spec/mailers/notifier_spec.rb
SaiVardhan: @jhass 
Updated commit as per your comments

spec/mailers/notifier_spec.rb
```diff
@@ -212,15 +212,8 @@
...
-    it "SUBJECT: has a snippet of the post contents" do
-      expect(@mail.subject).to eq(@cnv.subject)
-    end
-
-    it "SUBJECT: has 'Re:' if not the first message in a conversation" do
-      @cnv.messages << Message.new(text: "yo", author: eve.person)
-      @mail = Notifier.private_message(bob.id, @cnv.author.id, @cnv.messages.last.id)
-
-      expect(@mail.subject).to eq("Re: #{@cnv.subject}")
+    it "should include notifier translation data" do
```
jhass: `it "should use a generic subject"` perhaps? Also you should be fine using the `I18n.t` alias below. And I would still love a spec actually asserting that `@cnv.subject` is not in `@mail.subject` :)

jhass:
Thanks

SaiVardhan:
welcome :)

PR #6911, opened 2016-07-08
Remove unused methods

cmrd-senya:
While examining results of the analysis by the new fancy [coveralls](https://coveralls.io/github/diaspora/diaspora) tool, I noticed a couple ([1](https://coveralls.io/builds/6927615/source?filename=app%2Fmodels%2Fpost.rb#L56), [2](https://coveralls.io/builds/6927615/source?filename=app%2Fmodels%2Fstatus_message.rb#L128))  of methods that are marked as uncovered and that aren't referenced anywhere in the source cod. I suppose that these are just some methods left from the old code refactorings, so they could be safely deleted. Therefore, this PR should increase the test coverage rate.

### DIFF
--- a/app/models/status_message.rb
+++ b/app/models/status_message.rb
@@ -149,10 +149,5 @@ class StatusMessage < Post
   def self.tag_stream(tag_ids)
     joins(:taggings).where('taggings.tag_id IN (?)', tag_ids)
   end
-
-  def after_parse
-    # Make sure already received photos don't invalidate the model
-    self.photos = photos.select(&:valid?)
-  end
 end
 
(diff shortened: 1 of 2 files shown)


app/models/status_message.rb
```diff
@@ -149,10 +149,5 @@ def absence_of_content
   def self.tag_stream(tag_ids)
     joins(:taggings).where('taggings.tag_id IN (?)', tag_ids)
   end
-
-  def after_parse
```
SuperTux88: this is a leftover from the old federation and can be removed. Maybe there is more unused stuff, that I haven't found yet, sorry for that.

cmrd-senya: That what I thought.

`rgrep _parse app` gives nothing. But I guess with the coverage tool searching for unused code is an easy walk :)

svbergerem:
Thanks!

PR #6914, opened 2016-07-11
Add 'Be excellent to each other!' to the sidebar

skyschub:
Supersedes #6908 after some discussions via IRC.

screen shot 2016-07-11 at 23 10 48

### DIFF
--- a/app/assets/stylesheets/navbar_left.scss
+++ b/app/assets/stylesheets/navbar_left.scss
@@ -105,10 +105,18 @@
   margin-top: 20px;
   padding: 10px 20px 0;
 
+  .excellence-box,
   .info-links {
     border-top: 1px solid $border-grey;
-    margin-top: 10px;
     padding: 5px;
+  }
+
+  .excellence-box {
+    margin-top: 10px;
+
+    .content {
+      font-size: $font-size-base;
+    }
   }
 
   .section {
--- a/app/views/streams/main_stream.html.haml
+++ b/app/views/streams/main_stream.html.haml
@@ -149,6 +149,11 @@
               %p
                 = link_to t("aspects.index.help.mail_podmin"), "mailto:#{AppConfig.admins.podmin_email}"
 
+        .excellence-box
+          .content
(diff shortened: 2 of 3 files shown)


Flaburgan:
Just tested this locally, the code is fine and everything works well so it's a :+1: 
Although you can increase a little bit the font size if you want to, it's hardly readable on a 1920px width screen.

svbergerem:
Here are two versions. The left one is a screenshot with the original code, for the right one I added `font-size: $font-size-base;`. I'm unable to decide which one I'd prefer.

excellent

tchncs:
What about centered?
2016-07-13-100105_1920x1200_scrot

svbergerem:
And with the bigger font:
excellent centered

Since this would be the only centered text in the sidebar I fear that this will make too much “noise”.

tchncs:
Hmm but maybe it would be good to make something like this as a podmin-controllable motd.
In an own box: 


2016-07-13-110120_1920x1200_scrot


The current way is like a branding, maybe not every podmin wanna have this there and maybe not every one knows how to change it to own preference - just my personal thinking about.
To have this as a default motd which can be changed by the podmins, might be a more elegant way and brings a neat new feature .... and it's maybe more beautiful and more solid looking. :3

svbergerem:
@tchncs I think that this is out of scope for this PR and it's even more noise for the left sidebar which is really clean right now. There is also the “Welcome new users” section that you might call branding that not all podmins would like to have on their pods.

I think we should just make the font a little bit bigger and then this should be ready to merge.

svbergerem:
Thank you! :)

jpope777:
> "some discussions via IRC"
Anyone have a log of these discussions?

denschub:
IRC discussions were about the actual design of this element, not the contents or the usefulness. Discussions about adding a message as a start and following up with more community guidelines was/is done [on Loomio](https://www.loomio.org/d/F5zxTwiX/diaspora-statement) as usual.

PR #6547, opened 2015-11-15
Refactored photo extraction

samdvr:
I did this to improve the quality of user model and extract photo extraction responsibility to a dedicated object.

PR #6922, opened 2016-07-18
Refactored photo extraction

cmrd-senya:
supersedes #6547 

style issues fixes

### DIFF
--- a/lib/photo_exporter.rb
+++ b/lib/photo_exporter.rb
@@ -1,0 +1,44 @@
+class PhotoExporter
+  attr_reader :user
+
+  def initialize(user)
+    @user = user
+  end
+
+  def perform
+    temp_zip = Tempfile.new([user.username, "_photos.zip"])
+    begin
+      Zip::OutputStream.open(temp_zip.path) do |zip_output_stream|
+        user.photos.each do |photo|
+          export_photo(zip_output_stream, photo)
+        end
+      end
+    ensure
+      temp_zip.close
+    end
+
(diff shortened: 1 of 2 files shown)


lib/photo_exporter.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+  rescue Errno::ENOENT
+    user.logger.info "Export photos error: #{photo.unprocessed_image.file.path} not found"
+  end
+
+  def update_exported_photos_at(temp_zip)
+    user.update exported_photos_file: temp_zip, exported_photos_at: Time.zone.now if temp_zip.present?
+    user.restore_attributes if user.invalid? || temp_zip.present?
+    user.update exporting_photos: false
+  end
```
jhass: Upon taking a second look, the logic was actually changed here, the last two lines happened in an `ensure`. Though tbh I don't quite follow the logic of the `|| temp_zip.present?` here, do you?

cmrd-senya: So I guess `ensure` was lost unintentionally.

I don't follow it either, I guess it was quite wrong. The thing is that `temp_zip.present?` returns true everywhere here even if `temp_zip.close` or `temp_zip.unlink` called. Therefore `user.restore_attributes` is called everytime. However for the successful save case the  `user.restore_attributes` is no-ops because it restores attributes for dirty model, while `user.update` saves model and therefore makes the model instance clean. So I think we can safely remove `|| temp_zip.present?` condition.

jhass:
Thank you!

Issue #6642, opened 2016-01-16
Strip Markdown from conversation title.


Flaburgan:
![capture du 2016-01-16 11-27-58](https://cloud.githubusercontent.com/assets/930064/12371905/3be894f0-bc44-11e5-8d43-513bf741b418.png)

noplanman:
Similar thing applies to the notifications.
It doesn't show the whole markdown though, just the URL of a link.

**Post:**
Just a [link](http://somewhere).

**Notification popup:**
Just a link (http://somewhere).

*(GitHub is having issues with image upload, will update this comment and add a screenshot as soon as it works again)*

Flaburgan:
That's a choice, to not have an external link inside the notification.

noplanman:
Here the screenshot of the notification showing the URL:

![url-in-notifications](https://cloud.githubusercontent.com/assets/9423417/12861485/077f5c1e-cc64-11e5-8dd6-7f5931c942e9.png)

Since the link isn't clickable on the notification anyway, it would make sense to not have it there.

SaiVardhan:
@Flaburgan  can you elaborate to fix this issue

Flaburgan:
@SaiVardhan what do you need to know? Simply create a new conversation (private message) with some markdown in the subject and at the beginning of the message, and you will see in the right list that the markdown is not stripped (**subject** will appear `**subject**`). As in the notification panel, it should be stripped.

@noplanman to completely hide the link or to show it is another question, you can open an issue about it if you feel so. At the moment, I'd like to see the markdown stripped the same way in the conversation and in the notification.

PR #6923, opened 2016-07-19
[FIX] #6642 Strip Markdown from conversation title

SaiVardhan:
Strip Markdown from conversation title

### DIFF
--- a/app/views/conversations/_conversation.haml
+++ b/app/views/conversations/_conversation.haml
@@ -32,7 +32,7 @@
         .last_message
           - if conversation.messages.present?
             %em
-              = conversation.messages.last.text
+              = conversation.messages.last.message.plain_text_without_markdown
         - if other_participants.count > 1
           .participants
             - other_participants.drop(1).take(15).each do |participant|
--- a/spec/controllers/conversations_controller_spec.rb
+++ b/spec/controllers/conversations_controller_spec.rb
@@ -64,7 +64,7 @@ describe ConversationsController, :type => :controller do
         author:              alice.person,
         participant_ids:     [alice.contacts.first.person.id, alice.person.id],
         subject:             "not spam",
-        messages_attributes: [{author: alice.person, text: "cool stuff"}]
+        messages_attributes: [{author: alice.person, text: "**cool stuff**"}]
       }
       @conversations = Array.new(3) { Conversation.create(hash) }
       @visibilities = @conversations.map {|conversation|
@@ -101,6 +101,14 @@ describe ConversationsController, :type => :controller do
       sign_in :user, eve
       get :index, conversation_id: @conversations.first.id
       expect(assigns[:conversation]).to be_nil
+    end
+
+    it "retrieves a conversation message with out markdown content " do
+      get :index
+      @conversation = @conversations.first
+      expect(response).to be_success
+      expect(response.body).to match(%r/cool stuff/)
+      expect(response.body).not_to match(%r/<strong>cool stuff<\/strong>/)
     end
   end
 
--- a/spec/controllers/conversations_controller_spec.rb
+++ b/spec/controllers/conversations_controller_spec.rb
@@ -107,8 +107,8 @@ describe ConversationsController, :type => :controller do
       get :index
       @conversation = @conversations.first
       expect(response).to be_success
-      expect(response.body).to match(%r/cool stuff/)
-      expect(response.body).not_to match(%r/<strong>cool stuff<\/strong>/)
+      expect(response.body).to match(/cool stuff/)
+      expect(response.body).not_to match(%r{<strong>cool stuff</strong>})
     end
   end
 


jhass:
Tests :)

`Message` already has `#message` which returns a `MessageRenderer` for `#text`. `Conversation` should probably gain a similar convenience method for returning a wrapped `#subject`.

Though I'm not entirely sure we should render markdown there, especially in the subject, it's rendering links inside links, isn't it? Perhaps `MessageRenderer#title` is more appropriate?

Flaburgan:
>Though I'm not entirely sure we should render markdown there

SuperTux88:
Why is there markdown in the subject anyway? this field doesn't support markdown, so we maybe should strip/validate it before we save/federate it instead of strip it only for display?

svbergerem:
#6642 wasn't about the subject of a conversation but about the shortened last message we display in the left sidebar (inbox). Subjects don't support markdown so we don't have to change anything there. We just don't render it.

Messages however support markdown so when displaying the shortened last message we will have to strip it.

SaiVardhan:
@jhass,

so you don't want the subject to be rendered in markdown and the message should be stripped. right?
`#strip` you mean to remove or show them in markdown content

jhass:
I'd go for `MessageRenderer#plain_text_without_markdown` (notice the `truncate` option) or `MessageRenderer#title`, depending on what looks better.

SaiVardhan:
@jhass 

But ` MessageRenderer#plain_text_without_markdown ` and `MessageRenderer#title` is not working on Message

jhass:
What do you mean not working? Note you need to call them on the return value of `Message#message`, not on `Message#text` nor `Message` itself.

app/views/conversations/_conversation.haml
```diff
@@ -32,7 +32,7 @@
         .last_message
           - if conversation.messages.present?
             %em
-              = conversation.messages.last.text
+              = Diaspora::MessageRenderer.new(conversation.messages.last.text).markdownified
```
SaiVardhan: `Conversation.find(2).messages.last.plain_text_without_markdown` if I use this it is throwing error.
and if I use ` Diaspora::MessageRenderer.new(Conversation.find(2).messages.last).plain_text_without_markdown` it is returning `#<Message:0x0000000ae7d3d0>`

jhass: `Conversation.find(2).messages.last.message.plain_text_without_markdown`.

To explain both of your mistakes:

* `conversation = Conversation.find(2) #=> <Conversation ...>` 
* `messages  = conversation.messages # => [<Message ....>, <Message ...>, ...]` (well, `AR::Relation` really, but conceptually.
* `message = messages.last # => <Message ...> `, note *not* `Diaspora::MessageRenderer`, `Message` does not respond to `plain_text_without_markdown`, only `Diaspora::MessageRenderer` does.
* `message.text #=> "foobar..."`
* `renderer = message.message # => <Diaspora::MessageRenderer ....>` same as `Diaspora::MessageRenderer.new(message.text)`
* Your second snippet effectively did `Diaspora::MessageRenderer.new(message)` and `Message#to_s` is `"<Message ....>"`.

spec/controllers/conversations_controller_spec.rb
```diff
@@ -102,6 +102,14 @@
       get :index, conversation_id: @conversations.first.id
       expect(assigns[:conversation]).to be_nil
     end
+
+    it "retrieves a conversation message with out markdown content " do
+      get :index
+      @conversation = @conversations.first
+      expect(response).to be_success
+      expect(response.body).to match(/cool stuff/)
+      expect(response.body).not_to match(/<strong>cool stuff<\/strong>/)
```
diaspora-code-review: Use `%r` around regular expression.

spec/controllers/conversations_controller_spec.rb
```diff
@@ -102,6 +102,14 @@
       get :index, conversation_id: @conversations.first.id
       expect(assigns[:conversation]).to be_nil
     end
+
+    it "retrieves a conversation message with out markdown content " do
+      get :index
+      @conversation = @conversations.first
+      expect(response).to be_success
+      expect(response.body).to match(%r{cool stuff})
```
diaspora-code-review: Use `//` around regular expression.

spec/controllers/conversations_controller_spec.rb
```diff
@@ -102,6 +102,14 @@
       get :index, conversation_id: @conversations.first.id
       expect(assigns[:conversation]).to be_nil
     end
+
+    it "retrieves a conversation message with out markdown content " do
+      get :index
+      @conversation = @conversations.first
+      expect(response).to be_success
+      expect(response.body).to match(%r/cool stuff/)
```
diaspora-code-review: `%r`-literals should be delimited by `{` and `}`.

spec/controllers/conversations_controller_spec.rb
```diff
@@ -102,6 +102,14 @@
...
+      expect(response.body).not_to match(%r/<strong>cool stuff<\/strong>/)
```
diaspora-code-review: `%r`-literals should be delimited by `{` and `}`.

denschub:
Just saw the messages by the review bot and they might require some clarification.

Valid regex syntaxes are `%r{foo}` and `/foo/`. The style bot suggested using different syntaxes to avoid escaping. usually, `/foo/` is our preferred form, but if we can avoid having to escape a slash, `%{foo}` is fine, too. If you are looking for `foo/bar`, you'd have to escape the slash in `/foo\/bar/`, but not in `%{foo/bar}`, so the latter is to be preferred.

Feel free to ping someone if you got confused!

spec/controllers/conversations_controller_spec.rb
denschub: See my comment. This should be `/cool stuff/`.

spec/controllers/conversations_controller_spec.rb
denschub: See my comment. This should be `%r{<strong>cool stuff</strong>}`. (Note the unescaped slash!)

SaiVardhan:
A little bit confused on what is preferred form for specifying Regex.

Please specify.

jhass:
You can run the linter locally with `bin/pronto run -c develop`, see https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review Just apply the specific suggestions until it's happy.

You can ignore the cucumber failures, unfortunately we have some that randomly fail on CI.

spec/controllers/conversations_controller_spec.rb
SaiVardhan: Please refer [Rubocop](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RegexpLiteral) as it suggest to use either `%r{}` or `//`

jhass: The rule is to use `//` unless the regex contains a `/`. Just follow the suggestions the review bot makes, it's not broken.

jhass:
Merged as 18dbf02181a93af011c74ea6319357bedd3e35ce

Thank you!

PR #6932, opened 2016-07-28
Save relayable signature data

SuperTux88:
To be forward-compatible with future additional fields on relayables, we need to store the order used for the signature and also store data that we don't know yet.

This PR also cleans up the signature of the relayables and only stores `author_signature` for the relayables where we own the parent. So the DB gets much smaller again. 😃 

Also the signatures are in its own tables now, so they are only loaded when needed (for federation) and not for display on the frontend.

I tested the migration with a backup of my mysql DB, postgres is only tested with a small DB (2 posts and 3 comments or so), so the syntax is ok, but maybe someone could test this on a bigger postgres DB?


jhass:
This feels a bit redundant, is using a polymorphic association that much slower? Also is normalizing out the order into its own model really worth it, is it faster, does it save that much space?

SuperTux88:
> This feels a bit redundant, is using a polymorphic association that much slower?

I didn't do this really because of the speed. But signatures are what make our tables that big now, and I didn't want to create another big table with all signatures in it. I discussed this already with @CSammy and we decided that it would be better to have 3 less big tables, instead of one big. Also with 3 tables it is possible to have constraints/foreign-keys on the DB side, which is not possible with polymorphic association.

But the tables are smaller now, because I only store signatures where the pod owns the parent. So on my test-DB they are mid size (50 - 150MB), but i tested this only with a half year old backup. On my production-DB they would already be bigger and on a big pod like joindiaspora or geraspora, which own many parents, they are even bigger. So it might be possible to put it together to one table, but this table would also grow with time and get bigger and bigger. And I think smaller tables are better maintainable over time.

> Also is normalizing out the order into its own model really worth it, is it faster, does it save that much space?

It only saves a little space, but the order would many times be the same.

But if we add a new feature with a new field on a relayable, it would be added to the `additional_data` first on old pods. And when this pod gets updated and the table of the relayable gets this new column, we should migrate all already received relayables with this feature from `additional_data` to this column (or we must always look at  `additional_data` if the value in the new column is `NULL` in the future). And I think this migration could be done faster, if we find all affected relayables by finding all `signature_orders` that contain this field, and then selecting the signatures with this ids, instead of selecting with `LIKE` directly on the bigger `_signatures` table.

Flaburgan:
If it's only about the migration speed during the update, I'not sure it's

CSammy:
@SuperTux88 and I discussed a few parts of this.

**Regarding the split of the signatures tables.**
We had exploding tables before. While all signature tables share the same basic layout, semantically they belong to different datasets. If the signature to a comment is needed, it is definitely not in the signatures for participations or likes to find. If one of these tables grows stupendous again, the other tables will not be affected. Basically, this is content-aware partitioning and makes the design ready for a future with lots and lots of signatures. (If all three tables grow equally in size, we have thrice the time until we have the need to come up with an alternative design as we would have with all signatures in one table.) We thought about documenting the fact that it is partitioning, but at the moment we do not know where. For the sake of code readability, we determined that renaming the tables to signature_x in order to make the partitioning clearer does not justify the inconvenient names in the code or added complexity.

**Regarding the normalization of orders.**
The disadvantage is clear - added complexity.
The advantages are:

1. In any join that is not able to fall back on the index for the order column, an integer comparison is vastly faster than comparing the entire string.
2. In any query searching for a field in the orders column, only an integer comparison is required after getting the correct values from the orders table. (Even a text index may be used, which otherwise would not be sensible because it would need maintenance in the signature tables with lots of rows.)
3. There will be a little bit of space saved due to not repeatedly saving the same order.

Of course, if you find these advantages will not outweigh the disadvantage, then it should be reconsidered.

One may call this premature optimization. However, I believe these changes have huge advantages with very little complexity tradeoff.

jhass:
Well, alright then let's do this.

Thank you!

PR #6934, opened 2016-07-31
better error message if we webfinger a person with invalid guid

SuperTux88:
this happens if a remote person changed the diaspora-id manually (renamed
domain or something) in their database.


jhass:
Thanks!

PR #6940, opened 2016-08-04
Cleanup posts table

SuperTux88:
I removed all unused columns in the `posts` table:
- some columns are the same as in the `photos` table, but they are always `NULL`. I really don't know why these columns existed.
- some columns were used for cubbi.es in the past, they actually contain data, but the content was migrated to the `text` column with the `MigrateActivityStreamToStatusMessage` already, so it's safe to remove them (but most of the images were hosted on cubbi.es anyway, so they are gone).
- `frame_name` was used for the beta publisher/single-post-view, but I think the content of this column is not useful anymore.


app/models/photo.rb
```diff
@@ -141,6 +141,6 @@ def self.visible(current_user, person, limit=:all, max_time=nil)
              else
                Photo.where(author_id: person.id, public: true)
              end
-    photos.order("created_at desc")
+    photos.where(pending: false).order("created_at DESC")
```
SuperTux88: I thought about leaving out the `.where(pending: false)` here. That would make the difference, that the user can see his own pending photos (and can delete them). Other users can't see pending photos anyway, because they are default private and have no visibilities.

jhass: I think that's a design decision that's out of scope here.

SuperTux88: That's why I didn't change it here, with this PR everything works as before. 😃

jhass:
Thank you!

Issue #6898, opened 2016-06-29
Officially support ruby 2.3


Flaburgan:
I open this issue to discuss about the support of ruby 2.3 by diaspora*.

Do we want to support it in diaspora 0.6?

If yes, do we want to support ruby 2.1 in 0.6? We usually support only two ruby version. However, diaspora 0.5 supports ruby 2.0 and 2.1, so if 0.6 supports 2.2 and 2.3, that means podmin will **have to** upgrade ruby in the same time than diaspora*. Previously, we supported a ruby version in two major versions of diaspora*.

tchncs:
Not sure if this is on-topic enough but i noticed that people solved memory leaks by switching to 2.3.
See: https://github.com/puma/puma/issues/342#issuecomment-207044837

cmrd-senya:
May that help for #6763 then?

Flaburgan:
I'll use 2.3 on d-fr and we'll see ;)

Flaburgan:
Fyi, both debian and Ubuntu jumped from ruby 2.1 to ruby 2.3 without including 2.2. From that point of view, it would be a good idea to support 2.1 and 2.3 for 0.6 instead of 2.1 and 2.3. I don't know about fedora, redhat cent os etc, if someone wants to check.

cmrd-senya:
Another option is to support 3 ruby versions for a short period of time. E.g. support 2.2 up to the first minor of 0.6.x branch. So 2.1, 2.2, 2.3 for 0.6.0.0 and 2.1, 2.3 for 0.6.1.0 and upper. What do you think?

I'm not too deep in the realease making business, but that my thoughts :)

SuperTux88:
@cmrd-senya: The problem with supporting 3 versions is, that we need to add 4 more jobs to travis, and I think we have already enough jobs. But we can replace the 2.2 jobs with 2.3 and support 2.1 and 2.3 officially, and 2.2 would probably work too, but is not supported officially (as 2.3 works now, but is not supported officially).

denschub:
We discussed this on IRC a few days ago, but I lost track of the decision. Urgh. So let's discuss here again.

Supporting three versions is not really an option because of the reasons outlined by @SuperTux88, but I feel fine with supporting 2.1 and 2.3. That is, replacing 2.2 by 2.3.

Any other opinions?

PR #6942, opened 2016-08-05
Recommend Ruby 2.3, keep Ruby 2.1 as supported

Flaburgan:
Close https://github.com/diaspora/diaspora/issues/6898


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: ruby
 
 rvm:
-  - 2.2
+  - 2.3
```
SuperTux88: You need to use `2.3.1` for travis, I had the same problem with the [diaspora_federation-gem](https://github.com/diaspora/diaspora_federation/blob/develop/.travis.yml#L4)

svbergerem:
Thanks!

PR #6943, opened 2016-08-07
Fix duplicate activity stream posts

### LINKED ISSUE #6937: Own posts are repeated in mobile /activity
Hi, a video says more but athousant words ;D

https://vimeo.com/177424102

As you can see some of my own posts apear multible times in my mobile /activity/-stream. I think it apears once per interaction, so every like creates an incarnation of the same post.

I am using geraspora so this is about the version 0.5.99.0-pc5e151ff

<sub>(sorry if my English isn't that good, I didn't install a spell checker yet ;D)</sub>

svbergerem:
Fixes #6937.


jhass:
Thank you!

PR #6543, opened 2015-11-13
Do not disable submit button if comment is empty in mobile view

manuelVo:
Fixes #5485

PR #6944, opened 2016-08-07
Bump to jQuery 2

ghost:
Since no blocking point was raised in #6557, I propose we switch to jQuery 2.


spec/javascripts/mobile/mobile_comments_spec.js
```diff
@@ -82,14 +82,8 @@ describe("Diaspora.Mobile.Comments", function(){
   describe("createComment", function () {
     beforeEach(function() {
       spec.loadFixture("aspects_index_mobile_post_with_comments");
-      var commentBoxHtml = spec.fixtureHtml("comments_mobile_commentbox");
       var link = $(".stream .comment-action").first();
       Diaspora.Mobile.Comments.showCommentBox(link);
-      jasmine.Ajax.requests.mostRecent().respondWith({
```
denschub: Why did you remove that test? Is loading the recent comments covered somewhere else?

AugierLe42e: It's not a test, it's enclosed in the `beforeEach()` function. This line [is buggy](https://travis-ci.org/diaspora/diaspora/jobs/150391203#L813) with jQuery 2 and since it is not relevant for the test, I thought it was not a big deal to remove it.

denschub: True, it's a shim for the most recent comments ajax request. I'm surprised we don't break any tests by removing the shim. @svbergerem?

AugierLe42e: In this case, no ajax request should be performed during the test. So I'm actually not quite sure to which ajax call this line was supposed to answer.

svbergerem: This has been added in #6543 but I also don't see why. We spy on `jQuery.ajax` and make sure that it hasn't been called so this change looks good to me.

denschub: Thanks

svbergerem: @AugierLe42e You might also want to remove https://github.com/diaspora/diaspora/blob/develop/spec/controllers/jasmine_fixtures/comments_spec.rb then.

AugierLe42e: Sure, yes. Thx ;)

svbergerem:
Thank you!

PR #6945, opened 2016-08-07
Change eslint indentation rule for var, let and const

svbergerem:
Before:

``` javascript
var a,
  b;
let c,
  d;
const e,
  f;
```

After:

``` javascript
var a,
    b;
let c,
    d;
const e,
      f;
```


jhass:
:+1:

PR #6947, opened 2016-08-08
Various frontend fixes for the getting started page

svbergerem:
1. The avatar of the inviter was missing. I also made the text above bold.
   missing avatar
2. The profile image has been resized without keeping the aspect ratio.
   profile image resize
3. The margin for tag selections was wrong.
   tag selection margins


jhass:
Thanks!

PR #6864, opened 2016-06-15
Refactor  aspects membership dropdown to use handlebars template

cmrd-senya:
Aspect membership dropdown is used at the contacts, notifications, getting started pages, people hovercards and at a profile page header.

Before there was the [aspect_membership_dropdown](https://github.com/diaspora/diaspora/blob/develop/app/views/aspect_memberships/_aspect_membership_dropdown.haml) template which was used to render the dropdown serverside. Each update to the dropdown required to reload the dropdown using the [controller](https://github.com/diaspora/diaspora/blob/develop/app/controllers/people_controller.rb#L159-L170). Also, there was an AspectMembership backbone view which didn't use templates but rather attached to the DOM generated by serverside rendering.

This PR adds a handlebars template for the aspect_membership_dropdown element and refactors the AspectMembership view for using the template.

I started this job in order to fix #6120. I found out then, that it's not strictly necessary to do all these refactorings to fix #6120, but it is still a nice thing to do.

Currently this PR contains changes for profile, contacts and getting started pages, and notifications page is unchanged. I felt, that finishing this getting complicated, so I felt I need to get some feedback to proceed, to be sure that I'm doing the right thing before diving into more complex and large changes. Here is the list of thing left to do to finish the PR IMO:
- [x] Fix up notifications page to use handlebars template as well
- [x] Fix up peoples' hovercards
- [x] search pages should use it also
- [x] people/:guid/contacts
- [x] Fix up jasmine tests for the AspectMembership view component (there are more async stuff now, so it's not too straightforward)
- [x] Write some tests for the new code
- [x] Wipe everything left from the serverside rendering.

Also, there is a component called [AspectDropdown view](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/aspects_dropdown_view.js) which previously was a base for the AspectMembership view. The AspectDropdown view is used in the publisher and also renders serverside. This PR makes AspectMembership independent of the AspectDropdown. Maybe it worth making AspectDropdown template based as well and basing AspectMembership on AspectDropdown again since they are definetely related. I can do that as well, but it's some extra work and I'd like to avoid it for a while if it could be separated from this job. However, if you say it's wanted, I could make it.

PR #6948, opened 2016-08-08
Fix other user's profile photos stream

svbergerem:
Fixes a regression from #6864.


svbergerem:
I added another regression fix for hovercards in the notification dropdown on pages where we didn't load the aspects before. Now we load the aspects on all pages.

jhass:
Thank you!

PR #6950, opened 2016-08-09
Consolidate queues

cmrd-senya:
Rebase of #6530. Rake task added.

Didn't test that too much though.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -59,6 +59,24 @@ wiki](https://wiki.diasporafoundation.org/Integration/Chat#Vines_to_Prosody)
 for more information on how to migrate to Prosody if you've been using Vines
 before.
 
+## Sidekiq queue changes
+
+We've decreased the amount of sidekiq queues from 13 to 5 in PR [#6950](https://github.com/diaspora/diaspora/pull/6950).
+The new queues are organized according to priority for the jobs they will process. When upgrading please make sure to
+empty the sidekiq queues before shutting down the server for an update.
+
+If you run your sidekiq with a custom queue configuration, please make sure to update that for the new queues.
+
+The new queues are: `urgent, high, medium, low, default`.
+
+When you upgrade to the new version, some jobs may persist in the old queues. To ensure that jobs to be processed, launch
+job processing for old queues using command:
+```
+rake migrations:run_legacy_queues
+```
+
+The command will report queues that still have jobs and launch sidekiq process for that queues.
+
 ## Refactor
 * Improve bookmarklet [#5904](https://github.com/diaspora/diaspora/pull/5904)
 * Update listen configuration to listen on unix sockets by default [#5974](https://github.com/diaspora/diaspora/pull/5974)
@@ -118,6 +136,7 @@ before.
 * Extract relayable signatures into their own tables [#6932](https://github.com/diaspora/diaspora/pull/6932)
 * Remove outdated columns from posts table [#6940](https://github.com/diaspora/diaspora/pull/6940)
 * Remove some unused routes [#6781](https://github.com/diaspora/diaspora/pull/6781)
+* Consolidate sidekiq queues [#6950](https://github.com/diaspora/diaspora/pull/6950)
 
 ## Bug fixes
 * Destroy Participation when removing interactions with a post [#5852](https://github.com/diaspora/diaspora/pull/5852)
--- a/app/workers/clean_cached_files.rb
+++ b/app/workers/clean_cached_files.rb
@@ -1,6 +1,6 @@
 module Workers
   class CleanCachedFiles < Base
-    sidekiq_options queue: :maintenance
+    sidekiq_options queue: :low
 
     def perform
       CarrierWave.clean_cached_files!
--- a/app/workers/deferred_dispatch.rb
+++ b/app/workers/deferred_dispatch.rb
@@ -4,7 +4,7 @@
 
 module Workers
   class DeferredDispatch < Base
-    sidekiq_options queue: :dispatch
+    sidekiq_options queue: :high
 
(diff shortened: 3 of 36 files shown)


lib/tasks/upgrade.rake
```diff
@@ -0,0 +1,17 @@
+namespace :upgrade do
+  desc "Run sidekiq with old queues so it can finish deferred jobs"
+  task :run_legacy_queues do
+    system "bundle exec sidekiq --daemon -q maintenance -q dispatch -q delete_account -q http http_service -q export"\
```
Flaburgan: `-q http http_service`
If `-q` is needed before each queue, you're missing one here.

lib/tasks/upgrade.rake
```diff
@@ -0,0 +1,17 @@
+namespace :upgrade do
+  desc "Run sidekiq with old queues so it can finish deferred jobs"
+  task :run_legacy_queues do
+    system "bundle exec sidekiq --daemon -q maintenance -q dispatch -q delete_account -q http -q http_service"\
+           " -q export -q photos -q socket_webfinger -q mail -q receive_local -q receive -q receive_salmon"\
```
SuperTux88: `receive_salmon`: the worker of this queue was deleted, so if there are still jobs on this queue, they need to be run with the old code before the upgrade. So I think you can remove this queue from the rake-task.

Changelog.md
```diff
@@ -59,6 +59,16 @@ wiki](https://wiki.diasporafoundation.org/Integration/Chat#Vines_to_Prosody)
 for more information on how to migrate to Prosody if you've been using Vines
 before.
 
+## Sidekiq queue changes
+
+We've decreased the amount of sidekiq queues from 13 to 5 in PR [#6530](https://github.com/diaspora/diaspora/pull/6530).
```
SuperTux88: I don't know if you should link to this PR now?

lib/tasks/upgrade.rake
jhass: Not sure if we should daemonize here at all, let's just run in foreground and tell people to run it for 5 minutes, then Ctrl+C

cmrd-senya: Is it possible that some jobs are delayed to, say, hour?

jhass: Oh, I guess so. I wonder if we force retry them all whether they'll end up on the new queues when they fail again?

lib/tasks/upgrade.rake
```diff
@@ -0,0 +1,30 @@
+namespace :upgrade do
+  LEGACY_QUEUES = %w(
+    maintenance dispatch delete_account http http_service export photos socket_webfinger mail receive_local receive
+  ).freeze
+
+  desc "Run sidekiq with old queues so it can finish deferred jobs"
+  task :run_legacy_queues do
+    queues_with_jobs = LEGACY_QUEUES.select {|queue| Sidekiq::Queue.new(queue).size > 0 }
```
diaspora-code-review: Use `!empty?` instead of `size > 0`.

cmrd-senya: Sorry, @diaspora-code-review, but

lib/tasks/upgrade.rake
cmrd-senya: I pushed updated version, it still daemonizes, but it can check whether queues are empty and run only queues with jobs. If we don't want daemonizing we could tell podmins to check whether there are non-empty queues and keep the legacy sidekiq process.

jhass:
Should we just reuse the `migration` namespace?

lib/tasks/upgrade.rake
cmrd-senya: My explanation was a bit misleading.

I meant, it checks queues only on start, not in process. So podmin may restart the task and check whether there are any queues left.

cmrd-senya:
> Should we just reuse the migration namespace?

The reason why I didn't choose it was that it is related to the database migrations primarily. If nobody minds, I can reuse it for this rake test.

lib/tasks/upgrade.rake
SuperTux88: I'm not sure if we need this rake-task anyway. It is useful for people running already on develop, but when people upgrading from 0.5.x to 0.6.0.0 want to run queued jobs, they need to start sidekiq again before git pull, wait, and kill after 5 minutes and then pull and do the rest of the upgrade (like on [0.4 -> 0.5, step 4](https://wiki.diasporafoundation.org/Updating#Updating_diaspora.2A_0.4_to_diaspora.2A_0.5)). But this rake-task is not useful for them.

jhass: Some people will forget or skip that step when doing the upgrade, or do it incorrectly. We can't downgrade, so being able to just tell them to run this tasks to recover as much as possible is extremely helpful.

jhass: Queues can be confirmed via the webinterface and the task can be run again at any point. I think run, wait 5 min, Ctrl+C is the much easier and straightforward UX flow here.

SuperTux88: > We can't downgrade, so being able to just tell them to run this task to recover as much as possible is extremely helpful.

But 95% of the queued jobs probably are the old send/receive jobs, which wouldn't work anymore. So we can also tell people to run `bundle exec sidekiq -q ...` to recover if they didn't do that before? But most people should follow the guide and run sidekiq before the pull.

But since this task still is useful for people currently running develop, we should probably keep it, but I'd keep it as simple as possible: start sidekiq and Ctrl-C.

cmrd-senya:
Updated, please review.

jhass:
Thank you!

Issue #6787, opened 2016-04-07
Search bar triggeres a a request to load all contacts


denschub:
Regression of #6293.

Due to [app/assets/javascripts/app/views/search_base_view.js:26](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/search_base_view.js#L26), a request to `/contacts.json` is made every time the search view gets initialized. Since this json contains literally all the users contacts, users with a large contact base can cause quite a huge amount of traffic and database load...

@svbergerem why is the prefetching needed here?

denschub:
I guess the request is made to fix #4789, but uh... This is really a bummer. :/

Flaburgan:
@denschub I guess as for https://github.com/diaspora/diaspora/issues/6107 we should first get only the user id, name and handle to be able to make the search, and then get all the info we need for the few contacts displayed.

svbergerem:
AFAIR we already fetched this before for mentions in the publisher and now we definitely fetch it for both mentions and the search bar. We could create a server side contacts search and use that for mentions. For the search bar we could improve the current search by also searching for contacts. Then we would also get rid of contacts.json there and we would really fix #4789. I already tried to implement the second fix but found out that one would need some experience with our database to fix that.

AugierLe42e:
As discussed on IRC, I don't think #6951 is a nice fix to this. Triggering a new query to the searver each time the user wants to mention someone has at least two flaws:

1. the mentionning system is going to get much slower than before #6951 and probably _really_ slow on pod with high traffics (JD, d-fr or Framasphere),
2. the mentionning system will fail if the server or client temporarily goes offline while the user is typing a message.

I really don't think the gain is worth it. The search bar and the mentionning dropdown should, at the very least, be fed with the contacts the user is sharing with (the mentionning dropdown actually doesn't need more). Gon, could be used to feed Bloodhound with the these contacts once and for all when the page is loaded.

Flaburgan:
I'm not sure about the perf here. I know my use case is not everybody's one, but I have thousands of contacts and each call to contacts.json take 6 seconds before the server's answer. So if we want to do the search client side, we need to keep the contact list in a cache, but we cannot continue to make a that intensive call to the server each time we refresh the page.

denschub:
I'm reopening this so we have something open as a reminder to think about this.

AugierLe42e:
[...]

PR #6951, opened 2016-08-09
Modify search to include contacts

### LINKED ISSUE #4789: Search by name doesn't work for all contacts
As discussed on [Loomio](https://www.loomio.org/d/KPmd4Nng/should-search-display-non-searchable-profiles-if-they-are-contacts), currently if a contact selects not to be searchable throughout the network in his privacy settings he will not show up when searching by name. Since we can see the names of our contacts, after adding them (or being added by them) we should be able to search for them by name.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1415887-search-by-name-doesn-t-work-for-all-contacts?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #4789 and #6787.

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -13,13 +13,6 @@ class ContactsController < ApplicationController
 
       # Used by the mobile site
       format.mobile { set_up_contacts_mobile }
-
-      # Used to populate mentions in the publisher
-      format.json {
-        aspect_ids = params[:aspect_ids] || current_user.aspects.map(&:id)
-        @people = Person.all_from_aspects(aspect_ids, current_user).for_json
-        render :json => @people.to_json
-      }
     end
   end
 
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -13,17 +13,12 @@ class ContactsController < ApplicationController
 
       # Used by the mobile site
       format.mobile { set_up_contacts_mobile }
-    end
-  end
 
-  def search
-    @people = Person.search(params[:q], current_user, only_contacts: true).limit(15)
-
-    respond_to do |format|
-      format.json do
-        @people = @people.limit(15)
+      # Used for mentions in the publisher
+      format.json {
+        @people = Person.search(params[:q], current_user, only_contacts: true).limit(15)
         render json: @people
-      end
+      }
     end
   end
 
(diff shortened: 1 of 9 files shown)


app/models/person.rb
```diff
@@ -155,6 +155,18 @@ def self.search(query, user)
     ).includes(:profile).order(search_order)
   end
 
+  def self.contacts_search(query, user)
+    return where("1 = 0") if query.to_s.blank? || query.to_s.length < 2
```
jhass: http://devdocs.io/rails~4.2/activerecord/querymethods#method-i-none

`query.blank?`, it's defined everywhere. Use `size` over `length`.

app/models/person.rb
```diff
@@ -155,6 +155,18 @@ def self.search(query, user)
...
+
+    sql, tokens = search_query_string(query)
+
+    Person.joins(:contacts)
+          .where(contacts: {user_id: user.id})
+          .where(sql, *tokens)
+          .includes(:profile)
+          .order(search_order)
```
jhass: So the only difference here to `search` is that it doesn't include people with `searchable: true`. I feel like this could be less redundant, perhaps just a keyword argument to `search`?

```rb
def self.search(query, user, only_contacts: false)
# ...
  query = Person
  query = query.searchable(user) unless only_contacts
  #....

File: app/controllers/contacts_controller.rb [hunk 3]
```diff
@@ -13,13 +13,17 @@ def index
...
 
-      # Used to populate mentions in the publisher
-      format.json {
-        aspect_ids = params[:aspect_ids] || current_user.aspects.map(&:id)
-        @people = Person.all_from_aspects(aspect_ids, current_user).for_json
-        render :json => @people.to_json
-      }
+    respond_to do |format|
+      format.json do
+        @people = @people.limit(15)
```
Flaburgan: Does the rails limit works like the SQL limit (meaning it only truncates result at the end but the load is still heavy?)

jhass: http://devdocs.io/rails~4.2/activerecord/querymethods#method-i-limit

jhass:
Thank you!

PR #6952, opened 2016-08-09
Add collection for notification

ghost:
This supersedes #6493, solves #5780 and partly solves #3890.

### DIFF
--- a/app/assets/javascripts/app/collections/notifications.js
+++ b/app/assets/javascripts/app/collections/notifications.js
@@ -1,0 +1,118 @@
+app.collections.Notifications = Backbone.Collection.extend({
+  model: app.models.Notification,
+  // URL parameter
+  /* eslint-disable camelcase */
+  url: Routes.notifications({per_page: 10, page: 1}),
+  /* eslint-enable camelcase */
+  page: 2,
+  perPage: 5,
+  unreadCount: 0,
+  unreadCountByType: {},
+  timeout: 300000, // 5 minutes
+
+  initialize: function() {
+    this.pollNotifications();
+
+    setTimeout(function() {
+      setInterval(this.pollNotifications.bind(this), this.timeout);
+    }.bind(this), this.timeout);
+
+    Diaspora.BrowserNotification.requestPermission();
+  },
+
+  pollNotifications: function() {
+    var unreadCountBefore = this.unreadCount;
+    this.fetch();
+
+    this.once("finishedLoading", function() {
+      if (unreadCountBefore < this.unreadCount) {
+        Diaspora.BrowserNotification.spawnNotification(
+          Diaspora.I18n.t("notifications.new_notifications", {count: this.unreadCount}));
+      }
+    }, this);
+  },
+
+  fetch: function(options) {
+    options = options || {};
+    options.remove = false;
+    options.merge = true;
+    options.parse = true;
+    Backbone.Collection.prototype.fetch.apply(this, [options]);
+  },
+
+  fetchMore: function() {
+    var hasMoreNotifications = (this.page * this.perPage) <= this.length;
+    // There are more notifications to load on the current page
+    if (hasMoreNotifications) {
+      this.page++;
+      // URL parameter
+      /* eslint-disable camelcase */
+      var route = Routes.notifications({per_page: this.perPage, page: this.page});
+      /* eslint-enable camelcase */
+      this.fetch({url: route, pushBack: true});
+    }
+  },
+
+  /**
+   * Adds new models to the collection at the end or at the beginning of the collection and
+   * then fires an event for each model of the collection. It will fire a different event
+   * based on whether the models were added at the end (typically when the scroll triggers to load more
+   * notifications) or at the beginning (new notifications have been added to the front of the list).
+   */
+  set: function(items, options) {
+    options = options || {};
+    options.at = options.pushBack ? this.length : 0;
+
+    // Retreive back the new created models
+    var models = [];
+    var accu = function(model) { models.push(model); };
+    this.on("add", accu);
+    Backbone.Collection.prototype.set.apply(this, [items, options]);
+    this.off("add", accu);
+
+    if (options.pushBack) {
+      models.forEach(function(model) { this.trigger("pushBack", model); }.bind(this));
+    } else {
+      // Fires events in the reverse order so that the first event is prepended in first position
+      models.reverse();
+      models.forEach(function(model) { this.trigger("pushFront", model); }.bind(this));
+    }
+    this.trigger("finishedLoading");
+  },
+
+  parse: function(response) {
+    this.unreadCount = response.unread_count;
+    this.unreadCountByType = response.unread_count_by_type;
+
+    return _.map(response.notification_list, function(item) {
+      /* eslint-disable new-cap */
+      var model = new this.model(item);
+      /* eslint-enable new-cap */
+      model.on("change:unread", this.onChangedUnreadStatus.bind(this));
+      return model;
+    }.bind(this));
+  },
+
+  setAllRead: function() {
+    this.forEach(function(model) { model.setRead(); });
+  },
+
+  setRead: function(guid) {
+    this.find(function(model) { return model.guid === guid; }).setRead();
+  },
+
+  setUnread: function(guid) {
+    this.find(function(model) { return model.guid === guid; }).setUnread();
+  },
+
+  onChangedUnreadStatus: function(model) {
+    if (model.get("unread") === true) {
+      this.unreadCount++;
+      this.unreadCountByType[model.get("type")]++;
+    } else {
+      this.unreadCount = Math.max(this.unreadCount - 1, 0);
+      this.unreadCountByType[model.get("type")] = Math.max(this.unreadCountByType[model.get("type")] - 1, 0);
+    }
+    this.trigger("update");
+  }
+});
--- a/app/assets/javascripts/app/views/header_view.js
+++ b/app/assets/javascripts/app/views/header_view.js
@@ -12,12 +12,12 @@ app.views.Header = app.views.Base.extend({
     });
   },
 
-  postRenderTemplate: function(){
-    new app.views.Notifications({ el: "#notification-dropdown" });
-    this.notificationDropdown = new app.views.NotificationDropdown({ el: "#notification-dropdown" });
-    new app.views.Search({ el: "#header-search-form" });
+  postRenderTemplate: function() {
+    new app.views.Notifications({el: "#notification-dropdown", collection: app.notificationsCollection});
+    new app.views.NotificationDropdown({el: "#notification-dropdown", collection: app.notificationsCollection});
+    new app.views.Search({el: "#header-search-form"});
   },
 
-  menuElement: function(){ return this.$("ul.dropdown"); },
+  menuElement: function() { return this.$("ul.dropdown"); }
 });
 // @license-end
--- a/app/assets/javascripts/app/views/notification_dropdown_view.js
+++ b/app/assets/javascripts/app/views/notification_dropdown_view.js
@@ -6,16 +6,21 @@ app.views.NotificationDropdown = app.views.Base.extend({
   },
 
   initialize: function(){
-    $(document.body).click($.proxy(this.hideDropdown, this));
+    $(document.body).click(this.hideDropdown.bind(this));
 
-    this.notifications = [];
-    this.perPage = 5;
-    this.hasMoreNotifs = true;
     this.badge = this.$el;
     this.dropdown = $("#notification-dropdown");
     this.dropdownNotifications = this.dropdown.find(".notifications");
     this.ajaxLoader = this.dropdown.find(".ajax-loader");
     this.perfectScrollbarInitialized = false;
+    this.dropdownNotifications.scroll(this.dropdownScroll.bind(this));
+    this.bindCollectionEvents();
+  },
+
+  bindCollectionEvents: function() {
+    this.collection.on("pushFront", this.onPushFront.bind(this));
+    this.collection.on("pushBack", this.onPushBack.bind(this));
+    this.collection.on("finishedLoading", this.finishLoading.bind(this));
   },
 
   toggleDropdown: function(evt){
@@ -31,12 +36,11 @@ app.views.NotificationDropdown = app.views.Base.extend({
   },
 
   showDropdown: function(){
-    this.resetParams();
     this.ajaxLoader.show();
     this.dropdown.addClass("dropdown-open");
     this.updateScrollbar();
     this.dropdownNotifications.addClass("loading");
-    this.getNotifications();
+    this.collection.fetch();
   },
 
   hideDropdown: function(evt){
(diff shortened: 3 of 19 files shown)


app/assets/javascripts/app/views/header_view.js
```diff
@@ -12,12 +12,18 @@ app.views.Header = app.views.Base.extend({
...
   },
 
-  postRenderTemplate: function(){
-    new app.views.Notifications({ el: "#notification-dropdown" });
-    this.notificationDropdown = new app.views.NotificationDropdown({ el: "#notification-dropdown" });
-    new app.views.Search({ el: "#header-search-form" });
+  postRenderTemplate: function() {
+    if (!app.collections.Notifications.instance) {
+      app.collections.Notifications.instance = new app.collections.Notifications();
+    }
```
AugierLe42e: I'd like an advice on this. The collection needs to be shared with [the notification view of the `/notifications` page](https://github.com/diaspora/diaspora/pull/6952/files#diff-829437a96c9bf5dc88c42fa29c90e469R137) so I create a class instance but I'm not really satisfied with this solution.

svbergerem:
```javascript
setupHeader: function() {
  if(app.currentUser.authenticated()) {
    app.notificationsCollection = new app.collections.Notifications();
    app.header = new app.views.Header();
    $("header").prepend(app.header.el);
    app.header.render();
  }
},
```
in [`app/assets/javascripts/app/app.js`](https://github.com/diaspora/diaspora/blob/fa06eba/app/assets/javascripts/app/app.js#L91-L97)?

AugierLe42e: Ah, yes! It's nicer. Thx.

AugierLe42e:
Ping?

AugierLe42e:
Ping.

spec/javascripts/app/models/notification_spec.js
```diff
@@ -0,0 +1,94 @@
...
+    beforeEach(function() {
+      this.target = new app.models.Notification({"reshared": {id: 16}, "type": "reshared"});
+      spyOn(jQuery, "ajax").and.callThrough();
+      spyOn(app.models.Notification.prototype, "set").and.callThrough();
+    });
+
+    it("calls calls ajax with correct parameters and sets 'unread' attribute", function() {
+      this.target.setUnreadStatus(true);
+      jasmine.Ajax.requests.mostRecent().respondWith({status: 200, responseText: '{"guid": 16, "unread": true}'});
+      var call = jQuery.ajax.calls.mostRecent();
```
svbergerem: Have you tried
```javascript
var call = jasmine.Ajax.requests.mostRecent();
```
?

Then you wouldn't have to spy on `jQuery.ajax`.

spec/javascripts/app/models/notification_spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+      expect(jQuery.ajax).toHaveBeenCalled();
+      expect(call.args[0].url).toBe("/notifications/16");
```
svbergerem: With `jasmine.Ajax.requests.mostRecent()` this would be
```javascript
expect(jasmine.Ajax.requests.mostRecent().url).toBe('/notifications/16')
```
or
```javascript
expect(call.url).toBe('/notifications/16')
```

app/assets/javascripts/app/collections/notifications.js
```diff
@@ -0,0 +1,109 @@
...
+    this.pollNotifications();
+    Diaspora.BrowserNotification.requestPermission();
+  },
+
+  pollNotifications: function() {
+    var unreadCountBefore = this.unreadCount;
+    this.fetch();
+    if (unreadCountBefore < this.unreadCount) {
+      Diaspora.BrowserNotification.spawnNotification(Diaspora.I18n.t("notifications.new_notifications"));
+    }
```
svbergerem: At least while testing this feature it didn't work because the check `unreadCountBefore < this.unreadCount` was done before `fetch` finished. You can fix this by listening to the `finishedLoading` event:

```diff
   pollNotifications: function() {
     var unreadCountBefore = this.unreadCount;
+    this.once("finishedLoading", function() {
+      if (unreadCountBefore < this.unreadCount) {
+        Diaspora.BrowserNotification.spawnNotification(Diaspora.I18n.t("notifications.new_notifications"));
+      }
+    }, this);
     this.fetch();
-    if (unreadCountBefore < this.unreadCount) {
-      Diaspora.BrowserNotification.spawnNotification(Diaspora.I18n.t("notifications.new_notifications"));
-    }
```

This would also show a browser notification immediately after loading the page when there are unread notifications. To fix this call `this.fetch()` in `initialize` and call `pollNotifications` the first time after some timeout.

app/assets/javascripts/app/collections/notifications.js
```diff
@@ -0,0 +1,109 @@
...
+  /* eslint-disable camelcase */
+  url: Routes.notifications({per_page: 10, page: 1}),
+  /* eslint-enable camelcase */
+  page: 2,
+  perPage: 5,
+  unreadCount: 0,
+  unreadCountByType: {},
+
+  initialize: function() {
+    this.pollNotifications();
```
svbergerem: In combination with the browser notification stuff I wrote a few lines below the new code could be

```javascript
this.fetch();
this.timeout = 300000 // 5 minutes
setTimeout(setInterval(this.pollNotifications.bind(this), this.timeout).bind(this), this.timeout);
Diaspora.BrowserNotification.requestPermission();
```

svbergerem (changes requested):

app/assets/javascripts/app/collections/notifications.js
AugierLe42e: It has been in production on diaspote for 3 months now with no problem. Maybe @Flaburgan can merge it on d-fr before it is merged on develop to see what happens?

app/assets/javascripts/app/collections/notifications.js
```diff
@@ -0,0 +1,109 @@
...
+  set: function(items, options) {
+    options = options || {};
+    options.at = options.pushBack ? this.length : 0;
+
+    // Retreive back the new created models
+    var models = [];
+    var accu = function(model) { models.push(model); };
+    this.on("add", accu);
+    Backbone.Collection.prototype.set.apply(this, [items, options]);
+    this.off("add", accu);
```
AugierLe42e: Hmm... Nope, I didn't. Thanks for highlighting that.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -1,94 +1,78 @@
...
-      element.text(function(i, text){
-        return parseInt(text) + change;
-      });
-    });
+  updateView: function() {
+    var notificationsContainer = $("#notifications_container");
 
-    [allNotes, typeNotes].forEach(function(badge) {
-      if(badge.text() > 0) {
+    var updateBadge = function(badge, count) {
```
AugierLe42e: What do you mean?

config/locales/javascript/javascript.en.yml
```diff
@@ -246,6 +246,7 @@ en:
     notifications:
       mark_read: "Mark read"
       mark_unread: "Mark unread"
+      new_notifications: "You have new notifications"
```
AugierLe42e: Hmm... There is a problem using "count" as a variable name for localization. I suppose probably because of [this line](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/helpers/i18n.js#L57). I throws `Error: Missing translation: notifications.new_notifications.other  i18n.js:48:15`. I used `number` instead.

app/assets/javascripts/app/views/notifications_view.js
svbergerem: You assigned the function to a variable inside oft a function. I suggested to make this a function of the view.

One might want to use the function later for something else and one would be able to test the function.

config/locales/javascript/javascript.en.yml
svbergerem: Thanks for pointing this out. Actually we should be using `count` with `one` -> "... notification" `other` -> "... notifications".

AugierLe42e: Oh, so that's what the error was about!

app/assets/javascripts/app/collections/notifications.js
Flaburgan: I can, even if d-fr is not a big pod like frama or jd so I would tend to pick the careful way here and choose 1 or 2 minutes at the moment.

app/assets/javascripts/helpers/browser_notification.js
```diff
@@ -0,0 +1,33 @@
+Diaspora.BrowserNotification = {
+  requestPermission: function() {
+    if ("Notification" in window) {
+      if (Notification.permission !== "granted" && Notification.permission !== "denied") {
+        Notification.requestPermission();
+      }
+    } else {
+      /* eslint-disable no-console */
+      console.warn("Browser do not support notifications");
+      /* eslint-enable no-console */
```
denschub: Yeah, I'd rather not have console logging. Believe it or not, there are browsers who don't even have the `console` object, so this would raise. Also, this is not a hard requirement that would break anything, so remove it.

AugierLe42e:
It's ok on my side.

spec/javascripts/app/views/notification_dropdown_view_spec.js
```diff
@@ -1,125 +1,51 @@
...
-      this.view.dropdownScroll();
-      expect(this.view.getNotifications).toHaveBeenCalled();
-    });
-
-    it("Doesn't call getNotifications if is not at the bottom", function(){
-      this.view.isBottom = function(){ return false; };
-      this.view.hasMoreNotifs = true;
-      spyOn(this.view, "getNotifications");
+  describe("dropdownScroll", function() {
+    it("Calls collection#fetchMore if is at the bottom and has more notifications to load", function() {
```
svbergerem: > Calls collection#fetchMore if **it** is at the bottom ~~and has more notifications to load~~

(missing "it" and the check for more notifications is done in `fetchMore`)

spec/javascripts/app/views/notification_dropdown_view_spec.js
```diff
@@ -1,125 +1,51 @@
...
       this.view.dropdownScroll();
-      expect(this.view.getNotifications).not.toHaveBeenCalled();
+      expect(this.collection.fetchMore).toHaveBeenCalled();
     });
 
-    it("Doesn't call getNotifications if is not at the bottom", function(){
-      this.view.isBottom = function(){ return true; };
-      this.view.hasMoreNotifs = false;
-      spyOn(this.view, "getNotifications");
+    it("Doesn't call collection#fetchMore if is not at the bottom", function() {
```
svbergerem: > Doesn't call collection#fetchMore if **it** is not at the bottom

spec/javascripts/app/views/notifications_view_spec.js
```diff
@@ -42,40 +23,38 @@ describe("app.views.Notifications", function(){
...
-        this.view.updateView(this.readN.data("guid"), this.readN.data("type"), false);
-        expect(this.readN.hasClass("read")).toBeTruthy();
-        expect(this.readN.hasClass("unread")).toBeFalsy();
-        expect(this.readN.find(".unread-toggle .entypo-eye").attr("data-original-title")).toBe(
-          Diaspora.I18n.t("notifications.mark_unread")
-        );
+      it("hides badge count when notification count is zero", function() {
+        Object.keys(this.collection.unreadCountByType).forEach(function(notificationType) {
+          this.collection.unreadCountByType[notificationType] = 0;
+          this.collection.unreadCount = 0;
```
svbergerem: This should be outside of the `forEach` loop.

spec/javascripts/app/views/notifications_view_spec.js
```diff
@@ -92,78 +72,65 @@ describe("app.views.Notifications", function(){
...
-          expect(parseInt(badge1.text(), 10)).toBe(count);
+          this.collection.unreadCount++;
+          this.view.updateView();
+          expect(parseInt(badge1.text(), 10)).toBe(this.collection.unreadCount);
 
-          this.view.updateView(this.guid, this.type, true);
-          expect(parseInt(badge2.text(), 10)).toBe(count + 1);
+          this.view.updateView();
+          expect(parseInt(badge2.text(), 10)).toBe(this.collection.unreadCount);
+        });
```
svbergerem: Before we checked both increase and decrease of the unread counts. Calling `updateView` once should also be sufficient for changes to `badge1` and `badge2`, so just do "increase", `updateView`, "check badge1 and badge2", "decrease", `updateView`, "check badge1 and badge2".

spec/javascripts/app/views/notifications_view_spec.js
```diff
@@ -1,36 +1,17 @@
...
-        this.readN = $(".stream-element.read").first();
-        this.guid = this.readN.data("guid");
-      });
-
-      it("calls 'setUnread'", function() {
-        spyOn(this.view, "setUnread");
-        this.readN.find(".unread-toggle").trigger("click");
-
-        expect(this.view.setUnread).toHaveBeenCalledWith(this.guid);
-      });
```
svbergerem: Am I missing something or are there now no tests for marking notifications as read/unread via click?

app/assets/javascripts/app/views/notification_dropdown_view.js
```diff
@@ -6,16 +6,20 @@ app.views.NotificationDropdown = app.views.Base.extend({
...
-    this.hasMoreNotifs = true;
     this.badge = this.$el;
     this.dropdown = $("#notification-dropdown");
     this.dropdownNotifications = this.dropdown.find(".notifications");
     this.ajaxLoader = this.dropdown.find(".ajax-loader");
     this.perfectScrollbarInitialized = false;
+    this.bindCollectionEvents();
+  },
+
+  bindCollectionEvents: function() {
```
svbergerem: Some tests would be great.

app/assets/javascripts/app/views/notifications_view.js
```diff
@@ -1,96 +1,86 @@
...
+    "click #mark-all-read-link": "markAllRead"
   },
 
   initialize: function() {
     $(".unread-toggle .entypo-eye").tooltip();
     app.helpers.timeago($(document));
+    this.bindCollectionEvents();
+  },
+
+  bindCollectionEvents: function() {
```
svbergerem: Some tests would be great.

svbergerem (changes requested):

AugierLe42e:
@svbergerem : I don't understand why [this test](https://github.com/diaspora/diaspora/pull/6952/files#diff-c9a79d54cae2e13d797e0bad9c8366beR14) and [this one](https://github.com/diaspora/diaspora/pull/6952/files#diff-53aee2c4464cab10d51364a1473d86a9R18) won't pass. According to [the test failure report](https://travis-ci.org/diaspora/diaspora/jobs/176813639#L791) the spied methods seem not mocked at all and called anyway. I use strickly the same method in [this one](https://github.com/diaspora/diaspora/pull/6952/files#diff-c9a79d54cae2e13d797e0bad9c8366beR32) with no problem :/

svbergerem:
@AugierLe42e [Here](https://github.com/diaspora/diaspora/pull/6952/files#diff-c9a79d54cae2e13d797e0bad9c8366beR10) you already bind the events and you bind them [here](https://github.com/diaspora/diaspora/pull/6952/files#diff-c9a79d54cae2e13d797e0bad9c8366beR19) a second time with new functions. Thus both old and new functions are called and calling the old ones makes the tests fail.

```diff
diff --git a/spec/javascripts/app/views/notification_dropdown_view_spec.js b/spec/javascripts/app/views/notification_dropdown_view_spec.js
index 3b01319..f47e945 100644
--- a/spec/javascripts/app/views/notification_dropdown_view_spec.js
+++ b/spec/javascripts/app/views/notification_dropdown_view_spec.js
@@ -12,6 +12,9 @@ describe("app.views.NotificationDropdown", function() {
 
   describe("bindCollectionEvents", function() {
     it("binds collection events", function() {
+      this.view.collection.off("pushFront");
+      this.view.collection.off("pushBack");
+      this.view.collection.off("finishedLoading");
       spyOn(this.view, "onPushFront");
       spyOn(this.view, "onPushBack");
       spyOn(this.view, "finishLoading");
@@ -20,11 +23,11 @@ describe("app.views.NotificationDropdown", function() {
 
       this.collection.trigger("pushFront");
       this.collection.trigger("pushBack");
-      this.collection.trigger("finishLoading");
+      this.collection.trigger("finishedLoading");
 
       expect(this.view.onPushFront).toHaveBeenCalled();
       expect(this.view.onPushBack).toHaveBeenCalled();
-      expect(this.view.pushBack).toHaveBeenCalled();
+      expect(this.view.finishLoading).toHaveBeenCalled();
     });
   });
```

You should be able to fix your second problem in a similar way.

AugierLe42e:
JS is so tricky sometimes. Thx, @svbergerem.

AugierLe42e:
Done.

svbergerem (approved):

svbergerem:
Thank you! :)

Flaburgan:
I just pulled this on d-fr, looks like it works nicely. I'll report here the load usage if I see some significant changes. If not, I'll try to set it to 1 minute instead of 5 and then test again. This setting could probably be added in the config file btw.

AugierLe42e:
@svbergerem: Thanks for your careful reviewing.

PR #6955, opened 2016-08-10
Warden must be reset after each example with it

cmrd-senya:
see https://github.com/hassox/warden/wiki/Testing#requirements

### DIFF
--- a/spec/integration/mentioning_spec.rb
+++ b/spec/integration/mentioning_spec.rb
@@ -28,6 +28,7 @@ end
 
 describe "mentioning", type: :request do
   include MentioningSpecHelpers
+  include RequestsHelper
 
   before do
     @user1 = FactoryGirl.create :user_with_aspect
--- a/spec/integration/profile_spec.rb
+++ b/spec/integration/profile_spec.rb
@@ -1,10 +1,9 @@
 require "spec_helper"
-require "requests_helper"
 
 describe PeopleController, type: :request do
   context "for the current user" do
     before do
-      login alice
+      sign_in alice
     end
 
     it "displays the publisher for user profile path" do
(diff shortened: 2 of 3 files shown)


jhass:
Can't we do away with this and use only devise's helpers? https://github.com/plataformatec/devise/blob/master/lib/devise/test/controller_helpers.rb

spec/integration/mentioning_spec.rb
```diff
@@ -28,6 +28,7 @@ def users_connected?(user1, user2)
 
 describe "mentioning", type: :request do
   include MentioningSpecHelpers
+  include RequestsHelper
```
SuperTux88: Is this the only place where this is used? Why do we need a Helper then, when it is only used once?

jhass:
Porting everything to https://github.com/plataformatec/devise#test-helpers would be great I think.

cmrd-senya:
@jhass, these links are for devise master branch. We use 3.5.6 and it doesnt have `Devise::Test::IntegrationHelpers`. Instead, the documentation [says](http://www.rubydoc.info/github/plataformatec/devise/#Test_helpers):

>These helpers are not going to work for integration tests driven by Capybara or Webrat. They are meant to be used with functional tests only. It is undesirable even to include Devise::TestHelpers during integration tests. Instead, fill in the form or explicitly set the user in session;

jhass:
4.2 has them too, https://github.com/plataformatec/devise/tree/245b1f9de0b3386b7913e14b60ea24f43b77feb0#test-helpers

We really need to upgrade and it looks like the blocker is gone since a couple weeks ago! https://github.com/baschtl/devise-token_authenticatable/releases Mind giving updating it a shot?

cmrd-senya:
> Mind giving updating it a shot?

Not at all :)

cmrd-senya:
Superseded by #6959.

PR #6959, opened 2016-08-10
Upgrade devise to 4.2

cmrd-senya:
- Bump devise, devise-token_authenticatable
- Update deprecated devise references
- Remove `requests_helper.rb`

Note also, that devise 4.2 consideres "name@host" to be correct email, while previously it was different. Tests were corrected accordinlgy.

I'm setting this to [WIP] because the PR lacks translation updates for devise -- I wasn't sure what to do with it.
[Here](https://gist.github.com/cmrd-senya/dbbfb20577ff0fdec4e22958a4d54689) is the list of conflicting keys between ours and upstream version of locale file. The format of the file is:

```
key:
- our version
- upstream version
```

Mostly it seems to me, that we should pick the upstream versions. I would probably leave `registartion.destroyed` and `registration.signed_up` in our versions, since it feel more relevant for diaspora. IMO, everything else can be updated. Opinions?


jhass:
> Mostly it seems to me, that we should pick the upstream versions. I would probably leave registartion.destroyed and registration.signed_up in our versions, since it feel more relevant for diaspora. IMO, everything else can be updated. Opinions?

Going over the list I came to the same conclusion.

cmrd-senya:
Ok, I updated the locale. The PR is ready.

jhass:
Thank you!

PR #6960, opened 2016-08-10
Escape HTML in mentions box

svbergerem:
We didn't escape that before when the user wrote HTML in the publisher. (only on develop)


jhass:
Thank you!

Issue #6956, opened 2016-08-10
/contacts.json should use cache


AugierLe42e:
I open this issue to continue the discussion started in [#6787 (comment)](https://github.com/diaspora/diaspora/issues/6787#issuecomment-238660093).

#6951 removed Bloodhound contacts prefetching to solve #6787. That is a flaw to me. As stated in the discussion, I find allowing the mentionning system to fail if the pod or the user temporalily goes offline is bad UX. To solve the problem of speed when the stream load, I propose to use cache both on server-side (to save a DB query) and client-side (using @Flaburgan's [idea of HTTP 304](https://github.com/diaspora/diaspora/issues/6787#issuecomment-238821943)).

denschub:
Eh, no?

If anything, we should think about how to make the queries faster and how to reduce the amount of data in it so make the response smaller, but caching is not a good solution at all.

Probably, we shouldn't even be using such a method. Sending all contacts to the client just in case it could need them is insane. Instead, we should find a way to to filtered searches, similar to the issue we have open in #6107. That is, rewriting the mention autocomplete to start looking for contact suggestions after three-ish characters have been typed, requesting and presenting only a filtered subset of all contacts and thus reducing both the time we spend in the presenter and with the request itself.

AugierLe42e:
> Sending all contacts to the client just in case it could need them is insane. Instead, we should find a way to to filtered searches, similar to the issue we have open in #6107.

*I.e*: requesting the server everytime the user wants to mention someone, which is wrong to me.

denschub:
Wait. #6951 actually introduced a partial contact search, so if you type `@aug`, it will make a request to `/contact.json?q=aug`. So what exactly is your issue?

> I find allowing the mentionning system to fail if the pod or the user temporalily goes offline is bad UX.

Is this the only thing? If the pod is temporarily offline, the user cannot publish a post. So what's the use case here?

> requesting the server everytime the user wants to mention someone, which is wrong to me. 

No, that's actually required since the contact list changes regularly and adding a person just to mention them right afterwards is not an usual use case, so hard caching the contact list is actually an user visible "bug". The contacts search endpoint is fast, so I have troubles understanding your motivation.

AugierLe42e:
> Is this the only thing? If the pod is temporarily offline, the user cannot publish a post. So what's the use case here?

The use case is user should not be prevented to *only write* it.

denschub:
[...]

PR #6970, opened 2016-08-11
[WIP] Use 304 on contacts.json

ghost:
Here is my first proposition for #6956. There are obviously a lot of things to improve, but I managed to get a proper 304 to work on `/contacts.json` without reverting #6951. FYI, while doing the tests, requests on `/contacts.json` drop from 34ms when answering 200 to 9ms when answering 304.


AugierLe42e:
I won't have time to finish this, and it seems like nobody really cares. So, just closing it.

PR #6972, opened 2016-08-11
Add rake task to move jobs from any legacy queue to the default queue

jhass:


cmrd-senya:
Wow, so smart and simple!

denschub:
Thanks.

Issue #5748, opened 2015-03-09
Empty mention list


davidbaumann:
Hello,

some functions like the stream in Diaspora take some seconds to load.
My mention list is empty, and it just looks like it was still loading to me.

Maybe an empty list should be like "You have not been mentioned yet" or something like that.

What do you think?

David.

Flaburgan:
Yeah, we should add a message when streams are empty, especially on the profile user, it looks really weird.

davidbaumann:
I had a look into the source - but I have no idea where to fix this :D

PR #6974, opened 2016-08-12
Add 'no posts yet' message to empty streams

### LINKED ISSUE #6852: on /tags pages when there is no asociated post found, there sould be a message stating that.

svbergerem:
Fixes #6852 and #5748.

stream
tags
profile

### DIFF
--- a/app/assets/javascripts/app/views/infinite_stream_view.js
+++ b/app/assets/javascripts/app/views/infinite_stream_view.js
@@ -16,6 +16,7 @@ app.views.InfScroll = app.views.Base.extend({
     this.showLoader();
     this.bind("loadMore", this.fetchAndshowLoader, this);
     this.stream.bind("fetched", this.finishedLoading, this);
+    this.stream.bind("allItemsLoaded", this.showNoPostsInfo, this);
     this.stream.bind("allItemsLoaded", this.unbindInfScroll, this);
 
     this.collection.bind("add", this.addPostView, this);
@@ -47,6 +48,13 @@ app.views.InfScroll = app.views.Base.extend({
         this.prependedPosts.insertBefore(el, this.prependedPosts.firstChild);
     } else {
         this.appendedPosts.appendChild(el);
+    }
+  },
+
+  showNoPostsInfo: function() {
+    if (this.postViews.length === 0) {
+      var noPostsInfo = new app.views.NoPostsInfo();
+      this.$el.append(noPostsInfo.render().el);
     }
   },
 
--- a/app/assets/javascripts/app/views/no_posts_info_view.js
+++ b/app/assets/javascripts/app/views/no_posts_info_view.js
@@ -1,0 +1,3 @@
+app.views.NoPostsInfo = app.views.Base.extend({
+  templateName: "no_posts_info"
+});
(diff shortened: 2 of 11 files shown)


jhass:
Thank you!

paskalito:
thanks a lot :)

just wanted to ask if we really don't want to add the "( for you to see on this pod)"  to the message, since it is more correct and hopefully helps people to understand how federation works. (and could point people to search elswheren and then maybe find what theyre looking for.)

svbergerem:
@paskalito "There are no posts yet. (for you to see on this pod)" sounds to me like the user won't be able to find posts on the pod. The message appears in the main stream, the tags stream, the mentions stream, the profile stream and so on.

Maybe "There are no posts in this stream yet."? "There are no posts for you in this stream yet."?

jhass:
"There are no posts to display here yet."?

paskalito:
### i think if possible there sould be diffrent messages for diffrent streams

Main stream > No Posts found! [as there is already a message with more explanation and this one is only to assure that there has been a search done]
Tags stream > No posts shared with you on this pod that contains #tag yet
Mentions stream > You haven't been mentioned yet
Profile stream > no posts shared with you yet
Aspects stream > No posts found! [as there is already a message with more explanation and this one is only to assure that there has been a search done]

on /contacts there is already a message

### And if there has to be one for all...
no posts for you found here  < that's my new favourite (but my brain is already going nuts)

There are no posts to display you here yet.?


### And maybe
A good soulution would als be a simple

**No Posts Found** message and then one or two lines below in grey and small.
(this can have the following resons)
- No one postet something with #tag
- Post is only shared in Aspects you're not a poart of
- the post has not been federated to this pod yet ( you can try searching on other pods)
- post may have been deleted


this way the message, "we have searched search is finished but theres is nothing here for you" would be communicatet pretty fast and clear. and there is for interested users some more information about why that is..

(also the "No post Found" could appear as simple centered text right in the place where the "searching" animation took place instead of a box where normally the search result would appear. would maybe be more accurate design language.)

svbergerem:
There is also the liked, commented and activity stream and maybe others I didn't think about yet. Thus I prefered the "one message fits them all" solution and used the message @jhass proposed. You can see the changes in https://github.com/diaspora/diaspora/commit/1d58df13e34ec37b6b2bf77301b45eee41873823.

PR #6975, opened 2016-08-12
[API] don't store ID tokens in DB

### LINKED ISSUE #6857: [API] Cannot revoke an authorized application
If I hit "Revoke" in /api/openid_connect/user_applications:

```
ActiveRecord::StatementInvalid (Mysql2::Error: Cannot delete or update a parent row: a foreign key constraint fails (`diaspora_development`.`id_tokens`, CONSTRAINT `fk_rails_a461603753` FOREIGN KEY (`authorization_id`) REFERENCES `authorizations` (`id`)): DELETE FROM `authorizations` WHERE `authorizations`.`id` = 1):
  app/controllers/api/openid_connect/authorizations_controller.rb:44:in `destroy'
```

cmrd-senya:
fix #6857

### DIFF
--- a/lib/api/openid_connect/id_token.rb
+++ b/lib/api/openid_connect/id_token.rb
@@ -25,15 +25,12 @@ require "uri"
 
 module Api
   module OpenidConnect
-    class IdToken < ActiveRecord::Base
-      belongs_to :authorization
-
-      before_validation :setup, on: :create
-
-      default_scope { where("expires_at >= ?", Time.zone.now.utc) }
-
-      def setup
-        self.expires_at = 30.minutes.from_now
+    class IdToken
+      def initialize(authorization, nonce)
+        @authorization = authorization
+        @nonce = nonce
+        @created_at = Time.current
+        @expires_at = 30.minutes.from_now
       end
 
       def to_jwt(options={})
@@ -41,6 +38,8 @@ module Api
           jwt.kid = :default
         end
       end
+
+      private
 
       def to_response_object(options={})
         OpenIDConnect::ResponseObject::IdToken.new(claims).tap do |id_token|
@@ -54,17 +53,17 @@ module Api
         @claims ||= {
           iss:       AppConfig.environment.url,
           sub:       sub,
-          aud:       authorization.o_auth_application.client_id,
-          exp:       expires_at.to_i,
-          iat:       created_at.to_i,
-          auth_time: authorization.user.current_sign_in_at.to_i,
-          nonce:     nonce,
+          aud:       @authorization.o_auth_application.client_id,
+          exp:       @expires_at.to_i,
+          iat:       @created_at.to_i,
+          auth_time: @authorization.user.current_sign_in_at.to_i,
+          nonce:     @nonce,
           acr:       0
         }
       end
 
       def build_sub
-        Api::OpenidConnect::SubjectIdentifierCreator.create(authorization)
+        Api::OpenidConnect::SubjectIdentifierCreator.create(@authorization)
       end
     end
   end
--- a/app/controllers/api/openid_connect/authorizations_controller.rb
+++ b/app/controllers/api/openid_connect/authorizations_controller.rb
@@ -53,7 +53,6 @@ module Api
       def reset_auth(auth)
         return unless auth
         auth.o_auth_access_tokens.destroy_all
-        auth.id_tokens.destroy_all
         auth.code_used = false
         auth.save
       end
--- a/app/models/api/openid_connect/authorization.rb
+++ b/app/models/api/openid_connect/authorization.rb
@@ -12,7 +12,6 @@ module Api
       serialize :scopes, JSON
 
       has_many :o_auth_access_tokens, dependent: :destroy
-      has_many :id_tokens, dependent: :destroy
 
       before_validation :setup, on: :create
 
@@ -50,7 +49,7 @@ module Api
       end
 
       def create_id_token
-        id_tokens.create!(nonce: nonce)
+        IdToken.new(self, nonce)
       end
 
       def self.find_by_client_id_user_and_scopes(client_id, user, scopes)
--- a/db/migrate/20160813115514_remove_id_tokens.rb
+++ b/db/migrate/20160813115514_remove_id_tokens.rb
@@ -1,0 +1,7 @@
+require_relative "20150714055110_create_id_tokens"
+
+class RemoveIdTokens < ActiveRecord::Migration
+  def change
+    revert CreateIdTokens
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160810230114) do
+ActiveRecord::Schema.define(version: 20160813115514) do
 
   create_table "account_deletions", force: :cascade do |t|
     t.string   "diaspora_handle", limit: 255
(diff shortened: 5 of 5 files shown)


spec/models/api/openid_connect/authorization_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  context "with expired id token" do
+    before do
+      auth.create_id_token.update(expires_at: 1.minute.ago)
+      auth.reload
+    end
+
+    describe "#destroy" do
+      it { expect { auth.destroy }.not_to raise_error }
+    end
```
jhass: Also check the id token is actually gone?

jhass:
Ok, I thought a bit about this and talked to a couple of people, we actually should get rid of the `default_scope` instead, see http://www.innovationontherun.com/activerecord-default_scope-is-an-anti-pattern/ for example.

Could you turn the `default_scope` into a named scope and use it in the appropriate places when fetching `IdToken`s? So something like

```rb
class IdToken < AR::Base
  scope :valid, -> { ... }
end

class Authorization < AR::Base
  def self.valid_id_tokens
    id_tokens.valid
  end
end
```

cmrd-senya:
Looks like the `id_tokens` is actually never referenced besides this deletion, and the scope is useless. My guess is that ID tokens are used in API calls and we don't have any implemented.

cmrd-senya:
Shall I make the `valid` scope for future uses, or just remove the `default_scope`?

jhass:
I can only find creations either:

* https://github.com/diaspora/diaspora/blob/develop/lib/api/openid_connect/authorization_point/endpoint_confirmation_point.rb#L66
* https://github.com/diaspora/diaspora/blob/develop/lib/api/openid_connect/token_endpoint.rb#L29
* https://github.com/diaspora/diaspora/blob/develop/app/controllers/api/openid_connect/authorizations_controller.rb#L56
* https://github.com/diaspora/diaspora/blob/develop/app/models/api/openid_connect/authorization.rb#L53

Reading through the spec I can't find the client ever submitting the ID token back to the OP, @KentShikama, do you remember why we have to store it at all?

For now adding the `valid` scope anyway (and the convenience accessor for the association in `Authorization`) should be good though.

KentShikama:
Ahh thanks for taking care of this @cmrd-senya . Haven't had the time to get to this.

KentShikama:
@jhass Um no…lol. As far as I remember, the ID token is for the client to figure out whether the user is who they say they are and is not used by the OP after they issue it.

I'm currently checking whether the OP might want to send over the same ID token…but I doubt it.

cmrd-senya:
Doesn't that mean then that we don't need to store ID tokens at all and can remove the table?

KentShikama:
Yeah can't see why we would need to send over the same ID token. 

At the very least, I know I didn't question the necessity of the table when I originally created it: I most likely assumed that if I created a model, I would need to create the corresponding migration + table.

jhass:
Well, it just shouldn't be a model then but simply a small helper class.

jhass:
Might want to look through how https://github.com/nov/openid_connect_sample/blob/master/app/models/id_token.rb is used over there though first, since this was likely inspired of that.

Tbh. wouldn't be surprised if it's for no reason over there too though, given what I could see of nov's coding style so far.

jhass:
I can find one usage https://github.com/nov/openid_connect_sample/blob/master/app/controllers/authorizations_controller.rb#L27

Are we missing this check?

KentShikama:
Hmm…can't remember off the top of my head but I definitely considered those couple lines of code.

cmrd-senya:
We don't have `call_authorization_endpoint` method in our `AuthorizationCotroller`

KentShikama:
Ok well the purpose of those few lines is to ensure that once the user is shown the "approve this application" dialog, they click the approve button within max_age seconds.

KentShikama:
Ahh wait nvm

KentShikama:
The purpose seems to be to ensure the user logged in within max_age before hitting that AuthorizationsController endpoint. I believe I have made the equivalent check here: https://github.com/diaspora/diaspora/blob/develop/app/controllers/api/openid_connect/authorizations_controller.rb#L26

KentShikama:
My only question now is why Nov digs into the ID token to retrieve the max_age

jhass:
Can it be provided by the RP perhaps?

KentShikama:
@jhass Yeah the max_age is optionally specified by the RP on the initial authentication request

jhass:
So you carry it on through a form parameter? If so the issue is  that the client can change it.

KentShikama:
@jhass Yeah that is why the whole "Passing Request Parameters as JWTs" section exists, so you can sign your initial authentication request.

KentShikama:
So I guess Nov's code would make sense if it said

@request_object.try(:max_age)

(aka it is sending the max_age through some sort of signed body) but at the point of the initial authorization request, the ID token isn't created so I'm not sure why it says 

@request_object.try(:id_token).try(:max_age)

KentShikama:
Nvm again. The ID token might already be created by that point.

jhass:
I would have to read on a few things to completely follow tbh, so what's your recommendation here?

cmrd-senya:
Does `@request_object.try(:id_token)` actually reference the ID token (model) from the database, or something else?

jhass:
Good catch, I guess it actually does not? https://github.com/nov/openid_connect_sample/blob/master/lib/authorization_endpoint.rb#L17

KentShikama:
Ok, so what Nov is doing is the following:

1. Takes in the max_age param from inside the request object
2. Creates an ID token and throws in the max_age param
3. Digs back into the ID token and fetches the max_age param
4. Checks whether the user has logged in before the max_age param

So the difference between my code and Nov's is that I'm checking the max_age before I create the ID token and Nov is checking after he creates the ID token.

jhass:
There's also this https://github.com/nov/openid_connect_sample/blob/master/app/models/id_token_request_object.rb#L5 but I can't find a usage of `IdTokenRequestObject`.

jhass:
So we don't need to store it?

KentShikama:
So we don't need to store it?

Yeah…Pardon for taking so long to figure out

jhass:
No worries, thanks for doing it at all.

cmrd-senya:
Shall I remove the `id_tokens` table than?

jhass:
Yes and move the class to `lib`.

cmrd-senya:
To be reveiwed

lib/api/openid_connect/id_token.rb
```diff
@@ -25,15 +25,15 @@
...
-
-      default_scope { where("expires_at >= ?", Time.zone.now.utc) }
-
-      def setup
-        self.expires_at = 30.minutes.from_now
+      def initialize(opts={})
+        @created_at = Time.current
+        @expires_at = 30.minutes.from_now
+        opts.each do |key, value|
+          send("#{key}=", value)
```
jhass: I think I'd prefer explicitness here
```rb
def initialize(authorization:, nonce:)
  @authorization = authorization
  @nonce = nonce
  @expires_at = 30.minutes.from_now
  @created_at = Time.now
end
```

lib/api/openid_connect/id_token.rb
```diff
@@ -25,15 +25,15 @@
 
 module Api
   module OpenidConnect
-    class IdToken < ActiveRecord::Base
-      belongs_to :authorization
+    class IdToken
+      attr_accessor :authorization, :expires_at, :created_at, :nonce
```
jhass: `attr_reader`s should be enough, right? If we need to expose any at all that is.

cmrd-senya:
updated

jhass:
Thank you!

PR #6976, opened 2016-08-13
Cleanup invitations

### LINKED ISSUE #5116: Drop legacy invitations
We currently employ a series of fallbacks to handle legacy invitations. We should drop these and only expect sign ups through invitation codes.

While touching that changing the `settings.invitations.open` setting to actually disable invites and not just hide the corresponding views from the user would be good.

There also is probably a bug where we reduce the number of available invites on failed sign ups, since we do that before validating the model: https://github.com/diaspora/diaspora/blob/develop/app/controllers/registrations_controller.rb#L13

<del>Finally a button in the admin interface to increase a users invitation codes available invites would be nice.</del>

Maybe we can convert invitation_tokens to invitation codes and redirect the route? Or maybe we don't care about invites sent out that long ago anymore and can just drop them?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3707208-drop-legacy-invitations?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SuperTux88:
I deleted all legacy invitations (I think nobody tries to use an invitation after more than 4 years) and removed all unused columns from the users table. I also removed the old code used for the legacy invitations.

Then I did some improvements to the current invite-system:
- Fixed a bug where the available invites counter was decreased when there were validation-errors on the registration.
- The counter gets no longer decreased when registrations are open, because otherwise it could go into negative, because there is no limit at 0 when registrations are open. Also removed the "xxx invites left on this code" message when registrations are open.
- Prevent using and sending invitation-codes when invitations and registrations are closed.

Fixes #5116


jhass:
Perhaps the migration should also check whether there are any negative counters already and reset them to 0?

Otherwise LGTM :+1:

SuperTux88:
I now reset negative counters to 0 if registrations are closed, or to the configured default if registrations are open (so when registrations close, the user still has all his invitations left).

jhass:
Thank you!

PR #6978, opened 2016-08-13
fix migration for invitation-cleanup

SuperTux88:
delete invitations-table (and constraints) before deleting the users.

mysql ignored the constraints 👎 so I didn't see this before, sorry about that ...


jhass:
Thanks!

PR #6980, opened 2016-08-14
Display markdown images inline

svbergerem:
```
[![](imgurl.jpg)](example.org)[![](imgurl.jpg)](example.org)[![](imgurl.jpg)](example.org) <- look at my cat, my cat is amazing
```

**Before**
before

**After**
after


cmrd-senya:
Good one. One had to use tables before to achieve the same.

jhass:
Thanks!

PR #6983, opened 2016-08-14
Fixes the avatar on mobile multiphoto page

ghost:
Before:

screen shot 2016-08-14 at 14 30 56

After:

screen shot 2016-08-14 at 14 31 20


svbergerem:
Thanks!

PR #6985, opened 2016-08-14
Do not display chat on small devices

ghost:
It happens to me a lot, on mobile to accidentally open the chat. I think it should not be displayed on small screens.

### DIFF
--- a/app/assets/stylesheets/chat.scss
+++ b/app/assets/stylesheets/chat.scss
@@ -10,5 +10,9 @@ body > .container-fluid.chat-roster-hidden {
 // scss-lint:disable IdSelector, SelectorFormat
 #jsxc_roster {
   top: $navbar-height;
+
+  @media(max-width: $screen-sm) {
+    display: none;
+  }
 }
 // scss-lint:enable IdSelector, SelectorFormat


Flaburgan:
It would be nice to not load it at all to limit js download and execution on mobile.

AugierLe42e:
I guess it's time to bring @Zauberstuhl to the conversation.

Zauberstuhl:
> It would be nice to not load it at all to limit js download and execution on mobile.

https://github.com/diaspora/diaspora/blob/develop/app/views/streams/main_stream.html.haml#L7

If you toggle mobile view the jsxc should be not included.

AugierLe42e:
> If you toggle the mobile view jsxc is not included.

No, he meant, on desktop view, but seen from a mobile.

denschub:
Sorry, I'm going to close this one. The chat is not included in our mobile view, and the non-mobile responsive view is called "Desktop version" for a reason. Excluding features in the mobile version is the reason people switch over to the desktop version, and if we exclude features there, people will yell at us.

We could discuss about making a less obtrusive version for smaller screen resolutions, for example just a little speech bubble in the corner, but removing it entirely is not an option for me, and nobody else merged this yet, so.. closing.

PR #6986, opened 2016-08-14
Cleanup old rake tasks

SuperTux88:
- `accounts:run_deletions`: was added with 0.4.0.0 two years ago for one-time usage.
- `db:rebuild`: `db:reset` does basically the same
- `db:integration:preprare`: the integration environments are not used.
- `db:reset`: there is a default `db:reset`, no need to write our own.
- `db:drop_integration`: the integration environments are not used.
- `db:fix_diaspora_handle`: really old migration from 2010
- `db:move_private_key`: also old migration from 2010
- `maintenance:clear_carrierwave_temp_uploads`: tmp/uploads doesn't exist anymore. And we have `CleanCachedFiles` as cronjob.
- `maintenance:install_logrotate_config`: diaspora has built-in logrotate support now, and people who want to use logrotate instead can write their own configs with the values they want.
- `migrations:copy_hidden_share_visibilities_to_users`: old migration from 2012
- `migrations:invitations`: legacy invitations were removed with #6976
- `migrations:absolutify_image_references`: old migration from 2010
- `migrations:rewire_uppercase_hashtags`: old migration from 2012
- `migrations:remove_uppercase_hashtags`: old migration from 2012


jhass:
Thanks!

PR #6603, opened 2015-12-28
Activate hovercards for logged out users

svbergerem:

PR #6987, opened 2016-08-14
Allow hovercards of remote people when not logged in

SuperTux88:
In #6603 hovercards were activated for logged out users, but they only
worked for local people. Hovercards on remote-people are important to
get the diaspora-id to search the person on the local pod, if you don't
know who it is.


jhass:
I'm not sure we should return the tags here.

denschub:
👍  for showing the tags publicly on remote hovercards. The information is already public (on the users public profile view) and I see no reason to add a check here. As discussed on IRC, I'll take the blame if someone has serious concerns here.

denschub:
Merged, thanks.

Issue #6575, opened 2015-12-01
Cucumber test fails in non english firefox useragent


MartinDelille:
I tried to run the cucumber test but it fails because my OS is in french. I tried to change my Firefox language (*about:config*, then changing *general.useragent.locale* to *en*). It change the language against when I connect to the site launched with `rails s` but not in the cucumber run.

Since *Capybara* looks for some button using their value, it fails.

jhass:
https://github.com/diaspora/diaspora/blob/develop/features/support/env.rb#L6 is our best effort. If it doesn't work, I don't know.

svbergerem:
@MartinDelille We switched from Selenium Webdriver to Poltergeist so now the cucumber tests run in PhantomJS instead of Firefox. Could you give this another try and tell us if the tests still fail?

MartinDelille:
Ok it cloned again, ran `rails s` but I have the following error:

```
$rails s
Rails is not currently installed on this system. To get the latest version, simply type:

    $ sudo gem install rails

You can then rerun your "rails" command.
```

Running `sudo gem install rails` doesn't fix the problem.

jhass:
Run `bundle` and then `bin/rails s`.

MartinDelille:
Do you want to document this for you?

jhass:
It's documented in our installation guides.

MartinDelille:
Where is it?

jhass:
In our wiki.

MartinDelille:
The thing is when I see a github project, I want to get the local installation instruction within the README, do you want to add it for you?

jhass:
They're too complex and OS specific and too different between development and production for the readme.

MartinDelille:
Anyway, is diaspora on Liberapay, I made a donation promise on it: https://liberapay.com/on/twitter/joindiaspora/ (but maybe it is not the good account?)

MartinDelille:
I can sum it up in a CONTRIBUTING.md file: I'm a documentation ninja

jhass:
I'm not aware we ever authorized this server, I think it created that without our consent. We use bountysource.com to gather bounties.

MartinDelille:
As you wish, I'm only on Liberapay.

jhass:
If you want to contribute some documentation, the OS X installation guides in the wiki could use some serious love and still need to be ported to the format and templating system used by the other guides.

MartinDelille:
Send me the link, I'll fix it

[...]

PR #6990, opened 2016-08-15
OSX environment setup

MartinDelille:
After discussion here: #6575

This instruction have been extracted from here: https://wiki.diasporafoundation.org/Installation/Mac_OS/Mavericks

Anyway, it isn't fully working:
- [ ] Make sure the installation instruction are correct
- [ ] Remove the use of 'sudo' by using brew ruby

### DIFF
--- a/OSX.md
+++ b/OSX.md
@@ -20,7 +20,7 @@ To install MySQL, run the following:
 
 Launch the service so it will start automatically:
 
-    brew services mysql start
+    brew services start mysql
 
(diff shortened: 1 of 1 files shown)


jhass:
No, I was talking about making the wiki page conform to our updated installation guides. We won't add documentation for this to the repo.

MartinDelille:
Sorry I don't work on external wiki because I think the documentation must stick with the code.

jhass:
Thanks for your effort then.

PR #6991, opened 2016-08-15
Fix aspect dropdown when person is being fetched

### LINKED ISSUE #6989: Aspect membership dropdown doesn't show on search if ID is being fetched
When you make a search query for a yet unknown diaspora ID, the `PeopleController` initiates WebFinger query in the background. In this case, the search result list is populated after the discovery has been finished. In this case aspect membership dropdown doesn't get populated properly.

regression from #6864

cmrd-senya:
fix #6989

### DIFF
--- a/spec/controllers/people_controller_spec.rb
+++ b/spec/controllers/people_controller_spec.rb
@@ -472,19 +472,20 @@ describe PeopleController, :type => :controller do
                      :profile => FactoryGirl.build(:profile, :first_name => "Evan", :last_name => "Korth"))
     end
 
-    describe 'via json' do
-      it 'returns a zero count when a search fails' do
-        get :refresh_search, :q => "weweweKorth", :format => 'json'
-        expect(response.body).to eq({:search_count=>0, :search_html=>""}.to_json)
+    describe "via json" do
+      it "returns no data when a search fails" do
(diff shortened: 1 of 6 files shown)


spec/controllers/people_controller_spec.rb
```diff
@@ -472,19 +472,20 @@
...
+        get :refresh_search, q: "Korth"
+        expect(response.body).to eq({search_html: "", contacts: nil}.to_json)
       end
-      it 'returns with a found name' do
-        get :refresh_search, :q => @korth.diaspora_handle
-        expect(JSON.parse( response.body )["search_count"]).to eq(1)
+
+      it "returns with a found name" do
+        get :refresh_search, q: @korth.diaspora_handle
+        expect(JSON.parse(response.body)["contacts"].count).to eq(1)
```
jhass: `size` instead of `count`. While `Array#count` is obviously optimized to just call `Array#size` if given no block, there's a difference in intent, which is "enumerate this collection" vs "just give me the size you know", see `Enumerable#count`.

cmrd-senya: thanks, updated

svbergerem:
Thanks. :)

PR #6994, opened 2016-08-16
Capitalisation of colour themes

goobertron:
Just noticed that the themes don't follow Diaspora's capitalisation guideline. Now fixed.

Please note: Homebrew is currently kaput on my machine (going to try to fix this tomorrow) so I can't test this PR. As it's only changing capitalisation of strings, I hope it won't break anything but I'd be grateful if you'd confirm this before merging.

ps: shouldn't these theme names be localised?

### DIFF
--- a/config/color_themes.yml
+++ b/config/color_themes.yml
@@ -1,6 +1,6 @@
 available:
-  original: "Original Dark"
-  original_white: "Original White Background"
-  dark_green: "Dark Green"
+  original: "Original dark"
+  original_white: "Original white background"
+  dark_green: "Dark green"
   magenta: "Magenta"
-  egyptian_blue: "Egyptian Blue"
+  egyptian_blue: "Egyptian blue"
--- a/config/initializers/color_themes.rb
+++ b/config/initializers/color_themes.rb
@@ -10,10 +10,10 @@ if color_themes_file.exist?
     if color_themes["available"].length > 0
       color_themes["available"]
     else
-      {"original" => "Original Dark"}
+      {"original" => "Original dark"}
     end
 else
-  AVAILABLE_COLOR_THEMES = {"original" => "Original Dark"}
+  AVAILABLE_COLOR_THEMES = {"original" => "Original dark"}
 end
 # Get all codes from available themes into a separate variable, so that they can be called easier.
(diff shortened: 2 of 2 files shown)


jhass:
I'm on the fence, I think we can argue for them to be proper nouns. Which then would also mean their current capitalization is correct. Not sure yet what I think about it, have to sleep over it.

goobertron:
Don't try to sleep on the fence, that'll be very uncomfortable!

I consider them headings on a par with 'My aspects', My activity, #Followed tags', and so on; all of which were changed to have only the first letter capitalised.

I think they could only be consider proper nouns (in English, at least; and editing English text is where I do have some expertise) if they were recognised colour names, e.g. 'Cerulean Blue' (although I wouldn't capitalise 'blue' in that in any case). 'Egyptian Blue' *might* fall into this category, but not 'dark green' and certainly not 'original white background' (which is a description rather than a name, and should probably have a colon after 'Original', as should 'Original: dark').

Anyway, I won't argue about it if the team decision is otherwise, but at the moment these strings are inconsistent with the rest of the UI (in English, at least) and I think they would sit more happily in the UI if they were capitalised like the rest.

Sleep well!

goobertron:
If it's agreed that these theme names should be i18n-able, whether or not we change the capitalisation, I think I should be able to make that change. Let me know tomorrow morning if you want this included in 0.6.0.0 and I'll try to do it before end of business tomorrow.

goobertron:
Erk, didn't mean to close.

goobertron:
Ah, after looking at the code it's a little more complex than I expected and I doubt I'll be able to fix it properly tomorrow. I'll have a look at it properly tomorrow, but I suggest we leave that work for the next minor version, as it's a refinement to an existing feature which works fine as it is.

denschub:
Removing the milestone assignment on that one. @goobertron, feel free to ping whenever you are done. ;)

goobertron:
I'm afraid I'm not going to be able to make these localisable, certainly not for the time being. If it's worth merging this just to fix the capitalisation for the time being, please do, otherwise please add the 'orphan' label or close it. Thanks.

SuperTux88:
Superseded by #7369

PR #6995, opened 2016-08-16
only create mentions after creating statusmessages

SuperTux88:
otherwise it will try to fetch a mentioned person from an unhealthy pod every time the post loads, if it is the only mention in this post.

(@cmrd-senya: this will conflict with #6818, but you can simply overwrite this change there, because your refactoring already fixed this, this is only a quickfix now)


jhass:
Thanks!

PR #7000, opened 2016-08-18
Fix backbone router for URLs with trailing slashes

svbergerem:
https://stackoverflow.com/questions/9600390/backbone-js-route-optional-parameter/14329976#14329976


jhass:
Thank you!

PR #7001, opened 2016-08-18
Fix conversations autoSuggest showing non-mutual contacts

### LINKED ISSUE #6920: Able to send PM to someone who is sharing with me and who I'm not sharing with
I'm rising this issue because I've discovered while working on #6917 that the system accepts that a user A sends a private message to user B that is sharing with A even though A is not sharing with B. This is strange because when A go to the profile page of B, no private message icon is displayed, so I'm not sure where the bug is. 

#6917 uses [SearchBase](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/search_base_view.js) for the contacts autcomplete which only displays contacts you are sharing with (based on `/contacts.json`) while current PM dropdown uses [this method](https://github.com/diaspora/diaspora/blob/develop/app/controllers/conversations_controller.rb#L106) that also displays contacts sharing with you that you are not sharing with.

Should a user be able to send PM to contacts who are sharing with him and who he is not sharing with ?

svbergerem:
Fixes #6920.

### DIFF
--- a/spec/models/contact_spec.rb
+++ b/spec/models/contact_spec.rb
@@ -85,10 +85,15 @@ describe Contact, type: :model do
       it "returns contacts with sharing true" do
         expect {
           alice.contacts.create!(sharing: true, person: FactoryGirl.create(:person))
-          alice.contacts.create!(sharing: false, person: FactoryGirl.create(:person))
         }.to change {
           Contact.sharing.count
         }.by(1)
+
+        expect {
+          alice.contacts.create!(sharing: false, person: FactoryGirl.create(:person))
+        }.to change {
+          Contact.sharing.count
+        }.by(0)
       end
     end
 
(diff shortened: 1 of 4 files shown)


spec/models/contact_spec.rb
```diff
@@ -103,6 +103,19 @@
       end
     end
 
+    describe "mutual" do
+      it "returns contacts with sharing true and receiving true" do
+        expect {
+          alice.contacts.create!(receiving: true, sharing: true, person: FactoryGirl.build(:person))
```
Flaburgan: You could maybe test count == 1 right after this line, to be sure that mutual == sharing && receiving, and then add the other contacts and check again that the count didn't change.

svbergerem: Sounds good. Thank you.

jhass:
Thank you!

PR #7009, opened 2016-08-20
Remove started sharing notifications where the contact doesn't exist

SuperTux88:
Before #6873 we deleted contacts when someone blocks a person, but we didn't drop the notification for the started sharing event. In #6864 we try to get the contact for the notification, which is not there anymore.

So we need to remove the notifications for the contacts that don't exit anymore.


denschub:
Thaaank ya.

PR #7010, opened 2016-08-21
Use https for the pod_uri if require_ssl is true

SuperTux88:
inspired by https://nerdpol.ch/posts/19134cc049470134b634782bcb452bd5


denschub:
Thanks.

PR #7011, opened 2016-08-21
Use Person.by_account_identifier to find the author

SuperTux88:
This is needed if the author is not lowercase in the received message (hubzilla?). But relay will still fail, because we'll resend the message with a lowercase author, which makes the author_signature invalid. But at least receive will work with this.


denschub:
Thanks much!

PR #7012, opened 2016-08-21
Fix stream shortcuts on small screens

svbergerem:
This bug has been reported on https://nerdpol.ch and the test I added fails with the current code on the develop branch. The reason for this is that [`offsetTop`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetTop)

> returns the distance of the current element relative to the top of the offsetParent node.

This is fine on big screens since the offsetParent node is `div.col-md-9` which is at the top of the page. On small screens the node is below the sidebar which led to this bug.


denschub:
Thanks for the fix. ❤️

PR #7013, opened 2016-08-21
Move notifications and conversations links in collapsed header on small devices

svbergerem:
On devices smaller than 768px (phones) the links will be in the header dropdown, between 768px and 992px (tablets) they will be in the header but won't open the notification dropdown (that's the current behavior) and for a width >=992px you will have the normal header.

**Before (phones)**
phone_before

**After (phones)**
phone_after
phone_after_expanded

Tablet and desktop versions didn't change.


denschub:
Sure, thanks, merged!

PR #7014, opened 2016-08-21
Remove stream title from main stream

svbergerem:
I have been made aware of the fact that the stream title just tells you what you can already see in the left sidebar. For the multi stream you see the word "Stream" three times: in the header, in the sidebar and as the stream title. When selecting the aspects stream the stream title needs 3 lines for the account on my production pod.

This PR keeps the title for the tag stream (`/tag/diaspora`) since there is no stream selection in the left sidebar on that page and it tells you on which tag page you are right now.

**Before**
before_aspects_stream
before_stream

**After**
after_aspects_stream
after_stream

### DIFF
--- a/app/assets/javascripts/app/app.js
+++ b/app/assets/javascripts/app/app.js
@@ -108,12 +108,6 @@ var app = {
 
       evt.preventDefault();
       var link = $(this);
-      if(link.data("stream-title") && link.data("stream-title").length) {
-        $(".stream-title").text(link.data("stream-title"));
-      } else {
-        $(".stream-title").text(link.text());
-      }
-
       $("html, body").animate({scrollTop: 0});
 
       // app.router.navigate doesn't tell us if it changed the page,
--- a/app/assets/javascripts/app/views/aspects_list_view.js
+++ b/app/assets/javascripts/app/views/aspects_list_view.js
@@ -15,7 +15,6 @@ app.views.AspectsList = app.views.Base.extend({
 
   initialize: function() {
     this.collection.on("change", this.toggleSelector, this);
-    this.collection.on("change", this.updateStreamTitle, this);
     this.collection.on("aspectStreamFetched", this.updateAspectList, this);
     app.events.on("aspect:create", function(id) { window.location = "/contacts?a_id=" + id });
   },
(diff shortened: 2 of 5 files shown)


goobertron:
Great! I've been of the same mind as you.

Regarding the removal of aspect names when viewing all aspects, this is something I really wanted to change as well. However, I think a better solution would be to use the stream title 'All aspects', just to make clear that this is different from the main stream. It's possible someone might view this stream and then forget that they're not on the main stream, and not notice which item is selected in the sidebar menu. I think a visual reminder at the top of the stream is worthwhile.

svbergerem:
Did I get your proposal right that you'd still have a list of aspects once you deselect one of them? You'd also have the word "Stream" three times on this small amount of space while you can only see the stream title when you are at the top of your stream and then the sidebar tells you which stream you selected.

denschub:
I agree. Merged. ;)

goobertron:
> Did I get your proposal right that you'd still have a list of aspects once you deselect one of them?

Actually I thought that was your proposal; I thought you were only removing the aspect names when the stream contains all aspects.

> You'd also have the word "Stream" three times on this small amount of space 

No; I agree with removing the word 'Stream' when viewing the main content stream.

Perhaps I have misunderstood your proposal. I thought it was only (as per screenshots) removing the title 'Stream' when viewing the main content stream, and the list of aspects when view *all* aspects. If, as looks the case now I review the code – although I really can't be sure, as I don't understand it very well! – you are removing all stream titles (with the exception of individual tag streams, for the reason you gave above), then that sounds sensible, as no one would ever expect to see a title above the content stream. Apologies if I misunderstood.

svbergerem:
> If, as looks the case now I review the code [...] you are removing all stream titles (with the exception of individual tag streams, for the reason you gave above)

This is exactly what this PR did. Since you are the second one who wasn't sure about what this PR does (another one asked me on IRC) I guess my description just wasn't clear enough.

goobertron:
Ok, that sounds like a good change, then. Apologies for pushing my nose in, and thanks for your work on the code!

svbergerem:
There is no reason to apologize. Thank you for following the latest changes and giving feedback! :)

denschub:
❤️

goobertron:
<3 to all of you!

PR #7016, opened 2016-08-21
Fix misplaced publisher prefill on tags page

### LINKED ISSUE #7015: ui: strange duplication of tag on tag search
when i search for an "nonexisting" tag.

bildschirmfoto vom 2016-08-21 13-44-22

bildschirmfoto vom 2016-08-21 13-44-01

(the lower (white one) seems to be wrong)

svbergerem:
Fixes #7015.


denschub:
Thanks, merged!

PR #6998, opened 2016-08-17
6840 : meta tags update

dhovart:
Fixes #6840
- Add opengraph and general meta tags to posts, profile and tags views.
- Added a few methods to retrieve the meta data to the existing Person/Post presenter and created a TagPresenter (would like to get some feedback on my approach: I hope it makes sense for you to have a presenter  for a Stream::Tag, just the same as we have presenters for ActiveRecord models).
- Introduced a MetaDatahelper
- Moved the hardcoded contents already set for meta tags to config/defaults.yml

Hope I'm following your styleguides correctly and that my tests are good enough!

PR #7018, opened 2016-08-21
Fix regressions from #6998

SuperTux88:
- fixed old reshares of reshares
- fixed deleted root of a reshare


denschub:
Thanks for adding the tests!

PR #7024, opened 2016-08-22
Display images inline in profile bio

### LINKED ISSUE #7020: Markdown images in Bio on the profile are displayed as block elements
Similar to #6980. This might also be a problem in conversations which I haven't checked yet.

svbergerem:
Fixes #7020. We don't have the problem on conversation pages since we still use redcarpet on those.


denschub:
Merged as 5b5e3e2c2902307e2f58042dcb5c3e59fba2b07c, thanks.

PR #7025, opened 2016-08-22
Remove started sharing notifications where the contact doesn't exist

SuperTux88:
The same as #7009 but now all of them 😼 

Why did nobody tell me that I did it wrong? 🙀


denschub:
Merged as 6d8b7b49cbd77cf6e1773288fab3eac1956387ce.

> Why did nobody tell me that I did it wrong? 🙀

🙈

PR #7030, opened 2016-08-24
Only update interacted_at on new comments

SuperTux88:
I got feedback from many users, that they don't understand how the "my activity" stream is sorted, because they have posts on the top, but don't see why (including myself 😼). The user doesn't see if a like was added, and it is also not imported for the user to see the post again then. And we have notifications if someone likes a users own posts, so no need to move it on top of the "my activity" stream here too.

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -38,8 +38,9 @@ class Comment < ActiveRecord::Base
     self.text.strip! unless self.text.nil?
   end
 
-  after_commit :on => :create do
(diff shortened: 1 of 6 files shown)


denschub:
Postponing the merge of this after we've released 0.6.0.0 to keep the amount of possible conflicts with backporting freeze fixes down to a minimum.

denschub:
As promised, merged as 4fea926057aa291b25a9739dcd33d344f9d1de3a.

PR #7036, opened 2016-08-28
Fix links of more-button on mobile streams

### LINKED ISSUE #6999: More-button of mobile views' public activities stream points to my activities stream
When I'm on mobile view at /public, scroll to the bottom of the stream and click the "more"-button, I end up at /activity?max_time=XXXXXXXX instead of /public?max_time=XXXXXXXX

Is this intended?

SuperTux88:
Fixes #6999 (this only fixes the bug of #6999 and not #5037, which is an upstream bug)

### DIFF
--- a/app/helpers/stream_helper.rb
+++ b/app/helpers/stream_helper.rb
@@ -8,16 +8,8 @@ module StreamHelper
       tag_path(:name => @stream.tag_name, :max_time => time_for_scroll(@stream))
     elsif controller.instance_of?(PeopleController)
       local_or_remote_person_path(@person, :max_time => time_for_scroll(@stream))
-    elsif controller.instance_of?(PostsController)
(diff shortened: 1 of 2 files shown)


app/helpers/stream_helper.rb
```diff
@@ -29,6 +21,24 @@ def reshare?(post)
 
   private
 
+  def next_stream_path
+    if current_page?(:stream)
+      stream_path(max_time: time_for_scroll(@stream))
```
diaspora-code-review: - Assignment Branch Condition size for next_stream_path is too high. [21.84/20]
- Cyclomatic complexity for next_stream_path is too high. [7/6]
- Perceived complexity for next_stream_path is too high. [8/7]

denschub:
Merged as 8c58206e63507c9152b91b2e431a9a44e01080ce.

PR #7037, opened 2016-08-29
[WIP] Add federation behavior tests

### LINKED ISSUE #6479: Implement support for federation testing in the Diaspora automated testing system

cmrd-senya:
### What?

A change set that intoduces some real federation behaviour tests. This is a merge of code from my [diaspora_federation_behavior](https://github.com/cmrd-senya/diaspora_federation_behavior) project.

fix #6479
### Why?

One may think that the huge refactoring of federation is over. However there're going to be even more changes. Account migration, comment editing, groups/rsvp are all demanded features that require major federation protocol-level improvements. This is an attempt to create a tool that allows to test federation features of pods in a reproducible way.
### Features
- Brings up LXC containers, deploys diaspora inside and runs tests controlling the diaspora pods from client side
- Usage of rspec metadata to group different test examples forming specific scenarios. In theory, this could make it possible to create separates subsets, e.g. for regression testing, loading test, etc
- Pod count in the test farm, git revisions of every pod in the testfarm are configurable parameters
- Imitates transactional test with rspec by rolling back databases of each pod after each test
- Loads fixtures generated by FixtureBuilder to populate pods with some initial data (well known bob and his friends alice and eve, and others)
### Prerequisites
- LXC
- Vagrant -- this one isn't provided as a rubygem anymore, so one must install it explicitly
- About 3 Gb free space

On Ubuntu 16.04 LXC can be installed by

```
sudo apt-get install lxc cgroup-lite
```

Vagrant must be installed from the upstream package ([download page](https://www.vagrantup.com/downloads.html)), the version from repositories is buggy.
#### Fetch git submodules

```
git submodule update --init --recursive
```
#### Prepare vagrant

Install required vagrant plugins by

```
bin/rake replica:install_vagrant_requirements
```

Vagrant must be allowed to execute containers without asking password (must be run without `sudo`)

```
vagrant lxc sudoers
```

Add these lines to your `/etc/hosts`

```
192.168.11.6    pod1.diaspora.local
192.168.11.8    pod2.diaspora.local
192.168.11.10    pod3.diaspora.local
```

I also had to issue this command on Ubuntu 16.04. Not sure if it's really mandatory or just a requirement due to some my local misconfiguration.

```
sudo mount cgroup /sys/fs/cgroup//devices -t cgroup -o rw,relatime,devices,release_agent=/run/cgmanager/agents/cgm-release-agent.devices
```
#### Install bundle for capistrano

Within the diaspora folder run

```
cd replica/capistrano
bundle install
```
#### Run

As simple as

```
bin/rake replica:execute_tests["-t ~known_federation_bugs:true"]
```

It takes a while. Current settings have 3 pods. This requires 3 containers to be brought up, and a diaspora application along with the full gem bundle. So the first run may took **more than two hours**. Further launches will be faster (10-15 minutes) because it'll reuse the same gemset and will update bundle only if changed.
#### Clean

Remove the VMs

```
bin/rake replica:clean
```
### TODO
- [ ] Bring log file generation into order
- [ ] Push changes of diaspora-replica and puppet-diaspora to upstreams
- [ ] Think of the ways of reducing time/storage usage.
- [ ] Set CI system for this test system.


jhass:
This looks awesome indeed, I'm just not sure I'm following on the benefit of it living inside the main repository. We won't be running it on Travis and if we ever get a decent enough CI to run it regularly, we can just fetch it from a separate repository. So is there anything I miss that makes it beneficial for it to live here? WDYT of moving/creating a new repo under the `diaspora/` organization for it?

cmrd-senya:
> So is there anything I miss that makes it beneficial for it to live here?

It makes it easier to use. If you developing some feature, you may be too lazy to go outside your repository to clone another one and to set everything  just to run tests locally. If the required code is in the same repository, it takes less actions to run the tests. Also, the  structure of the test system is so, that it relies on the diaspora git repository copy. It uses diaspora sources to deploy them on pods. Having the replica code inside makes the directory layout fixed. The replica project code can then make an assumption that sources of diaspora is at `..` and don't require to set them explicitly.

Also, we're used to write tests for features at the same time as the features. If I implement something, I supply tests, so that a reviewer may check the implementation using the tests as well. If I have to provide tests for the features in one repository and tests in different, it makes logistics too complicated.  What do you push first? If you do PR tests first, they won't work okay unless the respective feature code is in its place. If you do PR for a feature first, the tests are not that useful. So if one develops a feature he must provide two PRs in two repositories and sync them all the time, which is unlikely going to be done. Just having two repositories result in a way too complex logistics.

As we don't have a separate CI for it now the only option is to run it locally. And if it is too complicated nobody will care to ever use it.

@denschub told me before that it is theoretically possible to run vagrant and containers inside travis. However, still it is heavy, and even if we manage to, we would need some extra jobs so that CI doesn't stuck everytime running these tests.

On the other hand, I don't like pushing not very related stuff to Gemfile of diaspora and also code in this  PR has no common points with the code of diaspora itself.

I'm going to think of different options now. And still I'd love to see some opinions.

jhass:
I think making assumptions on where the `diaspora` clone (or a symlink towards it) is, is okay, it just needs to be clearly documented and perhaps automated where possible. I share your concerns with it being a distinct codebase and bloating the `Gemfile` a bit, hence the question.

cmrd-senya:
Still, we have to make usage and contribution to the testsuite as easy as possible. Otherwise, chances are that it won't help when its time comes.

jhass:
Maybe a simple shell script/rake task in this repo that sets it up? (clone to `vendor/` or `..`, `cd`, `bundle`, run).

Issue #7032, opened 2016-08-26
share visibilties migration not working correctly 


Zauberstuhl:
```
--- a/db/migrate/20160225232049_link_share_visibilities_with_user.rb
+++ b/db/migrate/20160225232049_link_share_visibilities_with_user.rb

[...]
-    change_column :share_visibilities, :user_id, :integer, null: false
 
     ShareVisibility.joins("LEFT OUTER JOIN users ON users.id = share_visibilities.user_id")
       .delete_all("users.id is NULL")
 
+    change_column :share_visibilities, :user_id, :integer, null: false
[...]
```
https://github.com/diaspora/diaspora/blob/develop/db/migrate/20160225232049_link_share_visibilities_with_user.rb#L27-L30

In case you have entries with `user_id == null` this migration fails. deleting all entries with null before chancing the column should fix the problem.

no :?

SuperTux88:
I think this is probably a joindiaspora-only problem (big old DB, maybe manually edited?), because some other big pods didn't have a problem with this migration (geraspora, diasp.org). The user_id should never be null. How many share visibilities did you have with user_id null?

Zauberstuhl:
> How many share visibilities did you have with user_id null?

not many

its quite possible that I am the only one with the problem. still the migration is not correct, right?
I mean if there shouldn't be any null entries why bothering with deleting them in the first place?

jhass:
Let's fix the migration but doing a hotfix release for it doesn't seem necessary.

PR #7038, opened 2016-08-29
Delete all null entries before removing the column

ghost:
related to #7032


SuperTux88:
Merged as cf5a9d7e0467a998a18402c88b9e95d55cd23303

Thanks :)

PR #7041, opened 2016-08-30
make session-cookie HttpOnly

SuperTux88:
This was changed to `false` in 2010 to make websockets work, but diaspora doesn't have websockets anymore, so we can change this back to `HttpOnly`.


Flaburgan:
That will improve the grade set at https://observatory.mozilla.org/analyze.html?host=diaspora-fr.org

denschub:
Merged as b3453c494ebdf866a0e51cf76ce899efaba92159.

PR #7043, opened 2016-09-01
Add OpenGraph video support

### LINKED ISSUE #4325: Embedding Bandcamp audio
I've been looking into how we might embed audio from Bandcamp.

It seems that Bandcamp puts barriers in the way of doing a straight oembed.

However, I have noticed that there is link to an embedded player contained in the meta information for each Bandcamp page.

I've checked this out for several different artists, and it appears in the same format for each of them, so I'm assuming this is universal for Bandcamp. There is a separate https embed, which is the one we'd want.

See, for example, http://takedaband.bandcamp.com/album/edo-1

Under 
```meta property="og:video:secure_url" ```
there appears
```content="https://bandcamp.com/EmbeddedPlayer.swf/size=venti/album=55213112/" ```

Copying this URL and pasting it into a browser produces a nice little embedded player with artist and album titles, artwork, track listing and everything you need.

Could a way be found to embed this in Diaspora? If anyone wants to have a go, I'd be really grateful.

There are some artists who use Bandcamp's site and software but have a redirect to their own domain. See for example http://music.farblackfurlong.com/album/haidd-2

This is a Bandcamp page, but the domain is the band's own. If you could find a way to recognise it as a Bandcamp page (possibly by something in the header information) in spite of the domain being different, and still embed it, that would be fantastic.

I guess the presence of
```meta property="og:video:secure_url"content="https://bandcamp.com/EmbeddedPlayer.swf/ ```
in the header (which it is for this last example, even though the visible domain is different) would be a way to achieve this recognition.

I don't really understand the science of embedding, so you'll have to forgive me if I'm talking a load of nonsense here, but I was really pleased to find the existence of an embedded player hidden in Bandcamp's pages, and hope that something can be taken and used from this.

In hope, and with thanks in advance.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/654847-embedding-bandcamp-audio?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### PR DESCRIPTION
**Pull request by svbergerem**
Fixes #4325.

If the OpenGraphReader finds a `og_video:secure_url` on the page it will be saved in the database. If we consider this URL secure (currently the bandcamp player is enabled) we will add a play button to the open graph image and on click we will load the video in an iframe.

opengraph bandcamp

### CODE REVIEW

File: spec/javascripts/app/views/open_graph_view_spec.js [hunk 1]
```diff
@@ -1,30 +1,72 @@
...
-    "description": "Test description",
-    "image": "http://example.com/thumb.jpg",
-    "ob_type": "article"
-  };
-
-  beforeEach(function(){
-    this.statusMessage = factory.statusMessage({
-      "open_graph_cache": open_graph_cache
+  context("without a video_url", function() {
+    beforeEach(function() {
```
diaspora-code-review: Extra space after key 'model'.

spec/javascripts/app/views/open_graph_view_spec.js
```diff
@@ -1,30 +1,72 @@
...
+      this.openGraphCache = {
+        "url": "http://example.com/articles/123",
+        "title": "Example title",
+        "description": "Test description",
```
diaspora-code-review: Missing space before opening brace.

spec/javascripts/app/views/open_graph_view_spec.js
```diff
@@ -1,30 +1,72 @@
...
+        expect(html).toContain(this.openGraphCache.title);
+        expect(html).toContain(this.openGraphCache.description);
+        expect(html).toContain(this.openGraphCache.image);
+        expect(html).not.toContain("video-overlay");
+      });
+    });
   });
 
-  describe("rendering", function(){
-    it("shows the preview based on the opengraph data", function(){
```
diaspora-code-review: Extra space after key 'model'.

spec/javascripts/app/views/open_graph_view_spec.js
```diff
@@ -1,30 +1,72 @@
...
-      this.view.render();
-      var html = this.view.$el.html();
+  context("with a video_url", function() {
+    beforeEach(function() {
```
diaspora-code-review: Missing space before opening brace.

svbergerem:
I changed the behavior a bit: now we will only save the video_url if we consider it secure. (Before the check was done before displaying the video_url) Currently Bandcamp is the only secure provider so we can assume that the given URL should be embedded in an iframe.

In the future we might decide that we want to add more secure providers and in this case we might have to check `og_video:type` (`text/html` in case of Bandcamp) and save it in the database. Since we don't need it right now I didn't add it in this PR.

Flaburgan:
Did bandcamp switch to an HTML5 player or is it still using a flash one?

svbergerem:
Since it's working on my laptop I'm pretty sure that they don't use a flash player.

denschub:
This looks fine, but I'm unsure about the migration. Do we have some runtime estimations?

SuperTux88:
`== 20160901072443 AddVideoUrlToOpenGraphCache: migrated (16.9579s) ============`

with my production DB.

denschub:
r+, feel free to merge.

SuperTux88:
Merged as 4d51c0243252425179137b4c9633f0cc32b06760

Thanks :)

SuperTux88:
It was even faster on my real production DB :)

`== 20160901072443 AddVideoUrlToOpenGraphCache: migrated (4.7082s) =============`

PR #7049, opened 2016-09-02
Remove user_photo method from users controller

svbergerem:
The method seems to be unused and brakeman found [a possible unprotected redirect](https://codeclimate.com/github/diaspora/diaspora/app/controllers/users_controller.rb#security).


denschub:
Merged as 08794bc47bb1000398c98aaa1b9b6705d47d3910, thanks.

PR #7050, opened 2016-09-02
Sign out users with wrong CSRF tokens

svbergerem:
See http://guides.rubyonrails.org/security.html#cross-site-request-forgery-csrf


denschub:
Merged, thanks.

PR #7051, opened 2016-09-02
Cleanup invalid unconfirmed emails

### LINKED ISSUE #7048: 422 Validation failed: Unconfirmed email has already been taken
A framasphere user was not able to log in after the upgrade to 0.6, which contains a devise upgrade.
He has two accounts, and the same email address was used in the `email` of one and the `unconfirmed_email` of the other.
We should write a migration to clear all unconfirmed_email which are used as email by another user.

SuperTux88:
Also remove confirm_email_token when unconfirmed_email is removed.

The validation was added in #6905.

Fixes #7048

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -466,7 +466,7 @@ class User < ActiveRecord::Base
 
   # Whenever email is set, clear all unconfirmed emails which match
   def remove_invalid_unconfirmed_emails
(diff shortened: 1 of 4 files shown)


denschub:
Merged as 50e73145353ab0df76582de4292b1ea01a546966.

PR #7055, opened 2016-09-03
Fix conversations message overflow

svbergerem:
Related issue: https://github.com/twbs/bootstrap/issues/15958

**before**
before

**after**
after


denschub:
lgtm, merged!

PR #7061, opened 2016-09-04
Redirect logged in users to inviters page when following an invitation link

### LINKED ISSUE #7057: Invite your friends url logs out users that are logged in
Some people told me that the "invite your friends"-link I was posting them logs them out of diaspora. 
I tried to understand their problem and simulated what happened, but please excuse me when I am wrong on this, it is a little bit tiring to test with different browsers and cookies etc., however I have replayed this sequence now a few times with the same result:
* login to some pod in browser A with account A
* login to same pod in browser B with account B
* copy "invite your friends"-url from browser A to browser B
* follow that link in browser B (I mean copy to url bar and press enter, just to be clear)
* account B in browser B is now logged out and sees the login screen of the pod

I would expect this not to happen and if this behaviour of the "invite your friends"-url can be confirmed by others it should be classified as a bug. 

Thanks for your attention!

svbergerem:
Fixes #7057.

### DIFF
--- a/app/controllers/invitation_codes_controller.rb
+++ b/app/controllers/invitation_codes_controller.rb
@@ -5,9 +5,14 @@ class InvitationCodesController < ApplicationController
     redirect_to root_url, :notice => I18n.t('invitation_codes.not_valid')
   end
 
-  def show 
-    sign_out(current_user) if user_signed_in?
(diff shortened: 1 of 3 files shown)


denschub:
Merged as e27af6ee1a57485f70f0b6ab4f8c37f7bac76c6e, thanks.

PR #7062, opened 2016-09-05
[API] Override forgery settings in controllers

### LINKED ISSUE #7059: [API] Client registration doesn't work anymore
I'm not sure if this issue is valid.

#6182 contains client registration call:
```
curl http://localhost:3000/api/openid_connect/clients -d "redirect_uris[]=http://localhost:3000/&client_name=<your_client_name>"
```

After the recent commit 6ad9000f8c16f6758c39b59412a4b6978910d690 it stopped working. It returns the following instead of expected json.
```
<html><body>You are being <a href="http://development.diaspora.local/users/sign_in">redirected</a>.</body></html>
```

I guess it would work if we log in as some user, but do we need a logged user to register a client on pod?

cmrd-senya:
ClientsController and TokenEndpointController are called from the outside, so CSRF verification prevents them from normal operation.

fix #7059

### DIFF
--- a/app/controllers/api/openid_connect/clients_controller.rb
+++ b/app/controllers/api/openid_connect/clients_controller.rb
@@ -1,6 +1,8 @@
 module Api
   module OpenidConnect
     class ClientsController < ApplicationController
+      skip_before_action :verify_authenticity_token
+
       rescue_from OpenIDConnect::HttpError do |e|
         http_error_page_as_json(e)
       end
--- a/app/controllers/api/openid_connect/token_endpoint_controller.rb
+++ b/app/controllers/api/openid_connect/token_endpoint_controller.rb
@@ -1,6 +1,8 @@
 module Api
   module OpenidConnect
     class TokenEndpointController < ApplicationController
+      skip_before_action :verify_authenticity_token
+
       def create
         req = Rack::Request.new(request.env)
         if req["client_assertion_type"] == "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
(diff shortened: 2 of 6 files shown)


app/controllers/api/openid_connect/clients_controller.rb
```diff
@@ -1,6 +1,8 @@
 module Api
   module OpenidConnect
     class ClientsController < ApplicationController
+      protect_from_forgery
```
svbergerem: I'm wondering if we should add

```ruby
skip_before_action :verify_authenticity_token
```

instead.

If we keep the current line then I think we should describe in a comment why we added this here. Just from the code itself it's not that obvious (at least for me) that this is needed to override the forgery settings from the `ApplicationController` and use `with: :null_session` instead of `with: :exception`.

cmrd-senya: I feel `skip_before_action :verify_authenticity_token` to be more intelligible here, I would replace it.

Also I noticed there is one more place where it is required to suppress token verification. I want to build some simplified working web authentication client, before I update it, to make sure that I know all the changes required to make authentication working.

cmrd-senya:
updated

svbergerem (approved):
Looks good to me but it would be great if we could get a review of someone who was more involved in the OpenID implementation. @jhass @denschub?

denschub:
Hm.

The entire API should work without an csrf token, since we most likely won't be serving one. I wonder if there is a way to disable the check for the entire API namespace...

denschub:
Disregard my comment, merged as cdcf2d747e1fce8adaa540dd3774216ea73806a1.

PR #7066, opened 2016-09-05
Fix mapbox API URL

SuperTux88:
The old API URL doesn't work anymore.

### DIFF
--- a/spec/javascripts/app/views/location_stream_spec.js
+++ b/spec/javascripts/app/views/location_stream_spec.js
@@ -2,9 +2,9 @@ describe("app.views.LocationStream", function() {
   beforeEach(function(){
     this.post = factory.post();
     this.view = new app.views.LocationStream({model : this.post});
-    /* jshint camelcase: false */
-    gon.appConfig = {map: { mapbox: {enabled: true, id: "yourID", access_token: "yourAccessToken" }}};
-    /* jshint camelcase: true */
+    /* eslint-disable camelcase */
+    gon.appConfig = {map: {mapbox: {enabled: true, style: "mapbox/streets-v9", access_token: "yourAccessToken"}}};
+    /* eslint-enable camelcase */
   });
 
   describe("toggleMap", function() {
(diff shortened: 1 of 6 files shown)


spec/javascripts/app/views/location_stream_spec.js
```diff
@@ -3,7 +3,7 @@ describe("app.views.LocationStream", function() {
     this.post = factory.post();
     this.view = new app.views.LocationStream({model : this.post});
     /* jshint camelcase: false */
-    gon.appConfig = {map: { mapbox: {enabled: true, id: "yourID", access_token: "yourAccessToken" }}};
+    gon.appConfig = {map: { mapbox: {enabled: true, access_token: "yourAccessToken" }}};
     /* jshint camelcase: true */
```
diaspora-code-review: - There should be no space after '{'
- Identifier 'access_token' is not in camel case.
- There should be no space before '}'

denschub:
No. No. Nonono. That's not what I was talking about, not at all, sorry.

`https://api.mapbox.com/styles/v1/mapbox/streets-v9/tiles/256/{z}/{x}/{y}?access_token={accessToken}`

This URL is now always using the `mapbox/streets-v9` style without allowing podmins to choose. Mapbox does offer a few different styles, `mapbox/dark-v9` for example would be the same style with darker colors.

In addition, Mapbox allows everyone to customize map colors, which could be useful for pods that already have an established design color. A customized style could be `denschub/cisqp2kk2002q2xo5fuq8ux6m` for example, so the entire URL would be

`https://api.mapbox.com/styles/v1/denschub/cisqp2kk2002q2xo5fuq8ux6m/tiles/256/{z}/{x}/{y}?access_token={accessToken}`.

That's what the ID parameter is for. We can set `mapbox/streets-v9` as the default style, but removing the parameter is not an option.

SuperTux88:
I renamed the `id` parameter to `style`, because I didn't know what `id` should be and the API documentation didn't mention a `id` parameter.

skyschub (approved):

denschub:
Although this contains config file changes, it's most probably broken anyway, so I've merged this to `next-minor` as 0e5141dd678602f16a2adc8d9081e2dd7d95b41f. Thanks!

PR #7070, opened 2016-09-06
Prevent overflow of interaction avatars in the single post view

svbergerem:
**before**:

before

**after**:

after

(Don't be confused by the wrong likes count. I added the avatars manually.)


SuperTux88:
Merged as 16af93f8a240228d339c8479cc16d86570c067e4 

Thanks! 🍪

PR #7076, opened 2016-09-07
Remove remotipart gem and remotipart photo uploads

svbergerem:
This has been added in #3068. [Remotipart](https://github.com/JangoSteve/remotipart) adds support for photo uploads via rails remote forms and it looks like we don't do this anymore.


SuperTux88:
Merged as 521468986f8527b49f082c87ce3bbe06eaadc2df

Thanks

PR #7081, opened 2016-09-10
Fix ID token issuer field to match discovery data

cmrd-senya:
OpenID Connect spec says:

"The issuer returned by discovery MUST exactly match the value of iss in the ID Token."

https://openid.net/specs/openid-connect-core-1_0.html#IssuerIdentifier

### DIFF
--- a/lib/api/openid_connect/id_token.rb
+++ b/lib/api/openid_connect/id_token.rb
@@ -51,7 +51,7 @@ module Api
       def claims
         sub = build_sub
         @claims ||= {
-          iss:       AppConfig.environment.url,
+          iss:       Rails.application.routes.url_helpers.root_url,
           sub:       sub,
           aud:       @authorization.o_auth_application.client_id,
           exp:       @expires_at.to_i,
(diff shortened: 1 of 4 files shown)


lib/api/openid_connect/id_token.rb
```diff
@@ -65,6 +65,10 @@ def claims
       def build_sub
         Api::OpenidConnect::SubjectIdentifierCreator.create(@authorization)
       end
+
+      def root_url
+        Rails.application.routes.url_helpers.root_url(Rails.application.config.action_mailer.default_url_options)
```
cmrd-senya: I don't like referring `Rails.application.config.action_mailer` here. I thought of setting `Rails.application.routes.default_url_options` globally for production environment (currently it is set only in tests). However I'm unsure if it's okay.

denschub: No objections. Feel free.

skyschub (changes requested):
Looks good, feel free to change to `root_url` implementation. :)

cmrd-senya:
updated

denschub:
Merged as a80806ca58dc8ccd8ffc0e290e8b6963fa68cb98. Thanks.

PR #7086, opened 2016-09-10
Load jquery in the head on mobile

SuperTux88:
Some pages need jquery in the head and instead of loading it twice on these pages, it is better and easier to load it in the head on all pages. It should be in the cache after the first load anyway.

Also these pages still tried to load jquery 1 which didn't work anymore.


svbergerem:
Thanks! :)

PR #7090, opened 2016-09-12
Require installed optional groups with Bundler

svbergerem:
Let's see if this works.


denschub:
Merged and releasing a hotfix with that. Thanks much!

Issue #7067, opened 2016-09-06
Generate HTML / CSS compliant with w3c recommandations


Flaburgan:
Even if we will probably never be 100% compliant, there is room for improvements. I will try to do small PRs which improve our score at https://validator.w3.org. Any help is welcomed ;)

(I know @githubgnulinux will be happy about this one)

denschub:
Removed personal offenses/references from the original issue text.

Flaburgan:
I just wanted to ping him to tell him we're working on it, I didn't want that to sound like an offense. Sorry if it has been understood like that.

PR #7091, opened 2016-09-12
Use div instead of span for loader

svbergerem:
#7067


SuperTux88:
Merged as 2139f4b4c24fd878ec64fd2f9bc752e56eac2a61

Thanks.

Issue #4319, opened 2013-07-21
User setting for default post visibility


mlinksva:
I usually intend to make posts public, but forget to change from "all aspects" to "public" before posting. I'd enjoy a preference which set my default to "public".

Arcurus:
yea a default selection configuration would be great!

unity100:
This feature would be VERY helpful for those of us who use their diaspora account for sharing news and activism.

duncane:
Bumping cause this feature is definitly missing in Diaspora*.

nielspetersen:
This issue is part of the current issue (13th) of the RubyIssues. https://rubyissues.ongoodbits.com/. :gem:

hilkoc:
A [pull request](https://github.com/diaspora/diaspora/issues/7118) for this has now been opened.

PR #7101, opened 2016-09-14
Fix 4319 postdefault

hilkoc:
Please consider my contribution for issue #4319 that has been open for a long time.

In this change we add an aspect drop down selector on the user settings page. After selecting and saving these post_default aspects they will be used in the main stream's publisher for preselecting the aspects to post to.

In addition, we preselect the 'public' aspect in the publisher for the public activities stream. This makes the behavior analogous to the My Aspects stream, where we preselect the aspects of the aspect stream.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -6,9 +6,26 @@ class UsersController < ApplicationController
   before_action :authenticate_user!, except: %i(new create public user_photo)
   respond_to :html
 
+  # The publisher/aspect_dropdown partial requiers a @stream that responds to publisher().public()
+  class DummyStream
+    class HasPublic
+      attr_accessor :public
+    end
+
+    def publisher()
+      @h ||= HasPublic.new
+    end
+
+    def set_public!(b)
+      self.publisher().public = b
+    end
+  end
+
   def edit
     @user = current_user
     set_email_preferences
+    @stream ||= DummyStream.new
+    @stream.set_public! @user.post_default_public
   end
 
   def privacy_settings
@@ -153,6 +170,7 @@ class UsersController < ApplicationController
       :auto_follow_back,
       :auto_follow_back_aspect_id,
       :getting_started,
+      :post_default_public,
       email_preferences: %i(
         someone_reported
         also_commented
@@ -176,11 +194,14 @@ class UsersController < ApplicationController
       change_email(user_data)
     elsif user_data[:auto_follow_back]
       change_settings(user_data, "users.update.follow_settings_changed", "users.update.follow_settings_not_changed")
+    elsif user_data[:post_default_public]
+      change_post_default(user_data)
     elsif user_data[:color_theme]
       change_settings(user_data, "users.update.color_theme_changed", "users.update.color_theme_not_changed")
     else
       change_settings(user_data)
     end
+    edit
   end
 
   def change_password(user_data)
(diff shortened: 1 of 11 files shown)


cmrd-senya:
SInce you have a migration in your PR, please include the updated `db/schema.rb` file as well.

app/controllers/users_controller.rb
```diff
@@ -6,9 +6,26 @@ class UsersController < ApplicationController
   before_action :authenticate_user!, except: %i(new create public user_photo)
   respond_to :html
 
+  # The publisher/aspect_dropdown partial requiers a @stream that responds to publisher().public()
```
cmrd-senya: I'd prefer to change the publisher/aspect_dropdown partial rather than introducing a dummy class as a workaround.

app/controllers/users_controller.rb
```diff
@@ -176,11 +194,14 @@ def update_user(user_data)
...
     elsif user_data[:auto_follow_back]
       change_settings(user_data, "users.update.follow_settings_changed", "users.update.follow_settings_not_changed")
+    elsif user_data[:post_default_public]
+      change_post_default(user_data)
     elsif user_data[:color_theme]
       change_settings(user_data, "users.update.color_theme_changed", "users.update.color_theme_not_changed")
     else
       change_settings(user_data)
     end
+    edit
```
cmrd-senya: Do you make this call only to set `@stream` on the response page? If so, consider moving instance variables assignments to a separate method and calling it in the appropriate elsif block.

db/migrate/20160829170244_add_post_default_to_aspects.rb
```diff
@@ -0,0 +1,11 @@
+class AddPostDefaultToAspects < ActiveRecord::Migration
+  def self.up
```
cmrd-senya: I guess in this migration it is better to group changes using `self.change` instead of `self.up/self.down`.

cmrd-senya:
Also please follow the [style check guide](https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review). Run the check locally, fix the reported issues and update your PR.

I also feel that test coverage for this feature might be improved. Since the feature is mostly UX related, I think the best way is to write a few cucumber tests for it.

Thank you for your contribution. This feature is really demanded!

app/controllers/users_controller.rb
hilkoc: I agree that it would be much nicer if the dropdown does not depend on the presence of a @stream. I wanted to keep the impact of this change limited. It seems that more refactoring is required as a prerequisite for this change, otherwise the aspect_dropdown can not be used anywhere but inside a publisher in a stream.

app/controllers/users_controller.rb
hilkoc: The edit belongs in every elsif-block, hence outside of it. If we want to reuse the aspect_dropdown, we must have a @stream instance variable at any time, otherwise the page doesn't render.

hilkoc:
Thanks for your constructive feedback. I'll fix these items mentioned. I might split them up and make some smaller PRs first.

cmrd-senya:
Remember that you don't have to open a new pull request, you can just update the same one.

app/controllers/users_controller.rb
Flaburgan: The /contacts page uses an aspect_dropdown without being a stream IIRC, maybe you can have a look at how it's done there?

cmrd-senya: /contacts uses aspect_membership_dropdown which is unrelated to aspect_dropdown code used in publisher. they should probably be unified though

hilkoc:
I've rewritten this branch and tried to reopen this pull request, but it doesn't seem to pick up the new commits.

PR #7104, opened 2016-09-15
Jsxc automatic connection with opt-out

ghost:
Will add a new button in the privacy section to dis-/enable chat login on default

related to diaspora/jsxc#130

### DIFF
--- a/app/assets/javascripts/jsxc.js
+++ b/app/assets/javascripts/jsxc.js
@@ -44,8 +44,7 @@ $(document).ready(function() {
         });
 
         // auto login if the user hasn't disabled it
-        var onLogin = !(app.currentUser.get("disable_chat_login"));
-        if (onLogin) {
+        if (!(app.currentUser.get("chat_auto_login"))) {
           jsxc.gui.changePresence("online");
         }
       } else {
--- a/app/assets/javascripts/jsxc.js
+++ b/app/assets/javascripts/jsxc.js
@@ -44,7 +44,7 @@ $(document).ready(function() {
         });
 
         // auto login if the user hasn't disabled it
-        if (!(app.currentUser.get("chat_auto_login"))) {
+        if (app.currentUser.get("chat_auto_login")) {
           jsxc.gui.changePresence("online");
         }
       } else {
--- a/app/assets/javascripts/jsxc.js
+++ b/app/assets/javascripts/jsxc.js
@@ -42,6 +42,12 @@ $(document).ready(function() {
             onlogin: true
           }
         });
+
+        // auto login if the user hasn't disabled it
+        var onLogin = !(app.currentUser.get("disable_chat_login"));
+        if (onLogin) {
+          jsxc.gui.changePresence("online");
+        }
       } else {
         console.error('No token found! Authenticated!?');
       }
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -40,11 +40,7 @@ class UsersController < ApplicationController
   def update_privacy_settings
     privacy_params = params.fetch(:user)
     if privacy_params
-      if privacy_params[:strip_exif]
-        change_privacy_params(:strip_exif, privacy_params[:strip_exif])
-      elsif privacy_params[:chat_auto_login]
-        change_privacy_params(:chat_auto_login, privacy_params[:chat_auto_login])
-      end
+      change_settings(privacy_params, "users.update.settings_updated", "users.update.settings_not_updated")
     end
     redirect_to :back
   end
@@ -237,14 +233,6 @@ class UsersController < ApplicationController
 
     @user.user_preferences.each do |pref|
       @email_prefs[pref.email_type] = false
-    end
-  end
-
-  def change_privacy_params(key, value)
-    if current_user.update_attributes(key => value)
-      flash[:notice] = t("users.update.settings_updated")
-    else
-      flash[:error] = t("users.update.settings_not_updated")
     end
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -38,14 +38,14 @@ class UsersController < ApplicationController
   end
 
   def update_privacy_settings
-    privacy_params = params.fetch(:user).permit(:strip_exif)
-
-    if current_user.update_attributes(strip_exif: privacy_params[:strip_exif])
-      flash[:notice] = t("users.update.settings_updated")
-    else
-      flash[:error] = t("users.update.settings_not_updated")
+    privacy_params = params.fetch(:user)
+    if privacy_params
+      if privacy_params[:strip_exif]
+        change_privacy_params(:strip_exif, privacy_params[:strip_exif])
+      elsif privacy_params[:disable_chat_login]
+        change_privacy_params(:disable_chat_login, privacy_params[:disable_chat_login])
+      end
     end
-
     redirect_to :back
   end
 
@@ -145,6 +145,7 @@ class UsersController < ApplicationController
       :language,
       :color_theme,
       :disable_mail,
+      :disable_chat_login,
       :show_community_spotlight_in_stream,
       :auto_follow_back,
       :auto_follow_back_aspect_id,
@@ -236,6 +237,14 @@ class UsersController < ApplicationController
 
     @user.user_preferences.each do |pref|
       @email_prefs[pref.email_type] = false
+    end
+  end
+
+  def change_privacy_params(key, value)
+    if current_user.update_attributes(key => value)
+      flash[:notice] = t("users.update.settings_updated")
+    else
+      flash[:error] = t("users.update.settings_not_updated")
     end
   end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -42,8 +42,8 @@ class UsersController < ApplicationController
     if privacy_params
       if privacy_params[:strip_exif]
         change_privacy_params(:strip_exif, privacy_params[:strip_exif])
-      elsif privacy_params[:disable_chat_login]
-        change_privacy_params(:disable_chat_login, privacy_params[:disable_chat_login])
+      elsif privacy_params[:chat_auto_login]
+        change_privacy_params(:chat_auto_login, privacy_params[:chat_auto_login])
       end
     end
     redirect_to :back
@@ -145,7 +145,7 @@ class UsersController < ApplicationController
       :language,
       :color_theme,
       :disable_mail,
-      :disable_chat_login,
+      :chat_auto_login,
       :show_community_spotlight_in_stream,
       :auto_follow_back,
       :auto_follow_back_aspect_id,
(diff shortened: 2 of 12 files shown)


Zauberstuhl:
This one failed locally as well:
`  1) User close account #clearable_attributes returns the clearable fields`

```
       expected: ["authentication_token", "auto_follow_back", "auto_follow_back_aspect_id", "color_theme", "confirm_em...created_at", "reset_password_sent_at", "reset_password_token", "sign_in_count", "unconfirmed_email"]

            got: ["authentication_token", "auto_follow_back", "auto_follow_back_aspect_id", "color_theme", "confirm_em...created_at", "reset_password_sent_at", "reset_password_token", "sign_in_count", "unconfirmed_email"]
```

do I miss something? got and expected are identical

hilkoc:
I've got the same test failing  '#clearable_attributes returns the clearable fields...' for no apparent reason.
How did you fix it?

svbergerem:
I improved the failing spec a bit:
```diff
-        expect(user.send(:clearable_fields).sort).to eq(%w(
+        expect(user.send(:clearable_fields)).to match_array(%w(
```

and got this result:
```
  1) User close account #clearable_attributes returns the clearable fields
     Failure/Error:
       expect(user.send(:clearable_fields)).to match_array(%w(
         language
         reset_password_sent_at
         reset_password_token
         remember_created_at
         sign_in_count
         current_sign_in_at
         last_sign_in_at
         current_sign_in_ip
         hidden_shareables
     
       expected collection contained:  ["authentication_token", "auto_follow_back", "auto_follow_back_aspect_id", "color_theme", "confirm_em...created_at", "reset_password_sent_at", "reset_password_token", "sign_in_count", "unconfirmed_email"]
       actual collection contained:    ["authentication_token", "auto_follow_back", "auto_follow_back_aspect_id", "color_theme", "confirm_em...created_at", "reset_password_sent_at", "reset_password_token", "sign_in_count", "unconfirmed_email"]
       the extra elements were:        ["disable_chat_login"]
```

Flaburgan:
@Zauberstuhl could you please rebase that PR? I would love to test it on d-fr.

Zauberstuhl:
@Flaburgan done, travis fail should be resolved as well

app/assets/javascripts/jsxc.js
```diff
@@ -42,6 +42,12 @@ $(document).ready(function() {
             onlogin: true
           }
         });
+
+        // auto login if the user hasn't disabled it
+        var onLogin = !(app.currentUser.get("disable_chat_login"));
+        if (onLogin) {
+          jsxc.gui.changePresence("online");
+        }
```
svbergerem: I think `autoLogin` would be a better name for the variable. Or just drop the variable since it is only used in the next line.

app/controllers/users_controller.rb
```diff
@@ -247,4 +248,12 @@ def set_email_preferences
       @email_prefs[pref.email_type] = false
     end
   end
+
+  def change_privacy_params(key, value)
```
svbergerem: One could also use this to update other user params, so I'd use another name for the method.

app/models/user.rb
```diff
@@ -419,6 +419,7 @@ def seed_aspects
     self.aspects.create(:name => I18n.t('aspects.seed.friends'))
     self.aspects.create(:name => I18n.t('aspects.seed.work'))
     aq = self.aspects.create(:name => I18n.t('aspects.seed.acquaintances'))
+    aq.chat_enabled = true # enable chat aspect in acquaintances
```
svbergerem: Why? You didn't mention this in the description of the PR and also the commit message doesn't mention this. (Looks like a more or less unrelated change)

app/assets/javascripts/jsxc.js
```diff
@@ -42,6 +42,12 @@ $(document).ready(function() {
             onlogin: true
           }
         });
+
+        // auto login if the user hasn't disabled it
+        var onLogin = !(app.currentUser.get("disable_chat_login"));
```
svbergerem: How about using `chat_auto_login` or something similar everywhere? (jsxc.js, presenter, controller, database... default: `true`) I think it is easier to understand if you describe the “positive” event.

spec/controllers/users_controller_spec.rb
```diff
@@ -216,6 +216,12 @@
       get 'privacy_settings'
       expect(response.status).to eq(200)
     end
+
+    it "updates chat privacy settings" do
+      put(:update_privacy_settings, user: {disable_chat_login: true})
+      expect(request.flash[:notice]).to eql(I18n.t("users.update.settings_updated"))
+      expect(request.flash[:error]).to be_blank
+    end
```
svbergerem: Please check that this actually changes the privacy settings. (both true -> false and false -> true)

svbergerem (changes requested):

Flaburgan:
@Zauberstuhl do you plan to fix @svbergerem remarks soon?

app/controllers/users_controller.rb
Zauberstuhl: well then they can extend the function like in `update_user`, no?

app/models/user.rb
Zauberstuhl: mh not sure anymore. I think it was for new users only. If they log-in they have no contacts. with the default on acquaintances you would see the hq account and podmin accounts.

I can remove it, but in my opinion its more user-friendly?!

Flaburgan: I feel like activate the chat for an aspect can be a good idea but should be discussed first so it should not be part of this PR.

app/controllers/users_controller.rb
SuperTux88: I think you can simply use `change_settings` in `update_privacy_settings` and drop this method here.

app/assets/javascripts/jsxc.js
```diff
@@ -42,6 +42,11 @@ $(document).ready(function() {
             onlogin: true
           }
         });
+
+        // auto login if the user hasn't disabled it
+        if (!(app.currentUser.get("chat_auto_login"))) {
+          jsxc.gui.changePresence("online");
+        }
```
svbergerem: Now you changed the name of the variable but kept the meaning, so if `chat_auto_login` is true it will disable the auto login. That's super confusing. I first thought you made a mistake here and only by reading the translations I was able to find out that this code is actually correct.

Please change the meaning of the variable and the translations so that we will log users in if `chat_auto_login` is enabled.

app/controllers/users_controller.rb
svbergerem: Ping?

svbergerem (changes requested):
Do we have a decision somewhere about opt-in vs. opt-out? I see a possible privacy issue here but didn't think about the pros and cons too much.

Zauberstuhl:
@svbergerem it should be disabled on default now: https://github.com/diaspora/diaspora/pull/7104/files#diff-1acd2e7e27a227829d5d14a91c863bb6R641

app/controllers/users_controller.rb
```diff
@@ -38,14 +38,10 @@ def update
   end
 
   def update_privacy_settings
-    privacy_params = params.fetch(:user).permit(:strip_exif)
-
-    if current_user.update_attributes(strip_exif: privacy_params[:strip_exif])
-      flash[:notice] = t("users.update.settings_updated")
-    else
-      flash[:error] = t("users.update.settings_not_updated")
+    privacy_params = params.fetch(:user)
```
SuperTux88: We should still filter these params, otherwise the user can change stuff that he shouldn't be able to change.

config/locales/diaspora/en.yml
```diff
@@ -1207,6 +1207,8 @@ en:
       ignored_users: "Ignored users"
       stop_ignoring: "Stop ignoring"
       no_user_ignored_message: "You are not currently ignoring any other user"
+      chat_auto_login: "Disable chat autologin"
```
SuperTux88: This should be "Enable ..." now.

config/locales/diaspora/en.yml
```diff
@@ -1207,6 +1207,8 @@ en:
...
+      chat_auto_login_desc: "If you login the first time the chat will try to connect automatically. You can disable this behaviour here!"
```
SuperTux88: The default is `false`, to it is disabled by default.

app/controllers/users_controller.rb
```diff
@@ -38,14 +38,10 @@ def update
...
+    if privacy_params
+      change_settings(privacy_params, "users.update.settings_updated", "users.update.settings_not_updated")
```
SuperTux88: You need to change `change_settings` to use `current_user` or set `@user`.

spec/models/user_spec.rb
```diff
@@ -806,7 +806,13 @@
       [I18n.t('aspects.seed.family'), I18n.t('aspects.seed.friends'),
        I18n.t('aspects.seed.work'), I18n.t('aspects.seed.acquaintances')].each do |aspect_name|
         it "creates an aspect named #{aspect_name} for the user" do
-          expect(user.aspects.find_by_name(aspect_name)).not_to be_nil
+          aspect = user.aspects.find_by_name(aspect_name)
+          if aspect.eql? I18n.t("aspects.seed.acquaintances")
```
SuperTux88: This change isn't needed anymore.

SuperTux88 (changes requested):

db/schema.rb
```diff
@@ -638,6 +638,7 @@
     t.boolean  "exporting_photos",                                 default: false
     t.string   "color_theme",                        limit: 255
     t.boolean  "post_default_public",                              default: false
+    t.boolean  "chat_auto_login",                                  default: false
```
Flaburgan: Hm I personally feel like the default should be `true`. That's the reason we have so few users on the chat at the moment, they don't get they have to log in. So let's log them in by default, they can switch that off if they don't want to.

Zauberstuhl:
sorry guys.. should have updated the issue. I'll request a review

db/schema.rb
Flaburgan: Well, it depends how the "automatically log-in" is implemented here. IMO, what diaspora* should do is simply to remember your status between each page and each visit. It means, if last time I used diaspora*, I was online when I logged out, I should be online when I log back. Same for offline, or busy, or... Which means the default will only be used for the very first time the user will access the chat.

goobertron: Oh sorry, I misunderstood. I thought from your comment that you meant that every member of a pod will be automatically logged into chat when they next sign in to the pod after the software update, regardless of whether they wanted to use chat.

You're talking about only the people who were signed into chat when they signed out of diaspora\*. That's fine, of course.

SuperTux88: This PR is about "auto-login on every sign-in" and I agree with @goobertron here, that shouldn't be the default on.

There is no "remember my last chat-status and restore that on next sign-in"-feature yet. I think that could be useful, but I should also be "offline" on the first use.

Zauberstuhl: Well we could create a configuration option. Make it false on default and if the podmin likes to do it differently he can adjust the diaspora.yml, no?

SuperTux88:

PR #7107, opened 2016-09-17
Fix duplicate flash message on mobile profile edit

SuperTux88:
**before:**

screenshot_20160917_231858

**after:**

screenshot_20160917_233552


svbergerem:
Thanks!

PR #6917, opened 2016-07-14
Port conversations backbone, use Typeahead pt. 2 and Bootstrap tag

ghost:
- [x] Fix cukes and Jasmine
- [x] Write tests
- [x] Fix cucumber

screen shot 2016-07-14 at 01 07 01

PR #7108, opened 2016-09-18
Fully port conversations to Backbone and drop inbox.js

ghost:
Supersedes #6917


AugierLe42e:
Ready for review.

app/controllers/conversations_controller.rb
```diff
@@ -24,7 +24,7 @@ def index
     gon.contacts = contacts_data
 
     respond_with do |format|
-      format.html
+      format.html { render "index", locals: {no_contacts: current_user.contacts.empty?} }
```
svbergerem: I think we should check if the mutual contacts are empty since you are only allowed to start a conversation with mutual contacts.

app/views/conversations/index.haml
```diff
@@ -1,41 +1,36 @@
...
-        .stream_container.hidden
-          #conversation_show
-        .framed-content.clearfix#conversation_new
-          .new-conversation
-            %h3.text-center
-              = t("conversations.index.new_conversation")
-            = render "conversations/new"
+        #conversation-new{class: @conversation ? "framed-content clearfix hidden" : "framed-content clearfix"}
+          - if no_contacts
+            .well.text-center= t(".no_contacts")
```
svbergerem: `t("javascripts.conversation.new.no_contacts")` should work, too. (and you won't have to change these translations)

app/assets/javascripts/app/views/conversations_inbox_view.js
```diff
@@ -0,0 +1,87 @@
...
+    "click .new-conversation-btn": "displayNewConversation"
+  },
+
+  initialize: function() {
+    new app.views.ConversationsForm({contacts: gon.contacts});
+    this.setupConversation();
+  },
+
+  renderConversation: function(conversationId) {
+    if (parseInt("" + conversationId, 10)) { // Ensures that conversationId is a readable number
```
svbergerem: I think we should check this in the backbone router instead before calling this function.

app/views/conversations/_messages.haml
```diff
@@ -6,17 +6,12 @@
...
-            %label#messageLabel.sr-only{for: "message_text"}
-              = t("conversations.new.message")
-            = message.text_area :text,
-                                rows:     5,
-                                tabindex: 1,
-                                class:    "form-control form-group",
-                                aria:     {labelledby: "messageLabel"}
+            %label.sr-only#message-label{for: "response-message-text"}= t("conversations.new.message")
+            = message.text_area :text, rows: 5, tabindex: 1, id: "response-message-text",
+                class: "form-control form-group conversation-message-text", aria: {labelledby: "message-label"}
```
svbergerem: At least for me the readability was better before when we had one attribute per line.

app/views/conversations/_messages.haml
```diff
@@ -6,17 +6,12 @@
...
 
-          = message.submit t("conversations.show.reply"),
-                            "data-disable-with" => t("conversations.show.replying"),
-                            class: "btn btn-primary pull-right", tabindex: 2
+          = message.submit t("conversations.show.reply"), "data-disable-with" => t("conversations.show.replying"),
+              :class => "btn btn-primary pull-right", :tabindex => 2
```
svbergerem: For the sake of consistency I'd also prefer one attribute per line here.

app/views/conversations/_new.haml
```diff
@@ -1,22 +1,18 @@
...
       = text_field_tag "contact_autocomplete", nil, class: "form-control"
     .form-group
-      %label#subjectLabel{for: "conversation_subject"}
+      %label#subject-label{for: "conversation-subject"}
         = t(".subject")
-      = conversation.text_field :subject,
-                                class: "input-block-level form-control",
-                                aria:  {labelledby: "subjectLabel"}
+      = conversation.text_field :subject, id: "conversation-subject", class: "input-block-level form-control",
+        aria:  {labelledby: "subject-label"}, value: "", placeholder: t("conversations.new.subject_default")
```
svbergerem: One attribute per line

app/views/conversations/_new.haml
```diff
@@ -1,22 +1,18 @@
...
-      %label#messageLabel.sr-only{for: "conversation_text"}
-        = t(".message")
-      = text_area_tag "conversation[text]",
-                      "",
-                      rows:  5,
-                      class: "input-block-level form-control",
-                      aria:  {labelledby: "messageLabel"}
+      %label.sr-only#message-label{for: "new-message-text"} = t(".message")
+      = text_area_tag "conversation[text]", "", id: "new-message-text",
+        class: "conversation-message-text input-block-level form-control", rows: 5, aria: {labelledby: "message-label"}
```
svbergerem: One attribute per line

spec/javascripts/app/views/conversations_inbox_view_spec.js
```diff
@@ -0,0 +1,178 @@
+describe("app.views.ConversationsInbox", function() {
+  describe("initialize", function() {
+    beforeEach(function() {
+      spec.loadFixture("conversations_read");
+      $("#conversation-new").removeClass("hidden");
+      $("#conversation-show").addClass("hidden");
+    });
+
+    it("call app.views.ConversationsForm()", function() {
```
svbergerem:
```javascript
it("calls app.views.ConversationsForm.initialize", function() {
```
or even
```javascript
it("initializes the conversations form", function() {
```

spec/javascripts/app/views/conversations_inbox_view_spec.js
```diff
@@ -0,0 +1,178 @@
...
+      spyOn(app.views.ConversationsForm.prototype, "initialize");
+      new app.views.ConversationsInbox();
+      expect(app.views.ConversationsForm.prototype.initialize).toHaveBeenCalled();
+    });
+
+    it("call setupConversation()", function() {
```
svbergerem:
```javascript
it("calls setupConversation", function() {
```

spec/javascripts/app/views/conversations_view_spec.js
```diff
@@ -1,79 +0,0 @@
...
-      expect(this.submitCallback).toHaveBeenCalled();
-    });
-
-    it("shouldn't submit the form without the ctrl key", function(){
-      $("form#new_message").submit(this.submitCallback);
-      var e = $.Event("keydown", { which: Keycodes.ENTER, ctrlKey: false });
-      $("textarea#message_text").trigger(e);
-      expect(this.submitCallback).not.toHaveBeenCalled();
-    });
-  });
```
svbergerem: I think the `keyDown` specs should still exist for the conversations form view. (just move them to a new specs file for the conversations form view)

app/views/conversations/index.haml
```diff
@@ -1,41 +1,36 @@
...
             = render 'conversations/show', conversation: @conversation
-      - else
-        .stream_container.hidden
-          #conversation_show
-        .framed-content.clearfix#conversation_new
-          .new-conversation
-            %h3.text-center
-              = t("conversations.index.new_conversation")
-            = render "conversations/new"
+        #conversation-new{class: @conversation ? "framed-content clearfix hidden" : "framed-content clearfix"}
```
svbergerem: Before:
no contacts before

After:
no contacts after

svbergerem (changes requested):

app/views/conversations/index.haml
AugierLe42e: The two translations are different. The javascript one is used in the case the user tries to send a private message with no receipiant. This one is used in the case the user has no contact yet.

svbergerem: I guess you are talking about `javascripts.conversation.create.no_contacts` but I meant `javascripts.conversation.new.no_contacts`. You removed this translation in this PR and added a new one in `config/locales/diaspora/en.yml` instead.

AugierLe42e: Ah yes, I understand the mistake. I forgot a function when I wrote this PR that checks the recipient list is not empty from client side. I will rename `javascripts.conversation.create.no_contacts` to `javascripts.conversations.create.no_recipient` and `conversations.create.no_contacts` to `conversations.create.no_recipiant` to avoid confusion with `conversations.new.no_contacts`.

AugierLe42e:
Done.

svbergerem:
It looks like all of my remarks are still unresolved.

AugierLe42e:
Yes, they are, even though Github does not display so here. Just look in the *Files changed tab*. I only forgot the mutual contacts thing but it's corrected now.

config/locales/javascript/javascript.en.yml
```diff
@@ -239,9 +239,9 @@ en:
       contacts: "Contacts"
       posts: "Posts"
 
-    conversation:
-      new:
-        no_contacts: "You need to add some contacts before you can start a conversation."
```
svbergerem: You are still removing the translation here and adding it to `config/locales/diaspora/en.yml ` instead.

svbergerem (changes requested):

config/locales/javascript/javascript.en.yml
AugierLe42e: Yes. That's because that message:



was previously rendered client-side ([here](https://github.com/diaspora/diaspora/pull/7108/files#diff-bf05c23497eed4643bb8fe3148aa9f47L39)) and is now rendered server-side ([here](https://github.com/diaspora/diaspora/pull/7108/files#diff-bd1e796410e7b0390610775a446df10aR32)) and I can't use Javascript translation key from Ruby code (I tried).

svbergerem: I wrote [here](https://github.com/diaspora/diaspora/pull/7108#discussion_r80387041) that you could use `t("javascripts.conversation.new.no_contacts")`.

AugierLe42e: Hmm... Indeed. I must have failed somewhere. It didn't work the first time. Ok, I correct that and push.

app/assets/javascripts/app/views/conversations_inbox_view.js
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+      $("html").animate({scrollTop: 0});
+    }
+  },
+
+  displayConversation: function(evt) {
+    var $target = $(evt.target);
+    if (!$target.hasClass(".conversation-wrapper")) {
+      $target = $target.parents(".conversation-wrapper");
+    }
```
svbergerem:
```javascript
var $target = $(evt.target).closest(".conversation-wrapper");
```

config/locales/javascript/javascript.en.yml
```diff
@@ -239,7 +239,9 @@ en:
       contacts: "Contacts"
       posts: "Posts"
 
-    conversation:
+    conversations:
+      create:
+        no_recipient: "Hey, you need to add a recipient first!"
```
svbergerem: This would mean that we would have to translate `javascripts.conversations.new.no_contacts` again, which was `javascripts.conversation.new.no_contacts` before. I don't think that renaming “conversation” to “conversations” is important enough to throw away the existing translations. That was the reason for the whole discussion about the translations and I still think that we should keep the old translations. (and thus keep the “conversation” key)

spec/javascripts/app/views/conversations_form_view_spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+  describe("onSubmitNewConversation", function() {
+    beforeEach(function() {
+      spec.loadFixture("conversations_read");
+      $("#conversation-new").removeClass("hidden");
+      $("#conversation-show").addClass("hidden");
+      spyOn(app.views.ConversationsForm.prototype, "onSubmitNewConversation").and.callThrough();
+      this.target = new app.views.ConversationsForm();
+    });
+    it("calls onSubmitNewConversation", function() {
```
svbergerem: > onSubmitNewConversation **calls onSubmitNewConversation**

I'd rather say

> onSubmitNewConversation **is called when submitting the conversation form**

spec/javascripts/app/views/conversations_form_view_spec.js
```diff
@@ -0,0 +1,109 @@
...
+      spyOn(app.views.ConversationsForm.prototype, "getConversationParticipants").and.returnValue([]);
+      var event = jasmine.createSpyObj("event", ["preventDefault", "stopPropagation"]);
+
+      this.target.onSubmitNewConversation(event);
+
+      expect(app.views.FlashMessages.prototype.error)
+        .toHaveBeenCalledWith(Diaspora.I18n.t("conversations.create.no_recipient"));
+    });
+
+    it("does not flashe an error message when submitting a conversation with no recipient", function() {
```
svbergerem:
```javascript
it("does not flash an error message when submitting a conversation with recipients", function() {
```
(“flashe” -> “flash”, “with no recipients” -> “with recipients”)

spec/javascripts/app/views/conversations_inbox_view_spec.js
```diff
@@ -0,0 +1,164 @@
...
+
+  describe("selectConversation", function() {
+    beforeEach(function() {
+      spec.loadFixture("conversations_read");
+      this.conversationId = $("#conversation-inbox .stream-element").first().data("guid");
+      this.target = new app.views.ConversationsInbox();
+      $("#conversation-inbox .stream-element").addClass("selected");
+    });
+
+    it("unselect every conversation if called with no parameters", function() {
```
svbergerem: “unselect” -> “unselects”

spec/javascripts/app/views/conversations_inbox_view_spec.js
```diff
@@ -0,0 +1,164 @@
...
+  });
+
+  describe("displayConversation", function() {
+    beforeEach(function() {
+      spyOn(app.router, "navigate");
+      spec.loadFixture("conversations_read");
+      new app.views.ConversationsInbox();
+    });
+
+    it("call app.router.navigate with correct parameters", function() {
```
svbergerem: “call“ -> “calls”

spec/javascripts/app/views/conversations_form_view_spec.js
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  describe("onSubmitNewConversation", function() {
+    beforeEach(function() {
+      spec.loadFixture("conversations_read");
+      $("#conversation-new").removeClass("hidden");
+      $("#conversation-show").addClass("hidden");
+      spyOn(app.views.ConversationsForm.prototype, "onSubmitNewConversation").and.callThrough();
+      this.target = new app.views.ConversationsForm();
+    });
```
svbergerem: Please add an empty line between the `beforeEach` and the `it` block.

svbergerem (changes requested):

spec/javascripts/app/views/conversations_inbox_view_spec.js
svbergerem: Ping.

config/locales/javascript/javascript.en.yml
AugierLe42e: Ah, ok. I didn't have that translation problem in mind. Thx.

AugierLe42e:
Done.l

spec/javascripts/app/views/conversations_inbox_view_spec.js
svbergerem: Nevermind, I fixed it in https://github.com/diaspora/diaspora/pull/7108/commits/8579b974b79894eb2fbdfd65eead522fe9d2155a.

svbergerem:
Thank you.

PR #7115, opened 2016-09-20
Fix 20160906225138_fix_photos_share_visibilities.rb for postgres

SuperTux88:
Turns out that postgres can't handle "0" as boolean, but mysql can
handle "false".


skyschub (approved):

denschub:
Merged as b4455aaf25a484b72d4df1795cb1f118b24d3434.

PR #7118, opened 2016-09-21
Implement 4319 postdefault

hilkoc:
For your consideration, my contribution for issue #4319 

In this change we add an aspect drop down selector on the user settings page. After selecting and saving these post_default aspects they will be used in the main stream's publisher for preselecting the aspects to post to. To that end we refactor the aspect_dropdown view to not depend on a stream and publisher.
Selecting multiple aspects from the dropdown is not possible in mobile view. Therefore this setting is not available in mobile view.

In addition, we preselect the 'public' aspect in the publisher for the public activities stream. This makes the behavior analogous to the My Aspects stream, where we preselect the aspects of the aspect stream.

The schema migration preserves the existing behavior of preselecting 'All aspects'.
A cucumber test demonstrates the use of this new feature.

This is a follow up on https://github.com/diaspora/diaspora/pull/7101


app/views/aspects/_aspect_dropdown.html.haml
```diff
@@ -0,0 +1,44 @@
+-# locals: selected_aspects. Note: all_aspects is a global in the ApplicationController
+.btn-group.aspect_dropdown
+  %button.btn.btn-default.dropdown-toggle{!current_user.getting_started? ? {"data-toggle" => "dropdown"} : |
```
diaspora-code-review: Don't use the `|` character to split up lines. Wrap on commas or extract code into helper.

app/views/aspects/_aspect_dropdown.html.haml
```diff
@@ -0,0 +1,44 @@
...
+        - if all_aspects_selected?(all_aspects, selected_aspects)
+          = t("all_aspects")
+        - elsif selected_aspects.size == 1
+          = selected_aspects.first.name
+        - else
+          = t("shared.aspect_dropdown.toggle", count: selected_aspects.size)
+    %span.caret
+  %ul.dropdown-menu.pull-right{unSelectable: "on"}
+
+    %li.public.radio{"data-aspect_id" => "public", class: ("selected" if public_selected?(selected_aspects))}
```
diaspora-code-review: Don't mix styles in the same hash.

app/views/aspects/_aspect_dropdown.html.haml
```diff
@@ -0,0 +1,44 @@
...
+      %a
+        %span.status_indicator
+          %i.glyphicon.glyphicon-ok
+        %span.text
+          = t("public")
+    %li.all_aspects.radio{"data-aspect_id" => "all_aspects", class: ("selected" if (!public_selected?(selected_aspects) |
```
diaspora-code-review: - Line is too long. [121/120]
- Don't mix styles in the same hash.
- Don't use parentheses around the condition of an `if`.
- Don't use the `|` character to split up lines. Wrap on commas or extract code into helper.

app/views/aspects/_aspect_dropdown.html.haml
```diff
@@ -0,0 +1,44 @@
...
+      && all_aspects_selected?(all_aspects, selected_aspects)))} |
+      %a
+        %span.status_indicator
+          %i.glyphicon.glyphicon-ok
+        %span.text
+          = t("all_aspects")
+    %li.divider
+    - for aspect in all_aspects
```
diaspora-code-review: Prefer `each` over `for`.

app/views/aspects/_aspect_dropdown.html.haml
```diff
@@ -0,0 +1,44 @@
...
+      %li.aspect_selector{"data-aspect_id" => aspect.id, class: !all_aspects_selected?(all_aspects, selected_aspects) |
```
diaspora-code-review: - Don't mix styles in the same hash.
- Don't use the `|` character to split up lines. Wrap on commas or extract code into helper.

features/step_definitions/aspects_steps.rb
```diff
@@ -4,7 +4,11 @@ def click_aspect_dropdown
   end
 
   def toggle_aspect(a_name)
-    a_id = @me.aspects.where(name: a_name).pluck(:id).first
+    if "Public" == a_name
+      a_id = "public"
+    else
```
diaspora-code-review: Use the return of the conditional for variable assignment and comparison.

hilkoc:
Hold on your reviews, I've got some tests to fix...

CSammy:
You can [run the tests locally](https://wiki.diasporafoundation.org/Testing_workflow) if you prefer. This will speed up your testing process massively, trust me. Especially the sections "rspec" and "cucumber" might be interesting to you. If you want to avoid further style issues, you can try to [check against the style guide locally](https://wiki.diasporafoundation.org/Styleguide) locally as well. Thank you very much for contributing! ([Further information for contributors.](https://wiki.diasporafoundation.org/Getting_started_with_contributing))

denschub:
> Hold on your reviews, I've got some tests to fix...

Also, please add "[WIP]" to your title so we don't start reviewing the PR, remove the title prefix when you're done and leave a comment so we're aware that we can start review. :)

hilkoc:
This PR is now ready for review.
Thanks @cmrd-senya , @CSammy , @denschub  for your feedback.

svbergerem (changes requested):

svbergerem:
This PR only changes the default aspects for the main stream. I think it would be good to do the same for all other streams except for the aspect streams and the public stream.

When changing streams (except for changing between followed tags and other streams) we don't reload the page but reload the stream instead. In this case we keep the last aspect selection. I think we should set the aspect selection to the default aspect selection for the stream client side. (public stream -> public, aspect streams -> selected aspects, other streams -> default aspect selections) Otherwise the selected aspects would depend on the stream that you visited first and I think that would be confusing.

hilkoc:
Now the default aspects are used for all other streams except for the aspect stream and the public stream.

When changing streams aspect selection is preserved. This is the current behavior that users are used to. I'd be happy to change that, but I'd need some help. I have no idea how to do this (other than a full page reload).

Asgatoril:
Hi,
since I was searching for that exact change, I pulled out the tree in the state of your latest commit and did a full install in my testing environment with it. Unfortunately I get thrown an error 500 when trying to login.

Since it seems to work for you, do you have any ideas why this happens in my install?

Crashlog:
```
[2016-10-04T13:54:04] INFO  PID-21650 TID-9195760 ActionController::Base: Completed 500 Internal Server Error in 431ms (ActiveRecord: 14.3ms)
[2016-10-04T13:54:04] FATAL PID-21650 TID-9195760 Rails:
ActionView::Template::Error (undefined local variable or method 'post_default_public' for #<User:0x0000000560cd18>
Did you mean?  post_default_aspects):
     5: .container-fluid.main-stream-publisher
     6:   .pull-left.hidden-xs
     7:     = owner_image_link
     8:   = render "publisher/publisher", publisher_aspects_for(stream)
     9:
    10: - if current_user.getting_started?
    11:   .stream#main_stream{:title => popover_with_close_html("3. #{t('.stay_updated')}"),
  app/models/user.rb:256:in 'post_default_aspects'
  lib/stream/base.rb:65:in 'aspects'
  app/helpers/aspect_global_helper.rb:16:in 'publisher_aspects_for'
  app/views/aspects/_aspect_stream.haml:8:in '_app_views_aspects__aspect_stream_haml___1027612392941185520_78911060'
  app/views/streams/main_stream.html.haml:164:in '_app_views_streams_main_stream_html_haml___63589974447644214_50767600'
  app/controllers/streams_controller.rb:60:in 'block (2 levels) in stream_responder'
  app/controllers/streams_controller.rb:59:in 'stream_responder'
  app/controllers/streams_controller.rb:31:in 'multi'
  lib/rack/internet_explorer_version.rb:34:in 'call'
```

hilkoc:
This change involves a database migration. Try updating your dev database.
There should be a new column called 'post_default_public' appearing in the users table (and a new 'post_default' column in the aspects table).

svbergerem:
@hilkoc 

> When changing streams aspect selection is preserved. This is the current behavior that users are used to. I'd be happy to change that, but I'd need some help. I have no idea how to do this (other than a full page reload).

Well, I guess you are right. I'd say let's keep the current behavior for now. The solutions that I have in my mind aren't really nice. One could modify the aspect dropdown with some JavaScript each time we change the stream but that's not nice. For another solution we would have to render the aspect dropdown client side. Doing that would be great but that's out of scope of your PR.

Just a quick reminder: I added some remarks a few days ago: https://github.com/diaspora/diaspora/pull/7118#pullrequestreview-1898836. Apart from those this PR looked good to me.

hilkoc:
@svbergerem Thanks for getting back to this. The [last commit ](https://github.com/diaspora/diaspora/pull/7118/commits/50c9e55502e4226141f8403233a8b4d53972942d) ensures that the default aspects are used for all streams, except the public and aspect stream.
I'll rebease this branch on latest develop branch as soon as I find the time for that.

app/assets/javascripts/app/pages/settings.js
```diff
@@ -9,6 +9,12 @@ app.pages.Settings = Backbone.View.extend({
       preFill: gon.preloads.tagsArray
     });
     new Diaspora.ProfilePhotoUploader();
+
+
```
diaspora-code-review: More than 1 blank line not allowed.

svbergerem (approved):

svbergerem:
@denschub @SuperTux88 It would be great if you could decide if the inline ruby code is ok and if the migrations are suitable for a minor release. Everything else looks good to me.

SuperTux88:
The inline ruby looks better than before and I think it is OK for now, maybe we can refactor that sometime.

I tested the migration with a backup of the nerdpol.ch database:

`== 20160829170244 AddPostDefaultToAspects: migrated (2.3335s) =================`

I think that is OK for a minor release :)

I squashed your commits to a single one, so merged as b7791e6c9db0d202bceeb3d1c649874f9a66a5a3

Thank you very much for your contribution! This was a long awaited feature. Also don't forget to claim your bounty: https://www.bountysource.com/issues/618433-user-setting-for-default-post-visibility

hilkoc:
I'm happy to see that this is accepted and merged in. Thanks everyone!

Flaburgan:
@hilkoc hey what's your diaspora* Id btw I want to mention you to say thanks but can't find you.

hilkoc:
@Flaburgan , so far I've only created a dummy Diaspora account for testing. Once I've got my own pod setup I'll create a 'real' profile and post back here.

PR #7122, opened 2016-09-25
Add In-Reply-To and References headers to notification mails

skyschub:
This allows mail clients to collapse mails regarding the same object. Likes, comments and reshares will be grouped by the respective post, private messages will be grouped by the containing conversation.

### DIFF
--- a/app/mailers/notification_mailers/also_commented.rb
+++ b/app/mailers/notification_mailers/also_commented.rb
@@ -8,6 +8,7 @@ module NotificationMailers
 
       if mail?
         @headers[:from] = "\"#{@comment.author_name} (diaspora*)\" <#{AppConfig.mail.sender_address}>"
+        @headers[:in_reply_to] = @headers[:references] = "<#{@comment.parent.guid}@#{AppConfig.pod_uri.host}>"
         if @comment.public?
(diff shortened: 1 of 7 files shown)


app/mailers/notification_mailers/also_commented.rb
```diff
@@ -8,6 +8,7 @@ def set_headers(comment_id)
 
       if mail?
         @headers[:from] = "\"#{@comment.author_name} (diaspora*)\" <#{AppConfig.mail.sender_address}>"
+        @headers[:in_reply_to] = @headers[:references] = "<#{@comment.parent.guid}@#{AppConfig.bare_pod_uri}>"
```
SuperTux88: If the pod doesn't run on the default port (80/443) `AppConfig.bare_pod_uri` returns the domain with `:port`. But if I understand the [specs](https://tools.ietf.org/html/rfc2822#section-3.2.4) correctly, `:` is not a valid character. My suggestion is to only use the domain without the port.

denschub: Fair enough, changed.

SuperTux88:
Merged as dfb5b717b688207da2d56f6c7eac7842fd664e41

Thanks ❤️

PR #7125, opened 2016-09-25
Set redis url for sidekiq cleanup migration task

SuperTux88:
See: https://nerdpol.ch/posts/24d869e0655601349a46001d92b033b9


denschub:
Merged as 0a264c1fe0bc960e5aea091b5f66e48e4ef06b00, thanks.

SuperTux88:
@jaywink (and other people with multiple pods on the same server, but Jason is the only one that I know ;) ):

You probably want to run `RAILS_ENV=production bin/rake migrations:legacy_queues` again (after pulling this PR) on your second pod, because the second pod ignored the redis-url for this migration and probably cleaned up the sidekiq/redis of the first pod.

PR #7128, opened 2016-09-29
Add Content Security Policy header

SuperTux88:
CSP allows you to control from where resources are allowed to be loaded. This improves security, because it helps to detect and mitigate cross-site scripting and data injection attacks. (https://developer.mozilla.org/en/docs/Web/Security/CSP/CSP_policy_directives)

I tested this policy on my pod (https://nerdpol.ch) the last 3 weeks and it works really well.

I added options to the config to enable it (disable report-only mode) and set a report-uri.

All third party domains (e.g. oembed) that diaspora uses are allowed by default. The only thing that doesn't work is google analytics and piwik, because the gems that diaspora uses for this inject &lt;script&gt; tags without the `nonce`-attribute. (But since it is report-only by default, I don't see this as a problem, people who use google analitycs or piwik simply can't use CSP).

Sadly we can't remove the `unsafe-eval`, because underscore.js needs it (and backbone.js uses underscore.js).


app/views/conversations/new.mobile.haml
```diff
@@ -2,25 +2,26 @@
 -#   licensed under the Affero General Public License version 3 or later.  See
 -#   the COPYRIGHT file.
 
-:javascript
-  $(document).ready(function () {
```
diaspora-code-review: - Hash attribute should start with no space after the opening brace
- Hash attribute should end with no space before the closing brace

app/views/conversations/new.mobile.haml
```diff
@@ -2,25 +2,26 @@
...
+    $(document).ready(function () {
+      var data = $.parseJSON( "#{escape_javascript(@contacts_json).html_safe}" ),
+          autocompleteInput = $("#contact_autocomplete");
 
-    autocompleteInput.autoSuggest(data, {
-      selectedItemProp: "name",
-      searchObjProps: "name",
-      asHtmlID: "contact_ids",
-      retrieveLimit: 10,
-      minChars: 1,
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

config/diaspora.yml.example
```diff
@@ -551,6 +551,23 @@ configuration: ## Section
       #title: 'diaspora* social network'
       #description: 'diaspora* is the online social world where you are in control.'
 
+    ## CSP (Content Security Policy) header
+    ## CSP allows you to control from where resources are allowed to be loaded. This improves security,
+    ## because it helps to detect and mitigate cross-site scripting and data injection attacks.
+    ## The default policy of diaspora allows all third party domains that diaspora uses, so you can safely
+    ## activate it (set "report_only" to false). If you customized diaspora (edited templates or added own JS),
+    ## you maybe need additional work. You can test the policy with the "report_uri".
+    ## CSP doesn't work with google analytics or piwik, because they inject JS code that is blocked by CSP.
```
denschub: CSP allows limiting origins from where resources are allowed to be loaded. This
    improves security, since it helps to detect and mitigate cross-site scripting
    and data injection attacks. The default policy of diaspora* allows all third
    party domains from services that are included in diaspora*, like OEmbed
    scripts, so you can safely activate it by setting `report_only` to false. If
    you customized diaspora* (edited templates or added own JS), additional work
    may be required. You can test the policy with the "report_uri". Our default CSP
    does not work with Google analytics or Piwik, because they inject JS code that
    is blocked by CSP.

config/diaspora.yml.example
```diff
@@ -551,6 +551,23 @@ configuration: ## Section
...
+    csp:
+      ## Report-Only header (default=true)
+      ## By default diaspora adds only a "Content-Security-Policy-Report-Only" header. If you change
```
denschub:
```diff
-diaspora
+diaspora*
```

```diff
-If you change
+If you set
```

config/diaspora.yml.example
```diff
@@ -551,6 +551,23 @@ configuration: ## Section
...
+      ## this to false, the "Content-Security-Policy" header is added instead.
+      #report_only: false
+
+      ## CSP report uri (default=)
```
denschub:
```diff
-uri
+URI
```

skyschub (changes requested):
Looking good! I've nitpicked at your documentation, but the rest looks fine.

SuperTux88:
Thanks for the review, I updated the PR.

skyschub (approved):

denschub:
Config changes do not require podmin intervention, the defaults are sane. Merged to `next-minor`. Thanks much. ❤️

Flaburgan:
I guess this will only work correctly if camo is enabled, right? Or CSP applies only to script files and not to images so embedding with markdown is still fine?

denschub:
Images are allowed from all sources, regardless if it's http or https.

Issue #5091, opened 2014-07-28
Can not pick one of several contacts with the same name when sending a message


akerbos:
As it so happens, you can have multiple contacs with the same screen name: maybe a friend has old or multiple accounts, or two people just share a name.

However, when composing a message only the screen name is shown; neither avatar nor account name nor anything else are accessible. It is thus _impossible_ to pick the correct contact.

Please add (a small version of) the avatar and the account name to this and similar quicksearch result lists.
##

svbergerem:
Fixed by #7129.

Issue #7002, opened 2016-08-18
Contact status not validated when creating/receiving conversations


svbergerem:
We currently have [a validation in the conversation model](https://github.com/diaspora/diaspora/blob/c8d43a8/app/models/conversation.rb#L20) to check that a contact for the conversation author exists but we don't check that we are actually sharing with the author.

@SuperTux88 told me that he'd prefer to add a fix to the federation at the same time to make sure that we don't drop conversations without the author and other participants noticing it.

The code I had so far is the following:

```diff
diff --git a/app/models/conversation.rb b/app/models/conversation.rb
index 4802d24..36e90e6 100644
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -17,7 +17,7 @@ class Conversation < ActiveRecord::Base
   def local_recipients
     recipients.each do |recipient|
       if recipient.local?
-        unless recipient.owner.contacts.where(person_id: author.id).any? ||
+        unless recipient.owner.contacts.receiving.where(person_id: author.id).any? ||
             (author.owner && author.owner.podmin_account?)
           errors.add(:all_recipients, "recipient not allowed")
         end
diff --git a/spec/models/conversation_spec.rb b/spec/models/conversation_spec.rb
index 5bb08c2..4a64f9e 100644
--- a/spec/models/conversation_spec.rb
+++ b/spec/models/conversation_spec.rb
@@ -72,30 +72,56 @@ describe Conversation, :type => :model do
 
   describe "#invalid parameters" do
     context "local author" do
-      let(:invalid_hash) {
-        {author: peter.person, participant_ids: [peter.person.id, user1.person.id],
+      let(:conversation_hash) {
+        {author: peter.person, participant_ids: [peter.person.id, @recipient.person.id],
           subject: "cool stuff", messages_attributes: [{author: peter.person, text: "hey"}]}
       }
 
-      it "is invalid with invalid recipient" do
-        invalid_conversation = Conversation.create(invalid_hash)
+      it "is invalid if the recipient is not a contact" do
+        @recipient = alice
+        invalid_conversation = Conversation.create(conversation_hash)
         expect(invalid_conversation).to be_invalid
       end
+
+      it "is invalid if the recipient is not sharing with the author" do
+        @recipient = FactoryGirl.create(:user)
+        peter.share_with(@recipient.person, peter.aspects.first)
+        invalid_conversation = Conversation.create(conversation_hash)
+        expect(invalid_conversation).to be_invalid
+      end
[...]

PR #7129, opened 2016-09-30
Use typeahead on conversations

ghost:
Typeahead results:
screen shot 2016-09-30 at 14 15 46
Conversation with one recipiant:
screen shot 2016-09-30 at 14 09 30
When recipient tag is hovered:
screen shot 2016-09-30 at 14 09 47

### DIFF
--- a/app/assets/stylesheets/conversations.scss
+++ b/app/assets/stylesheets/conversations.scss
@@ -183,10 +183,60 @@
 }
 // scss-lint:enable SelectorDepth
 
-#new_conversation_pane {
+.new-conversation {
   ul.as-selections { width: 100% !important; }
+
   input#contact_ids { box-shadow: none; }
+
   label { font-weight: bold; }
+
+  .twitter-typeahead,
+  .tt-menu {
+    width: 100%;
+  }
+}
+
+.recipients-tag-list {
+  .conversation-recipient-tag {
+    background-color: $brand-primary;
+    border-radius: $btn-border-radius-base;
+    display: inline-flex;
+    margin: 0 2px $form-group-margin-bottom;
+    padding: 8px;
+
+    &:first-child { margin-left: 0; }
+
+    &:last-child { margin-right: 0; }
+
+    div {
+      align-self: center;
+      justify-content: flex-start;
+    }
+  }
+
+  .avatar {
+    height: 40px;
+    margin-right: 8px;
+    width: 40px;
+  }
+
+  .name-and-handle {
+    color: $white;
+    margin-right: 8px;
+    text-align: left;
+
+    .diaspora-id { font-size: $font-size-small; }
+  }
+
+  .entypo-circled-cross {
+    color: $white;
+    cursor: pointer;
+    font-size: 20px;
+    height: 22px;
+    line-height: 22px;
+
+    &:hover { color: $light-grey; }
+  }
 }
 
 .new-conversation.form-horizontal .form-group:last-of-type { margin-bottom: 0; }
--- a/app/assets/templates/conversation_recipient_tag_tpl.jst.hbs
+++ b/app/assets/templates/conversation_recipient_tag_tpl.jst.hbs
@@ -1,0 +1,12 @@
+<div class="conversation-recipient-tag clearfix" data-diaspora-handle="{{ handle }}">
+  <div href="{{ url }}">
+    <img src="{{ avatar }}" class="avatar img-responsive center-block">
+  </div>
+  <div class="pull-left clearfix name-and-handle" href="{{ url }}">
+    <div class="name">{{ name }}</div>
+    <div class="diaspora-id">{{ handle }}</div>
+  </div>
+  <div class="remove pull-right clearfix">
+    <i class="entypo-circled-cross"></i>
+  </div>
+</div>
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -30,12 +30,12 @@ class ConversationsController < ApplicationController
   end
 
   def create
-    contact_ids = params[:contact_ids]
-
-    # Can't split nil
-    if contact_ids
-      contact_ids = contact_ids.split(',') if contact_ids.is_a? String
-      person_ids = current_user.contacts.where(id: contact_ids).pluck(:person_id)
+    # Contacts autocomplete does not work the same way on mobile and desktop
+    # Mobile returns contact ids array while desktop returns person id
+    # This will have to be removed when mobile autocomplete is ported to Typeahead
+    recipients_param, column = [%i(contact_ids id), %i(person_ids person_id)].find {|param, _| params[param].present? }
+    if recipients_param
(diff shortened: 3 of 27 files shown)


Flaburgan:
Those are design propositions?

AugierLe42e:
Yep. We can discuss the design of the tags.

Flaburgan:
I'll be in favour of keeping the design as close as possible of the mention one. Unifying, you know :p

AugierLe42e:
Ready for review.

svbergerem:
I'd prefer a representation of the recipients that is similar to the representation in the dropdown: name and handle left aligned, the handle smaller than the name. The margins for the avatar should all be the same. (Currently the left margin is bigger.) I'd also like a smaller close icon that is in the right upper corner and uses an Entypo icon.

AugierLe42e:
Yes, you're right, that's better:

screen shot 2016-10-01 at 08 51 54

AugierLe42e:
BTW, the pills also show hovercard when hovered. I though it would be handy.
screen shot 2016-10-03 at 10 16 45

spec/controllers/conversations_controller_spec.rb
```diff
@@ -115,203 +124,323 @@
...
+          expect(response).not_to be_success
+          expect(response.body).to eq(I18n.t("javascripts.conversation.create.no_recipient"))
+        end
       end
 
-      it "responds with an error message" do
-        post :create, @hash
-        expect(response).not_to be_success
-        expect(response.body).to eq(I18n.t("conversations.create.fail"))
+      context "with nil contact"  do
```
diaspora-code-review: Unnecessary spacing detected.

AugierLe42e:
Rebased and ready for another review.

app/assets/stylesheets/conversations.scss
```diff
@@ -183,10 +183,56 @@
...
+    }
+  }
+
+  .avatar {
+    height: 40px;
+    margin-right: 8px;
+    width: 40px;
+  }
+
+  .pseudo-and-handle {
```
svbergerem: `.name-and-handle`?

app/assets/templates/conversation_recipient_tag_tpl.jst.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="conversation-recipient-tag clearfix btn btn-primary" data-diaspora-handle="{{ handle }}">
```
svbergerem: This is not a button so I don't like using the `btn` class here. Changing the color on hover and showing the cursor as a pointer gives me the impression that you can click on it to change something. I'd rather just have the same background as the primary buttons.

app/views/conversations/_new.haml
```diff
@@ -2,24 +2,29 @@
...
+      .recipients-tag-list.clearfix#recipients-tag-list
+      = text_field_tag "contact_autocomplete", nil, id: "contacts-search-input", class: "form-control"
+      - unless defined?(mobile) && mobile
+        = text_field_tag "person_ids", nil, id: "as-values-contact-ids", type: "hidden",
+                                            aria: {labelledby: "to-label"}
+
     .form-group
-      %label#subject-label{for: "conversation-subject"}
-        = t(".subject")
+      %label#subject-label{for: "conversation-subject"}= t(".subject")
```
svbergerem: I think this was more readable before.

app/views/conversations/_new.haml
```diff
@@ -2,24 +2,29 @@
...
     class: "new-conversation form-horizontal"}, remote: true do |conversation|
     .form-group
-      %label#toLabel{for: "contact_ids"}
-        = t(".to")
-      = text_field_tag "contact_autocomplete", nil, id: "contact-autocomplete", class: "form-control"
+      %label#to-label{for: "contacts-search-input"}= t(".to")
+      .recipients-tag-list.clearfix#recipients-tag-list
+      = text_field_tag "contact_autocomplete", nil, id: "contacts-search-input", class: "form-control"
+      - unless defined?(mobile) && mobile
+        = text_field_tag "person_ids", nil, id: "as-values-contact-ids", type: "hidden",
```
svbergerem: Why `as-values-contact-ids`? I thought `as` would be `autoSuggest`.

app/controllers/conversations_controller.rb
```diff
@@ -91,15 +91,20 @@ def new
...
 
-    @contacts_json = contacts_data.to_json
-    @contact_ids = ""
-
     if params[:contact_id]
-      @contact_ids = current_user.contacts.find(params[:contact_id]).id
+      gon.push conversation_prefill: [current_user.contacts.find(params[:contact_id]).person.as_json]
     elsif params[:aspect_id]
-      @contact_ids = current_user.aspects.find(params[:aspect_id]).contacts.map{|c| c.id}.join(',')
+      gon.push conversation_prefill: current_user.aspects.find(params[:aspect_id]).contacts.map {|c| c.person.as_json }
```
svbergerem: Do we need these for the mobile view, too? Otherwise you could move this down into the `else` branch.

app/controllers/conversations_controller.rb
```diff
@@ -30,12 +30,12 @@ def index
...
-    # Can't split nil
-    if contact_ids
-      contact_ids = contact_ids.split(',') if contact_ids.is_a? String
-      person_ids = current_user.contacts.where(id: contact_ids).pluck(:person_id)
+    # Contacts autocomplete does not work the same way on mobile and desktop
+    # Mobile returns contact ids array while desktop returns person id
+    # This will have to be removed when mobile autocomplete is ported to Typeahead
+    recipients_param, column = [%i(contact_ids id), %i(person_ids person_id)].find {|param, _| params[param].present? }
+    if recipients_param
+      person_ids = current_user.contacts.where(column => params[recipients_param].split(",")).pluck(:person_id)
```
svbergerem: Shouldn't we use `current_user.contacts.mutual...` here?

svbergerem (changes requested):
Looks quite good so far. I reviewed everything except for the tests.

app/controllers/conversations_controller.rb
AugierLe42e: Dunno. @jhass wrote this.

svbergerem: This is related to #7002. The UI already doesn't allow you to create a conversation with non-mutual contacts and doing the same in the controller would be the next logical step.

But that would be another change of behavior that would need some additional tests. So I'd say let's keep it like it is and change it to mutual contacts later.

AugierLe42e:
Done.

svbergerem (approved):

SuperTux88:
Merged.

Thanks @AugierLe42e and @svbergerem for the work!

Flaburgan:
Awesome work guys :D

SansPseudoFix:
Nice!
Fixes #5091, no?

svbergerem:
@SansPseudoFix Yes, thank you!

PR #7130, opened 2016-09-30
Link to diaspora* admin dashboard

SuperTux88:


svbergerem:
> You can find instructions in the diaspora* wiki. This should add an “Admin” link to your user menu in the header when you are logged in. It gives you stuff like user search and stats for your pod. For intense detail on the operational aspects of your pod, go to the admin panel.

We explain the admin pages with the admin dashboard first and then we talk about “intense detail on the operational aspects”. I also have my doubts about linking to the admin panel but if we change the link then we should also change the text.

SuperTux88:
Thanks for the hint, I updated the PR.

svbergerem:
Thank you!

PR #7133, opened 2016-09-30
Unify new conversation margins

svbergerem:
**before**
before

**after**
after


svbergerem:
I added another commit that removes the extra padding from the comment form in the stream. (Thanks @SuperTux88 for the hint)

SuperTux88:
Merged as 471af747b1dcd016e0cc7bb8e665741782bf8a08 2a31c2a6dbd4194c0390090d8b52b4e775b1ed5b

Thanks ❤️

PR #7137, opened 2016-10-02
Fix message modal on profile page

### LINKED ISSUE #7136: Private Message gets deleted if you click 'send' and didn't enter a recipient
If I click on the 'message' button on a diaspora profile of one of my contacts, there is no recipient entered automaticly on the appearing private message form. And there is no auto-search for contacts, like on the 'private message'-page.

If I write some text in "message" and "subject", but don't enter a recipient and click on 'send', the whole message + subject gets deleted.

[Maybe we could fix that, by linking the "private message"-button to the private message page and enter the profile owner as a recipient automaticly, so then we have to support only one private message editor/publisher.]

svbergerem:
Fixes  #7136.

### DIFF
--- a/app/views/conversations/new.html.haml
+++ b/app/views/conversations/new.html.haml
@@ -2,6 +2,7 @@
   $(document).ready(function () {
     var data = $.parseJSON( "#{escape_javascript(@contacts_json)}" );
     new app.views.ConversationsForm({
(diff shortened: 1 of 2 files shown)


SuperTux88:
Merged as 74fff52e512e4342d6f7529927ae66eddb520792

Thanks

PR #7143, opened 2016-10-11
Added button to mobile stream. 

Muhannes:
Button to stream is now available in the dropdown menu in mobile view.
Issue: https://github.com/diaspora/diaspora/issues/7087

PR #7144, opened 2016-10-11
Added mobile buttons for streams

Muhannes:
Added two buttons in the mobile drawer. One to stream, and one to public stream.

### DIFF
--- a/app/assets/stylesheets/mobile/header.scss
+++ b/app/assets/stylesheets/mobile/header.scss
@@ -149,7 +149,7 @@ $mobile-navbar-height: 46px;
 
     li {
       font-size: 1.8rem;
-      line-height: 25px;
+      line-height: 2rem;
       color: $light-grey;
       border-bottom: solid $navbar-inverse-border 1px;
     }
@@ -179,7 +179,7 @@ $mobile-navbar-height: 46px;
     .avatar {
       height: 35px;
       width: 35px;
-      margin: -5px 5px;
+      margin: -7px 5px;
       float: right;
     }
   }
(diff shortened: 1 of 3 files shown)


Flaburgan:
Don't worry about your branch, it was a hint for the next time. Since you opened this one from another branch, let's continue here and I'll close #7143 

So, I tested this branch on my Sony Z3C running Firefox OS (I don't have any Android device). To reduce the font-size is not the way to go, sorry, because the line-height is set just [the line under](https://github.com/diaspora/diaspora/blob/1fdc17d96cf5190a9e9ac069510c1985608a7a58/app/assets/stylesheets/mobile/header.scss#L152). So you can keep the font-size to 1.8rem but set the `line-height` to `2rem` (which means 20px on my phone, instead of 25) and it will look fine.
Doing that you also need to modify the [avatar margin](https://github.com/diaspora/diaspora/blob/1fdc17d96cf5190a9e9ac069510c1985608a7a58/app/assets/stylesheets/mobile/header.scss#L182) setting -5 to -7 to keep it aligned. This code is a little bit dirty I guess we should clean it. Apart from that the code looks good to me, so it should fine to merge once you corrected these remarks.

Muhannes:
Ok, now it is fixed @Flaburgan.

Flaburgan (approved):

Flaburgan:
@svbergerem this looks okay to me, do you want to merge it?

svbergerem:
Merged as c5b133d3e86fb0a6a822588e31facdc6db6d37db. Thank you! :)

PR #7145, opened 2016-10-15
Remove empty pod and make pod host NOT NULL

SuperTux88:
Somehow this empty pod is being created, and I don't see where it comes from (it doesn't have any connections). So a `NOT NULL` prevents creating it again.


Flaburgan:
So with that change you will have a trace in the postgres log when the constraint will block the insert or update. Did that happen to your pod yet? If so, did you find from where this come from?

SuperTux88:
> Did that happen to your pod yet?

No, but I deleted the empty pod over a month ago, and saw now, that it is here again after multiple weeks (that's why I created this PR). So it only happens rarely, but it happens. Maybe it happens more often on a bigger pod. But if the `NOT NULL` blocks something, I can find the cause with the stacktrace, because I didn't find what could do this with reading the code.

But regardless of this, the `NOT NULL` should be there anyway.

SuperTux88:
Finally it happened again on my pod and I had a trace \o/ 

It came from the `:update_pod` callback while sending a public post to `https://despora.de/receive/public`, so somehow `response.effective_url` sometimes returns an invalid url without a host (which shouldn't happen and i still can't reproduce it). But I refactored the federation-gem to handle this case better and the callback shouldn't be called with an empty host now.

denschub:
Thanks for digging into that. Merged.

PR #7151, opened 2016-10-18
Bump secure_headers

SuperTux88:
Version 3.5 supports two CSP headers (enforced/report-only).

Also remove the `frame-src` hack.


denschub:
Sweet. Merged.

PR #7152, opened 2016-10-19
Dark theme

SuperTux88:
I created a dark theme for diaspora:

diaspora-dark

(If you have an account on nerdpol.ch you can also test it there.)

I also refactored the default theme a bit, to use more color-variables and less overrides (makes diaspora better themeable if you want to change more than the header color). And I renamed the default theme to "Original Gray" instead of "Original Dark" because I found it weird, because the default theme isn't "dark" (only the header is dark).

### DIFF
--- a/app/assets/stylesheets/color_themes/_color_theme_override_dark.scss
+++ b/app/assets/stylesheets/color_themes/_color_theme_override_dark.scss
@@ -1,0 +1,170 @@
+// Only overriding existing selectors here, so disable some lint rules
+// scss-lint:disable IdSelector, SelectorFormat, NestingDepth, SelectorDepth, QualifyingElement
+body {
+  .navbar.navbar-fixed-top #user_menu .dropdown-menu > li > a {
+    color: $text-color;
+    &:hover { color: $white; }
+  }
+
+  .publisher {
+    .mentions-input-box { background-color: $gray; }
+    form {
+      #publisher_textarea_wrapper { background-color: $gray; }
+      .btn.btn-link.question_mark:hover .entypo-cog { color: $gray-light; }
+    }
+
+    .write-preview-tabs > li.active * { color: $text-color; }
+    .md-preview { background-color: $gray; }
+    .md-cancel:hover .entypo-cross { color: $gray-light; }
+    .publisher-buttonbar .btn.btn-link:hover i { color: $gray-light; }
+  }
+
+  .aspect_dropdown li a .text { color: $dropdown-link-color; }
+
+  .info .tag { background-color: $gray-light; }
+
+  .poll_form .progress {
+    background-color: $gray-dark;
+    .bar { background-color: $gray-light; }
+  }
+
+  .stream-element .collapsible {
+    .markdown-content hr { border-top: 1px solid $hr-border; }
+    .expander {
+      @include linear-gradient(transparent, $gray-light, 0%, 95%);
+      border-bottom: 2px solid $gray-light;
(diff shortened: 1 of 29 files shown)


AugierLe42e:
Wow... That's... Really dark !

goobertron:
I think 'Default' might be a better name instead of 'Original [Dark/Gray]'. What do others think?

ps: you should make the text and images black as well for a *really* dark theme...

SuperTux88:
> Wow... That's... Really dark !

Yes, thats the intention of a dark theme ;)

> I think 'Default' might be a better name instead of 'Original [Dark/Gray]'. What do others think?

I don't know. Podmins can set another default-theme and I think "Default" as name can be weird then?

> ps: you should make the text and images black as well for a *really* dark theme...

No, I think  it's dark enough ;)

AugierLe42e:
> Yes, thats the intention of a dark theme ;)

Sad. I would really love to have a dark gray theme. But this one misses some contrast for me :(

SuperTux88:
I made the backgrounds a bit darker and the text-color lighter, so it have more contrast now.

screen shot 2016-11-05 at 19 46 16

I also updated nerdpol.ch with the new theme.

app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -176,7 +176,7 @@ footer {
     border-radius: 3px 3px 0 0;
 
     border: {
-      bottom: 1px solid #ccc;
+      bottom: 1px solid $border-medium-grey;
```
denschub: When you are already changing and adding variables... can we try to get a more generic naming? This border may not be grey and I'd rather have a generic name for the variables. Also, we should find a consent between grey and gray, I guess...

denschub:
This is nice.

app/assets/stylesheets/mobile/mobile.scss
SuperTux88: `$border-medium-grey` existed before, I'm only using it instead of the hardcoded color.

But I agree with you about the naming. I tried to use better names for new variables I created, but I didn't rename existing variables. I didn't want to change too much here, the PR is already big enough. I only used the existing variables instead of hardcoded colors, so I need less css-hacks here. I think the renaming should be done in another PR?

denschub: > I think the renaming should be done in another PR?

Well. This probably means it will never be done, but fair point. Okay. :)

SuperTux88: I love cleanups, so I'll do it sometime (if nobody else does it), but I'd rather not do it now as drive-by for a theme. ;)

app/assets/stylesheets/color_themes/_color_theme_override_dark.scss
```diff
@@ -0,0 +1,170 @@
...
+  #welcome-to-diaspora { background: $orange; }
+
+  .block-form fieldset .form-control:focus { border-color: $input-border; }
+
+  &.page-registrations.action-new,
+  &.page-registrations.action-create {
+    .ball { filter: invert(100%); }
+  }
+
+  .spinner { border-color: $gray-light transparent $gray-light $gray-light; }
```
diaspora-code-review: Shorthands of length `4` are not allowed. Value was `$gray-light transparent $gray-light $gray-light`

denschub:
Sweet stuff, merged. Thank you!

PR #7156, opened 2016-10-24
[API] pass "state" parameter back to openid client

### LINKED ISSUE #7052: [API] With authorization code flow the state parameter is missing in the callback URI
I'm playing around with omniauth with an aim to build an omniauth strategy for diaspora. I'm trying to reuse this library: https://github.com/jjbohn/omniauth-openid-connect.

The request to diaspora I managed to get generated was following:

```
GET "/api/openid_connect/authorizations/new?client_id=df90d24ecc6dde9d8bae32af2ac7fdff&nonce=ec1e2c533cea5ca445a82e78ceda6aac&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fusers%2Fauth%2Fdiaspora%2Fcallback&response_type=code&scope=openid+name+profile&state=d3f0755568c16544937a52a76fa1a130
```

There is `state=d3f0755568c16544937a52a76fa1a130` among the parameters.

The OpenID Connect specification in [this section](https://openid.net/specs/openid-connect-core-1_0.html#AuthResponse) says

> When using the Authorization Code Flow, the Authorization Response MUST return the parameters defined in Section 4.1.2 of OAuth 2.0 [RFC6749] by adding them as query parameters to the redirect_uri specified in the Authorization Request using the application/x-www-form-urlencoded format, unless a different Response Mode was specified.

Section [4.1.2](https://tools.ietf.org/html/rfc6749#section-4.1.2) of the OAuth 2.0 spec says that the `state` parameter is

> REQUIRED if the "state" parameter was present in the client authorization request.

However when I push the "Approve" button at the 

```
/api/openid_connect/authorizations/new?client_id=37c5044f1491004aaeb6f6e89c27c7cb&nonce=723fe2fc78594dbed50f9134d5a3306b&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fusers%2Fauth%2Fdiaspora%2Fcallback&response_type=code&scope=openid+name+profile&state=996af52c79f3ff5136c3ae92c776df44
```

I'm redirected to this callback URI

```
http://localhost:3000/users/auth/diaspora/callback?code=7fd09699ced2c27455533399e9c195d0ffaed0e7a3ffe265cbba9220c4cb517f
```

The `state`  parameter is clearly missing, and because of this [omniauth-openid-connect](https://github.com/jjbohn/omniauth-openid-connect) panics and returns 401 response to a client.

cmrd-senya:
fixes #7052 

The issue described in #7052 was supposed to be covered with a test at [this place](https://github.com/diaspora/diaspora/blob/9a09515955dd5629c5038969d3bd2d9a0bbbace8/spec/controllers/api/openid_connect/authorizations_controller_spec.rb#L340). However, controller functional testing approach based on `ActionDispatch::TestRequest` (as rspec `type: controller`) was used in this test, which is only valid when you make a single action query per example, and here it was different: `#new` was followed by `#create`. That's why the test didn't catch the issue: the test flow went quite different from the real word execution. I rewrote this test to use `ActionDispatch::IntegrationTest` based approach (as rspec `type: request`) which allows multiple queries per example.

Request helpers for `ActionDispatch::Integration` require usage of paths instead of action names (see [reference](http://api.rubyonrails.org/classes/ActionDispatch/Integration/RequestHelpers.html)). Though everything works as expected, I'm a little worried that these examples got less readable because of changing from action names to path names like

``` ruby
# was
            get :new, client_id: client.client_id, redirect_uri: "http://localhost:3000/", response_type: "id_token"
# now is
            get new_api_openid_connect_authorization_path, client_id: client.client_id,
                redirect_uri: "http://localhost:3000/", response_type: "id_token"
```

Is it an issue?

### DIFF
--- a/app/controllers/api/openid_connect/authorizations_controller.rb
+++ b/app/controllers/api/openid_connect/authorizations_controller.rb
@@ -125,6 +125,7 @@ module Api
         session[:response_type] = @response_type
         session[:redirect_uri] = @redirect_uri
         session[:scopes] = scopes_as_space_seperated_values
+        session[:state] = params[:state]
         session[:nonce] = params[:nonce]
       end
 
@@ -149,6 +150,7 @@ module Api
         session.delete(:response_type)
         session.delete(:redirect_uri)
         session.delete(:scopes)
+        session.delete(:state)
         session.delete(:nonce)
       end
 
@@ -162,6 +164,7 @@ module Api
         req.update_param("redirect_uri", session[:redirect_uri])
         req.update_param("response_type", response_type_as_space_seperated_values)
         req.update_param("scope", session[:scopes])
+        req.update_param("state", session[:state])
         req.update_param("nonce", session[:nonce])
       end
 
--- a/spec/controllers/api/openid_connect/authorizations_controller_spec.rb
+++ b/spec/controllers/api/openid_connect/authorizations_controller_spec.rb
@@ -1,6 +1,6 @@
 require "spec_helper"
 
-describe Api::OpenidConnect::AuthorizationsController, type: :controller do
+describe Api::OpenidConnect::AuthorizationsController, type: :request do
   let!(:client) { FactoryGirl.create(:o_auth_application) }
   let!(:client_with_xss) { FactoryGirl.create(:o_auth_application_with_xss) }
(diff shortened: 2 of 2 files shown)


denschub:
Merged as 31a31ecbe16eecf2709bc343f031e85b87e2cd87, thanks.

PR #7120, opened 2016-09-24
Don't federate to pods that are offline for more than two weeks

SuperTux88:
I think two weeks are enough, because if a pod is offline for more than 16 hours it misses stuff anyway.

Fixes #6220

PR #7158, opened 2016-10-25
Schedule a connection-check when receiving a message from an offline pod

SuperTux88:
This is an addition to #7120 and should help to bring pod faster back online if they were offline and we receive a message from them.

But I was unsure how to implement it. My first idea was to start an immediate check if the last check was older than an hour (or so), but sometimes a pod sends multiple messages in parallel (limited post to multiple recipients), and so we could start several checks in parallel for the same pod.

So I decided to create only a flag for the next check and another cron-job that checks all flagged pods.


denschub:
Merged, thanks!

PR #7160, opened 2016-10-27
Remove ability to authenticate with auth_token on the frontend

SuperTux88:
Also moved the token out of the API scope, because this token is only used for the chat, it isn't an official API.

Remove `devise-token_authenticatable` gem and only generate a token to be used by the chat.


denschub:
Gnah, good catch. I was supposed to do that before we release... 🙀 Merged.

PR #7162, opened 2016-10-28
Fix tag page when default aspect is "public" or a single aspect

SuperTux88:
And: Use default visibility also for bookmarklet


denschub:
Merged as 433d952eb5145845ad59c59d5df91241b1506c48, thank you *very much*.

PR #7163, opened 2016-10-28
Suppress deprecation warning: disable csp in report-only mode

SuperTux88:
Explicitly opt-out  `Content-Security-Policy` in report-only mode.

uplift? @denschub :)


denschub:
r+, uplift+ ;).

Merged.

PR #7164, opened 2016-10-30
Remove unused /user/getting_started_completed route

SuperTux88:
The used one is `/getting_started_completed`


denschub:
Merged!

PR #7175, opened 2016-11-06
[WIP] Add ES6 support

ghost:
This is a technical proposition to add support for ES6 features in diaspora using Babel transpiler, showing one real-life example. The goal is to progressively transpose some Backbone classes into ES6 classes. The main idea behind this move is to increase the quantity of code shared between mobile and desktop. For instance the [search_base_view](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/search_base_view.js), [search_view](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/search_view.js) and [mention_view](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/publisher/mention_view.js) could then be shared as well as some templates like [flash_messages](https://github.com/diaspora/diaspora/blob/develop/app/assets/templates/flash_messages_tpl.jst.hbs) and [mobile_alert](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/mobile/mobile_alert.js) could be merged.

The idea of this PR is to discuss pros and cons of this proposition.


app/assets/javascripts/helpers/markdown_editor.js.es6
```diff
@@ -1,11 +1,5 @@
-Diaspora.MarkdownEditor = function(element, opts) {
```
diaspora-code-review: Parsing error: The keyword 'class' is reserved

cmrd-senya:
The code definitely looks more readable with ES6!

Does it have performance concerns? I suppose the compilation from ES6 to
usual javascript happens during initial assets generation, so it
shouldn't have a big impact, right?

AugierLe42e:
I'm not quite sure. I suppose so as it relies on sprockets like, AFAIK, the JS, images ans SCSS asset generation.

AugierLe42e:
@svbergerem, @denschub I'd like to have your opinion on this.

svbergerem:
Well, I don't see us moving away from backbone.js right now. I talked to @denschub about this some time ago and I think the first step would be getting a proper API. Then we can do some frontend rewrites using our own API. Currently I'm thinking about React + Redux on top of ES2015+Babel. I never worked with React+Redux but @denschub seems to have a lot of experience working with it. (Maybe because of the Firefox debugger rewrite with React+Redux)

We decided to render desktop views on the client and mobile views on the server and I don't think we should render stuff on the client without a proper view library. Yes, there is some code that we can use for both desktop and mobile views right now and yes, I think we could start using ES2015 for that to make it easier to write code.

AugierLe42e:
The goal is not to move away from Backbone at all. I currently don't see ES6 as a replacement for Backbone but as a solution to the mobile view problem. The mobile view really lacks some important features that already exist and tested on the desktop view. I think some parts of the desktop code that are not tightly linked to Backbone could be easily ported to ES6 and shared with mobile view with almost no effort.

I don't know about React and Redux but I think adopting ES6 with Babel can be an immediate easy step that don't need to wait for an API. It will only affect a few components at first.

denschub:
I don't see how using modern language features has anything to do with the portability between mobile and desktop, or with the framework we use. No objections against adding babel as a preprocessor.

In any way, you updated sprockets from 2 to 3 as if that was no big deal. There's a [pretty large UPGRADEING doc](https://github.com/rails/sprockets/blob/3.x/UPGRADING.md). Have you checked it? Also, why not Sprockets 4? v4 ships with babel support embedded, rendering `sprockets-es6` obsolete.

Cukes are failing.

Gemfile
```diff
@@ -92,6 +92,11 @@ gem "jquery-ui-rails",   "5.0.5"
 gem "js_image_paths",    "0.1.0"
 gem "js-routes",         "1.2.9"
 
+# ES6
+gem "sprockets",            ">=3.0.0"
+gem "sprockets-es6",        "0.9.2"
+gem "babel-transpiler",     "0.7.0"
```
denschub: If we, for whatever reason, stick with Sprockets 3: `babel-transpiler` is a dependency of `sprockets-es6`. Why the manual fix?

AugierLe42e: I just followed a tutorial. This can probably be removed, then.
But, from what I've read, ES6 is fully supported by sprocket 4. So we wouldn't need `sprockets-es6` at all. The downside is that Sprockets 4 seems not backward compatible.

AugierLe42e:
> I don't see how using modern language features has anything to do with the portability between mobile and desktop, or with the framework we use.

Template strings can be used for simple HTML templates (like showed in the real-life exemple), and inheritence features can be used to decouple search_base_view, search_view and mention_view from Backbone, allowing us to use it on mobile view.

> In any way, you updated sprockets from 2 to 3 as if that was no big deal. There's a pretty large UPGRADING doc. Have you checked it? 

Nope. It's not the point of this PR ;)

> Also, why not Sprockets 4?

Isn't sprocket 4 still in alpha?

denschub:
As the dependency support looked different some years later, es6 support has been added in the latest `develop` branch as part of #7581. Closing this.

PR #7180, opened 2016-11-10
Report view

SansPseudoFix:
Before:
old_report_view

This PR:
new_report_view

I think it's more logical to display **reason** and **reported by** on the same line. Besides sometimes **Post** sections are displayed on 2 or 3 lines, and break headers.

I also now display full message on **Post** section, because, as a podmin, sometimes one and half line is not enough to understand report without clicking on it. I think it can be usefull for spam, for instance: you have the full message and don't have to click on every spam message to delete them.

I was thinking to truncate messages (20 lines?) but my ruby skills are very limited. Any help appreciated :)

### DIFF
--- a/app/assets/stylesheets/report.scss
+++ b/app/assets/stylesheets/report.scss
@@ -1,5 +1,9 @@
 #reports {
-  .reason {
+  .reason-label {
+    font-weight: bold;
+  }
+
+  .content {
     padding-bottom: 20px;
   }
   form input {
--- a/app/views/report/index.html.haml
+++ b/app/views/report/index.html.haml
@@ -17,11 +17,14 @@
               .panel-heading
                 .reporter.pull-right
                   = raw t("report.reported_label", person: link_to(username, user_profile_path(username)))
-                .title
-                  = report_content(report)
-              .panel-body
                 .reason
-                  = t("report.reason_label", text: report.text)
+                  %span.reason-label
+                    = t("report.reason_label")
+                  %span
+                    = report.text
+              .panel-body
+                .content
+                  = report_content(report)
 
                 = button_to t("report.reported_user_details"),
                   user_search_path(admins_controller_user_search: {guid: report.reported_author.guid}),
--- a/app/helpers/report_helper.rb
+++ b/app/helpers/report_helper.rb
@@ -6,7 +6,7 @@ module ReportHelper
   def report_content(report)
     case (item = report.item)
     when Post
-      raw t("report.post_label", title: link_to(post_page_title(item), post_path(item.id)))
+      raw t("report.post_label", content: link_to(post_message(item), post_path(item.id)))
     when Comment
       raw t("report.comment_label", data: link_to(
         h(comment_message(item)),
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -946,10 +946,10 @@ en:
 
   report:
     title: "Reports overview"
-    post_label: "<b>Post</b>: %{title}"
-    comment_label: "<b>Comment</b>:<br>%{data}"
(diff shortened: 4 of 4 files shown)


SuperTux88:
I think the Post/Comment should stay the "title" (because it is the main-item of the report). On bigger pods the same spam-post gets reported multiple times, it would be better to move the reporter down to the body and combine all reports for the same post/comment and make a list like:
* reporter 1: reason 1
* reporter 2: reason 2

SansPseudoFix:
> On bigger pods the same spam-post gets reported multiple times, it would be better to move the reporter down to the body and combine all reports for the same post/comment and make a list like:

Would be definitely great but totally out of my skills :)

AugierLe42e:
I think this is a good improvement. The post can be potentially long and then, it feels weird that it is the title rather than the content.

goobertron:
I agree with @SuperTux88, but combining reports on the same post is perhaps out of scope of this PR. It is important that the entire post be visible to the podmin, however. Perhaps a useful first step would be to add a roll-over behaviour so that the entire post appears in a pop-up?

denschub:
Oh god no please no popup.

I don't think it's needed to have a large area for the report reason. It's a small input box anyway, so people don't write essays there. The post is usually larger than the report reason, so using the larger area for the actual content seems reasonable. I'm fine with this from a UX perspective, but I have not done any code review.

AugierLe42e:
Code looks fine to me.

goobertron:
When I said 'I agree with SuperTux' above, I actually meant that I agree that combining reports on the same post would be good. The new UI, with the post in the body of the report, looks more natural to me, but I don't maintain a pod so can't comment whether it's as useful to podmins.

In order to combine reports as suggested, you could simply add extra lines in the grey header bar of the new UI, such as:

> Reason: offensive content         Reported by minus
> Reason: spam                           Reported by fred
> Reason: offensive content         Reported by Jane

with the post text in the white area below.

Would that be acceptable?

AugierLe42e:
I think it's a bit difficult for @SansPseudoFix and out of scope of this PR.

SansPseudoFix:
> I think it's a bit difficult for @SansPseudoFix and out of scope of this PR.

I can't agree more. This PR is about unification and "UX", that's all.

app/views/report/index.html.haml
```diff
@@ -17,11 +17,11 @@
               .panel-heading
                 .reporter.pull-right
                   = raw t("report.reported_label", person: link_to(username, user_profile_path(username)))
-                .title
-                  = report_content(report)
-              .panel-body
                 .reason
-                  = t("report.reason_label", text: report.text)
+                  = raw t("report.reason_label", text: report.text)
```
denschub: Huge XSS issue via the report reason.

skyschub (changes requested):

SansPseudoFix:
Ok, no more XSS issue :)

app/assets/stylesheets/report.scss
```diff
@@ -1,5 +1,8 @@
 #reports {
-  .reason {
+  .reason-label {
+    font-weight: bold;
+  }
```
diaspora-code-review: Rule declaration should be followed by an empty line

SansPseudoFix: oO

Flaburgan: There should be an empty line between `}` and the following `.content {`

skyschub (approved):

Flaburgan (approved):
Tested locally with Firefox, works fine.

SansPseudoFix:
Ping?

svbergerem:
Merged as 0f60b9fb949f51abde379e34e66ffabfca1a38df. Thank you!

PR #7181, opened 2016-11-11
Use string-direction gem for rtl detection

svbergerem:
https://pod.omgsrsly.net/posts/209854 has been detected as rtl before because it starts with a BOM. This PR fixes the problem.


SuperTux88:
Merged as 1863137161303e9460883daa3a4bb66d515e16ed

Thanks :)

PR #7183, opened 2016-11-12
Only reload profile header when changing aspect memberships

### LINKED ISSUE #7072: When I add someone to an aspect the page refreshes and I lose my work
Hello,

Today I was commenting on a post and I had written about 500 words. Before I finish my comment I thought it would be nice to add this person to my aspects so I can follow similar posts of him but doing so I lost my comment.

Maybe a solution is not to refresh the page when we add someone to an aspect but I guess it can be solved in many other ways.

Thanks.

svbergerem:
Fixes #7072.


SuperTux88:
Merged as a951c40ba0bb8fd2cae94959af607cc80290fa8a and 3930069e675253e3237b491cb8cb7be9751f91e1

Thanks for fixing this 👍

PR #7184, opened 2016-11-12
Reduce i18n.load side effects

svbergerem:


app/assets/javascripts/helpers/i18n.js
```diff
@@ -18,7 +18,7 @@ Diaspora.I18n = {
   },
 
   updateLocale: function(locale, data) {
-    locale.data = $.extend(locale.data, data);
+    locale.data = $.extend({}, locale.data, data);
```
Flaburgan: I'm curious, what's the improvement caused by this?

svbergerem: So the improvement is the one I checked in this test: https://github.com/diaspora/diaspora/pull/7184/files#diff-bc361aa440611f178379efdd1012b25cR63.

SuperTux88:
Merged as cdce25374ff1464296aa3e91d69740d51a9ba99e

Thank you.

PR #7191, opened 2016-11-17
fixing modal visibility issue

ghost:
As a user viewing my stream, when I click on "You can search or invite more contacts.", the modal backdrop appears but the modal content is invisible. This is because the element containing the modal in the sidebar is hidden until the "invite" section is uncollapsed.

This PR moves the modal to `:after_content`.

### DIFF
--- a/features/desktop/invitations.feature
+++ b/features/desktop/invitations.feature
@@ -39,10 +39,19 @@ Feature: Invitations
     And I click on selector "#invitations-button"
     Then I should see one less invite
 
-  Scenario: sends an invitation
+  Scenario: sends an invitation from the sidebar
     When I sign in as "alice@alice.alice"
     And I click on "Invite your friends" navbar title
     And I click on selector "#invitations-button"
+    And I fill in the following:
+      | email_inviter_emails         | alex@example.com    |
+    And I press "Send an invitation"
+    Then I should have 1 Devise email delivery
+    And I should not see "change your notification settings" in the last sent email
+
+  Scenario: sends an invitation from the stream
+    When I sign in as "alice@alice.alice"
+    And I press the first "a.invitations-link" within "#no_contacts"
     And I fill in the following:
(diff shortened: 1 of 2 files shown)


svbergerem:
Nice catch! Could you add a test for this? One similar to [this test](https://github.com/diaspora/diaspora/blob/b319e0caa24c7387d1040668284bec747da08f9f/features/desktop/invitations.feature#L42) would be great, where you just click on the other link to open the invitation modal.

features/desktop/invitations.feature
```diff
@@ -49,6 +49,15 @@ Feature: Invitations
     Then I should have 1 Devise email delivery
     And I should not see "change your notification settings" in the last sent email
 
+  Scenario: sends an invitation from a page link
```
svbergerem: `sends an invitation from the stream`

features/desktop/invitations.feature
```diff
@@ -49,6 +49,15 @@ Feature: Invitations
...
+    When I sign in as "alice@alice.alice"
+    And I click on selector "a.invitations-link"
```
svbergerem: I'd prefer `And I click on "invite" within "#no_contacts"`. In the sidebar we also have the `invitations-link` class but in that case it is a `div`. This could be confusing.

svbergerem (changes requested):
I just had two minor remarks.

Please squash your commits so we only have one commit at the end. Putting different behavior changes in different commits is a great thing but in this case it is just one change and a test for it. As long as you are still working on something you may of course have commits fixing/extending stuff from earlier commits.

svbergerem (approved):

svbergerem:
Merged as c7a0b053fa0382649672564ca25b65ac3f8d2404. Thank you!

Flaburgan:
Thanks and congrats for these first pull requests :D

PR #7192, opened 2016-11-17
Update typeahead.js, use the version from rails-assets.org

svbergerem:


SuperTux88:
Yay! Merged as 75f95faebefaa4ec3987ec76eba0caf5011c46ac :)

PR #7198, opened 2016-11-18
Add avatar fallback on tags page

### LINKED ISSUE #6969: Load default avatars when they are 404 on tag page
https://diaspora-fr.org/tags/diaspora

See movilla's image which is broken, for example.

svbergerem:
Fixes #6969.


SuperTux88:
Merged as 8e345c0335b5e58a9ab00a8168689f0f8bddb48c

Thanks for fixing this! :)

PR #7201, opened 2016-11-18
only require "fog/aws"

ghost:
This slightly reduces memory footprint by only requiring `fog/aws` instead of the full `fog` gem.

```
Before: 87.1523 MiB
After: 84.7461 MiB
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,8 +73,8 @@ gem "activerecord-import", "0.15.0"
 
 # File uploading
 
+gem "fog",         "1.38.0", require: "fog/aws"
 gem "carrierwave", "0.11.2"
-gem "fog",         "1.38.0"
 gem "mini_magick", "4.5.1"
 
 # GUID generation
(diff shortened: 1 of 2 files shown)


Flaburgan:
I was even thinking about dropping aws support completely, as far as I know joindiaspora was the only pod using it and I'm not sure it still works as we didn't test it for a while.

jramos:
I'm using it. I've noticed old profile photos aren't deleted when they're replaced, but otherwise it seems to work.

svbergerem:
> If there's a metagem available for your cloud provider, e.g. fog-aws, you should be using it instead of requiring the full fog collection to avoid unnecessary dependencies.

(https://github.com/fog/fog#dependency-notice)

SuperTux88:
We use `asset_sync` and that gem has still `fog` as dependency :(

See: AssetSync/asset_sync#325

svbergerem:
@SuperTux88 Thanks. :(

SuperTux88:
Merged as 43b1632148c17106752753abb0ee40337b478a16

Thank you!

PR #7202, opened 2016-11-18
Refactor flash messages on ajax errors for comments, likes, reshares and aspect memberships

### LINKED ISSUE #7200: Incorrect message when attempt to reshare with connection down
If I attempt to reshare a post when my internet connection is down, I get the flash alert 'That good, eh?  You've already reshared that post!'

Steps to reproduce: 1) switch off internet connection; 2) try to reshare a post!

I think the above message should be displayed *only* if the pod's server replies that the signed-in user has already reshared that post, if it's possible to test that. If the browser can't get a response from the server, a general alert 'Unable to connect to the pod at this moment' would be more appropriate (for all cases; not just a failed reshare attempt); or if, it's not possible to determine that as the cause of the error, a more general alert such as 'Unable to reshare this post at this moment'.

svbergerem:
For failing requests this will show
* “Unable to connect to the server.” if `response.code` is `0`
* the `responseText` received from the server for other responses

Fixes #7200.

### DIFF
--- a/app/controllers/likes_controller.rb
+++ b/app/controllers/likes_controller.rb
@@ -26,7 +26,7 @@ class LikesController < ApplicationController
         format.json { render :json => @like.as_api_response(:backbone), :status => 201 }
       end
     else
-      render :nothing => true, :status => 422
+      render text: I18n.t("javascripts.failed_to_like"), status: 422
     end
   end
 
--- a/app/controllers/likes_controller.rb
+++ b/app/controllers/likes_controller.rb
@@ -26,7 +26,7 @@ class LikesController < ApplicationController
         format.json { render :json => @like.as_api_response(:backbone), :status => 201 }
       end
     else
-      render text: I18n.t("javascripts.failed_to_like"), status: 422
+      render text: I18n.t("likes.create.error"), status: 422
     end
   end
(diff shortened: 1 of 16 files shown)


app/controllers/likes_controller.rb
```diff
@@ -26,7 +26,7 @@ def create
         format.json { render :json => @like.as_api_response(:backbone), :status => 201 }
       end
     else
-      render :nothing => true, :status => 422
+      render text: I18n.t("javascripts.failed_to_like"), status: 422
```
SuperTux88: `failed_to_like` is not used from javascript anymore. So I'm unsure if we should remove it from the javascript translations. We don't need to render it to the <head> on every page then, but we also need to retranslate it?

svbergerem: > but we also need to retranslate it

I would have restructured the translation files in a lot of PRs if we wouldn't have to retranslate the stuff. And using JavaScript translations in the Rails code is easy. The other way round isn't.

goobertron:
Oh, fabulous - thanks for such a quick fix.

SuperTux88:
Merged, thanks for your work! <3

PR #7083, opened 2016-09-10
Replace fileuploader-custom with FineUploader

Fensterbank:
This PR replaces the customized and unmaintained `fileuploader-custom.js` with the current version of [FineUploader](https://github.com/FineUploader/fine-uploader).

Yet no new functionality is added, this PR is just a prerequisite for #5621.

The new library is located in `/vendor/assets/javascripts` just like the old file.
I talked with @svbergerem and I think we will find a solution for this later. Unfortunately FineUploader doesn't support rails-assets.org at the moment.

PR #7204, opened 2016-11-19
[WIP] Improve file upload button

svbergerem:
I had this PR lying around for some time now and was waiting for #7083 at the beginning. Since that took some time I decided to open it now.

Currently the upload buttons in the publisher (desktop and mobile) have an `input` element inside of them. When hovering them you have a normal cursor most of the time and when the cursor is a pointer nothing happens when you click the button. This PR moves the `input` element somewhere else so you now have some normal buttons behaving like every other button.

### DIFF
--- a/app/assets/javascripts/app/views/publisher/uploader_view.js
+++ b/app/assets/javascripts/app/views/publisher/uploader_view.js
@@ -37,6 +37,7 @@ app.views.PublisherUploader = Backbone.View.extend({
     this.publisher.wrapperEl.before(this.info);
 
     this.publisher.photozoneEl.on("click", ".x", _.bind(this._removePhoto, this));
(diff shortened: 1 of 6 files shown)


SuperTux88:
It is currently possible to drop an image to the upload button, and with this PR this doesn't work anymore. :(

But anyway, I would prefer the whole publisher as drop-zone, because it is always difficult to aim for the tiny upload button...

svbergerem:
> It is currently possible to drop an image to the upload button, and with this PR this doesn't work anymore.

Indeed, I didn't know that that's possible. Well, then I think it's not worth the extra effort. I'll close this PR and after #7083 we can check if this fix is still needed.

> I would prefer the whole publisher as drop-zone, because it is always difficult to aim for the tiny upload button.

With FineUploader we can define additional dropzones. \o/ http://docs.fineuploader.com/features/drag-and-drop.html#drag-and-drop-standalone-module

PR #7207, opened 2016-11-19
Always show mobile reaction counts

### LINKED ISSUE #5829: [Mobile] Reaction count not consistent
It seems that the number of reactions within the mobile version are the sum of real reactions + likes. This seems not really consistent. See screenshot below, where there are 3 counted reactions, 1 real reaction and 2 likes.

screenshot_reactionscount
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10286371-mobile-reaction-count-not-consistent?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #5829.

### **Before**

mobile_reactions_before_1
mobile_reactions_before_2
mobile_reactions_before_3

### **After**
mobile_reactions_after_1
mobile_reactions_after_2
mobile_reactions_after_3

### DIFF
--- a/app/assets/stylesheets/mobile/comments.scss
+++ b/app/assets/stylesheets/mobile/comments.scss
@@ -38,29 +38,25 @@
   }
 
   .post-stats {
-    top: -5px;
     float: right;
     position: relative;
     display: flex;
-    margin-bottom: -15px;
 
     .count {
-      background-color: $background-grey;
-      text-align: center;
+      line-height: 22px;
+      margin-left: 5px;
       z-index: 2;
     }
 
     .icon-count-group {
       display: flex;
-      flex-flow: column;
-      justify-content: center;
       margin: 0 7px;
     }
 
     [class^="entypo"] {
       color: $text-grey;
       font-size: 24px;
-      height: 28px;
+      line-height: 24px;
       margin: 0;
       width: 100%;
     }
@@ -77,21 +73,23 @@
   }
 
   .post-action {
-    height: 28px;
-
     .disabled { color: $medium-gray; }
   }
 
   .add-comment-switcher { padding-top: 10px; }
 
-  &.inactive .post-stats .count,
-  &.inactive .comment-container {
-    display: none;
+  &.inactive {
+    padding-bottom: 8px;
+
+    .comment-container {
+      display: none;
+    }
   }
 }
 
 .stream-element .comments {
   margin: 0;
+  margin-top: 10px;
   padding: 0;
   width: 100%;
 
--- a/config/locales/javascript/javascript.en.yml
+++ b/config/locales/javascript/javascript.en.yml
@@ -281,6 +281,11 @@ en:
         one: "<%= count %> Reshare"
         other: "<%= count %> Reshares"
 
+      comments:
+        zero: "<%= count %> comments"
+        one: "<%= count %> comment"
+        other: "<%= count %> comments"
+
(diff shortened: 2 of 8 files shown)


goobertron:
Hey you're on fire at the moment. Thanks for your work!

Flaburgan:
Thanks for this, we definitely have the room to always display it.

app/assets/stylesheets/mobile/comments.scss
```diff
@@ -38,29 +38,25 @@
   }
 
   .post-stats {
-    top: -5px;
```
Flaburgan: :heart:

config/locales/javascript/javascript.en.yml
```diff
@@ -282,6 +282,11 @@ en:
         one: "<%= count %> Reshare"
         other: "<%= count %> Reshares"
 
+      comments:
+        zero: "<%= count %> comments"
```
Flaburgan: why do you move from "No" to "0"?

Flaburgan (approved):
Thanks for this PR it nicely improves the UX on mobile.
I tested it with Firefox and Chromium using the responsive mode and everything is working well.
I just feel like tapping on the comment icon should have the same behavior than tapping on the "2 comments" link (which now replace the reactions like) == opening the comment textarea **and** showing the existing comments. Being able to comment without displaying the other comments feels weird. This behavior is the same in the SPV.
You can change that here or in another PR. If so, this looks ready to be merged to me, the code also looks fine.

Flaburgan:
Also please rebase your code, I had to do it to do the review ;)

config/locales/javascript/javascript.en.yml
svbergerem: I didn't “move” from “No” to “0”. The old translations for the reaction counts in `config/locales/diaspora/en.yml` were using “No” instead of “0” but the removed translations in `config/locales/javascript/javascript.en.yml` as well as the translations for reshares and likes were using “0”, so I did the same when adding the comment count translations.

svbergerem:
Rebased, although there were no conflicts.

svbergerem:
> I just feel like tapping on the comment icon should have the same behavior than tapping on the "2 comments" link (which now replace the reactions like) == opening the comment textarea and showing the existing comments. Being able to comment without displaying the other comments feels weird.

Done.

Flaburgan:
:+1: tested again locally, it works well. Another small suggestion would be to focus on the textarea when the comment icon is tapped, but that can be done in another PR. So from my point of view, this is ready to be merged.

svbergerem:
> Another small suggestion would be to focus on the textarea when the comment icon is tapped, but that can be done in another PR

If there are more than 3 comments then we scroll down to the comment textarea. I agree that we could focus the textarea instead, but let's do that in another PR.

SuperTux88 (approved):

SuperTux88:
Merged as 513a3e217d1180d15646ae6ff5241fd537a4738c and daa4aec6146b268d8c23f68901c38a44767a1414

Thank you ❤️ (and thanks to @Flaburgan for the review)

PR #7211, opened 2016-11-21
Don't show browser notifications on pageload

svbergerem:
by calling `fetch` instead of `pollNotifications`. When loading the page you already see that you have new notifications and I don't think that a browser notification is needed here.

I also removed the `setTimeout` call when initializing the notifications collection. `pollNotifications` will be called the first time after `this.timeout` and we don't need an additional timeout here. (See https://html.spec.whatwg.org/multipage/webappapis.html#timer-initialisation-steps)


SuperTux88 (approved):

SuperTux88:
Merged as 00ce187b07de10409dfb90924db95c727b36b349

Thanks :)

PR #7213, opened 2016-11-22
only show community spotlight link if it is enabled

ghost:
This hides a community spotlight link if it's disabled in the `AppConfig`.

### DIFF
--- a/app/views/contacts/index.html.haml
+++ b/app/views/contacts/index.html.haml
@@ -1,5 +1,5 @@
 - content_for :page_title do
-  = t('.title')
+  = t(".title")
 
 .container-fluid#contacts_container
   .row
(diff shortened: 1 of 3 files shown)


app/views/contacts/index.html.haml
```diff
@@ -22,9 +22,10 @@
           .no_contacts
             %h3
               = t('.no_contacts')
-            %p
-              != t('.no_contacts_message',
-                   :community_spotlight => link_to(t('.community_spotlight'), community_spotlight_path))
+            - if AppConfig.settings.community_spotlight.enable?
+              %p
+                != t('.no_contacts_message',
```
diaspora-code-review: - Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
- Use the new Ruby 1.9 hash syntax.

cmrd-senya (approved):
Looks good to me.

IMO, we could also conditionally disable `/community_spotlight` route or render a stub message saying it is disabled at that page. However I don't think it is necessary within this PR

svbergerem:
I'd prefer a simple integration test instead of the cuke since the only thing we have to test is the existence of the community spotlight link on the contacts page.

You can use https://github.com/diaspora/diaspora/blob/edd3ddef5284f51a82d8f0bc97a9f4d89822d033/spec/integration/profile_spec.rb as an example. Just make sure that the logged in user doesn't have any contacts (e.g. by creating a new one with `FactoryGirl`).

spec/integration/contacts_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+
+          it_behaves_like "community spotlight information is not present on the page"
+        end
+
+        it_behaves_like "community spotlight information is not present on the page"
+      end
+    end
+  end
+end
```
diaspora-code-review: Final newline missing.

jramos:
This led me to another minor display issue: when invites are closed, the invite modal link is still displayed on `/contacts` and in the contacts sidebar.

spec/integration/contacts_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+          it "displays a community spotlight link" do
+            get "/contacts"
+
+            expect(response.status).to eq(200)
+            expect(response.body).to match(/a href="#{community_spotlight_path}"/)
+          end
+        end
+
+        it_behaves_like "community spotlight information is not present on the page"
```
svbergerem: For the sake of clarity I'd prefer moving this into a new context "community spotlight is disabled".

spec/integration/contacts_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        context "community spotlight is enabled" do
+          before do
+            AppConfig.settings.community_spotlight.enable = true
+          end
+
+          it_behaves_like "community spotlight information is not present on the page"
+        end
+
+        it_behaves_like "community spotlight information is not present on the page"
```
svbergerem: The same thing would be great here.

svbergerem (approved):

svbergerem:
Merged as 6c6f9bb982b67a9a8ae6b27356037332b866682b. Thank you!

PR #7228, opened 2016-11-27
Remove the duplicate "stream" container…

noplanman:
…which already gets added in the "messages" partial.

This is to remove the duplicate `<div class="stream">` container in the conversations view.

### DIFF
--- a/app/views/conversations/_show.haml
+++ b/app/views/conversations/_show.haml
@@ -25,4 +25,4 @@
   - for participant in conversation.participants
     = person_image_link(participant, :size => :thumb_small)
 
-= render partial: 'messages', locals: { conversation: conversation }
+= render partial: "messages", locals: {conversation: conversation}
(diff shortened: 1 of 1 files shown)


app/views/conversations/_show.haml
```diff
@@ -25,5 +25,4 @@
   - for participant in conversation.participants
     = person_image_link(participant, :size => :thumb_small)
 
-.stream
-  = render partial: 'messages', locals: { conversation: conversation }
+= render partial: 'messages', locals: { conversation: conversation }
```
diaspora-code-review: - Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
- Space inside { detected.
- Space inside } detected.

noplanman: Hmm, now do I just change these coding style problems in this single file or all .haml ones?
This PR was actually supposed to be a simple fix, not a rewrite of coding standards 😁

SuperTux88:
Merged as 60d49eb42335199ff5121a3fa82c79f2c73ef034

Thank you :)

PR #7230, opened 2016-11-29
Remove a dangling "lola" object creation

cmrd-senya:
It's never used and possibly was left unnoticed by some refactoring.


SuperTux88:
RIP Lola

Merged as c547e66eb662dd7cabdc101bf3d378932cf50e52

Thanks :)

PR #7203, opened 2016-11-19
add missing indexes across several tables

ghost:
There are undoubtedly others, but these are the ones that jumped out at me.

Before:
```
$ bundle exec rspec
Finished in 5 minutes 5 seconds (files took 11.38 seconds to load)
```

After:
```
$ bundle exec rspec
Finished in 4 minutes 30.6 seconds (files took 10.77 seconds to load)
```

PR #7234, opened 2016-12-01
add missing indexes on several tables

skyschub:
#7203 was closed by accident, reopening for further reviews.

### DIFF
--- a/db/migrate/20161118235506_add_missing_id_indexes_on_several_tables.rb
+++ b/db/migrate/20161118235506_add_missing_id_indexes_on_several_tables.rb
@@ -1,0 +1,22 @@
+class AddMissingIdIndexesOnSeveralTables < ActiveRecord::Migration
+  def change
+    add_index :account_deletions, :person_id
+    add_index :aspects, %i(user_id order_id)
+    add_index :blocks, :person_id
+    add_index :blocks, :user_id
+    add_index :invitation_codes, :user_id
+    add_index :notifications, %i(id type)
+    add_index :photos, :author_id
+    add_index :poll_participations, :author_id
+    add_index :poll_participations, :poll_answer_id
+    add_index :posts, :facebook_id
+    add_index :posts, :o_embed_cache_id
+    add_index :posts, :open_graph_cache_id
+    add_index :reports, :user_id
+    add_index :reports, %i(item_id item_type)
+    add_index :roles, :person_id
+    add_index :user_preferences, :user_id
+    add_index :users, :auto_follow_back_aspect_id
+    add_index :users, :invited_by_id
+  end
+end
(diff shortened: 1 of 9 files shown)


denschub:
(this is @jramos's work, not mine. Just opening this PR so we have a tracking issue.)

jramos:
👍

CSammy:
WIP: Indices I'd like to keep

- [x] `:account_deletions, :person_id, unique: true` (the entries should be unique, but aren't due to a unique constraint missing and technical difficulties/legacy; delete `:account_deletions, :diaspora_handle`; delegate to `person` table using `delegate :diaspora_handle, to :person` in the model `AccountDeletion`)
- [ ] `:aspects, %i(user_id order_id)` (order_id is used solely for the sorted display)
- [ ] `:blocks, :person_id` (a lookup will always start at the `user_id`, never at `person_id`, due to semantics)
- [x] `:blocks, :user_id` (good one!)
- [ ] `:invitation_codes, :user_id` (small table, rarely used)
- [ ] `:notifications, %i(id type)` (id is already unique and implicitly has an index)
- [x] `:photos, :author_id` (good one!)
- [ ] `:poll_participations, :author_id` (not necessary -- polls are looked up by `poll_id`)
- [ ] `:poll_participations, :poll_answer_id` (_may_ do good, but only for large polls, and only as combined index with `poll_id`)
- [ ] `:posts, :facebook_id` (there is no lookup by this ID)
- [ ] `:posts, :o_embed_cache_id` (there is no lookup by this ID)
- [ ] `:posts, :open_graph_cache_id` (there is no lookup by this ID)
- [ ] `:reports, :user_id` (there is no lookup by this ID)
- [ ] `:reports, %i(item_id item_type)` (there is already an index on `item_id`)
- [ ] `:roles, :person_id` (small table)
- [ ] `:user_preferences, :user_id` (a bit useful, a lookup will benefit from the index and it is used often enough to warrant the added complexity)
- [ ] `:users, :auto_follow_back_aspect_id` (the semantics of this field - a setting attached to the user - do not require an index)
- [ ] `:users, :invited_by_id` (there is no lookup using this field)

jramos:
These were a few more I came up with on another effort before I got sidetracked. Take what you will.

```
add_index :blocks, %i(user_id person_id)
add_index :locations, :status_message_id
add_index :roles, %i(person_id name)
add_index :users, :unconfirmed_email
add_index :user_preferences, %i(user_id email_type)
```

Flaburgan:
The freeze of the next major version is closer, @jramos and @CSammy any chance you can have a look at that this summer so we can ship it in 0.7?

SuperTux88:
All indexes that @CSammy approved are now in 0.7.0.0, but everywhere where he wrote he is undecided (https://github.com/diaspora/diaspora/pull/7234#issuecomment-267187943) is still open:

- [ ] `:authorizations, %i[user_id o_auth_application_id], unique: true`
- [ ] `:ppid, %i[identifier user_id], unique: true`
- [ ] `:ppid, :guid, unique: true`
- [ ] `:services, %i[type uid], unique: true`

denschub:
The parts of this work which we agreed on having a positive impact are merged, and the other pieces are undecided on. This PR is large, has conflicts, and probably does not help overall. It would be good that, if we are still missing important indexes on tables, but let's close this PR and open new ones if that is the case.

PR #7235, opened 2016-12-01
Markdown editor final chapter: add editor on publisher, comments and conversation of mobile view

ghost:
Publisher:
screen shot 2016-12-01 at 11 44 43

Comments:
screen shot 2016-12-01 at 12 03 47

New conversation:
screen shot 2016-12-01 at 12 07 20

Existing conversation:
screen shot 2016-12-01 at 12 07 12

### DIFF
--- a/app/assets/javascripts/mobile/mobile_comments.js
+++ b/app/assets/javascripts/mobile/mobile_comments.js
@@ -11,6 +11,8 @@
     initialize: function() {
       var self = this;
 
+      new Diaspora.MarkdownEditor(".comment_box");
+
       this.stream().on("tap click", "a.show-comments", function(evt){
         evt.preventDefault();
         self.toggleComments($(this));
--- a/app/assets/stylesheets/markdown-editor.scss
+++ b/app/assets/stylesheets/markdown-editor.scss
@@ -42,6 +42,7 @@
 
   .btn-group {
     margin-bottom: 8px;
+    &:first-child { margin-left: 0; }
 
     [class^="entypo-"],
     [class*="entypo-"] {
--- a/app/assets/stylesheets/mobile/markdown_editor.scss
+++ b/app/assets/stylesheets/mobile/markdown_editor.scss
@@ -1,0 +1,22 @@
+.md-editor {
+  border: 1px solid $light-grey;
+  border-radius: $btn-border-radius-base;
+  overflow: hidden;
+
+  &.active { border-color: $text-grey; }
+
+  &::after,
+  &::before {
+    clear: both;
+    content: ' ';
+    display: table;
+  }
+
+  textarea {
+    border: 0;
+    border-radius: 0;
+    min-height: 6rem;
+    resize: none;
+    width: 100%;
+  }
+}
--- a/app/assets/stylesheets/mobile/markdown_editor.scss
+++ b/app/assets/stylesheets/mobile/markdown_editor.scss
@@ -1,22 +1,6 @@
 .md-editor {
   border: 1px solid $light-grey;
   border-radius: $btn-border-radius-base;
-  overflow: hidden;
 
   &.active { border-color: $text-grey; }
-
-  &::after,
-  &::before {
-    clear: both;
-    content: ' ';
-    display: table;
-  }
-
-  textarea {
-    border: 0;
-    border-radius: 0;
-    min-height: 6rem;
-    resize: none;
-    width: 100%;
-  }
 }
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -16,6 +16,8 @@
 @import "mobile/stream_element";
 @import "mobile/comments";
 @import 'mobile/openid_connect_error_page';
+@import 'markdown-editor';
+@import 'mobile/markdown_editor';
 
 @import 'typography';
 
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -44,8 +44,6 @@ h3 {  margin-top: 0; }
 .clear { clear: both; }
 #main { padding: 56px 10px 0 10px; }
 
-textarea { resize: vertical; }
-
 .avatar {
   border-radius: 4px;
 }
@@ -405,11 +403,6 @@ footer {
   .counter {
     font-size: 14px;
   }
-
-  textarea {
-    min-width: 100%;
-    max-width: 100%;
-  }
 }
 
 select {
@@ -538,13 +531,6 @@ form#new_conversation.new_conversation {
     font-weight: normal;
     line-height: 20px;
   }
-}
-
-textarea#conversation_text {
-  font-size: larger;
-  width: 100%;
-  min-width: 100%;
-  max-width: 100%;
 }
 
 .registrations_error,
@@ -714,33 +700,6 @@ select#aspect_ids_ {
   background: $framed-background;
   position: relative;
   padding-right: 10px;
-
-  #hide_publisher  {
-    opacity: 0.3;
-    z-index: 5;
-    padding: 3px;
-    position: absolute;
-    right: 6px;
-    top: 0;
-
-    &:hover {
-      opacity: 1;
-    }
-  }
-
-  &.active {
-    border: 1px solid $border-dark-grey;
-  }
-
-  textarea {
-    z-index: 2;
-    border: none;
-
-    &:focus {
-      outline: 0;
-      background: transparent;
-    }
-  }
 
   &.with_attachments {
     padding-bottom: 55px;
--- a/app/assets/javascripts/mobile/mobile.js
+++ b/app/assets/javascripts/mobile/mobile.js
@@ -1,9 +1,9 @@
-// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-v3-or-Later
-
 /*   Copyright (c) 2010-2011, Diaspora Inc.  This file is
  *   licensed under the Affero General Public License version 3 or later.  See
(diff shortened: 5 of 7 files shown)


app/assets/javascripts/mobile/mobile_comments.js
```diff
@@ -11,6 +11,8 @@
     initialize: function() {
       var self = this;
 
+      new Diaspora.MarkdownEditor(".comment_box");
```
cmrd-senya: Maybe it is out of scope for this PR, but are we okay with the object design where constructor actually performs some job besides the object creation? This decision was made before this PR, and if it was fine back then, then these calls are fine either.

Perhaps it would be a little cleaner to have something like
```js
new Diaspora.MarkdownEditor().renderOn(".comment_box");
```

Perhaps it is a kind of nit picking, but I felt I have to notice that.

AugierLe42e: I followed the idioms of Backbone where the HTML element is passed to the constructor.

cmrd-senya: Then it can be done this way:
```
new Diaspora.MarkdownEditor(".comment_box").render();
```

My main point is to split object creation and its actual usage.

I don't think it major (and moreover out of the PR scope), but what do others think?

AugierLe42e: I'm not sure this is big deal. Furthermore, we aren't constructing the object without rendering it. So that's a bit redundant.

svbergerem: @cmrd-senya What you write seems legit to me, but I'd also like to hear @denschub s opinion on this topic. But _please_ let's do this somewhere else. This is out of scope of this PR. Just open a new code refactor issue or so.

denschub: Ideally, you'd have the constructor doing nothing besides internal state. There are multiple reasons for that, but mainly there are very valid reasons to have the render function called again under some circumstances, or to build objects early when the DOM is not even finishing loading, dispatching rendering to a later point in time to increase the apparent performance.

I agree that this is something we could change, but as @svbergerem said, let's not do that here.

Flaburgan (approved):
Alright, this looks good! The code is pretty simple, I didn't check if there is a simpler way to write the CSS but at least on Firefox everything is displayed fine.
I will probably merge that on d-fr to test it on production more deeply.
A small remarks, I would have preferred to see the list icons displayed instead of the code one as I feel code is mostly for tech people (who probably directly type markdown) but this should be changed in an other PR as this one is pretty clean and easy to review.

Thanks for your work!

cmrd-senya:
I did some tests of the markdown editor and found some issues with it.

In the desktop publisher when the publisher is full with lines and I press `H` the publisher text area stretches to fit the text. In the mobile publisher when it is full and I press `H` it doesn't stretch, but the text is inserted.

See video:
https://b2aeaa58a57a200320db-8b65b95250e902c437b256b5abf3eac7.ssl.cf5.rackcdn.com/media_entries/13845/out-3.ogv

Also, it is not a bug of this PR, but when I consequently push `H` button it adds and removes `###` on the same line (toggles the heading) but it always adds a newline, so on removing of `###` it doesn't delete the newline, so pressing the `H` button continuosly with the same line makes the text area filled with newlines. It happens both on mobile and desktop.

Another thing which is not directly related to this PR, but related to publisher - is that for some reason I can't insert any other link type besides `http://` and `https://`. User may want to instert any link type, including `ftp://`, `bitcoin:`, `xmpp:`, `mailto:`, `magnet:`, and many others. I see no reason to limit input link types for user. But it different for images, since we won't be able to embed any links besides `http` and `https`.

Flaburgan:
Points 2 and 3 don't belong to this PR.

What you're saying in point 1 is that the publisher doesn't expand correctly when we press the "insert title" button. I'll have a look but I don't have an idea what's causing that on the top of my head right now.

cmrd-senya:
> Points 2 and 3 don't belong to this PR.

That's true, but I think it would be nice to fix them before we actually make a release of this feature.

> What you're saying in point 1 is that the publisher doesn't expand correctly when we press the "insert title" button. I'll have a look but I don't have an idea what's causing that on the top of my head right now.

Well, that doesn't happen in desktop, it expands there correctly and also expansion works in mobile in other cases, so must be something wrong with the styles I guess. Do you feel okay for an investigation?

cmrd-senya:
Okay, I checked points 2 and 3, these are upstream issues of `bootstrap-markdown`. I created issues https://github.com/toopay/bootstrap-markdown/issues/292 and https://github.com/toopay/bootstrap-markdown/issues/293.

app/assets/stylesheets/markdown-editor.scss
```diff
@@ -27,6 +27,7 @@
 
   .btn-group {
     margin-bottom: 8px;
+    &:first-child { margin-left: 0; }
```
svbergerem: It looks to me like there is no margin set anywhere else, so this line looks unnecessary to me.

app/assets/stylesheets/mobile/markdown_editor.scss
```diff
@@ -0,0 +1,22 @@
...
+  overflow: hidden;
+
+  &.active { border-color: $text-grey; }
+
+  &::after,
+  &::before {
+    clear: both;
+    content: ' ';
+    display: table;
+  }
```
svbergerem: I removed these lines and I can't spot any difference for comments, conversations and new status messages.

app/assets/stylesheets/mobile/markdown_editor.scss
```diff
@@ -0,0 +1,22 @@
+.md-editor {
+  border: 1px solid $light-grey;
+  border-radius: $btn-border-radius-base;
+  overflow: hidden;
```
svbergerem: `overflow: hidden` is already added to the textarea by autosize and this line here looks unnecessary.

app/assets/stylesheets/mobile/markdown_editor.scss
```diff
@@ -0,0 +1,22 @@
...
+  &::after,
+  &::before {
+    clear: both;
+    content: ' ';
+    display: table;
+  }
+
+  textarea {
+    border: 0;
+    border-radius: 0;
```
svbergerem: The desktop markdown editor CSS already removes the border, so one can remove these two lines.

app/assets/stylesheets/mobile/markdown_editor.scss
```diff
@@ -0,0 +1,22 @@
...
+    min-height: 6rem;
```
svbergerem: The size is set by autosize and this min-height doesn't seem to make any difference.

app/assets/stylesheets/mobile/markdown_editor.scss
```diff
@@ -0,0 +1,22 @@
...
+    resize: none;
```
svbergerem: Already set by autosize.

app/assets/stylesheets/mobile/markdown_editor.scss
```diff
@@ -0,0 +1,22 @@
...
+    width: 100%;
```
svbergerem: Already set in comments, conversations and new status messages by `.form-control`.

svbergerem (changes requested):
I'll add the requested changes in a commit myself.

app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -42,8 +44,6 @@ h3 {  margin-top: 0; }
 .clear { clear: both; }
 #main { padding: 56px 10px 0 10px; }
 
-textarea { resize: vertical; }
```
svbergerem: The same code is a few lines above.

app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -403,11 +403,6 @@ footer {
   .counter {
     font-size: 14px;
   }
-
-  textarea {
-    min-width: 100%;
-    max-width: 100%;
-  }
```
svbergerem: `width: 100%` already set for `.form-control`.

app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -538,13 +533,6 @@ form#new_conversation.new_conversation {
   }
 }
 
-textarea#conversation_text {
-  font-size: larger;
-  width: 100%;
-  min-width: 100%;
-  max-width: 100%;
-}
-
```
svbergerem: The element doesn't exist anymore.

app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -713,33 +701,6 @@ select#aspect_ids_ {
...
-  textarea {
-    z-index: 2;
-    border: none;
-
-    &:focus {
-      outline: 0;
-      background: transparent;
-    }
-  }
-
```
svbergerem: The `#publisher_textarea_wrapper` only contains the photodropzone.

svbergerem (approved):

svbergerem:
Would be great if someone else could review my changes.

app/assets/stylesheets/markdown-editor.scss
SuperTux88: This line was to move the first group to the beginning (same margin as from the top/bottom):



Without this line:

SuperTux88:
> Would be great if someone else could review my changes.

Done, see https://github.com/diaspora/diaspora/pull/7235#discussion_r132806833

Otherwise looks good to me 👍

app/assets/stylesheets/markdown-editor.scss
svbergerem: Thanks!

svbergerem:
Thanks!

PR #7236, opened 2016-12-01
Remove user_methods require line

cmrd-senya:
No need to explicitly require user_methods.rb since it is already required in spec_helper.rb


SuperTux88:
Merged as dd9bda24f4b81cce1e0c9ee99b1480a00e116b3d

Thanks :)

PR #7241, opened 2016-12-07
Reformulate CSRF e-mail

Flaburgan:
I still feel like we should put the important information (you have been logged out but can now log back safely) first. What do you think?

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -715,9 +715,11 @@ en:
       body: |-
         Hello %{name},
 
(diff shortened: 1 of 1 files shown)


goobertron:
I agree, and I thought that was the version that was going to be merged.

Flaburgan:
@goobertron is the english fine in this one?

goobertron:
Perfect.

(I'm still not 100% happy about 'innocent', but I still can't think of a better word that would be easily understood. It's the best I could come up with, and it's good enough for now.)

SuperTux88 (approved):

SuperTux88:
Merged as f785de74840ca83416d8e3eefdcf00288ba6e81c to the 0.6.2.0 release branch. Thanks.

Flaburgan:
Thanks for merging after the freeze ;)

PR #7242, opened 2016-12-07
new PostController#mentionable_in_comment action

cmrd-senya:
This PR introduces `/posts/:id/mentionable_in_comment` route that can be used by the frontend to suggest people to a user in the comment publisher.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -3,7 +3,7 @@
 #   the COPYRIGHT file.
 
 class PostsController < ApplicationController
-  before_action :authenticate_user!, only: :destroy
+  before_action :authenticate_user!, only: %i(destroy mentionable)
   before_action :set_format_if_malformed_from_status_net, only: :show
 
   respond_to :html, :mobile, :json, :xml
@@ -48,6 +48,21 @@ class PostsController < ApplicationController
       }
       format.any { render nothing: true, status: 406 }
     end
+  end
+
+  def mentionable
+    respond_to do |format|
+      format.json {
+        if params[:id].present? && params[:q].present?
+          render json: post_service.mentionable_in_comment(params[:id], params[:q])
+        else
+          render nothing: true, status: 204
+        end
+      }
+      format.any { render nothing: true, status: 406 }
+    end
+  rescue ActiveRecord::RecordNotFound
+    render nothing: true, status: 404
   end
 
   def destroy
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -62,7 +62,8 @@ class Person < ActiveRecord::Base
   scope :remote, -> { where('people.owner_id IS NULL') }
   scope :local, -> { where('people.owner_id IS NOT NULL') }
   scope :for_json, -> {
-    select('DISTINCT people.id, people.guid, people.diaspora_handle')
+    select("people.id, people.guid, people.diaspora_handle")
+      .distinct
       .includes(:profile)
   }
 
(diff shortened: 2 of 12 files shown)


cmrd-senya:
Please review

SuperTux88 (changes requested):
Looks quite good so far, only some small things.

SuperTux88:
@CSammy I added you as reviewer. If you have the time, can you please have a quick look at the sql-queries? (I think they look ok, but it doesn't harm to have a second opinion)

app/controllers/posts_controller.rb
```diff
@@ -50,6 +50,21 @@ def interactions
     end
   end
 
+  def mentionable_in_comment
+    respond_to do |format|
+      format.json {
+        if params[:id].present? && params[:q].present?
```
cmrd-senya: We can do this without making a query, by reusing data we already have in the frontend. This could be more complex for the frontend, but will simplify and unload the backend.

Maybe we should ask @AugierLe42e for his opinion here?

SuperTux88: I would prefer a solution with the available data from the frontend (because everything should already be there), but if that isn't possible or too complicated, then a query without q parameter to the backend should be easy.

AugierLe42e: > I still think, that for comments you should be able to mention the post-author and the last few comment-authors without typing anything (only the @), so without q.

It's not possible the way the mention class is written right now. This mean we should change the logic and add an exception case and I'm not for at all. Plus, on a personnal point of view, I find this counter intuitive to be able to mention people without a `@` on comments and not in publisher.

> I would prefer a solution with the available data from the frontend (because everything should already be there)

Same problem: no data is available on the front-end right now. There was, not long ago, but #6951 changed that behaviour. This mean refactor (again) the search system.

ghost (approved):

app/controllers/posts_controller.rb
SuperTux88: > It's not possible the way the mention class is written right now.

I don't see how this shouldn't be possible with the current mention system, you just remove the need to type 2 additional characters.

> Plus, on a personnal point of view, I find this counter intuitive to be able to mention people without a @ on comments and not in publisher.

I don't understand what you mean here? There will be an @. So you type the @ and you already see the last few commenters and the post author as suggestion, but without the need to type the actual name (so you can select them with arrow-keys). I don't see where this should be "counter intuitive"?

> Same problem: no data is available on the front-end right now.

That's why I suggested to use the data from the backend (because it is available there and can be returned easily), and refactor the frontend later to use the data the frontend already know. If the refactoring is easy, then I'd prefer to directly use the data from the frontend, but I can't estimate that.

svbergerem: @AugierLe42e It looks like there is some misunderstanding.

> I find this counter intuitive to be able to mention people without a @ on comments and not in publisher.

That's not what @SuperTux88 wrote. He wrote

> you should be able to mention the post-author and the last few comment-authors without typing anything (only the @)

So you type the “@” and you will see the post author and the last comment authors immediately.

> no data is available on the front-end right now. There was, not long ago, but #6951 changed that behaviour.

That's wrong. The data you need is the post author and the comment authors. And this information is already available in the front end.

SuperTux88: I talked to @svbergerem via IRC, and he meant that it is doable with normal effort to use the data already there in the frontend (you maybe need to convert the author objects from the post to another structure, but then can use it for typeahead as local data. But that's a simple mapping and no problem).

@cmrd-senya: so there is no need to change the controller here and the check for the `q` param presence is OK. If you need help for doing this in the frontend, it is possibly the best to talk to @svbergerem directly.

AugierLe42e: >  don't understand what you mean here? There will be an @. So you type the @ and you already see the last few commenters and the post author as suggestion, but without the need to type the actual name (so you can select them with arrow-keys). I don't see where this should be "counter intuitive"?

Ah, misunderstanding. My bad, sry.

> That's wrong. The data you need is the post author and the comment authors. And this information is already available in the front end.

Where is it available? If so, it's easilly doable, then, yes.

> That's wrong. The data you need is the post author and the comment authors. And this information is already available in the front end.

As stated before, I can take care of the front-end.

svbergerem: Here is a small example for the comment stream view and the author of the post:

```diff
diff --git a/app/assets/javascripts/app/views/comment_stream_view.js b/app/assets/javascripts/app/views/comment_stream_view.js
index 05ba192db..81ab9a5ec 100644
--- a/app/assets/javascripts/app/views/comment_stream_view.js
+++ b/app/assets/javascripts/app/views/comment_stream_view.js
@@ -16,6 +16,16 @@ app.views.CommentStream = app.views.Base.extend({
   initialize: function() {
     this.CommentView = app.views.Comment;
     this.setupBindings();
+    var author = this.model.get("author");
+    var person = {
+      id: author.id,
+      guid: author.guid,
+      handle: author.diaspora_id,
+      name: author.name,
+      avatar: author.avatar.medium,
+      url: Routes.person(author.guid)
+    };
+    console.log(person);
   },
 
   setupBindings: function() {
```

config/routes.rb
```diff
@@ -32,6 +32,7 @@
   resources :posts, only: %i(show destroy) do
     member do
       get :interactions
+      get :mentionable_in_comment
```
cmrd-senya: Do I rename `PostService#mentionable_in_comment` also? I think we should leave it, since `PostService#mentionable` would be ambigous.

SuperTux88: I'm unsure about it. Since it also needs a `post_id` as parameter it should be clear that it means mentionable in the context of the post with this id. But if you think it's ambiguous I think `PostService#mentionable_in_comment` is also OK, or you can add a documentation for the method.

cmrd-senya: I would leave `PostService#mentionable_in_comment` then.

SuperTux88 (approved):

app/models/person.rb
```diff
@@ -94,6 +94,71 @@ def downcase_diaspora_handle
...
+  # Sorts people in the following priority: post author > commenters > likers > contacts > non-contacts
+  # @param [Post] post for which the mentionable in comment people list is requested
+  # @param [User] user who requests the people list
+  # @return [Person::ActiveRecord_Relation]
+  scope :sort_for_mention_suggestion, ->(post, user) {
+    left_join_visible_post_interactions_on_authorship(post.id)
+      .joins("LEFT OUTER JOIN contacts ON people.id = contacts.person_id AND contacts.user_id = #{user.id}")
+      .joins(:profile)
+      .order(<<-SQL
+        people.id = #{post.author.try(:id).to_i} DESC,
```
SuperTux88: Oh, I didn't see this before, but this generates this additional query including the search-term:
```sql
SELECT  `people`.* FROM `people` INNER JOIN `profiles` ON `profiles`.`person_id` = `people`.`id` WHERE (`people`.`id` != 73) AND (
      profiles.full_name LIKE '%te%' OR
      people.diaspora_handle LIKE 'te%'
) AND `people`.`id` = 74 LIMIT 1
```
(that's why you needed the `try` here)

So inside a `scope` rails does weird things (it also adds the scope to other queries). When you call `post.author_id` outside the `scope` first (in the `PostService`), it doesn't do that and you don't need a `try`.

SuperTux88 (changes requested):
Sorry, didn't saw this before, and I'm also discussing the SQL-queries with @CSammy currently, maybe we can improve something there.

app/models/person.rb
cmrd-senya: We can write it this way then:
```diff
diff --git a/app/models/person.rb b/app/models/person.rb
index 7a200d0..8a30c53 100644
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -149,7 +149,7 @@ class Person < ActiveRecord::Base
       .joins("LEFT OUTER JOIN contacts ON people.id = contacts.person_id AND contacts.user_id = #{user.id}")
       .joins(:profile)
       .order(<<-SQL
-        people.id = #{post.author.try(:id).to_i} DESC,
+        people.id = #{unscoped { post.author_id }} DESC,
         comments.id IS NOT NULL DESC,
         likes.id IS NOT NULL DESC,
         contacts.id IS NOT NULL DESC,
```

SuperTux88: Oh, didn't know that :) yes that fixes that 👍

SuperTux88 (approved):
Look's good to me, but maybe @CSammy has some feedback about the SQL-queries. So let's wait until he has the time for it.

CSammy (approved):
The SQL part looks ok to me!

cmrd-senya:
@CSammy, why not changing to `UNION` after all? Isn't that possible in Rails?

CSammy:
@cmrd-senya It turned out to be not as profitable and not as clear as it seemed at first. Also, Rails does not directly support `UNION` to my knowledge, so you'd have to inject plain SQL for a large part of the queries which didn't seem worth it.

SuperTux88:
Thanks <3

cmrd-senya:
> Thanks <3

No u

Issue #7243, opened 2016-12-08
Support for S3-compatible APIs


rev138:
There are several products on the market that emulate Amazon S3 (swift, ceph, minio), but the diaspora.yml config doesn't let you configure a custom endpoint. It assumes s3.amazon.com. Can a new parameter please be added to enable this?

Thanks

emmaly:
This is needed for Google Cloud Storage, which uses the endpoint `storage.googleapis.com`.

HankG:
> This is needed for Google Cloud Storage, which uses the endpoint `storage.googleapis.com`.

Is that using an S3-compatible call structure like Digital Ocean's storage system?

emmaly:
GCS is API compatible, yes.

HankG:
@emmaly if I were to pick up the story in the near future would you be able to help with testing?  I don't have a pod setup that would allow me to test that and it would help expedite bringing it across the finish line faster.

emmaly:
Yes, I certainly could.  I'm all set up and ready to go.

PR #7244, opened 2016-12-08
Added support for 3rd party s3-compatible object storage

rev138:
In regards to issue #7243, I was able to modify the carrierwave settings to enable my pod to use minio as an s3 backend. For some reason using the 'bucketname.example.com' form of uploading was throwing an SSLv3 error, so I had to enable `:path_style = true` to get around this, which uses 'example.com/bucketname' instead.

### DIFF
--- a/config/defaults.yml
+++ b/config/defaults.yml
@@ -26,6 +26,9 @@ defaults:
       bucket:
       region:
       cache: true
+      host: "s3.amazonaws.com"
+      endpoint: "https://s3.amazonaws.com"
+
     image_redirect_url:
     assets:
       serve: false
--- a/config/diaspora.yml.example
+++ b/config/diaspora.yml.example
@@ -116,6 +116,8 @@ configuration: ## Section
       #secret: 'change_me'
       #bucket: 'my_photos'
       #region: 'us-east-1'
+      #host: 's3.amazonaws.com'
+      #endpoint: 'https://s3.amazonaws.com'
 
       ## Use max-age header on Amazon S3 resources (default=true).
       ## When true, this allows locally cached images to be served for up to
--- a/config/initializers/carrierwave.rb
+++ b/config/initializers/carrierwave.rb
@@ -12,7 +12,10 @@ CarrierWave.configure do |config|
         provider:              'AWS',
         aws_access_key_id:     AppConfig.environment.s3.key.get,
         aws_secret_access_key: AppConfig.environment.s3.secret.get,
-        region:                AppConfig.environment.s3.region.get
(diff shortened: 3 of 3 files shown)


SuperTux88:
Does the `path_style` also affect/change the current s3 behavior? Should we also create an option for that in the diaspora.yml?

rev138:
AFAIK it should still work, it just accesses the bucket by an alternate method. Someone with an s3 account should probably verify though.

rev138:
I appears that AssetSync isn't so easy to override the defaults on, so while this patch makes uploads work, assets still try to go to s3.

SuperTux88:
What I meant with my comment above was: Is the `path_style` only used for upload? Or does that also change the URL how the files are accessed by the users?

rev138:
Just uploads, I think. Doesn't D* get the url path from configuration > assets > host?

SuperTux88:
I think `environment.assets.host` is only for assets. What I meant was, does `path_style` affect the URL generated for uploaded photos when using S3?

Also, if this PR doesn't work with AssetSync, we should add a comment somewhere, that you couldn't enable asset upload if you change the S3 host. And probably also add a check that it only enables when the `host` is still default.

rev138:
Yes, I believe it does change the urls from 'bucket.example.com' to 'example.com/bucket', but AFAIK Amazon supports both styles. Would you like me to add comments and resubmit, or do you want that done on your end?

SuperTux88:
> Yes, I believe it does change the urls from 'bucket.example.com' to 'example.com/bucket', but AFAIK Amazon supports both styles.

I think we should expose this option in the config then and use the previous behavior as default.

> Would you like me to add comments and resubmit, or do you want that done on your end?

Please add a comment and also add a check that blocks if someone tries to enable asset upload with a changed S3 host. You can also print a warning then.

rev138:
Ok. What's the best way to deal with that? Throw an exception?

SuperTux88:
I think so, because it wouldn't work anyway.

HankG:
I'm researching this topic to potentially pick it up.  Looking at the documentation for Fog it seems that the credentials is highly system dependent.   I'm surprised the Minio thing worked but perhaps since it is S3-compatible it's just taking the same access key/secret/region pairs and "just works".  In the Fogs library the logins for Digital Ocean and GCS are different, and in the CarrierWave source code it actually shows the different credentials for GCS and Rackspace.  We'll have to look at this more carefully.

jhass:
@HankG How's the progress on this, are you still planning on getting this work in? :)

weex:
Bump on conflicting files

PR #7259, opened 2016-12-29
Bump ruby to 2.4

Flaburgan:
Just checking how travis will react to that.

[Ruby 2.4 changelog](https://www.ruby-lang.org/en/news/2016/12/25/ruby-2-4-0-released/)


Flaburgan:
Well, looks like it's not available yet :p

SuperTux88:
Yes, I've already tried it too, and it also failed locally, so I didn't try travis ;)

SuperTux88:
I rebased this on `develop` after merging #7345, lets see what travis says now :) (local tests worked)

SuperTux88:
Travis still doesn't like it 😿 (I only tested rspec and starting a pod and clicking a bit locally)

Flaburgan:
Thanks @SuperTux88 for the webmock bump. Travis is happier now, the failing one looks like a randomly failing one.

SuperTux88 (approved):
Looks good to me now 👍

SuperTux88:
Just for the record: [Support of Ruby 2.1 has ended](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/)

SuperTux88:
ruby 2.3.x doesn't compile with libssl-1.1 (which is default on debian stable ... another sign to upgrade to 2.4 when debian stable is too new for something :P )

Also: Sidekiq 5 only supports 2.2.2+, so if we want to upgrade sidekiq we need to switch too :)

cmrd-senya (approved):

skyschub (approved):
Oh well, let's do it!

SuperTux88:
I rebased this and added a second commit to also install production gems on travis (to check compatibility with ruby versions).

skyschub (approved):
👍

SuperTux88:
Let's do this 🎉

PR #7263, opened 2017-01-02
Fix year bg color on /notifications

Flaburgan:
**Before**
notif
**After**
notif2

If we really want to be able to set a specific color to the year background we should use another variable than `$white` but I feel this is useless.


SuperTux88:
Merged as 747ea161bf1ab8817cd5737dd7b1ee69da55382b

Nice catch, thanks :)

PR #7214, opened 2016-11-22
Add markdown editor on conversations and comments

ghost:
On comments :

markdown editor on comments

On a new conversation :

markdown editor on new conversation

On a conversation reply :

markdown editor on conversation reply

Just needs some tests.

PR #7275, opened 2017-01-13
[WIP] Comment mentions front-end

ghost:
Front-end over #7242. CSS is a little broken, but it works ;)
screen shot 2017-01-13 at 12 03 58

### DIFF
--- a/app/assets/templates/comment-stream_tpl.jst.hbs
+++ b/app/assets/templates/comment-stream_tpl.jst.hbs
@@ -26,7 +26,14 @@
 
     <div class="bd">
       <form accept-charset="UTF-8" action="/posts/{{id}}/comments" class="new_comment" id="new_comment_on_{{id}}" method="post">
-        <textarea class="comment_box form-control" id="comment_text_on_{{id}}" name="text" rows="1" required placeholder="{{t "stream.comment"}}" />
+        <div class="mentions-input-box">
+          <div class="mentions-box">
+            <div class="mentions"/>
+          </div>
+          <textarea class="comment_box form-control mention-fake-textarea"
+                    id="comment_text_on_{{id}}" name="text" rows="1" required placeholder="{{t "stream.comment"}}" />
+          <input class="typeahead-mention-box hidden" type="text">
+        </div>
         <div class="submit_button">
           <input class="btn btn-primary" id="comment_submit_{{id}}" name="commit" type="submit" value="{{t "stream.comment"}}" />
         </div>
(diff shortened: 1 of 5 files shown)


cmrd-senya:
@AugierLe42e, if you need any help in completing this, just tell me, I'm ready to share that work ;)

Flaburgan:
What's the status of this? We talked about Simplify the mention syntax in https://github.com/diaspora/diaspora/issues/7276 and https://github.com/diaspora/diaspora/pull/7277 is now integrated. We also talked about displaying the syntax directly instead of rendering wysiwyg in the publisher write tab. This should simplify your work here.

svbergerem:
@Flaburgan 
Quote from #7276:

> What we came up with is adding support for the new syntax (display, create notifications, etc.) but still converting everything to the old syntax when saving the post/comment.

Since #7277 we support displaying the new syntax but everything else is missing. The most important step is adding some code that converts the new syntax to the old one when saving posts or comments.

This has to be done before we simplify the mentions front end code.

cmrd-senya:
I can pick this one if @AugierLe42e is not willing to complete it.

SuperTux88:
@cmrd-senya based on what @AugierLe42e wrote [here](https://github.com/diaspora/diaspora/pull/7214#issuecomment-287564370) (and since there was no reaction here since 2 weeks), I think you can pick this one.

cmrd-senya:
Yeah, thanks, I saw that comment and I plan to take this one.

app/assets/templates/comment-stream_tpl.jst.hbs
```diff
@@ -26,7 +26,14 @@
 
     <div class="bd">
       <form accept-charset="UTF-8" action="/posts/{{id}}/comments" class="new_comment" id="new_comment_on_{{id}}" method="post">
-        <textarea class="comment_box form-control" id="comment_text_on_{{id}}" name="text" rows="1" required placeholder="{{t "stream.comment"}}" />
+        <div class="mentions-input-box">
+          <div class="mentions-box">
+            <div class="mentions"/>
+          </div>
+          <textarea class="comment_box form-control mention-fake-textarea"
```
cmrd-senya: I need some help from frontend people here. Since `textarea` went down in the hierarchy here, `submit_button` is not a sibling anymore, and therefore [this](https://github.com/diaspora/diaspora/blob/dba8de54888b50291c5b7b4e64af1fd6c03e2e19/app/assets/stylesheets/comments.scss#L63) style doesn't apply anymore. Does it mean that the only option I have is to implement that behavior (submit button hiding) via JS (like we do for the publisher)?

svbergerem: (Discussed on IRC. `mentions-input-box` is not needed anymore. Also see #7380)

svbergerem:
Superseded by #7386.

PR #7281, opened 2017-01-15
Fix script/server when bin/ is not in PATH

spixi:
This fixes an issue, when the bin/ directory of diaspora is not in the PATH variable


SuperTux88 (approved):

SuperTux88:
Merged as fa0bd14dfbbab7658e2fdbd2ad14b4439967ebd0

Thanks

PR #7285, opened 2017-01-18
Bump json

svbergerem:
Json 1.8.3 was incompatible with Ruby 2.4. Rails 4.2 is still incompatible but that will be fixed in the next release. I tried https://github.com/rails/rails/issues/27670#issuecomment-272433797 and it looks like that's is the only thing left we need for Ruby 2.4 support.


SuperTux88 (approved):

SuperTux88:
Merged as 8f95e973b20791e4b2a0d79a6325a0f2a7d83462

Thanks 👍

PR #7289, opened 2017-01-21
Bump asset_sync

SuperTux88:
This new version now has `fog-core` instead of `fog` as dependency. We can now only add `fog-aws` as dependency and remove all other fog provider (30 dependencies less! 🎉 ).

Since I have no AWS, maybe somebody with AWS can test if everything still works. Pinging @jramos because he did #7201.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,7 +72,7 @@ gem "activerecord-import", "0.15.0"
 
 # File uploading
 
-gem "fog",         "1.38.0", require: "fog/aws"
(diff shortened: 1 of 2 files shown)


jramos:
👍  Tested okay for me.

SuperTux88:
Thanks ❤️

denschub:
Merged as b910c1730ccfe5283a045e4cfa9f61bf90095f28, thanks.

Issue #7293, opened 2017-01-25
Serve security headers


strugee:
There are a couple security-related headers that diaspora* should be serving:

 - [ ] [Content Security Policy](https://content-security-policy.com/)
 - [ ] [X-XSS-Protection](https://helmetjs.github.io/docs/xss-filter/) (note that this one should be disabled in older IE versions because it causes even worse problems there)
 - [ ] [X-Content-Type-Options](https://stackoverflow.com/q/18337630/1198896)
 - [ ] [X-Download-Options](https://helmetjs.github.io/docs/ienoopen/) (mentioning for completeness; probably not needed)

These are mostly nice for a defense-in-depth approach.

denschub:
![](https://media.giphy.com/media/E7iIfUlWHBmh2/giphy.gif)

> Content Security Policy

Already available [at the podmins discretion](https://github.com/diaspora/diaspora/blob/a025a23c71bd1649af57fd5cb44c5d5abf77c0d6/config/diaspora.yml.example#L563-L572).

> X-XSS-Protection

Already always set to `1; mode=block`.

> X-Content-Type-Options

Already always set to `nosniff`.

> X-Download-Options

Already always set to `noopen`.

---

What's this? An issue that you copy and paste to every project that seems appropriate? 😒

strugee:
Ugh! So sorry. I definitely checked my own pod and didn't find these served... :/

I'm terribly embarrassed. 

> What's this? An issue that you copy and paste to every project that seems appropriate?

Really? No. Of course not.

strugee:
Though CSP should really be on by default...

denschub:
We serve most of those headers for multiple years now?! Either you haven't updated in quite a while, or your reverse proxy strips out the headers. In any way, this is something you should fix. ;)

> CSP should really be on by default...

Nah, not really. First of all, we added CSP support in a minor version. We might think about switching the default in the next major, but we're not sure if this is going to happen since it may cause issues with embedded third-party resources (what a surprise).

denschub:
Also, just checked `pod.strugee.net` and realized you're running the pod via Phusion Passenger. Note that this is nothing we support. In fact, we highly discourage doing so for a wide range of reasons (like, for example, we can't control all headers!)

strugee:
Yeah, I don't know. I'm _sure_ I didn't see these when I looked before I filed this, but I can't reproduce anymore. It was an honest mistake. \*sigh*

[...]

PR #7294, opened 2017-01-25
Enable CSP in .diaspora.yml.example

strugee:
Users don't change defaults, so shipping this defaulting to off isn't really enough. It can't be enabled by default for everyone because it might break people's installs unexpectedly, so we enable it only for new installs by putting it in the example config.

Context: https://github.com/diaspora/diaspora/issues/7293#issuecomment-275014255

### DIFF
--- a/config/diaspora.yml.example
+++ b/config/diaspora.yml.example
@@ -574,7 +574,7 @@ configuration: ## Section
       ## Report-Only header (default=true)
       ## By default diaspora* adds only a "Content-Security-Policy-Report-Only" header. If you set
       ## this to false, the "Content-Security-Policy" header is added instead.
-      #report_only: false
+      report_only: false
 
       ## CSP report URI (default=)
       ## You can set an URI here, where the user agent reports violations as JSON document via a POST request.


SuperTux88:
grumpy-no-because-no

The defaults are set in the `defaults.yml` and the `diaspora.yml.example` doesn't set anything, it is only there that podmins can change it.

This setting was added in a minor release and could cause problems for podmins with customizations when enabled by default (so we kept the old behavior). So it at least needs a major release to change the default.

strugee:
@SuperTux88 was the angry cat really necessary...? I really hope you don't respond that way to people brand-new to open source.

And did you actually read the PR description? And make an effort to understand it? I'm _fully_ aware that defaults are set in `defaults.yml`. This patch is about enabling CSP by default **for new installs _only_**. This PR didn't change `defaults.yml` **on purpose** to avoid impacting existing installs which, as you mention, could be broken by the behavior changing in a minor release. Instead, the patch changed `diaspora.yml.example` so that **_new podmins_** would have it on when they copied the example, but existing podmins wouldn't (because they would've already done the copy).

I won't be offended if you reject my patch, but I am offended that you appear to have rejected it without actually having any understanding of how or why it works. Please reopen.

denschub:
All values inside `diaspora.yml.example` are always a representation of the default values. There is not a single setting commented in and we'd like to keep it that way, simply because of the fact we provide sane defaults and we ask podmins to read the available options and make changes if they want to.

When changing the default values in `defaults.yml`, existing installations would be upgraded as well, which is a good thing. However, changes to the defaults require a major release and a note in the changelog by convention. Feel free to submit a PR that changes the default and adds a Changelog note, we have the ability to merge this into our branch for major releases, not the one for minor releases.

> I really hope you don't respond that way to people brand-new to open source.

I would totally have you written that if you didn't open this PR from GitHubs web-editor within seconds after writing a comment elsewhere, rushing there without even giving us a chance to respond. ;)

strugee:
> All values inside diaspora.yml.example are always a representation of the default values. There is not a single setting commented in and we'd like to keep it that way, simply because of the fact we provide sane defaults and we ask podmins to read the available options and make changes if they want to.

OK, that makes sense.

> I would totally have you written that if you didn't open this PR from GitHubs web-editor within seconds after writing a comment elsewhere, rushing there without even giving us a chance to respond. ;)

I would've waited had it been non-trivial. But I didn't bother in this case since it was a one-character change :)

PR #7227, opened 2016-11-27
Refactor mobile alerts for error responses

svbergerem:
first error, failed to unlike
second error, connection

PR #7299, opened 2017-01-28
Fix a jasmine test issue

cmrd-senya:
The issue from a jasmine spec introduced by #7227 caused the testsuite that runs in a browser to be redirected out of the jasmine runner and therefore not finished.


svbergerem:
Thank you for fixing things I broke. ❤️

Issue #7276, opened 2017-01-13
Simplify mention syntax


svbergerem:
Currently our syntax looks like this: `@{FirstName LastName; user@pod.tld}`.
I'd like to change this to `@{user@pod.tld}` and display the name from the person object. This would fix #3526.

This would also make mentions in the publisher a lot easier since the syntax is easy enough to expose it to our users. So instead of having some overlay for the textarea that adds a blue background to the mentioned name and having to distinguish between the “real” value of the textarea (including the mention syntax) and the “fake” value that only shows the name, we could just display the actual textarea. We still have the preview so users can see the rendered markdown.

I talked about this with @SuperTux88 and we also talked about the needed federation changes. What we came up with is adding support for the new syntax (display, create notifications, etc.) but still converting everything to the old syntax when saving the post/comment. As soon as enough pods support the new syntax we could start saving and sending mentions with the new syntax. These changes would be documented in the federation gem.

We also talked with @annando about this, so he also knows that there might be some changes and he didn't see any problems for friendica. (Given that we announce the changes which we would definitely do.)

I played with the new syntax in the develop branch of the markdown-it-diaspora-mention plugin and now it already supports the new syntax: https://github.com/diaspora/markdown-it-diaspora-mention/commit/52281e10d6811bd6a8344bf29104e14d289e6c9d

### TODOs:

- [x] support new syntax in `markdown-it-diaspora-mention` #7277
- [x] support new syntax in backend (and convert to old syntax for backward compatibility) #7300,  #7305
- [x] simplify publisher and use new syntax #7302
- [ ] remove syntax-backporting in the backend (should be done after most pods support the new syntax)

cmrd-senya:
I would prefer, that along with the new syntax we keep a support of the old one, so that some more advanced users could use it to mention a person with any string as a display name. It is useful in some cases. For example, in inflectional languages the word ending changes in non-nominative case. If you say in Russian "I met Senya today", you'll have to change Senya to "Senyu". Most of the social networks ignore that, and people mention other users in nominative case all the time, which looks and feels very weird. People are getting used to it, but only because they've got no choice. But we aren't them, so we can leave this possibility.

svbergerem:
I never said that we should drop the old syntax. ;) All I'd drop is that overlay in the publisher displaying the user's name.

Flaburgan:
[...]

PR #7300, opened 2017-01-29
New mention syntax - backend changes

SuperTux88:
This are the changes needed in the backend for #7276.

It adds support for the new syntax and converts new syntax to the old syntax on create (to be backward compatible with old pods when changing the frontend to the new syntax).

### DIFF
--- a/lib/diaspora/mentionable.rb
+++ b/lib/diaspora/mentionable.rb
@@ -29,8 +29,8 @@ module Diaspora::Mentionable
     people = [*people]
 
     msg_text.to_s.gsub(REGEX) {|match_str|
-      name, handle = mention_attrs(match_str)
-      person = people.find {|p| p.diaspora_handle == handle }
+      name, diaspora_id = mention_attrs(match_str)
+      person = people.find {|p| p.diaspora_handle == diaspora_id }
 
       ERB::Util.h(MentionsInternal.mention_link(person, name, opts))
     }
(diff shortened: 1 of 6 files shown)


lib/diaspora/mentionable.rb
```diff
@@ -1,23 +1,20 @@
...
 
-  # regex for finding mention markup in plain text
-  # ex.
+  # regex for finding mention markup in plain text:
+  #   "message @{user@pod.net} text"
+  # it can also contain a name:
   #   "message @{User Name; user@pod.net} text"
   #   will yield "User Name" and "user@pod.net"
-  REGEX = /(@\{(.+?; [^\}]+)\})/
+  REGEX = /@\{(?:([^\}]+?); )?([^\} ]+)\}/
```
denschub: @svbergerem, you might want to use this RegEx in your `markdown-it` plugin as well!

lib/diaspora/mentionable.rb
```diff
@@ -1,23 +1,20 @@
 
 module Diaspora::Mentionable
 
-  # regex for finding mention markup in plain text
-  # ex.
+  # regex for finding mention markup in plain text:
+  #   "message @{user@pod.net} text"
+  # it can also contain a name:
```
denschub: Nit: `it can also contain a name, which gets used as the link text`, just to be sure.

skyschub (approved):
Looks good!

lib/diaspora/mentionable.rb
SuperTux88: @annando, you might also want to use this RegEx when parsing mentions in friendica :) (It handles the optional name from #7276)

PR #7303, opened 2017-01-30
Use jQuery 3

### LINKED ISSUE #7179: Update jquery to version 3
Its slimmer and faster https://blog.jquery.com/2016/06/09/jquery-3-0-final-released/ It does break diaspora, so we need to make some changes.

cmrd-senya:
fixes #7179 

There are some breaking changes in jQuery 3. They are described in the [upgrade guide](https://jquery.com/upgrade-guide/3.0/).

The deprecated methods that was used in diaspora and had to be replaced:
- `.selector` property of jQuery object has been removed. This was largerly used in jasmine specs in order to test object type. Instead of the deprecated method I used [`.is`](https://api.jquery.com/is/) method. Since this approach requires real objects in DOM to be instanciated, I had to update examples code where it wasn't so.
- `error`, `unload`, `success` event helpers has been replaced with the non-deprecated aliases.

Also there was an issue, when bootstrap tried to use "#" as a selector. "#" is not a valid selector in jQuery 3 anymore and it produces an exception if anyone tries to use it so. It happened when bootstrap [called](https://github.com/twbs/bootstrap/blob/0b9c4a4007c44201dce9a6cc1a38407005c26c86/js/dropdown.js#L25) `attr('data-target')` on an object and received "#". In bootstrap v4 which is a current dev version it [can't happen](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/util.js#L118), but since it's not released yet we can use a hack here. Later f91731d55bfd890f6918336cc44da472d3b395a9 can be just reverted.

Tests are green for me (let's see what travis will say). I didn't do deep manual testing. @pravi, could you test it please, since you were interested in having this implemented?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -324,7 +324,7 @@ GEM
       rake
     jasmine-core (2.5.2)
     jasmine-jquery-rails (2.0.3)
-    jquery-rails (4.2.1)
+    jquery-rails (4.2.2)
       rails-dom-testing (>= 1, < 3)
       railties (>= 4.2.0)
       thor (>= 0.14, < 2.0)
@@ -540,7 +540,7 @@ GEM
     rails-assets-jasmine (2.5.2)
     rails-assets-jasmine-ajax (3.3.1)
       rails-assets-jasmine (~> 2)
-    rails-assets-jquery (2.2.4)
+    rails-assets-jquery (3.1.1)
     rails-assets-jquery-colorbox (1.6.4)
       rails-assets-jquery (>= 1.3.2)
     rails-assets-jquery-fullscreen-plugin (0.5.0)
@@ -831,7 +831,7 @@ DEPENDENCIES
   i18n-inflector-rails (= 1.0.7)
   jasmine (= 2.5.2)
   jasmine-jquery-rails (= 2.0.3)
-  jquery-rails (= 4.2.1)
+  jquery-rails (= 4.2.2)
   jquery-ui-rails (= 5.0.5)
   js-routes (= 1.3.3)
   js_image_paths (= 0.1.0)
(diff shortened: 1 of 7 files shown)


svbergerem:
@cmrd-senya Does https://github.com/diaspora/diaspora/pull/7303/commits/f052d6aecdbe65b8ce0747f511afa4badc749621 also work with jQuery 2? If so, then please move this commit to another PR. Once the PR has been merged it would be great if you could rebase the present PR and only keep the first commit which switches to jQuery 3. All other commits should have been covered by other PRs.

cmrd-senya:
f052d6a works, I didn't move it in a separate PR, because it is somewhat hackish and there is few sense in it as in a separate commit. Do you really think that I should do it? Besides there is a 20b8c92d9c1b9695102f4e2fe086a1a94f912f0f which is the only commit that doesn't work with the old code.

cmrd-senya:
Rebased

Gemfile.lock
```diff
@@ -875,10 +875,7 @@ DEPENDENCIES
   rails-assets-bootstrap-markdown (= 2.10.0)!
   rails-assets-corejs-typeahead (= 1.1.1)!
   rails-assets-diaspora_jsxc (= 0.1.5.develop.7)!
-  rails-assets-fine-uploader (= 5.13.0)!
-  rails-assets-highlightjs (= 9.9.0)!
-  rails-assets-jasmine-ajax (= 3.3.1)!
```
svbergerem: I guess you didn't want to remove these 3 lines.

cmrd-senya:
> I guess you didn't want to remove these 3 lines.

fixed

svbergerem (approved):

svbergerem:
Thank you! :)

PR #7305, opened 2017-01-31
Improve mention syntax

SuperTux88:
I discussed with @svbergerem that we shouldn't generate invalid mentions (that can't be parsed again) automatically in the backend. So I filter `{` and `}` now (they are not supported in mention-texts), an alternative would be to replace them with `(` and `)` or so, but I think remove is better (Only `}` is not supported, but it would be weird to only remove one). This is a temporary solution anyway, only until we support the new syntax everywhere and can remove the server-side backport on create.

And I added the fallback to display the diaspora ID when person is not found. And while doing this I also fixed `filter_people` for that case (regression from #6818).

### DIFF
--- a/lib/diaspora/mentionable.rb
+++ b/lib/diaspora/mentionable.rb
@@ -75,7 +75,7 @@ module Diaspora::Mentionable
     text.to_s.gsub(NEW_SYNTAX_REGEX) do |match_str|
       _, diaspora_id = mention_attrs(match_str)
       person = find_or_fetch_person_by_identifier(diaspora_id)
-      old_syntax = "@{#{person.name}; #{diaspora_id}}" if person
+      old_syntax = "@{#{person.name.delete('{}')}; #{diaspora_id}}" if person
       old_syntax || match_str
     end
   end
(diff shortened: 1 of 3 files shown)


lib/diaspora/mentionable.rb
```diff
@@ -75,7 +75,7 @@ def self.backport_mention_syntax(text)
     text.to_s.gsub(NEW_SYNTAX_REGEX) do |match_str|
       _, diaspora_id = mention_attrs(match_str)
       person = find_or_fetch_person_by_identifier(diaspora_id)
-      old_syntax = "@{#{person.name}; #{diaspora_id}}" if person
+      old_syntax = "@{#{person.name.tr('{}', '')}; #{diaspora_id}}" if person
```
svbergerem: Maybe someone with more Ruby skills should have a look at this but `person.name.delete('{}')` looks more obvious to me.

SuperTux88: I changed it, both do the same and are equally fast, but you're right, `delete` is probably better readable 👍

skyschub (approved):

denschub:
Merged in 29e712dfc8ccdd9b8d54e1584e21dfab7c40cf42.

Issue #7297, opened 2017-01-27
Podmin privileges (private message - block)


SansPseudoFix:
I faced to an user that spam some comments, and I wanted to send him a private message, but he doesn't share with podmin (users automatically share with podmin but he stops sharing). So I wanted to post a comment on one of his posts and… He blocks podmin. So, unless sending him an email or make a public post, I can't contact him.

Proposal: allow podmin to send a private message to his users even if they don't share with him and/or disallow users from blocking podmin.

PR #7309, opened 2017-02-05
Issue#7297 Disallow blocking admins

spixi:
This fixes a part of issue #7297 . It is now impossible to block admins.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,7 @@
 ## Features
 * Add links to the aspects and followed tags pages on mobile [#7265](https://github.com/diaspora/diaspora/pull/7265)
 * diaspora\* is now available in Gàidhlig, Occitan, and Schwiizerdütsch
+* admins cannot be blocked anymore [#7279](https://github.com/diaspora/diaspora/issues/7297)
 
 # 0.6.2.0
 
--- a/app/models/block.rb
+++ b/app/models/block.rb
@@ -8,10 +8,15 @@ class Block < ActiveRecord::Base
   validates :person_id, :presence => true, :uniqueness => { :scope => :user_id }
 
   validate :not_blocking_yourself
+  validate :not_blocking_podmin
 
   def not_blocking_yourself
     if self.user.person.id == self.person_id
       errors[:person_id] << "stop blocking yourself!"
     end
+  end
+
+  def not_blocking_podmin
+    errors[:person_id] << t("blocks.create.failure.podmin") if Role.exists?(person_id: person_id, name: "admin")
   end
 end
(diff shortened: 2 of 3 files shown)


app/models/block.rb
```diff
@@ -8,10 +8,15 @@ class Block < ActiveRecord::Base
...
+  validate :not_blocking_podmin
 
   def not_blocking_yourself
     if self.user.person.id == self.person_id
       errors[:person_id] << "stop blocking yourself!"
     end
   end
+
+  def not_blocking_podmin
+    errors[:person_id] << t('blocks.create.failure.podmin') if Role.exists?(person_id: self.person_id, name: "admin")
```
diaspora-code-review: - Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
- Redundant `self` detected.

spec/models/block_spec.rb
```diff
@@ -4,5 +4,11 @@
       block = alice.blocks.create(person: alice.person)
       expect(block.errors[:person_id].size).to eq(1)
     end
+
+    it "doesnt allow you to block an admin" do
+      Role.add_admin(bob.person)
+      block = alice.blocks.create(person: bob.person)
+      expect(block.errors[:person_id].size.to eq(1)
+    end
```
diaspora-code-review: unexpected token kEND
(Using Ruby 2.2 parser; configure using `TargetRubyVersion` parameter, under `AllCops`)

spixi:
Alice is an admin, so it seems that I broke ```features/desktop/blocks_user.feature``` :(

SuperTux88 (changes requested):
First: Please use pronto locally and review your PR before submitting it, you again needed 3 tries to create it and every push starts a travis run (luckily the first two failed fast).

And I think we shouldn't disallow to block podmins, because my personal account has also the admin role, but I still want to allow people to ignore me. So it would be better to just allow all podmins to send messages, even when they don't share or are blocked, like also described in #7297.

Changelog.md
```diff
@@ -18,6 +18,7 @@
 ## Features
 * Add links to the aspects and followed tags pages on mobile [#7265](https://github.com/diaspora/diaspora/pull/7265)
 * diaspora\* is now available in Gàidhlig, Occitan, and Schwiizerdütsch
+* admins cannot be blocked anymore [#7279](https://github.com/diaspora/diaspora/issues/7297)
```
spixi: Ok, just drop the PR. It was a try. I realize that this is the wrong way to implement this. I see you can still unshare the admin.

SuperTux88: You can still use this PR to fix it another way, or you can close it yourself if you don't want to work at this issue anymore.

denschub:
16 days since

> You can still use this PR to fix it another way, or you can close it yourself if you don't want to work at this issue anymore.

and no reaction. Closing this.

PR #7311, opened 2017-02-06
Make the #newhere post public again

SuperTux88:
This fixes a regression from #7118.

I also did some refactorings after discussing with @svbergerem how to fix this the best. We decided to move the helper methods to the `PublisherHelper` because they are used from the publisher, and the settings for the publisher.

I also removed the `aspects` parameter and used the `all_aspects` helper directly. I think that it is better readable to have less parameters. Also I was first confused whats the difference between `aspects` and `selected_aspects` is.

### DIFF
--- a/app/helpers/aspect_global_helper.rb
+++ b/app/helpers/aspect_global_helper.rb
@@ -11,7 +11,7 @@ module AspectGlobalHelper
     options
   end
 
-  def publisher_aspects_for(stream=nil)
+  def publisher_aspects_for(stream)
     if stream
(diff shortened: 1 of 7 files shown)


svbergerem (approved):

svbergerem:
Merged as d9a12972714a861f88e1780c598728d74e9954f5, 8ebf9c4a4c7eece58f8150bc115d61c91cde6684 and 44f712a2e611a104df262b6b95ca8c385f35d74b. Thank you!

PR #7319, opened 2017-02-11
Increase time to wait before showing the hovercard

svbergerem:
At least for me the current wait time is not long enough so it happens quite often that I open hovercards accidentally. Opening them after a seconds should still be fast enough.


SuperTux88:
Merged as 9558e747e4e9f75b97b7b73e4d4aa05f665e13d8

Thanks

PR #7322, opened 2017-02-11
Create asterisk.png without digest after precompile

### LINKED ISSUE #7317: 404 in assets/branding
Hello,

When a post does not include an image, it gets default value as followed in diaspora/app/helpers/meta_data_helper.rb :
```
def default_image_url
      asset_url "assets/branding/logos/asterisk.png"
end
```
Which is rendered as:
`<meta property="og:image" content="/assets/branding/logos/asterisk.png" />
`

However, URL http(s)://pod.diaspora.url/assets/branding/logos/asterisk.png returns a 404.

http(s)://pod.diaspora.url/assets/branding/asterisk.png does work.

SuperTux88:
Fixes #7317

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -10,9 +10,32 @@ namespace :assets do
     BookmarkletRenderer.compile
   end
 
+  desc "Create non digest assets"
+  task non_digest_assets: :environment do
+    logger = ::Logging::Logger["assets:non_digest_assets"]
+
+    non_digest_assets = Diaspora::Application.config.assets.non_digest_assets
+    manifest_path = Dir.glob(File.join(Rails.root, "public/assets/manifest-*.json")).first
+
+    JSON.load(File.new(manifest_path))["assets"].each do |logical_path, digested_path|
+      logical_pathname = Pathname.new(logical_path)
+      next unless non_digest_assets.any? {|testpath| logical_pathname.fnmatch?(testpath, File::FNM_PATHNAME) }
+
+      full_digested_path     = File.join(Rails.root, "public/assets", digested_path)
+      full_non_digested_path = File.join(Rails.root, "public/assets", logical_path)
+
+      next unless FileUtils.uptodate?(full_digested_path, [full_non_digested_path])
+
+      logger.info "Copying #{full_digested_path} to #{full_non_digested_path}"
+
+      FileUtils.copy_file(full_digested_path, full_non_digested_path, true)
+    end
+  end
+
   # Augment precompile with error page generation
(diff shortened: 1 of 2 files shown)


jhass:
Shouldn't we rather fix linking to a non-digest version?

svbergerem:
Like #7317 mentions this is linked as the OpenGraph image:

```
<meta property="og:image" content="/assets/branding/logos/asterisk.png" />
```

Are the old digest versions still available after creating a few new ones? We only look up the image once in our own implementation.

jhass:
Yes, they're kept until you run `rake assets:clean`.

svbergerem:
https://wiki.diasporafoundation.org/Updating#Post_update_cleanup

jhass:
Yeah, I'm aware, I'm just not too much of fan of side-stepping the cache bursting mechanism here. Perhaps we can rather define some routes with temporary redirects to the digest version?

lib/tasks/assets.rake
```diff
@@ -10,9 +10,32 @@ namespace :assets do
     BookmarkletRenderer.compile
   end
 
+  desc "Create non digest assets"
+  task non_digest_assets: :environment do
+    logger = ::Logging::Logger["assets:non_digest_assets"]
+
+    non_digest_assets = Diaspora::Application.config.assets.non_digest_assets
+    manifest_path = Dir.glob(File.join(Rails.root, "public/assets/manifest-*.json")).first
```
jhass: This feels rather fragile, are we sure there's never more than one manifest?

SuperTux88:
The asterisk.png didn't change since two year, so it still is the same digest version. But every cached opengraph-result now links to the non-digest version.

I copied it once by hand to create the non-digest version, but that doesn't help other pods (and also it isn't there anymore, so I deleted it once by accident, probably when updating to 0.6, because I deleted the whole assets folder to have a clean start).

There is also a gem doing redirects from the non-digest versions to the latest digest version, but then every time somebody loads it, it needs two requests.

lib/tasks/assets.rake
SuperTux88: I only have one file, and it gets updated when I do a precompile. If somebody has two files, it's probably a problem anyway, because sprocket also simply takes the first file: https://github.com/rails/sprockets/blob/v2.12.4/lib/sprockets/manifest.rb#L47-L55

skyschub (approved):
Given the fact that this image is referenced externally and the digest-version could disappear, I am fine with using the non-digest version here!

svbergerem:
Thank you :)

PR #7324, opened 2017-02-12
Display @ before mentions

### LINKED ISSUE #7269: Display @ before mentions
Before mentions in comments were there, I wrote @Person by hand when I answer to multiple persons in one comment, that looks like this (I don't make space between @ and the name normally, but did it here, because I didn't want to mention Alice and Bob on github ;) ):

> @ Alice: answer to alice
>
> @ Bob: answer to bob

Now with mentions in comments we can use mentions for this, that would look like this:

> [Alice](https://example.org/u/alice): answer to alice
> 
> [Bob](https://example.org/u/bob): answer to bob

But I like the @ before a name in the comment, so I would add it manually which would generate a code like this: `@@{Alice; alice@example.org}` and look like this: @[Alice](https://example.org/u/alice)

And friendica supports diaspora-mentions now, but they display an @ before mentions, so if I add an extra @ before the mention, there would be two @@ on friendica (@@[Alice](https://example.org/u/alice)). They could add a hack do remove one of the double @@, but I don't know if they want to do that. ;)

I actually like the idea to mark a mention with an @ (so you know it is hoverable) and I suggest to also add the @ in the frontend on diaspora, and I also think that it doesn't annoy when it is in the middle of a sentence:

> I did something cool with @[Alice](https://example.org/u/alice) today.

What do you think about this idea?

SuperTux88:
Fixes #7269.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,7 +96,7 @@ source "https://rails-assets.org" do
 
   gem "rails-assets-markdown-it",                         "8.2.2"
   gem "rails-assets-markdown-it-hashtag",                 "0.4.0"
-  gem "rails-assets-markdown-it-diaspora-mention",        "1.1.1"
+  gem "rails-assets-markdown-it-diaspora-mention",        "1.2.0"
   gem "rails-assets-markdown-it-sanitizer",               "0.4.3"
   gem "rails-assets-markdown-it--markdown-it-for-inline", "0.1.1"
   gem "rails-assets-markdown-it-sub",                     "1.0.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -551,7 +551,7 @@ GEM
       rails-assets-jquery (>= 1.6)
     rails-assets-markdown-it--markdown-it-for-inline (0.1.1)
     rails-assets-markdown-it (8.2.2)
-    rails-assets-markdown-it-diaspora-mention (1.1.1)
+    rails-assets-markdown-it-diaspora-mention (1.2.0)
     rails-assets-markdown-it-hashtag (0.4.0)
     rails-assets-markdown-it-sanitizer (0.4.3)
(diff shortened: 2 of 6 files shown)


svbergerem (approved):
LGTM

svbergerem:
Thanks! :)

Flaburgan:
Just upgraded d-fr, I find weird that the @ is not part of the link to the user profile.

SuperTux88:
So we have two different ways to do it: as part of the link (like github or twitter does it), or before the link (as google+ does it (with a "+" instead of an "@"), or as friendica did it, when sending mentions as markdown-link in the past).

The difference between github/twitter and diaspora/google+ is, that github/twitter display usernames, and diaspora/google+ displays the realname. Because the realname also could contain (or start with) an "@" and the realname can contain spaces, I thought it would be better to not include the "@" in the link and stick it to the first part of the realname.

For #tags we add the "#" also to the link, but #tags also don't allow special chars or spaces. 

But since @denschub also told me, that he would include it in the link, we should probably change it, and add the "@" to the link? I can do it this evening or we can wait if there are other opinions?

Flaburgan:
Oh okay, nice to know the thinking behind the choice. Well it looks better in my opinion but I understand the point "display name" vs "username".

SuperTux88:
So what do you think? Should we change it? Or should we keep it for a week as it is now, and wait if people get used to it, or if there is more feedback about it? For me both ways are OK, but I preferred the current solution because of the reasons above.

svbergerem:
This has been merged to the develop branch, so I'd say it's ok to keep it for a while and re-evaluate later.

denschub:
Yeah, it's landed, let's leave it there for a bit. If it turns out it is particularly annoying, we can always change it later on.

PR #7318, opened 2017-02-11
Unify blue color variables

svbergerem:
Fixes #6656. All links now use `$link-color` which is the same as `$blue`. `$blue` is only used to color reshare icons. We still have `$brand-primary` which we use for things like buttons or backgrounds.

There are three pronto errors which I'm not planning to fix.

PR #7325, opened 2017-02-12
Remove some unused color-theme overrides

SuperTux88:
After #7318 `$link-color` is used by default, no need to override anymore.

### DIFF
--- a/app/assets/stylesheets/color_themes/_color_theme_override.scss
+++ b/app/assets/stylesheets/color_themes/_color_theme_override.scss
@@ -1,18 +1,5 @@
 /* Raw CSS */
 body {
-  a,
-  a.tag,
-  .btn-link,
-  #main_stream .stream-element > .media a.author-name,
-  #hovercard h4 a,
-  .stream-element .from a.self {
-    color: $link-color;
-
-    &:hover, &:focus {
-      color: darken($link-color, 10%);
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/color_themes/_color_theme_override.scss
```diff
@@ -1,18 +1,5 @@
...
-  a.tag,
-  .btn-link,
-  #main_stream .stream-element > .media a.author-name,
-  #hovercard h4 a,
-  .stream-element .from a.self {
-    color: $link-color;
-
-    &:hover, &:focus {
-      color: darken($link-color, 10%);
-    }
```
svbergerem: This is still used in the themes but it looks different in the original, original white and dark theme. So removing it unifies the themes. 👍

svbergerem (approved):
Looks good to me.

svbergerem:
Thank you!

app/assets/stylesheets/color_themes/_color_theme_override.scss
SuperTux88: Yes, the theme changed minimal, but it's now unified with other themes. I think that some of the effects that are now removed were done "by accident", because simply all links were overridden the same way.

PR #7326, opened 2017-02-13
Change color of author-name on hover

svbergerem:
This already happened before #7325 in some color themes and I don't see why author-name should look different from other links.

For those who review this PR: `author-name` elements are only created via [handlebars-helper/linkToAuthor](https://github.com/diaspora/diaspora/blob/99351bd152f7e4672c77df8d8326995d416525c3/app/assets/javascripts/app/helpers/handlebars-helpers.js#L26) (`grep -Rn "author-name" app`) and this is used in the single post view, stream elements, comments, likes and reshares (`grep -Rn "linkToAuthor" app/`) that are rendered client side.

### DIFF
--- a/app/assets/stylesheets/single-post-view.scss
+++ b/app/assets/stylesheets/single-post-view.scss
@@ -7,7 +7,6 @@
     border-bottom: 1px solid $border-grey;
     padding: 10px 0;
     #post-info {
-      .author{ color: $grey; }
       .info {
         color: lighten($text-grey,10%);
         font-size: 12px;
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/single-post-view.scss
```diff
@@ -22,6 +21,9 @@
         padding-left: 10px;
       }
     }
+
+    .author-name { color: $grey; }
+
```
svbergerem: This change is needed because `.author-name` no longer inherits the color of `.author`.

SuperTux88 (approved):

SuperTux88:
Merged as 5f8c838750c02afa607c2f00d3c5a06ee62ed9c5

Thanks, that makes totally sense :) And it also fixed the blue hover on the single-post-view with the dark theme 👍

PR #7331, opened 2017-02-16
Fix mentioned in comments notifications

SuperTux88:
After upgrading my pod to develop (with mentions in comments) I found some stuff:
* I received multiple notifications when I already commented multiple times on that post
* The notification mails were not added to the thread of the post

And while fixing this, I saw that we add the `From` header again for some notification mails (it is added already in the [`default_headers`](https://github.com/diaspora/diaspora/blob/27b4a44e4a2c2013794eaabc3aaa319aa6909894/app/mailers/notification_mailers/base.rb#L42)), so I cleaned this up too.

### DIFF
--- a/app/models/notifications/mentioned_in_comment.rb
+++ b/app/models/notifications/mentioned_in_comment.rb
@@ -11,7 +11,7 @@ module Notifications
     end
 
     def self.filter_mentions(mentions, mentionable, _recipient_user_ids)
-      mentions.includes(:person).merge(Person.allowed_to_be_mentioned_in_a_comment_to(mentionable.parent)).distinct
+      mentions.includes(:person).merge(Person.allowed_to_be_mentioned_in_a_comment_to(mentionable.parent))
     end
 
     def mail_job
--- a/app/models/notifications/mentioned_in_comment.rb
+++ b/app/models/notifications/mentioned_in_comment.rb
@@ -11,7 +11,7 @@ module Notifications
     end
 
     def self.filter_mentions(mentions, mentionable, _recipient_user_ids)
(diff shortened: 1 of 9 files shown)


Flaburgan:
>Fix receiving multiple mentioned-in-comment notifications

Nice catch, when I received two notifications for you comment, I thought you deleted one first because of a typo :p

SuperTux88:
I received 3, and first thought the same, but then I realized, that they all link to the same guid :)

app/models/notifications/mentioned_in_comment.rb
```diff
@@ -11,7 +11,7 @@ def deleted_translation_key
     end
 
     def self.filter_mentions(mentions, mentionable, _recipient_user_ids)
-      mentions.joins(:person).merge(Person.allowed_to_be_mentioned_in_a_comment_to(mentionable.parent))
+      mentions.includes(:person).merge(Person.allowed_to_be_mentioned_in_a_comment_to(mentionable.parent)).distinct
```
cmrd-senya: Why change from `joins` to `includes`? Both seem to work fine here, so `includes` should be okay. But is there anything better in using `includes`? I checked how it changes the SQL request and there seem to be nothing semantically different.

SuperTux88: `joins` simply adds a `INNER JOIN` for the query, `includes` includes the person in the result. With `joins` a followed `mention.person` queries the person again, while with `includes` it is already included.

cmrd-senya (approved):
Looks good to me!

svbergerem (approved):

svbergerem:
Thanks :)

PR #7344, opened 2017-02-26
Swap sidebar to rightside of view

robertdavid010:
This is simply a swapping of the left sidebar to the right side of the tags view.

The information of the stream in the tags view is most likely of greater priority than the sidebar, and feels more comfortable to the user if this is the case.

### DIFF
--- a/app/views/tags/show.haml
+++ b/app/views/tags/show.haml
@@ -10,15 +10,6 @@
 
 .container-fluid#tags_show
   .row
-    .col-md-3.hidden-xs
-      .sidebar
-        .sidebar-header.clearfix
-          %h3
-            = t(".tagged_people", count: @stream.tagged_people_count, tag: @stream.display_tag_name)
-
-        .side_stream.stream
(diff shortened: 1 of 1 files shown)


Flaburgan:
Even if I agree that the posts are probably more important than the list of users, I don't feel like we should move the column, that's how the whole interface is, on the stream page, on the user page, on the private message page... So I would prefer to keep the interface consistent here.

robertdavid010:
Ok thanks. Would it be considerable to change this layout app wide?

—
R

Flaburgan:
Almost every website out there have the menu on the left. I think the users will be more disturbed than anything. Opinions?

goobertron:
Left seems natural to me.

denschub:
Based on our interface flow, left makes the most sense, yeah. However, for RTL language, we'd have to put it on the right side. But eh... maybe one day if we're easily able to rewrite our frontend... ;)

For now, I agree with @Flaburgan and @goobertron - and I will close this PR. For future reference, drastic UI changes like this one are probably best discussed on Discourse first. :)

PR #7345, opened 2017-02-28
Update rails

svbergerem:

PR #7351, opened 2017-03-04
Pin thor to prevent accidental upgrades

SuperTux88:
Thor was upgraded again in #7345, but there are still "warnings" looking
like errors when using "db:migrate" for example. So lets pin it until it is
safe to upgrade.


svbergerem (approved):

Flaburgan:
Is the `develop` branch broken right now because of that or is it safe to update?

SuperTux88:
It is safe to update, this only prints a warning (`Expected string default value for '--serializer'; got true (boolean)`) that looks like an error and can be weird for podmins.

svbergerem:
Thanks!

PR #7353, opened 2017-03-05
Disable Rails/HttpPositionalArguments cop

svbergerem:
See http://rubocop.readthedocs.io/en/latest/cops_rails/#railshttppositionalarguments.

It only applies to Rails >= 5


SuperTux88:
Merged as 69175bc47b4a1eb0fd8b15ba47017ea9d5539130

Thanks

PR #7355, opened 2017-03-06
Stabilize cuke features/desktop/signs_up.feature:46

cmrd-senya:


features/support/publishing_cuke_helpers.rb
```diff
@@ -27,6 +27,8 @@ def submit_publisher
     txt = find("#publisher #status_message_text").value
     find("#publisher .btn-primary").click
     # wait for the content to appear
+    expect(find("#publisher")).not_to have_css(".btn-primary")
```
SuperTux88: This is still failing:
`expected not to find css ".btn-primary", found 1 match: "Share". Also found "", which matched the selector but not all filters. (RSpec::Expectations::ExpectationNotMetError)`

cmrd-senya:
I tryed to catch the issue localy by reducing the resources of the virtual machine I ran the tests on, but had no luck.

Will it be useful for random cukes failure debugging if we make screenshot and save HTML when the exceptions are raised (expectations aren't met)? Beside making screenshots we'll also need to save this data after travis builds, which can be achieved by uploading generated files (screenshots, htmls) to S3. I think it can be useful because it's quite hard to reproduce this random failures locally. So getting some more information from faulty runs might be useful.

Flaburgan:
I had a similar approach in https://github.com/diaspora/diaspora/pull/7384 but it was not enough apparently. In any case this has to be rebased if still useful.

cmrd-senya:
Closing, since it doesn't seem very useful.

PR #7356, opened 2017-03-06
Remove jQuery deprecations

cmrd-senya:
These are commits from #7303 which work with jQuery 2 and which are replacing deprecated features in jQuery.


svbergerem (approved):

svbergerem:
Thanks!

PR #7358, opened 2017-03-08
Update LISENCE

PiumiHR:
Corrected some grammar mistakes


SuperTux88:
The `LICENSE` file is copied from the original AGPL-3.0 license, see http://www.gnu.org/licenses/agpl-3.0.txt. So I think we shouldn't change it ourselves.

Also (for next time): please don't create PRs from your `develop` branch. Instead of merging your `PiumiHR-patch-1` branch into your `develop` you can create a PR directly from `PiumiHR-patch-1`.

PR #7367, opened 2017-03-15
Improve CSP to avoid confilcts with cloudflare

Amadren:
CSP was causing a lot of troubles in my pod (I use Cloudflare) so I decided to remove it and I thought that it could be a good idea to make it easier to remove csp.

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -38,7 +38,7 @@ SecureHeaders::Configuration.default do |config|
 
   if AppConfig.settings.csp.report_only?
     config.csp = SecureHeaders::OPT_OUT
-    config.csp_report_only = csp
+    config.csp_report_only = csp if AppConfig.settings.csp.report_uri.present?
   else
     config.csp = csp
   end
--- a/config/diaspora.yml.example
+++ b/config/diaspora.yml.example
@@ -565,6 +565,7 @@ configuration: ## Section
     ## does not work with Google analytics or Piwik, because they inject JS code that
     ## is blocked by CSP.
     csp:
+
       ## Report-Only header (default=true)
       ## By default diaspora* adds only a "Content-Security-Policy-Report-Only" header. If you set
       ## this to false, the "Content-Security-Policy" header is added instead.


SuperTux88 (changes requested):
First: Thanks for your contribution.

Just out of interest: Can you tell what troubles you had with Cloudflare? Because the current default shouldn't make any trouble, it only print's a warning sometimes in browsers (but only warning, not error).

But the warning was the reason why I also tried to disable CSP completely, but I wasn't able to do it, because I always got trouble when setting the nonce with disabled CSP. But that was an older version of the `secure_headers`-gem back then, I didn't realize that it is now possible with the new version, so thanks for the pull-request. 👍

But I wouldn't add a new setting to disable/enable it, because the current default doesn't make any sense with that. We add it with report-only by default, but without an report-uri, report-only is useless, so I would only enable report-only CSP when also a uri is set:

```diff
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -38,7 +38,7 @@ SecureHeaders::Configuration.default do |config|

   if AppConfig.settings.csp.report_only?
     config.csp = SecureHeaders::OPT_OUT
-    config.csp_report_only = csp
+    config.csp_report_only = csp if AppConfig.settings.csp.report_uri.present?
   else
     config.csp = csp
   end
```

That's all I would change in the secure_headers.rb. With that it is disabled by default (like you want it), and enables automatically once you change either `settings.csp.report_only` to `false` or set a `settings.csp.report_uri`.

And we should probably also change the comment above `report_only` in the `diaspora.yml.example` that you need to set `report_only` to `false` to enable CSP.

Amadren:
Hello @SuperTux88 

I had problems with cloudflare's js scripts. When some were turned on (like auto-minify etc..), the report-only policy not only reported but completly disabled every scripts, turning my pod's aesthetic into garbage (there was no js nor css loaded). It wasn't a problem with my computer/ip/cache/whatever because I tried it in several places with several computers before figuring out that it was because of the report_only policy that it wasn't working.

I'll try your changes tonight and will change my pull request if it's working correctly on my pod :)

denschub:
Just a side note to you as a podmin, not really related to this PR (but also not worth opening a second communication channel since I was not able to find you on Discourse):

> When some were turned on (like auto-minify etc..), the report-only policy not only reported but completly disabled every scripts

That is, to be fair, not a bug. Cloudflare changes the contents of the asset files, thus invalidating them and we block them for good (it could very well be that someone tried to MitM the connection there). Also, since we already minify all assets in the production env, you should not really need that.

Amadren:
Sorry everyone for the late reply but it took me a long time to be able to test @SuperTux88's changes.

Anyway:
@denschub The point is that cloudflare's settings are domain-wide and for people like me who have pods in subdomains (pod.kakise.xyz), it is great to be able to not change any of my previous settings for one subdomain.

@SuperTux88 Maybe I did something wrong but your changes didn't work for me. Same problem as before. Can you try them on your pod or something? 
Thank you

SuperTux88:
Can you tell me what the exact problem is? Is the header still added? Why is it blocking, because by default it is added as `Content-Security-Policy-Report-Only`.

So you removed all your other changes and added only my change, and you have the default config:
```
       report_only: true
       report_uri:
```

It would set `config.csp` to `SecureHeaders::OPT_OUT` because `report_only` is true by default, and it wouldn't set `config.csp_report_only` because `report_uri` is empty by default (and the default for `csp_report_only` is `OPT_OUT`). That works perfectly here and no header is added. So either you have changed your settings, or you have changed something else or cloudflare adds a header by its own?

Amadren:
@SuperTux88 Ok. It seems to be cloudflare's cache that added the headers on its own. After a second try and emptying the cache, it worked perfectly.

config/initializers/secure_headers.rb
```diff
@@ -35,11 +35,16 @@
   csp[:form_action] << "www.paypal.com" if AppConfig.settings.paypal_donations.enable?
 
   csp[:report_uri] = [AppConfig.settings.csp.report_uri] if AppConfig.settings.csp.report_uri.present?
-
-  if AppConfig.settings.csp.report_only?
-    config.csp = SecureHeaders::OPT_OUT
-    config.csp_report_only = csp
+
```
diaspora-code-review: Trailing whitespace detected.

config/initializers/secure_headers.rb
```diff
@@ -35,11 +35,16 @@
   csp[:form_action] << "www.paypal.com" if AppConfig.settings.paypal_donations.enable?
 
   csp[:report_uri] = [AppConfig.settings.csp.report_uri] if AppConfig.settings.csp.report_uri.present?
-
-  if AppConfig.settings.csp.report_only?
-    config.csp = SecureHeaders::OPT_OUT
-    config.csp_report_only = csp
+  
+  if AppConfig.settings.csp.enabled?
```
diaspora-code-review: Trailing whitespace detected.

Amadren:
@SuperTux88: Here, I commited the requested changes ;)

config/initializers/secure_headers.rb
```diff
@@ -35,11 +35,11 @@
...
 
   csp[:report_uri] = [AppConfig.settings.csp.report_uri] if AppConfig.settings.csp.report_uri.present?
-
-  if AppConfig.settings.csp.report_only?
-    config.csp = SecureHeaders::OPT_OUT
-    config.csp_report_only = csp
-  else
-    config.csp = csp
-  end
+
```
diaspora-code-review: Trailing whitespace detected.

config/initializers/secure_headers.rb
```diff
@@ -35,11 +35,11 @@
...
   csp[:report_uri] = [AppConfig.settings.csp.report_uri] if AppConfig.settings.csp.report_uri.present?
-
-  if AppConfig.settings.csp.report_only?
-    config.csp = SecureHeaders::OPT_OUT
-    config.csp_report_only = csp
-  else
-    config.csp = csp
-  end
+  
+   if AppConfig.settings.csp.report_only?
```
diaspora-code-review: Inconsistent indentation detected.

Flaburgan:
Thanks for your contribution @Amadren, can you please fix the pronto indication and squash your commits into one with `git rebase -i`?

config/initializers/secure_headers.rb
```diff
@@ -38,7 +38,7 @@
 
   if AppConfig.settings.csp.report_only?
     config.csp = SecureHeaders::OPT_OUT
-    config.csp_report_only = csp
+    config.csp_report_only = csp if AppConfig.settings.csp.report_uri.present?
```
diaspora-code-review: Trailing whitespace detected.

SuperTux88:
As @Flaburgan already wrote: can you please squash your commits to a single one. But since you created the pull-request from your develop branch `git rebase -i develop` wouldn't work. If you added the upstream-repo as `upstream`, you can use `git rebase -i upstream/develop`. You can also use the latest commit from the upstream-develop: `git rebase -i 78359d69d85e1a64864f665fe2d07def1d9b1636`

For the next time create a new branch before working on a PR.

SuperTux88:
You can ignore the last comment from @diaspora-code-review, you pushed so many commits so quickly that the comment was added after you changed it again.

Amadren:
The rebase did something weird

SuperTux88:
I have no idea what exactly you did ... but somehow you managed to rebase another commit from me, and you still have all other commits oO

Amadren:
This is super weird :/ I only used `git commit rebase -i upstream/develop` and it did this

SuperTux88:
hmm, then maybe your `upstream/develop` wasn't up to date. Can you do:
`git reset --hard dc09e9c509f6d4b15e0728f552548dd0bb5e5676` (to undo your rebase)

Then do:
* `git rebase -i 78359d69d85e1a64864f665fe2d07def1d9b1636`
* Mark all but the first commit with `squash` (the first should stay `pick`)
* Save and exit the editor
* Write a new commit message for your commit
* and force-push again.

Amadren:
Here, sorry for the trouble, I'll make a branch next time :)

SuperTux88 (approved):

SuperTux88:
Merged as 30fad8279b4f0df645d0b025cc78296c07e66472

Thanks for your contribution 👍 

Also: You should now reset your develop branch back to the upstream develop:
* `git reset --hard 78359d69d85e1a64864f665fe2d07def1d9b1636`
* `git pull upstream develop`

PR #7371, opened 2017-03-16
Remove custom jquery input event

svbergerem:
Native browser support looks good to me for IE >= 9: https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oninput


Flaburgan:
Yep, if we ever want to care about IE users it should not be for something < 11.

SuperTux88 (approved):

SuperTux88:
Merged as aa55a7a023ea1e96d699a3e4140b5be35049f3f5

Thanks 👍

PR #7372, opened 2017-03-17
Use empty selector where "#" was used as a selector before (jQuery 2 -> 3 upgrade)

cmrd-senya:
In jQuery 3 "#" is disallowed as a selector. Here is a somewhat hacky way to force use an empty selector where "#" was used as a selector before.

See #7303


svbergerem (approved):

svbergerem:
Merged as 526ebcfa13a799cae6bfbcc822c958ebecfa4d2b. Thanks!

PR #7375, opened 2017-03-19
Disable autocomplete for the conversation form recipient input

svbergerem:
We use typeahead for the text field so the browser autocomplete should be turned off.


SuperTux88 (approved):

SuperTux88:
Merged as d3de97b244b0b09b7dc9a753c9c17b45793d56bb

Thanks :)

Issue #7373, opened 2017-03-17
Fix randomly failing specs


svbergerem:
**RSpec**
* [ ] Notifications::Mentioned.notify calls filter_mentions on self 
[spec](https://github.com/diaspora/diaspora/blob/ca6ce912d390a8beb44e45ffd6d5f34272f5de8f/spec/models/notifications/mentioned_spec.rb#L15), [example](https://travis-ci.org/diaspora/diaspora/jobs/212061463#L655)

**Cucumber**
* [ ] features/desktop/blocks_user.feature, Scenario: Blocking a user from the profile page
[spec](https://github.com/diaspora/diaspora/blob/ca6ce912d390a8beb44e45ffd6d5f34272f5de8f/features/desktop/blocks_user.feature#L18), [example](https://travis-ci.org/diaspora/diaspora/jobs/212061462#L678)
* [ ] features/desktop/comments.feature, Scenario: comment on a status show page
[spec](https://github.com/diaspora/diaspora/blob/ca6ce912d390a8beb44e45ffd6d5f34272f5de8f/features/desktop/comments.feature#L45), [example](https://travis-ci.org/diaspora/diaspora/jobs/212071615#L820)
* [ ] features/desktop/invitations.feature, Scenario: sends an invitation from the stream
[spec](https://github.com/diaspora/diaspora/blob/ca6ce912d390a8beb44e45ffd6d5f34272f5de8f/features/desktop/invitations.feature#L52), [example](https://travis-ci.org/diaspora/diaspora/jobs/212071570#L1328)
* [ ] features/desktop/mentions_from_profile_page.feature, Scenario: mentioning while posting to all aspects
[spec](https://github.com/diaspora/diaspora/blob/ca6ce912d390a8beb44e45ffd6d5f34272f5de8f/features/desktop/mentions_from_profile_page.feature#L22). [example](https://travis-ci.org/diaspora/diaspora/jobs/212061465#L1655)
* [ ] features/desktop/mentions_from_profile_page.feature, Scenario: mentioning while posting to just one aspect
[spec](https://github.com/diaspora/diaspora/blob/ca6ce912d390a8beb44e45ffd6d5f34272f5de8f/features/desktop/mentions_from_profile_page.feature#L36), [example](https://travis-ci.org/diaspora/diaspora/jobs/208337921#L1654)
* [ ] features/desktop/reshare.feature, Scenario: Resharing a post from a single post page that is reshared
[...]

PR #7379, opened 2017-03-20
Fix block user failing feature

Flaburgan:
Alright so I was annoyed by the "randomly" failing features and especially the `features/desktop/blocks_user.feature:18 # Scenario: Blocking a user from the profile page` one so I decided to have a quick look. And... I discovered that the failure was not random at all. In fact, the test was failing for a good reason, and it is the local green which was a bug. Since https://github.com/diaspora/diaspora/pull/5027 we immediately remove the posts from a stream when the author is ignored. And since https://github.com/diaspora/diaspora/pull/6617 we still display posts of an author on his profile. This works fine on F5, but **when you ignore a user on his profile pages, the posts are removed**, making the test fails (as it checks the post is displayed). There is a transition, so phantomJS is too quick and the post is still there when the check is done locally, but on Travis, the post is removed before the check is done, so the test failed.

In this PR, I fixed the problem first by checking in the test that the dropdown becomes "stop ignoring" when a user is ignored, so locally it has another step to do before it checks if the post is there. This makes the test becomes red like on travis. Then I fixed the code by checking in the StreamPost view that we are not on a user profile, based on the URL, to activate the "remove post immediately" feature. Making the test green again, locally and, hopefully, on Travis.

There is probably a better way to know if we're on the author profile than to use the URL. Suggestions? @svbergerem

### DIFF
--- a/app/assets/javascripts/app/views/stream_post_views.js
+++ b/app/assets/javascripts/app/views/stream_post_views.js
@@ -28,9 +28,11 @@ app.views.StreamPost = app.views.Post.extend({
                      ".permalink"].join(", "),
 
   initialize : function(){
-    var personId = this.model.get("author").id;
-    app.events.on("person:block:"+personId, this.remove, this);
-
+    // If we are on a user page, we don't want to remove posts on block
+    if (!app.page.model.has("profile")) {
+      var personId = this.model.get("author").id;
+      app.events.on("person:block:" + personId, this.remove, this);
+    }
     this.model.on("remove", this.remove, this);
     //subviews
     this.commentStreamView = new app.views.CommentStream({model : this.model});
--- a/features/desktop/blocks_user.feature
+++ b/features/desktop/blocks_user.feature
@@ -18,6 +18,7 @@ Feature: Blocking a user from the stream
   Scenario: Blocking a user from the profile page
     When I am on "alice@alice.alice"'s page
     And I confirm the alert after I click on the profile block button
-    Then "All your base are belong to us!" should be post 1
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/app/views/stream_post_views.js
```diff
@@ -28,9 +28,12 @@ app.views.StreamPost = app.views.Post.extend({
                      ".permalink"].join(", "),
 
   initialize : function(){
-    var personId = this.model.get("author").id;
-    app.events.on("person:block:"+personId, this.remove, this);
-
+    var baseUrl = Backbone.history.fragment.split("/")[0];
+      // If we are on a user page, we don't want to remove posts on block
+    if (baseUrl !== "people" && baseUrl !== "u") {
```
SuperTux88: Maybe you can check the `page-people` class on the `<body>`? That's the only way I know to detect on which page you are, but I don't really know the frontend, so maybe @svbergerem still has a better idea? ;)

Flaburgan:
This solves the most annoying of #7373 btw.

Flaburgan:
Well, at least the one I try to solve is green on travis \o/

app/assets/javascripts/app/views/stream_post_views.js
svbergerem:
```javascript
if(!app.page.model.has("profile")) {
  ...
}
```
should also work.

Flaburgan:
Steffen suggestion works, PR updated.

svbergerem (approved):

svbergerem:
Thank you! ❤️

PR #7380, opened 2017-03-21
Refactor mentions-input-box css

svbergerem:
Most of the textarea styles are already covered by publisher.scss and `.mentions-autocomplete-list` doesn't exist anymore. I only had to move the textarea background to publisher.scss and add `position: relative` to the textarea wrapper.


SuperTux88 (approved):

SuperTux88:
Thanks for the cleanup 🍪

PR #7383, opened 2017-03-22
Increase maximal height of large thumbnail

### LINKED ISSUE #6248: :thumb_large size should only resize width to 300px
This is related to @Flaburgan 's comment [here](https://github.com/diaspora/diaspora/pull/6244#issuecomment-124727525):

> So, here is the real problem: thumb_large is not resized to be 300px width. It is its larger dimension, either height or width, which is resized to 300px. So, for images which have a big height and a small width, the result is catastrophic. Here is a good example: https://diaspora-fr.org/posts/901677

Furthermore, I think networks today are fast enough to support a little increase of that size. I think `thumb_large` size can be insreased to 400 or 500px.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24909741-thumb_large-size-should-only-resize-width-to-300px?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
Fixes #6248, supersedes #6244.

Here is a test with a 500x3000px image:

**before**
screen shot 2017-03-22 at 12 58 22

**after**
screen shot 2017-03-22 at 12 57 43

### DIFF
--- a/app/uploaders/processed_image.rb
+++ b/app/uploaders/processed_image.rb
@@ -18,16 +18,15 @@ class ProcessedImage < CarrierWave::Uploader::Base
   end
 
   version :thumb_small do
-    process :resize_to_fill => [50,50]
+    process resize_to_fill: [50, 50]
   end
   version :thumb_medium do
-    process :resize_to_limit => [100,100]
+    process resize_to_limit: [100, 100]
   end
   version :thumb_large do
-    process :resize_to_limit => [300,300]
+    process resize_to_limit: [300, 1500]
   end
   version :scaled_full do
-    process :resize_to_limit => [700,nil]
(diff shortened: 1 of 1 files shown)


SuperTux88 (approved):

SuperTux88:
Merged as 6f3c61d8f2c2dcbfae0c2998ac2aa681ec6d12f7

That's the solution @denschub suggested at https://github.com/diaspora/diaspora/pull/6244#issuecomment-260892412 👍 

Thank you :)

PR #7384, opened 2017-03-22
Attempt to mitigate new user message failing test

Flaburgan:
#7373

- new user without any tags posts first status message
- new user with some tags posts first status message

### DIFF
--- a/features/support/publishing_cuke_helpers.rb
+++ b/features/support/publishing_cuke_helpers.rb
@@ -26,6 +26,8 @@ module PublishingCukeHelpers
   def submit_publisher
     txt = find("#publisher #status_message_text").value
     find("#publisher .btn-primary").click
+    # wait for the publisher to be closed
+    expect(find("#publisher")["class"]).to include("closed")
     # wait for the content to appear
     expect(find("#main_stream")).to have_content(txt)
   end
--- a/features/desktop/signs_up.feature
+++ b/features/desktop/signs_up.feature
@@ -48,7 +48,9 @@ Feature: new user registration
     When I confirm the alert after I follow "awesome_button"
     Then I should be on the stream page
     And the publisher should be expanded
-    When I submit the publisher
+    When I wait for the popovers to appear
+    And I click close on all the popovers
+    And I submit the publisher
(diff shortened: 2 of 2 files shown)


features/support/publishing_cuke_helpers.rb
```diff
@@ -26,6 +26,8 @@ def make_post(text)
   def submit_publisher
     txt = find("#publisher #status_message_text").value
     find("#publisher .btn-primary").click
+    # wait for the publisher to be closed
+    find("#publisher")["class"].should include("closed")
```
Flaburgan: Interesting, this now [fails](https://travis-ci.org/diaspora/diaspora/jobs/213826386#L2763) as `expected "mention_popup publisher row" to include "closed"`, what's the link with "mention_popup" here... ? Is that the infobulle which is open with when getting_started is on?

svbergerem: We add the class [when the publisher should be opened on page load](https://github.com/diaspora/diaspora/blob/55fee86a95665446b197e16606be81378664e773/app/views/publisher/_publisher.html.haml#L1). (However the class is not needed anymore)

Maybe the getting started popovers are a problem. You could close them before doing anything with the publisher (like submitting the publisher).

```
And I wait for the popovers to appear
And I click close on all the popovers
```

Flaburgan: This is taken from `https://github.com/diaspora/diaspora/blob/55fee86a95665446b197e16606be81378664e773/features/step_definitions/custom_web_steps.rb#L72` and I confirm that even if the page is loaded with the publisher open, the `closed` class is added to it when the post is submitted (this test would fail otherwise, but it was green with Ruby 2.3 and locally).

I thought about removing the popovers and I'll try that if it helps, however this changes the test behavior, it means we do not test anymore that the publisher can be submitted even if the popovers are there.

svbergerem (approved):

svbergerem:
Merged as 30c2d5d9ac7825e3a4c70a01418da4c3c0e80c34 and ce15f41a041fc739dd388a8fc92f42f3ace6cb03. Thank you! :)

PR #7385, opened 2017-03-23
Remove posts after response

### LINKED ISSUE #5445: Post removed from stream even if 'destroy' fails
When deleting a post it is removed from the stream even if the request to the server fails.

Removing `this.model.on('remove', this.remove, this);` from `app.views.StreamPost.initialize` (https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/views/stream_post_views.js#L35) would solve that issue but would break updating a post preview and could also break hiding posts.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6640439-post-removed-from-stream-even-if-destroy-fails?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

svbergerem:
With this PR the opacity of posts is decreased after clicking the delete button. If the server response shows a successful removal the post is removed from the stream. Otherwise it goes back to the initial state.

post_remove_before

post_remove_after

Fixes #5445.

### DIFF
--- a/app/assets/javascripts/app/views.js
+++ b/app/assets/javascripts/app/views.js
@@ -132,19 +132,20 @@ app.views.Base = Backbone.View.extend({
 
   destroyModel: function(evt) {
     evt && evt.preventDefault();
-    var self = this;
     var url = this.model.urlRoot + "/" + this.model.id;
 
     if( confirm(_.result(this, "destroyConfirmMsg")) ) {
       this.$el.addClass("deleting");
-      this.model.destroy({ url: url })
-        .done(function() {
-          self.remove();
-        })
-        .fail(function() {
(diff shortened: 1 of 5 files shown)


SuperTux88 (approved):

SuperTux88:
Merged as 8a98cd4517e599724b3ca3b68e233c3e6038e848 and 3ce4bba3831dcbcb1ed683c93c25025d7e660e65

Thank you 👍

PR #7386, opened 2017-03-23
Comment mentions front-end

cmrd-senya:
Supersedes #7275 

It is missing some tests still.

There is also a CSS issue I couldn't solve easily, described [here](https://github.com/diaspora/diaspora/compare/develop...cmrd-senya:comment-mention-front-end?expand=1#diff-56e750d910393e15f705b54a78a6ea33R75). I made a workaround, which is probably fine for now, since otherwise it probably need some deeper CSS refactoring.

### DIFF
--- a/app/assets/javascripts/app/views/comment_mention_view.js
+++ b/app/assets/javascripts/app/views/comment_mention_view.js
@@ -1,0 +1,10 @@
+// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-v3-or-Later
+//= require ./publisher/mention_view
+
+app.views.CommentMention = app.views.PublisherMention.extend({
+  initialize: function(opts) {
+    opts.url = Routes.mentionablePost(opts.postId);
+    app.views.PublisherMention.prototype.initialize.call(this, opts);
+  }
+});
+// @license-end
--- a/app/assets/javascripts/app/views/comment_stream_view.js
+++ b/app/assets/javascripts/app/views/comment_stream_view.js
@@ -27,6 +27,7 @@ app.views.CommentStream = app.views.Base.extend({
     this.model.comments.each(this.appendComment, this);
     this.commentBox = this.$(".comment_box");
     this.commentSubmitButton = this.$("input[name='commit']");
+    new app.views.CommentMention({el: this.$el, postId: this.model.get("id")});
   },
 
   presenter: function(){
--- a/app/assets/stylesheets/mentions.scss
+++ b/app/assets/stylesheets/mentions.scss
@@ -65,3 +65,21 @@
     }
   }
 }
+
+.typeahead-mention-box-wrap .twitter-typeahead {
+  left: -1px;
+  width: calc(100% + 2px);
+
+  .tt-menu {
+    // Override inline rule of Typeahead
+    // If this is not overridden (`position: absolute` by default in Typeahead) then
+    // the box is cut when opened because of the `overflow: hidden` from parent classes of comment form styles. By
+    // having `position: relative` here we make it visible by inserting it in the flow.
+    // This has a side effect of "Comment" button move down when box is open, but it feels like the least evil.
+    // scss-lint:disable ImportantRule
+    position: relative !important;
+    // scss-lint:enable ImportantRule
+
+    width: 100%;
+  }
+}
--- a/app/assets/stylesheets/publisher.scss
+++ b/app/assets/stylesheets/publisher.scss
@@ -19,12 +19,6 @@
 
   .container-fluid{ padding: 0; }
 
-  .twitter-typeahead {
-    width: calc(100% + 2px);
-
-    .tt-menu { width: 100%; }
-  }
-
   form {
     margin: 0;
     #fileInfo { display: none !important; }
@@ -245,7 +239,6 @@
   }
 
   .twitter-typeahead {
-    left: -1px;
     // Override inline rule of Typeahead
     // scss-lint:disable ImportantRule
     position: absolute !important;
--- a/app/assets/javascripts/app/views/publisher/mention_view.js
+++ b/app/assets/javascripts/app/views/publisher/mention_view.js
@@ -6,15 +6,16 @@ app.views.PublisherMention = app.views.SearchBase.extend({
   mentionSyntaxTemplate: function(person) { return "@{" + person.handle + "}"; },
 
   events: {
-    "keydown #status_message_text": "onInputBoxKeyDown",
-    "input #status_message_text": "updateTypeaheadInput",
-    "click #status_message_text": "onInputBoxClick",
-    "blur #status_message_text": "onInputBoxBlur"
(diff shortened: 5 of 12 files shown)


svbergerem (changes requested):
This PR is looking good so far. Just write some tests and fix the remark, then I'll do a proper review.

app/assets/stylesheets/publisher.scss
```diff
@@ -243,14 +237,6 @@
     .hide-location { display: inline-block; }
     .locator { display: none; }
   }
-
-  .twitter-typeahead {
-    left: -1px;
-    // Override inline rule of Typeahead
-    // scss-lint:disable ImportantRule
-    position: absolute !important;
```
svbergerem: Please keep `position: absolute` for the publisher.

publisher relative typeahead

Having `position: relative` everywhere else isn't perfect but I wasn't able to find a better solution.

svbergerem (changes requested):

app/assets/stylesheets/publisher.scss
cmrd-senya: If I keep this line, width for the comment publisher gets improper:



However I don't see any issues without this line in the main publisher (exacly like at your screenshot):



So what's wrong with the CSS I proposed there?

cmrd-senya: On the second though, I think that it is more clear to fix the width issue instead of removing `position: absolute`. I'll try to do that now.

cmrd-senya: Oh, width suffers from exactly the same problem: if the box is absolutely positioned, then it's width is defined by the closest relatively positioned parent. If this parent is outside the comment input box, the we have wrong width. If it is inside the comment input box, then suggestions box get cut because of the `overflow: hidden` setting of the comment widget. Maybe it can be solved by removing `overflow: hidden` (there is no `overflow: hidden` property used in the post publisher), but it breaks the layout this way, so this must be solved also.

svbergerem: I'm unable to reproduce your issues with the width of the comment mentions, but maybe there is just some misunderstanding. This is the diff between your current version and the one I'm talking about:

```diff
diff --git a/app/assets/stylesheets/publisher.scss b/app/assets/stylesheets/publisher.scss
index 70e89b3d0..f95480c94 100644
--- a/app/assets/stylesheets/publisher.scss
+++ b/app/assets/stylesheets/publisher.scss
@@ -237,6 +237,13 @@
     .hide-location { display: inline-block; }
     .locator { display: none; }
   }
+
+  .twitter-typeahead {
+    // Override inline rule of Typeahead
+    // scss-lint:disable ImportantRule
+    position: absolute !important;
+    // scss-lint:enable ImportantRule
+  }
 }
 
 .publisher-buttonbar {
```

This only adds styles to the publisher (inside `.publisher`), so I don't see how that could break the comment mentions styles.

The problem with the current version is the fact that mention suggestions are between the textarea and photos / location / polls / the submit button. This means that when you start mentioning someone everything below the suggestions might move with every single character you type.

cmrd-senya: > This only adds styles to the publisher (inside .publisher), so I don't see how that could break the comment mentions styles.

I see what you mean, I'll fix it now.

> The problem with the current version is the fact that mention suggestions are between the textarea and photos / location / polls / the submit button. This means that when you start mentioning someone everything below the suggestions might move with every single character you type.

Yes, and I don't see any simple solutions, since if I make it overlapping, it gets cut because of `overflow: hidden` property of the box. A possible solution could be to remove this property from the styles, but it requires to fix layout then, which is unobvious for me.

svbergerem: Well, adding the above code to the publisher css fixes the problem. It doesn't fix it for the comment textarea, but as I wrote above I also don't have a nice solution here, so the current behavior for the comment textarea looks ok to me.

cmrd-senya:
Do I squash it now?

app/assets/javascripts/app/views/comment_mention_view.js
```diff
@@ -0,0 +1,8 @@
+//= require ./publisher/mention_view
```
svbergerem: Please add
```javascript
// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-v3-or-Later
```
at the top of the file.

app/assets/javascripts/app/views/comment_mention_view.js
```diff
@@ -0,0 +1,8 @@
...
+
+app.views.CommentMention = app.views.PublisherMention.extend({
+  initialize: function(opts) {
+    opts.url = Routes.mentionablePost(opts.postId);
+    app.views.PublisherMention.prototype.initialize.call(this, opts);
+  }
+});
```
svbergerem: And add
```javascript
// @license-end
```
at the end.

app/assets/javascripts/app/views/comment_mention_view.js
```diff
@@ -0,0 +1,8 @@
+//= require ./publisher/mention_view
+
+app.views.CommentMention = app.views.PublisherMention.extend({
+  initialize: function(opts) {
+    opts.url = Routes.mentionablePost(opts.postId);
+    app.views.PublisherMention.prototype.initialize.call(this, opts);
```
svbergerem: Please check with a simple jasmine test that `app.views.PublisherMention.prototype.initialize` is called with the correct url.

app/assets/javascripts/app/views/comment_stream_view.js
```diff
@@ -27,6 +27,7 @@ app.views.CommentStream = app.views.Base.extend({
     this.model.comments.each(this.appendComment, this);
     this.commentBox = this.$(".comment_box");
     this.commentSubmitButton = this.$("input[name='commit']");
+    new app.views.CommentMention({el: this.$el, postId: this.model.get("id")});
```
svbergerem: Please add a jasmine test checking that `app.views.CommentMention` is initialized here.

app/assets/stylesheets/mentions.scss
```diff
@@ -65,3 +65,21 @@
...
 }
+
+.typeahead-mention-box-wrap .twitter-typeahead {
+  left: -1px;
+  width: calc(100% + 2px);
+
+  .tt-menu {
+    // Override inline rule of Typeahead
+    // If this is not overridden (`position: absolute` by default in Typeahead) then
+    // the box is cut when opened because of  the `overflow: hidden` from parent classes of comment form styles. By
```
svbergerem: There are two whitespaces between “of” and “the”. Please remove one of them.

svbergerem (changes requested):

cmrd-senya:
Update. Please tell me if I can squash.

svbergerem:
> Please tell me if I can squash.

Yes, the code is looking good. 👍

svbergerem (approved):

svbergerem:
Thank you! ❤️

Issue #7392, opened 2017-03-25
Mention syntax shouldn't be automatically replaced inside `` blocks


cmrd-senya:
As I made a comment:
<img width="910" alt="2017-03-25 15-48-27" src="https://cloud.githubusercontent.com/assets/10187586/24322778/5d473676-1173-11e7-8ca3-79068b5cbaf7.png">

As it is displayed on `develop` version of diaspora:
<img width="915" alt="2017-03-25 15-48-38" src="https://cloud.githubusercontent.com/assets/10187586/24322779/5d8335cc-1173-11e7-9d56-67b93102296e.png">

svbergerem:
Fixing #7276 would also fix this issue.

SuperTux88:
I don't know if it's worth fixing the replacement in that case (because as @svbergerem mentioned, it will be fixed automatically when we finish #7276).

But the problem is maybe bigger (haven't tested that): The fact that the mention was replaced maybe shows that it would also be detected as mention and send a notification for it. Because we do the replacement and the detection with a simple regex, without knowing if it's inside a code-block. So if we fix the detection, we can also fix it for the replacement.

SuperTux88:
@cmrd-senya: I just fixed what happened for your comment in #7394: Your pod runs a really old version of the develop branch, so it doesn't change the syntax before sending it (so your pod still have the original version that you wrote), but other pods with a newer develop version backported the syntax after receiving it. With that it should be consistent over all pods, in the format the sender-pod sent it.

Flaburgan:
#7394 is merged, I deployed d-fr and the bug is fixed, I close this.

svbergerem:
No, this hasn't been fixed. #7394 just made the syntax consistent across pods but it is still replaced on the current develop branch.

![mention_code_publisher](https://cloud.githubusercontent.com/assets/3798614/24355525/8aaef2a0-12f6-11e7-8aa9-98edfdf23c27.png)
![mention_code_preview](https://cloud.githubusercontent.com/assets/3798614/24355524/8aad1a66-12f6-11e7-839a-9e8633b4297a.png)
![mention_code_post](https://cloud.githubusercontent.com/assets/3798614/24355523/8a8e6850-12f6-11e7-820c-8797036f9542.png)

SuperTux88:
Fixed by #7788

PR #7394, opened 2017-03-25
Only backport mention syntax when the author is local

SuperTux88:
We don't need to change new to old syntax when we receive a post from a
newer pod, since we can handle the new syntax. This is only needed when
sending it to older pods.

related to #7392


svbergerem (approved):

svbergerem:
Thank you! :)

PR #7395, opened 2017-03-25
Disable rack session cookie for sidekiq

SuperTux88:
Since 4.2.3 sidekiq allows to disable its own session cookie for the case where the app provides a session already.

Always when I opened `/sidekiq` after long time (with an old `rack.session`-cookie) I was logged out from diaspora. With this setting sidekiq doesn't use this cookie anymore, so lets hope that fixes it :)


denschub:
Merged as 35f4d61d112dde19d1ffee0aa48f89bbb0a1e516, danke. 🍪

PR #7397, opened 2017-03-27
[WIP] Include the @ in the mention link

Flaburgan:
It feels more natural to me.

This is bugged at the moment, I have to get why it doesn't work :P

### DIFF
--- a/spec/lib/diaspora/mentionable_spec.rb
+++ b/spec/lib/diaspora/mentionable_spec.rb
@@ -45,7 +45,8 @@ STR
 
         [people, names].transpose.each do |person, name|
           link = person_link(person, class: "mention hovercardable", display_name: name)
-          expect(fmt_msg).to include "@#{link}"
+          expect(link).to include ">@"
+          expect(fmt_msg).to include link
         end
       end
 
@@ -54,7 +55,8 @@ STR
 
         people.each do |person|
           link = person_link(person, class: "mention hovercardable", display_name: person.name)
-          expect(fmt_msg).to include "@#{link}"
+          expect(link).to include ">@"
+          expect(fmt_msg).to include link
         end
       end
 
@@ -182,7 +184,7 @@ STR
         user_a.aspects.where(name: "generic").first.contacts.map(&:person_id)
       )
 
-      expect(txt).to include("@[user C](#{local_or_remote_person_path(user_c.person)}")
+      expect(txt).to include("[user C](#{local_or_remote_person_path(user_c.person)}")
       expect(txt).not_to include("href")
       expect(txt).not_to include(mention)
     end
@@ -204,7 +206,7 @@ STR
       mention = "@{non_existing_user@example.org}"
       txt = Diaspora::Mentionable.filter_people("mentioning #{mention}", [])
 
-      expect(txt).to eq "mentioning @non_existing_user@example.org"
+      expect(txt).to eq "mentioning non_existing_user@example.org"
     end
   end
 
--- a/app/helpers/people_helper.rb
+++ b/app/helpers/people_helper.rb
@@ -29,7 +29,7 @@ module PeopleHelper
     opts[:class] << " hovercardable" if defined?(user_signed_in?) && user_signed_in? && current_user.person != person
     remote_or_hovercard_link = Rails.application.routes.url_helpers.person_path(person).html_safe
     "<a data-hovercard='#{remote_or_hovercard_link}' href='#{remote_or_hovercard_link}' class='#{opts[:class]}'>"\
-      "#{html_escape_once(opts[:display_name] || person.name)}</a>"\
+      "@#{html_escape_once(opts[:display_name] || person.name)}</a>"\
       .html_safe
   end
 
--- a/lib/diaspora/mentionable.rb
+++ b/lib/diaspora/mentionable.rb
@@ -32,7 +32,7 @@ module Diaspora::Mentionable
       name, diaspora_id = mention_attrs(match_str)
       person = people.find {|p| p.diaspora_handle == diaspora_id }
 
-      "@#{ERB::Util.h(MentionsInternal.mention_link(person, name, diaspora_id, opts))}"
+      ERB::Util.h(MentionsInternal.mention_link(person, name, diaspora_id, opts))
     }
   end
 
@@ -64,7 +64,7 @@ module Diaspora::Mentionable
       if person && allowed_people.include?(person.id)
         match_str
       else
-        "@#{MentionsInternal.profile_link(person, name, diaspora_id)}"
+        MentionsInternal.profile_link(person, name, diaspora_id)
       end
     }
   end
@@ -101,10 +101,10 @@ module Diaspora::Mentionable
     # @param [String] display name
     # @param [Hash] formatting options
     def self.mention_link(person, display_name, diaspora_id, opts)
-      return display_name || diaspora_id unless person.present?
+      return "@#{display_name || diaspora_id}" unless person.present?
 
       if opts[:plain_text]
-        display_name || person.name
+        "@#{display_name || person.name}"
       else
         person_link(person, class: PERSON_HREF_CLASS, display_name: display_name)
       end


spec/lib/diaspora/mentionable_spec.rb
```diff
@@ -182,7 +184,7 @@
         user_a.aspects.where(name: "generic").first.contacts.map(&:person_id)
       )
 
-      expect(txt).to include("@[user C](#{local_or_remote_person_path(user_c.person)}")
+      expect(txt).to include("[user C](#{local_or_remote_person_path(user_c.person)}")
```
Flaburgan: The advantage with that is we visually have a difference between the users who have been mentioned (so notified) and the ones you have a link to but can't access the post.

SuperTux88:
Still not sure about that. After a while I get used to the @ outside the link now (first it looked "different" than before, but now I'm pretty happy with it), and I still think that [@user@pod.tld](https://pod.tld/u/user) looks weird.

Flaburgan:
Well that's why I waited a bit since the @ has been added, I wanted to see if I get used to it but I still feel I prefer this way. But it's not a strong opinion, I can live with the current situation ;)

goobertron:
I still don't like the preceding @ at all. I think it looks ungainly and unnecessary. Sorry about that.

SuperTux88:
But I think we need more discussion then, or at least more opinions. I can also live with the @ inside the link ;) But while it looks OK for [@Fla]() (and other simple single-word names), we have many cases, where the @ inside the link looks weird for me:
* when there is only the diaspora-ID we have 2 @ inside the link: [@user@pod.tld]()
* @svbergerem found [this profile](https://sechat.org/u/dzx) yesterday: [@&#φφφqwertyuiop[]\789zxcvbnm,./ /λολ000asdfghjkl'456helloworld]()
* and I'm also not sure about realnames, because the @ is then attached to the first name: [@Benjamin Neff]()

cmrd-senya:
I feel okay with the current implementation (@ is not a part of link).

Actually, if it is possible to make it adjustable? So that each user sets the rendering of mentions to what he/she prefers?

SuperTux88:
Making that configurable isn't easy (and not worth it in my eyes), because the @ is added in may different places. So it would make the code much more complex.

Flaburgan:
Should we continue to discuss it here, or move to discourse, or can we take a decision? I'm ready to implement what we decide but I need a spec.

Flaburgan:
Update, can we make a decision? Mastodon is using @user@pod.tld and it looks like users are fine with it.

goobertron:
If the argument is 'If we don't display an @, but I want to have the @ in my comment ...' the question is:

1. 'how many people using Diaspora want to see an @ symbol in front of every link?' So far we have two for, and two against.
2. If the problem lies with federation to Friendica, and it's creating a problem on that network rather than this one, perhaps they want to solve it for their UI? It shouldn't be difficult for them to suppress a double @ before mentions.
3. Also, is this a real problem? How many diaspora users are manually adding @ symbols before mentions, and how many times therefore is the @@ displayed in Friendica?

> To quickly differentiate someone from a normal link? 

> The other advantage is, that you can see which link is a mention (and has a hovercard)

You can do this at the moment, as mentions are a different shade of blue, and are bold rather than the normal weight of links; and if you mouse over them, the hovercard appears. And yes, you can 'spoof' the look of a mention by making a link bold, but it'll be just as easy to spoof a mention by placing an @-symbol at the front of a link.

And what is the answer to the cases you raised above:

> - when there is only the diaspora-ID we have 2 @ inside the link: @user@pod.tld
> - @svbergerem found this profile yesterday: @&#φφφqwertyuiop[]\789zxcvbnm,./ /λολ000asdfghjkl'456helloworld
> - and I'm also not sure about realnames, because the @ is then attached to the first name: @Benjamin Neff

This seems to be forcing something on everyone, which at least some users don't want, and creating some problem cases (SuperTux's list above) because Fla likes it and possibly to solve an issue with Friendica that might not be a real problem and, if it is, is a problem that is probably best solved by Friendica if they feel they need to solve it.

Flaburgan:
I honestly don't have a strong opinion on this. It just became the standard in many places, but I will not block if we have a consensus about removing it ;)

denschub:
Okay, I've spent ~some~ a lot of time thinking about this.

Initially, I felt a lot like @Flaburgan, suggesting to link the @ as well (somewhere here on GitHub, there is this discussion linked). However, I quickly realized this is not the way to go forward. Our names, unlike twitter's or GitHub's, can contain special characters. Since we link the real name, not the username, we are a bit like Google+, which, funny enough, also does not link the +.

I played around with linking the @ and whenever there are spaces in the link, it looks weird. Really weird. In addition to my feeling, there are a lot of good explanations for having the @ not linked, including those at https://github.com/diaspora/diaspora/pull/7397#issuecomment-289438112. Since @SuperTux88, @svbergerem and me agree here, ... let's close this PR.

(sorry @goobertron for somewhat ignoring you here - I've read your point. However, I clearly want to have a visual distinction between a link to someones profile and something that actually does something else. And mentioning someone is definitely a different thing than linking someone.)

I will not spread hate if someone feels strong about this and disagrees with us here, but if that's the case, please open a discussion in Discourse.

goobertron:
I'm very happy to have a more clear differentiation between standard links and mentions, as long as it doesn't involve something ugly like slapping an @ symbol in front of them! If anyone comes up with an elegant solution, I'll be open to supporting it.

Flaburgan:
Like a different color for the link?

goobertron:
We already have a different colour for the link (subtly)*, plus mentions are bold whereas links are not.

*Actually, this might only be the case in the default theme.

SuperTux88:
The different blue was only on the default theme, not on the dark theme and not on the other color themes. It is also removed from the default theme now (#7318), because it was only on some places (I just checked an old pod, and at least on the tag-search and on the SPV it has the same color as a normal link) and it was only visible when you had two different links next to each other. And mentions aren't bold (currently, and old on older pods). So there is no way to differentiate mentions from other links and the @ fixes that problem. There are also already some pods running with the `develop` branch (geraspora for example, and that's one of the biggest pods) and I saw nobody complaining about the @.

cmrd-senya:
Can't we make it a part of user theme?

For example if we have the following HTML for the mention:
```html
<a href="/person/..." class="mention"> <span class="at-sign">@</span>User Name</a>
```

Then we can make it non-clickable with CSS:
```scss
a.mention {
  span.at-sign {
    pointer-events: none;
    cursor: default;
    text-decoration: none;
    text-color: 0;
  }
}
```

So different themes may enable or disable `span.at-sign` style and so we can adjust the desirable representation by changing user themes.

SuperTux88:
No, I don't want to duplicate all themes, because they already take forever to precompile.

cmrd-senya:
We don't have to duplicate them. We can leave only the default themes with `@` non-linked. Podmins can change the themes themselves if they want to. If we change the html/css to what I proposed we can leave it as the only supported thing out-of-the-box, but this way it would be easily adjustable (e.g. @Flaburgan can add a theme on his pod by himself). I think this doesn't cost too much for us (just a little heavier html+css) and at the same time it makes it possible for people to get the experience they prefer.

However I don't care too much about that. I'm okay with how it works now. I'm just trying to propose a better solution to get more people satisfied. If there is strong opposition to my idea, I'm okay with not implementing it.

denschub:
> I think this doesn't cost too much for us (just a little heavier html+css)

This, honestly, sounds like an incredibly large rabbit hole to fall down.

cmrd-senya:
> This, honestly, sounds like an incredibly large rabbit hole to fall down.

I don't get what exactly you mean by this :/

denschub:
I am sorry.

If we start exposing that one change in the user themes, there always will be other stuff that people want to have in user themes as well (let's say, for example, the sidebar position) and we'd eventually end up with a huge number of variations to maintain. So I'd rather not start this trend.

cmrd-senya:
Okay, that's a question of our UI approach. I won't fight for it, but I'm not agaist of some of customizations at least. The discussion of this will get offtopic, so I'm getting off. If anyone feels supportive about my propositions the discussion can be continued on Discourse.

PR #7399, opened 2017-03-27
Remove statistics.json

svbergerem:


Flaburgan:
Podupti.me and the-federation.info both use nodeInfo, so we should be fine there.
In the last 24h I have GET requests on  `/statistics.json` in my log from https://f.left.social/ https://kirgroup.com/ carina.uberspace.de https://hub.libranet.de and several other friendica and hubzilla pods. I don't know why these nodes need to do that, but this PR is dropping this route so it will soon answer 404. Those are up-to-date servers so I'm warning @annando and @zotlab here.

annando:
Friendica is polling for the statistics.json for its server type detection. The development version now uses nodeinfo as well, but release date will be somewhere in summer, I guess.

We have to poll statistics.json in the future as well since we want to detect different versions of the different systems. But this shouldn't create problems, I guess.

SuperTux88:
@annando so this PR would break federation with friendica? Then I think we should wait.

SuperTux88 (approved):
Code looks good, but let's wait until friendica doesn't need it anymore.

annando:
It doesn't really break the communication. It's just a small inconvenience. As long as there is a ````X-Diaspora-Version:```` header, the detection will still work.

SuperTux88:
Yes, the `X-Diaspora-Version` header is still there. So nothing breaks? It's just an additional request that result in a 404 and then fallback to another request reading `X-Diaspora-Version`?

annando:
The header probing is done all the time since it is needed for other network systems as well.

SuperTux88:
Ok, then I think we can safely merge it?

annando:
Feel free.

SuperTux88:
Merged, thanks @svbergerem for the work and @annando for the feedback. 👍

PR #7408, opened 2017-03-31
Fix submitting publisher in test by waiting for the closed class

Flaburgan:
New try, let's see if the wait works this way.


SuperTux88:
What should we do with this? This still failed at least once on travis. Do you want to use this PR to test other solutions? Or should we close it? Or merge it? (It's basically the same as before, so I don't know if it's worth merging it, if it doesn't fix it?)

Flaburgan:
That's apparently not a real fix.

PR #7414, opened 2017-04-06
Add avatar fallback for typeahead and conversations

svbergerem:


app/assets/javascripts/app/views.js
```diff
@@ -55,8 +55,7 @@ app.views.Base = Backbone.View.extend({
       .html(this.template(presenter))
       .attr("data-template", _.last(this.templateName.split("/")));
 
-    // add avatar fallback if it can't be loaded
-    this.$el.find(this.avatars.selector).on("error", this.avatars.fallback);
+    this.setupAvatarFallback();
```
SuperTux88: Can we also use this here?
* [app/assets/javascripts/app/views/tags_view.js](https://github.com/diaspora/diaspora/blob/1b9ab16039160aefe16dc255487fb739b97f0a55/app/assets/javascripts/app/views/tags_view.js#L9)
* [app/assets/javascripts/app/views/notification_dropdown_view.js](https://github.com/diaspora/diaspora/blob/1b9ab16039160aefe16dc255487fb739b97f0a55/app/assets/javascripts/app/views/notification_dropdown_view.js#L99)

svbergerem: It was possible with some small refactoring which even improved the code. The conversations view now adds the fallback to the added recipient instead of adding it every time to all recipients.

SuperTux88 (approved):

SuperTux88:
Merged as 3971a24d682bad183eccba98ec289712280ab59d, bf105db2972f5c59fc6a626cd5219a1cde41bf22, bcb75e308218c1c452d2b0bea7957d3cc6d4aab8 and 1c1af74e9f3cfc5f54ba07ecec78ccc1a6379308

Thank you 👍

PR #7419, opened 2017-04-10
Bump timeago to fix mobile timestamps regression

### LINKED ISSUE #7396: Relative time on mobile broken
Relative timestamps in the mobile version are broken and only show a date:

screen shot 2017-03-27 at 02 53 56

This is on the latest develop, no idea when that regression popped up.

cmrd-senya:
fixes #7396

### DIFF
--- a/app/assets/javascripts/widgets/timeago.js
+++ b/app/assets/javascripts/widgets/timeago.js
@@ -28,6 +28,10 @@
       }
 
       $.timeago.settings.autoDispose = false;
+
+      $(function() {
+        $("time[data-time-ago]").timeago();
+      });
     });
   };
 })();
--- a/app/assets/javascripts/widgets/timeago.js
+++ b/app/assets/javascripts/widgets/timeago.js
@@ -26,6 +26,8 @@
           }
         });
       }
+
+      $.timeago.settings.autoDispose = false;
     });
   };
 })();
--- a/Gemfile
+++ b/Gemfile
@@ -183,8 +183,8 @@ gem "typhoeus",           "1.1.2"
 gem "gon",                     "6.1.0"
 gem "hamlit",                  "2.8.0"
 gem "mobile-fu",               "1.3.1"
+gem "rails-timeago",           "2.16.0"
 gem "will_paginate",           "3.1.5"
-gem "rails-timeago",           "2.11.0"
 
 # Logging
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -568,7 +568,7 @@ GEM
     rails-i18n (4.0.8)
       i18n (~> 0.7)
       railties (~> 4.0)
-    rails-timeago (2.11.0)
+    rails-timeago (2.16.0)
       actionpack (>= 3.1)
       activesupport (>= 3.1)
     rails_admin (0.8.1)
@@ -734,7 +734,7 @@ GEM
       unf (~> 0.1.0)
     typhoeus (1.1.2)
       ethon (>= 0.9.0)
-    tzinfo (1.2.2)
+    tzinfo (1.2.3)
       thread_safe (~> 0.1)
     uglifier (3.1.2)
       execjs (>= 0.3.0, < 3)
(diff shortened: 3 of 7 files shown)


SuperTux88:
The code looks good and it fixes the issue. But after playing with it a bit I found another "problem": After page load it always displays the timeago in englisch and after waiting 1 minute for the first refresh it switches to the correct language (german in my case). This also only happens on mobile and didn't happen with the old version of timeago. So maybe you can also have a look at this?

SuperTux88:
I had a closer look at this problem and found, that the reason for the wrong language was caused by [the change for the fix for jquery 3](https://github.com/jgraichen/rails-timeago/commit/7e3e74f02f7632c1c19c018d3ffad4a1d7515035) in the new version ;) With the new version, `.timeago()` was called before the language was set. To fix that, I copied the code from `rails-timeago.js` to `widgets/timeago.js`, so that it is called after the language is set.

cmrd-senya:
Thanks for having a look and fixing this. I was terribly busy last week, so I unfortunately didn't have time for this.

cmrd-senya:
Don't we need a test (a cuke perhaps) for the locale bug you've found?

SuperTux88:
There are currently no tests for the timeago widget at all, and I have no idea how to write any (to test if the init function is called after the language is set). And I don't know if we should add cukes based on translations. Since your cuke also tests if the init function is called (but not the order), I think that's OK for now. I could copy your cuke and change the language in the settings and then test for the language, but I don't know if that's worth it.

app/assets/javascripts/widgets/timeago.js
```diff
@@ -26,6 +26,14 @@
...
         });
       }
+
+      $.timeago.settings.autoDispose = false;
+
+      var initTimeAgo = function() {
+        $("time[data-time-ago]").timeago();
+      };
+      $(initTimeAgo);
+      $(document).on("turbolinks:load page:load ajax:success", initTimeAgo);
```
denschub: This is racy. Basically, the handler for the ajax success here races with [the one after loading the comments](https://github.com/diaspora/diaspora/blob/caaea50cf61d1c54e876e16f4464af087dbf1750/app/assets/javascripts/mobile/mobile_comments.js#L106), basically assuming that the latter runs first so that the `initTimeAgo()` call here is able to find the new html.

One possible workaround would be calling `$("time[data-time-ago]").timeago();` after the `insertAfter` in the comment ajax success handler to make sure it gets called *after* the html was added or to add and dispatch an arbitrary event (`diaspora:pagecontentschanged` for example) instead, making sure we dispatch that every time when we are done adding stuff to the document.

skyschub (changes requested):

app/assets/javascripts/widgets/timeago.js
```diff
@@ -26,6 +26,14 @@
           }
         });
       }
+
+      $.timeago.settings.autoDispose = false;
+
+      if (typeof(Diaspora.Mobile) !== "undefined") {
```
diaspora-code-review: Unary word operator 'typeof' must be followed by whitespace.

SuperTux88:
I removed the handles like @denschub suggested in https://github.com/diaspora/diaspora/pull/7419#discussion_r114439855 and I also added an check to only execute it on mobile, because on desktop timeago is called from backbone.js, so we don't need to call it twice.

I also updated to the now latest version of rails-timeago :)

SuperTux88:
OK, I removed the check for mobile again, because otherwise the desktop interface is broken (wrong language) on some pages (/notifications for example), because backbone.js calls `timeago()` before the `widgets/timeago.js` sets the language. With calling `timeago()` again after that, it corrects the language. This isn't a clean solution, because we scan the DOM again on every page (which theoretically isn't needed) ... but it's not worse than the current situation, because we do this currently too. :(

So we probably need some bigger refactorings here, but that's out of scope here. This PR fixes #7396 without doing anything worse than before.

svbergerem (approved):

svbergerem:
Merged as 9e36f0201ce8de0d7cd6f83ffe2a224c6dfd6bd3, 87ed561d451c853ff7705f2a7f90df53c8fe30e5, a880de28a8bbf3fd08e043d58d3d2dc8ef58e134, and 4621c54421444394bd755ae7ff6e77c9fed4d4c3. Thank you! :)

PR #7420, opened 2017-04-11
Use post page title in atom feed

### LINKED ISSUE #7249: 500 in a user atom feed
```
[2016-12-13T10:17:05] INFO  PID-11028 TID-8215938281780 Rails: Started GET "/public/fla.atom" for x.x.x.x at 2016-12-13 10:17:05 +0100
[2016-12-13T10:17:05] INFO  PID-11028 TID-8215938281780 ActionController::Base: Processing by UsersController#public as ATOM
[2016-12-13T10:17:05] INFO  PID-11028 TID-8215938281780 ActionController::Base:   Parameters: {"username"=>"fla"}
  Rendered users/public.atom.builder (16.0ms)
[2016-12-13T10:17:05] INFO  PID-11028 TID-8215938281780 ActionController::Base: Completed 500 Internal Server Error in 61ms (ActiveRecord: 22.8ms)
[2016-12-13T10:17:05] FATAL PID-11028 TID-8215938281780 Rails: 
ActionView::Template::Error (undefined method `lstrip' for nil:NilClass):
    30:     feed.entry post, :url => "#{@user.url}p/#{post.id}",
    31:       :id => "#{@user.url}p/#{post.id}" do |entry|
    32: 
    33:       entry.title post.message.title
    34:       entry.content post.message.markdownified(disable_hovercards: true), :type => 'html'
    35:       entry.tag! 'activity:verb', 'http://activitystrea.ms/schema/1.0/post'
    36:       entry.tag! 'activity:object-type', 'http://activitystrea.ms/schema/1.0/note'
  lib/diaspora/message_renderer.rb:214:in `title'
  app/views/users/public.atom.builder:33:in `block (3 levels) in _app_views_users_public_atom_builder__3305162908281538328_8216011566200'
  app/views/users/public.atom.builder:30:in `block (2 levels) in _app_views_users_public_atom_builder__3305162908281538328_8216011566200'
  app/views/users/public.atom.builder:29:in `each'
  app/views/users/public.atom.builder:29:in `block in _app_views_users_public_atom_builder__3305162908281538328_8216011566200'
  app/views/users/public.atom.builder:1:in `_app_views_users_public_atom_builder__3305162908281538328_8216011566200'
```

d-fr is running last develop. I tried an other user on d-fr and it worked, so I guess a specific post is faulty here.

svbergerem:
Fixes #7249 

I also wasn't able to reproduce the issue but now it uses the same code as the posts presenter.

### DIFF
--- a/spec/helpers/posts_helper_spec.rb
+++ b/spec/helpers/posts_helper_spec.rb
@@ -17,6 +17,13 @@ describe PostsHelper, :type => :helper do
         post_page_title(post)
       end
     end
+
+    context "with a reshare" do
+      it "returns 'Reshare by...'" do
+        reshare = FactoryGirl.create(:reshare, author: alice.person)
+        expect(post_page_title(reshare)).to eq I18n.t("posts.show.reshare_by", author: reshare.author_name)
+      end
+    end
   end
 
 
--- a/app/helpers/activity_streams_helper.rb
+++ b/app/helpers/activity_streams_helper.rb
@@ -1,0 +1,12 @@
+module ActivityStreamsHelper
+  def add_activitystreams_author(target, person)
+    target.author do |author|
+      author.name person.name
+      author.uri local_or_remote_person_path(person, absolute: true)
+
+      author.tag! "activity:object-type", "http://activitystrea.ms/schema/1.0/person"
+      author.tag! "poco:preferredUsername", person.username
+      author.tag! "poco:displayName", person.name
+    end
+  end
+end
--- a/app/views/users/public.atom.builder
+++ b/app/views/users/public.atom.builder
@@ -17,21 +17,16 @@ atom_feed("xmlns:thr"       => "http://purl.org/syndication/thread/1.0",
 	    'media:height' => '100', :href => "#{@user.image_url}"
   feed.tag! :link, :href => "#{AppConfig.environment.pubsub_server}", :rel => 'hub'
 
-  feed.author do |author|
-    author.name @user.name
-    author.uri local_or_remote_person_path(@user.person, :absolute => true)
-
-    author.tag! 'activity:object-type', 'http://activitystrea.ms/schema/1.0/person'
-    author.tag! 'poco:preferredUsername', @user.username
-    author.tag! 'poco:displayName', @user.name
-  end
+  add_activitystreams_author(feed, @user.person)
 
   @posts.each do |post|
     feed.entry post, :url => "#{@user.url}p/#{post.id}",
       :id => "#{@user.url}p/#{post.id}" do |entry|
 
-      entry.title post.message.title
+      entry.title post_page_title(post)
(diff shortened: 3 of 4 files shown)


Flaburgan:
@svbergerem I was also wondering about how reshares are handled. I thought about using https://github.com/diaspora/diaspora/blob/develop/app/helpers/posts_helper.rb#L6 but it makes sense in the atom feed to keep the original title as the title. And maybe to precise it's a reshare? There is no difference between a reshare and a post actually as far as I see.

svbergerem:
@Flaburgan We could use the original title if it's present and use the `post_page_title` otherwise. What do you think?

SuperTux88 (changes requested):
Can you please add a test for that in `users_controller_spec.rb`?

SuperTux88:
We can set an `author` if it's a reshare and/or prefix it with "Reshare:"?

svbergerem:
This PR now uses the `post_page_title` and sets the author for all entries.

Flaburgan:
I didn't test it at the moment but the code looks good to me.

SuperTux88 (approved):
The code looks OK and I tested it with my feedreader and it displays the correct author.

spec/helpers/posts_helper_spec.rb
```diff
@@ -17,6 +17,13 @@
         post_page_title(post)
       end
     end
+
+    context "with a reshare" do
+      it "returns 'Reshare by...'" do
```
SuperTux88: A side note (off topic here, because it has nothing to do with the feed): I don't know why you added this test here, because it is never used in the feed, the feed contains always the original root post with the original title. But it tests the current behavior of `post_page_title` and more tests are never bad, so it's OK ;)

I don't know if we should change the `post_page_title` for reshares, because for notifications  it's sometimes suboptimal. 
* Likes for example look like this:
  "[[person x]]() has liked your post [Reshare by [myname]]()."
  (I know that it's my reshare, but would be useful to know which one)
* For comments on reshares of another person it looks like this:
  "[[person x]]() also commented on [person y]’s post [Reshare by [person y]]()."
  (It's a bit redundant, "... on [person y]’s reshare of [person y]'s post" would be better.)

And for the single post view of a reshare the "Reshare by ..." is only displayed shortly, but some javascript replaces it immediately with the title of the original title of the post.

Should we create a refactoring issue for `post_page_title`?

SuperTux88:
Merged as 23d3d9dc3f4aaa1e8838cbd0f81bbb1384f35188 and 6dc4493685ee3058f4b4d9bfe35bd1c04a343e92

Thanks 👍

svbergerem:
> I don't know why you added this test here, because it is never used in the feed, the feed contains always the original root post with the original title.

While playing with `post_page_title` I added the test and kept it for the PR since it improves the coverage.

PR #7423, opened 2017-04-13
Fix lstrip used on nil variable

ghost:
the @text var is not defined with this profile
joindiaspora.com/public/kostasvoulazeris.atom

### DIFF
--- a/lib/diaspora/message_renderer.rb
+++ b/lib/diaspora/message_renderer.rb
@@ -210,6 +210,8 @@ module Diaspora
     # @option opts [Integer] :length (70) Truncate the title to
     #   this length. If not given defaults to 70.
     def title opts={}
+      @text = "" if @text.nil?
+
       # Setext-style header
       heading = if /\A(?<setext_content>.{1,200})\n(?:={1,200}|-{1,200})(?:\r?\n|$)/ =~ @text.lstrip
                   setext_content


Flaburgan:
There is also https://github.com/diaspora/diaspora/pull/7420 to solve https://github.com/diaspora/diaspora/issues/7249 I don't know what's the best approach.

lib/diaspora/message_renderer.rb
```diff
@@ -210,6 +210,8 @@ def markdownified opts={}
     # @option opts [Integer] :length (70) Truncate the title to
     #   this length. If not given defaults to 70.
     def title opts={}
+      @text = "" if @text.nil?
+
       # Setext-style header
       heading = if /\A(?<setext_content>.{1,200})\n(?:={1,200}|-{1,200})(?:\r?\n|$)/ =~ @text.lstrip
```
cmrd-senya: I would write
```ruby
@text.try(:lstrip)
```
instead but that's just a question of style

jhass:
I think I quite prefer Steffen's approach :) The message render shouldn't handle nil itself but fail hard on it IMO, since swallowing it at this point makes finding the origin only harder.

cmrd-senya:
> I think I quite prefer Steffen's approach :) The message render shouldn't handle nil itself but fail hard on it IMO, since swallowing it at this point makes finding the origin only harder.

Then it's probably better to have a check there and raise an exception, so that it is explicit that the `title` call was made improperly with `@text == nil`.

Zauberstuhl:
ups haven't searched for it. just hot-fixed it on JD and wanted to show you guys that this could be a problem. thanks anyway

PR #7428, opened 2017-04-22
Add translation for zh-CN

qinwenshi:
Add translation or missing keys with original EN text. This will stop the failure when which default locale as zh-CN.


SuperTux88:
Thanks for your contribution, but we use [WebTranslateIt](https://webtranslateit.com/en/projects/3020-Diaspora) for the translations, see: https://wiki.diasporafoundation.org/Contribute_translations

PR #7435, opened 2017-05-01
Bump diaspora_federation

SuperTux88:
The previous version logged the encrypted private messages of the new protocol, which only makes the logfiles bigger without any benefit, because it's encrypted anyway ...

This is the only change: [8859c96](https://github.com/diaspora/diaspora_federation/commit/8859c960ac2b771399ad42ccf795043aea4ec9a5)


cmrd-senya (approved):
LGTM

skyschub (approved):

SuperTux88:
Merged as e7cfaa7e42acd568017ab0c0a3ca25031c597d77

Thanks for the reviews @cmrd-senya and @denschub ❤️

PR #7436, opened 2017-05-02
Federation gem update

SuperTux88:
This contains all changes needed for the upcoming 0.2.0 release of the federation gem. I marked it as [WIP], because the gem is not released yet (and I maybe make more changes to the gem, so there are more changes needed here). But the commits already in this PR are pretty finished, so they are ready for review.

* diaspora/diaspora_federation#27: Send only `Retraction` instead of `SignedRetraction` and `RelayableRetraction`
* diaspora/diaspora_federation#30: Send the raw magic envelope and new encrypted magic envelope with crypt-json-wrapper
* diaspora/diaspora_federation#31: Remove old fetch route
* diaspora/diaspora_federation#32: Send `Contact` entity for start/stop sharing
* diaspora/diaspora_federation#35: Make `Participation` entity non-relayable
* diaspora/diaspora_federation#36: Make `Message` entity non-relayable
* diaspora/diaspora_federation#39: Remove `guid` and `public_key` from `WebFinger`
* diaspora/diaspora_federation#52: Different refactorings needed
* diaspora/diaspora_federation#58: Switch to Fabricate for federation factories

The changes for the federation gem for the switch to the new protocol are here: diaspora/diaspora_federation#62

I tested these changes between different versions (0.6.0.0, latest-develop and this branch), and everything seams to work 🎉. This also contains the fix for #4269, so we break with pods older than 0.6.3.0.


SuperTux88:
@annando found some problems with friendica while testing against this PR, but he fixed them already (https://github.com/friendica/friendica/pull/3443). There should be a release [soon](https://github.com/friendica/friendica/milestone/10) including this fixes and he agreed that it should be OK to merge this PR (which will bring the new protocol to pods running `develop`) a few weeks after that friendica release. That would also be after the 0.6.6.0 release of diaspora (which also includes [a small improvement for federation logging](https://github.com/diaspora/diaspora/pull/7435)). Since this PR isn't ready yet anyway it would only mean a small extra delay and I think that's a good solution.

Gemfile
```diff
@@ -13,7 +13,7 @@ gem "unicorn-worker-killer", "0.4.4"
 
 # Federation
 
-gem "diaspora_federation-rails", "0.1.8"
+gem "diaspora_federation-rails", github: "SuperTux88/diaspora_federation", branch: "protocol-next-step"
```
denschub: Note: Replace this with a replaced version.

(I know that's your plan, I just want to have a note below this line so it's marked as addressed when it is.)

skyschub (commented):
Adding a tracking note for the version so I get notified when you update it.

cmrd-senya:
Will the final diaspora_federation version in this PR include changes from https://github.com/diaspora/diaspora_federation/pull/65 and https://github.com/diaspora/diaspora_federation/pull/63?

SuperTux88:
Yes, I want to include these 2 PRs too in the 0.2.0 release of the gem. That's why this PR is still "WIP" (New commits will be added, but the commits already there are ready for review).

lib/diaspora/federation/entities.rb
```diff
@@ -200,37 +177,29 @@ def self.reshare(reshare)
         )
       end
 
-      def self.retraction(target)
+      def self.retraction(retraction)
+        case retraction.data[:target_type]
+        when "Contact"
```
cmrd-senya: Do we really need to use the `Retraction` class as a mediator between `Contact` model and `Contact` federation entity? Can't we map `Contact` model directly to `Contact` entity and pass the `Contact` model instance to the dispatcher instead of using `Retraction` for that?

In the current implemenation you have to keep "Contact" type as a special case everywhere you deal with the retraction. And I think the original point was that we don't use the `Retraction` entity for the stop sharing event, so I don't think `Contact` entity dispatching has anything to do with retractions anymore.

cmrd-senya: Also, if you're going to change anything in the retraction implemenation, perhaps the retraction/contact related changes could be extracted to a different pull request. If I'm not mistaken, these changes don't depend of the federation gem update and could work both with the 0.1 and 0.2 versions of diaspora_federation. So we can process these changes out of the federation gem update flow. (correct me if I'm wrong)

SuperTux88: Yes, I tried that, but the problem is that all other entities (not `Retraction`) need the object still in the database for `Workers::DeferredDispatch` while Retractions use `Workers::DeferredRetraction` and have all date needed in the job parameters. So we need a special case for contact retractions somewhere, and I think that the current solution was the one with the least specials (compared to create it's own special sidekiq worker), since we can still use `Workers::DeferredRetraction` and also I think it's OK since it is some kind of "retraction".

One other alternative I had was, that we could use `Diaspora::Federation::Dispatcher.build(...).dispatch` instead of `Diaspora::Federation::Dispatcher.defer_dispatch(...)` [here](https://github.com/SuperTux88/diaspora/blob/e994c31590aef3b4afadbec6aa51f7e68440c8cb/app/models/user/connecting.rb#L38) and build the federation-salmon in the request context before deleting the `Contact` in the database, so we don't need the `Contact` in the database afterwards.

What solution do you prefer? Or do you have a third cleaner solution that I didn't think of?

SuperTux88: And yes, this specific change (use `Contact` for stop sharing) doesn't depend on the gem, but it depends on other changes in this PR and other changes here depend on this change. So I'd like to keep it here if possible, because I think there would be many conflicts otherwise. Also I mixed it with the "use `Contact` for start sharing" change (which is the same issue in the federation repo, so I didn't split that to different commits), and the start sharing does depend on the gem, because I blocked the old `Request` entity in the gem.

cmrd-senya: Okay, so we need an object to hold the data before the worker constructs the entity. Then we could make a specific `ContactRetraction` object that encapsulates all of the specifics of contact retraction and avoid branching in entity dispatching process. It's pretty much the same thing as what you do, but it uses inheritance instead of branching which is a bit cleaner IMO.

Example: https://gist.github.com/cmrd-senya/90a4bc2aaf7370b24caa91e5487fa763

What do you think?

cmrd-senya: I also thought that the `Retraction` class actually has two different responsibilities: first is submitting a job to sidekiq and second is passing data from sidekiq worker to entity constructor. Actually this responsibilities aren't too bound IMO, so I think they could actually be separated to different objects. Maybe you don't want to make that big refactorings, but if you want, I can try to make some code changes to show my idea.

SuperTux88: > It's pretty much the same thing as what you do, but it uses inheritance instead of branching which is a bit cleaner IMO.

Yes, looks good to me, I took your diff and added it as it's own commit (I wasn't able to add it as fixup for any commit, because all other changes around `Retraction`s are in different commits, so it would have had to many conflicts) and used you as author (since you did the work for that, I hope that's OK for you).

I changed only one little thing: I moved the `retraction_class` as second parameter, so `opts` is the last parameter again. Please tell me what you think about it. (We could let it as the last parameter and set a default value to stay backward compatible with jobs still in the queue, but since this is a major release we can do the "run sidekiq for extra 5 minutes to process queued jobs"-migration again, so I don't think we need to stay compatible with old jobs.)

I also added some tests for it.

> Actually this responsibilities aren't too bound IMO ...

Yes, they aren't directly bound, but I don't think it's a problem to have both in the same class. To pass the data to the sidekiq worker we basically only need the `data` and `subscribers` parameters.

And with both things in the same class we can now do something like this:

```ruby
Diaspora::Federation::Dispatcher.build(sender, Retraction.for(target)).dispatch
```

(Even when we don't really need this, but I think that's still nice)

> I can try to make some code changes to show my idea.

If you think you can do something better, go ahead and show me your idea 👍 I don't see how it could help to separate the `data` and `subscriber` yet, I think it gets more complicated. But if you already have an idea, I'm interested.

SuperTux88:
I also rebased the PR on the latest `develop` and use now the latest `develop` of `diaspora_federation` and added a commit to use the RFC 7033 WebFinger from the gem with OpenID.

lib/diaspora/federation/entities.rb
cmrd-senya: Well, it's something like this: https://gist.github.com/cmrd-senya/a8263ec0394f7b24ce3b989ef296b274

It can look more complicated since it now contains more classes, but I think that it's actually better since different behaviors are split into different classes now. This way it is explicit which methods are used for submitting a job to sidekiq and which methods are used for entity dispatching. And if we change these classes in future it is less likely to get into hardly maintainable code by mixing different responsibilities. So I think it is better in terms of maintainability.

I'm fine if you decide to reject this code though. I think can live with how we have it now. But I think that this change slightly improves the implementation.

SuperTux88: I thought a bit more about this now after viewing your diff. I think we should keep that how it is for now and here is why:

We use the model as "dispatchable" for all other entities to, so the models contain some federation methods (`subscribers` and `public?`), so the same object is used as model and for dispatching. It's the same for all entities and retraction is no "special case" here.

The only thing special about retractions is, that it's not a database model, it has only a short lifespan and the data is stored in redis instead of postgres/mysql. So the difference is only in `Workers::DeferredRetraction` where the "dispatchable" isn't loaded from the database, but after that everything works the same as for all other entities.

So if we want to split federation logic from the models, we should do that for all entities, and that is in my eyes completely out of scope for this PR. And even then I'm not sure if the overhead of many more classes is worth it or if we should just keep it as it is, since for example `public?` can be used directly from the model now for some entities.

cmrd-senya: That's not about splitting federation logic from models. Following to this logic `Retraction::Dispatchable` is actually something like model. So we kinda always use models for federation entity generation & dispatch. Maybe I just picked wrong naming and `Retraction::Dispatchable` should be called `Retraction` and what is called `Retraction` in my diff should be called `RetractionDispatcher`. For usual models the part which does the job submission is in `Diaspora::Federation::Dispatcher` and if we follow that design for retractions then we should probably make a separate `RetractionDispatcher`. Other models don't do their job submission/dispatching themselves, they use dispatchers for this. It makes more sense this way now, and it means that I can try to make another diff for that :) Maybe you will like it more.

But do you think it worth to continue discussion within this PR? It feels somewhat out-of-scope. On the other hand. We're at it now, and we refactor the retractions code in this PR anyway so maybe it isn't bad time to make some more changes.

SuperTux88: Hmm, so maybe we need a `RetractionDispatcher` or so?

But I still think that's out of scope here, this PR is for the protocol change and 0.2.0 version of the gem. The `ContactRetraction` refactoring was about something changed in this PR, so I added it here. But for the "retraction dispatching" refactoring I'd rather do/discuss it independent of this PR.

cmrd-senya: Okay, so we'll probably get to this later, lets move along :)

cmrd-senya:
947f8066deb50d3b1a86729bea0addcdd50628c3 this commit looks strange, the description doesn't match the change. Aren't there any lost changes?

SuperTux88:
No, that's correct. Since we need this method for diaspora/diaspora_federation#64 in the future, I only removed the `@deprecated` warning.

cmrd-senya:
I think you should also remove `parent` setting for `Participation` here: https://github.com/diaspora/diaspora/blob/4d9cb1f401fe574b4ba13707abbcaf4b2dedbaa1/lib/diaspora/federation/entities.rb#L101-L101

spec/lib/diaspora/federated/contact_retraction_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  describe ".for" do
+    it "creates a retraction for a contact" do
+      expect(ContactRetraction).to receive(:retraction_data_for).with(contact)
+      ContactRetraction.for(contact)
+    end
+
+    it "sets receiving to false" do
+      expect(contact).to receive(:receiving=).ordered.with(false)
```
cmrd-senya: Maybe it's better to check resulting retraction data value instead of checking for the function call?

The important thing here is that we don't create contact data with sharing/following with `true`, it's not that important how exactly we achieve this.

Also our `contact` object here has `sharing` and `receiving` values with `false` which is not a real world case, so this should probably be changed at least for this example or even maybe for the whole `ContactRetraction` test.

cmrd-senya (commented):
So I reviewed the code for this PR. The code looks fine to me exept of the couple of notes I left here and at https://github.com/diaspora/diaspora/pull/7436#issuecomment-309226058.

I think that I also want to conduct some tests with changes from here and https://github.com/diaspora/diaspora_federation/pull/67 and I can approve it afterwards.

SuperTux88:
I updated the PR with the released version of `diaspora_federation` 0.2.0. I also changed the second last commit for https://github.com/diaspora/diaspora/pull/7436#discussion_r122572872 and added the last commit for https://github.com/diaspora/diaspora/pull/7436#issuecomment-309226058

For what scenarios do you want more tests? diaspora/diaspora_federation#67 didn't change anything here, this change is only in the gem.

cmrd-senya (approved):
Okay, I tested some federation features and found no regressions. LGTM.

skyschub (approved):

denschub:
Thanks very much for your reviews, @cmrd-senya! <3

cmrd-senya:
A big step for diaspora project! Thanks @SuperTux88 for your work!

SuperTux88:
And thanks @cmrd-senya and @denschub for the review and testing :)

Issue #6413, opened 2015-09-19
500 error in rails_admin when creating a new user


axilleas:
For starters I'm on 5.3, haven't tested the develop branch.

Visit `admin_panel/user/new`

```
[2015-09-20T00:12:16] INFO  PID-10706 TID-6889100 ActionController::Base: Processing by RailsAdmin::MainController#new as HTML
[2015-09-20T00:12:16] INFO  PID-10706 TID-6889100 ActionController::Base: Parameters: {"model_name"=>"user"}
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_field.html.haml (0.9ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_text.html.haml (0.9ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_boolean.html.haml (0.8ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_boolean.html.haml (0.7ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_field.html.haml (0.5ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_field.html.haml (0.5ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_field.html.haml (0.6ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_field.html.haml (0.5ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_field.html.haml (0.5ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_datetime.html.haml (3.8ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_datetime.html.haml (3.6ms)
  Rendered vendor/bundle/ruby/gems/rails_admin-0.7.0/app/views/rails_admin/main/_form_field.html.haml (0.6ms)
[...]

PR #7440, opened 2017-05-03
Remove rails_admin

### LINKED ISSUE #7027: undefined method `per_page_kaminari'  in admin_panel
I know we discourage podmin to use the admin_panel but it's still a very useful tool. This doesn't look like a timeout to me, so I wanted to know if an easy fix exists ([we're not the only ones with the problem](https://github.com/sferik/rails_admin/issues?q=is%3Aissue+undefined+method+%60per_page_kaminari%27+is%3Aclosed)). This looks like a pagination problem with kaminari, we currently use the 0.6.13 version.

```
[2016-08-24T12:18:01] INFO  PID-10352 TID-8116976800060 Rails: Started GET "/admin_panel/message" for X.X.X.X at 2016-08-24 12:18:01 +0200
[2016-08-24T12:18:01] INFO  PID-10352 TID-8116976800060 ActionController::Base: Processing by RailsAdmin::MainController#index as HTML
[2016-08-24T12:18:01] INFO  PID-10352 TID-8116976800060 ActionController::Base:   Parameters: {"model_name"=>"message"}
[2016-08-24T12:18:02] WARN  PID-28946 TID-364036367760 DiasporaFederation::Federation::Sender::HydraWrapper: success=false sender=951 obj=StatusMessage:20dd14604c110134437f1b5eeb325e42 url=https://www.diasporasf.org/receive/public message=couldnt_resolve_host code=0 time=1.454728
[2016-08-24T12:18:03] INFO  PID-10352 TID-8116976800060 ActionController::Base: Completed 500 Internal Server Error in 1754ms (ActiveRecord: 1.1ms)
[2016-08-24T12:18:03] FATAL PID-10352 TID-8116976800060 Rails: 
NoMethodError (undefined method `per_page_kaminari' for #<Message::ActiveRecord_Relation:0x000ec3d2fe40f8>):
```

SuperTux88:
This is now broken since more than half a year, so I think we can just remove it.

Podmins shouldn't use this anyway, because it does bad things, like replacing NULL with an empty string when editing something. If we really need an admin interface for something, we should add it to our own admin panel at /admins/dashboard.

I have already removed this from my own pod since many years now, because less dependencies are always good 👍 

Fixes #7027 and #6413


Flaburgan:
Not too sure honestly. I often need to go there because the moderation tool is currently very limited, it only allows to delete a message, a comment or a user. I'm fine doing that if we really apply:

>If we really need an admin interface for something, we should add it to our own admin panel at /admins/dashboard.

Meaning we should first at least list the actions podmins need. Then we will see what should be implemented, and if it's wise or not to remove the panel even if the feature is not implemented yet.

- The first thing which comes to my mind is the ability to change the user e-mail. I had multiple time users who can't reset their (forgotten) password because they don't have the e-mail used anymore. I then manually check their identity and change the e-mail. I can do that with the Rails console or SQL of course but the admin feels safer to me.
- I also use it to see links between reported accounts. It has been useful during the ISIS crisis for example. I used it to read PM between account, see who was sharing with whom...

I guess we need to do a poll with diaspora HQ to warn about the planning removal and ask for the current podmins usage. However that could advertise it... So maybe a generic message asking about needs for moderation?

SuperTux88:
> because the moderation tool is currently very limited

That's why I wrote we should make our own admin interface better.

> I had multiple time users who can't reset their (forgotten) password because they don't have the e-mail used anymore.

This can be added to the user search, where we already display the email address, so we can add an edit button, I'll created #7442 for that.

> I used it to read PM between account, see who was sharing with whom...

I'm not sure if we should add tools to spy users, and I think we shouldn't make it easy to read PMs (that's in fact also one reason why I want to remove the rails admin). I don't mind what people write in PMs, because it's private. If people do bad things publicly (or somebody reports a limited post), then I'll ban them because of that. If you really need to read PMs I think it's OK to need to go to the database for that.

> I guess we need to do a poll with diaspora HQ to warn about the planning removal and ask for the current podmins usage.

Since it is broken since 0.6, I don't think many are using it. Maybe there is a small group who have fixed it their own, but then they would also be able to re-add the dependency after this is merged. But for everybody else it's already broken, so I don't think we need a poll.

cmrd-senya:
>  I don't mind what people write in PMs, because it's private

It's a bit off topic, but we need to do something about the PM, since having non-e2e PM in 2017 is a quite bad feature. Perhaps if we had good enough XMPP integration, we could just drop our PM and use XMPP+OMEMO instead which supports e2e and offline messages which would be right enough for us. That's another discussion, but we should do something about that at some point.

SuperTux88:
@cmrd-senya yes, that's pretty off topic here. But even without e2e-encryption (where podmins theoretically can read all PMs), I still think that we shouldn't have a one-click solution for podmins to read other PMs.

(and since I mostly write PMs with other podmins, it's e2e enough for me :P )

denschub:
> I guess we need to do a poll with diaspora HQ to warn about the planning removal and ask for the current podmins usage.

Not really. We never made it an actual function in the first place, it's not linked anywhere. Also, it is broken now for a long time and literally noone complained, so the usage seems to be not there. Using rails_admin actually is a dangerous thing for reasons outlined here, so I have to agree with @SuperTux88, I doubt we'd even need to make a poll.

Also, just as a side note, if you really have to dig into the data, you can use the `rails c` for that. It's much easier than to look into the database, and it's faster than rails_admin. As a plus, you don't destroy data integrity by accident...

Flaburgan:
>Also, it is broken now for a long time and literally noone complained

It's not exactly true, I already redirected several podmins to #7027 on IRC. But you and Benjamin are right, we should not promote it doing a poll. So let's remove it ;)

dmorley:
I use rails_admin every day to troll and remove kitty porn, without it I am unsure how I would have a quick preview of the photos posted every day :/

skyschub (approved):

denschub:
Based on the discussion here, I'll approve this PR.

@dmorley you should not remote anything from within rails_admin, never. If anything, report the post yourself and delete it in the reports view. If you delete something via rails_admin, you only purge it from your database, that change will not get federated.

Removing it has several pros. In addition to the one mentioned above, we'll also remove a bunch of deps in our Gemfile, which makes it less pain to maintain in the long term. Also, it IS broken by default in our current distribution, so it's a useless feature anyway.

If you're able and willing to maintain your own fixes (which you apparently have done), then you'll also be able to revert the commit here after it get's removed, so you're free to continue using it. However, we do not have to care about it and there is no need to pull it around forever.

SuperTux88:
Merged as 7d9c6c85bd38d5703d17d0ed3d0721783d13b456 (as @denschub said, you can revert this commit, if you really need it)

PR #7443, opened 2017-05-04
Bump linters

SuperTux88:
The new pronto version has a new `github_pr_review` formatter, that's probably useful when somebody forgets to run pronto locally, so we only receive one review mail instead of 50 single comments ;) ([diaspora_federation](https://github.com/diaspora/diaspora_federation) already has the new pronto version, so it can be changed for @diaspora-code-review there too)

For rubocop the defaults for `Style/PercentLiteralDelimiters` have changed, I changed it back to the old defaults `()` for now, but we maybe want do discuss if we want to change to the new default `[]`, because the [ruby style guide](https://github.com/bbatsov/ruby-style-guide#percent-literal-braces) suggests to use `[]`. Any opinions about that?


cmrd-senya:
> For rubocop the defaults for Style/PercentLiteralDelimiters have changed, I changed it back to the old defaults () for now, but we maybe want do discuss if we want to change to the new default [], because the ruby style guide suggests to use []. Any opinions about that?

I don't have any special opinion about this specific style rule. I think following the ruby style isn't a bad choice if we don't have any strong reasons of not doing this in special cases.

skyschub (approved):

denschub:
I don't have any specific opinion on the style guide changes yet, but let's move that to a seperate discussion. Merged as c2eb53e8275649e6d79662c6001b4bd59acb45d8 and 153569e29753435a4387ff337b4a3a09a6563df3 after our bot just got blocked again...

PR #7446, opened 2017-05-09
Add link to discourse in footer and left side bar

Flaburgan:
See http://discourse.diasporafoundation.org/t/place-a-link-to-the-diaspora-discourse-instance-on-each-diaspora-pod/729/1

Do we want to put a "s" at Forum? Do we need to make that translatable? (I see wiki is not. I know Forum is the same in french...)


app/views/shared/_links.haml
```diff
@@ -1,5 +1,6 @@
 %li= link_to 'diasporafoundation.org', "https://diasporafoundation.org"
 %li= link_to 'Wiki', "https://wiki.diasporafoundation.org"
+%li= link_to 'Forum', "https://discourse.diasporafoundation.org"
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/views/shared/_links.haml
```diff
@@ -1,8 +1,9 @@
-%li= link_to 'diasporafoundation.org', "https://diasporafoundation.org"
-%li= link_to 'Wiki', "https://wiki.diasporafoundation.org"
-%li= link_to t('layouts.application.whats_new'), changelog_url
-%li= link_to t('layouts.header.code') + " " + pod_version, "#{source_url}", {:title => t('layouts.application.source_package')}
+%li= link_to "diasporafoundation.org", "https://diasporafoundation.org"
+%li= link_to "Wiki", "https://wiki.diasporafoundation.org"
+%li= link_to "Forum", "https://discourse.diasporafoundation.org"
+%li= link_to t("layouts.application.whats_new"), changelog_url
+%li= link_to t("layouts.header.code") + " " + pod_version, "#{source_url}", {:title => t("layouts.application.source_package")}
```
diaspora-code-review: - Line is too long. [127/120]
- Prefer `to_s` over string interpolation.
- Redundant curly braces around a hash parameter.
- Use the new Ruby 1.9 hash syntax.

SuperTux88:
I'm not sure if we want to simply name it "Forum", I think discourse should have a small description (that it's for discussions and support) in the "Need help?" section in the sidebar.

Flaburgan:
I was thinking about a generic "Forum" link but also another link directly to the "Support" category in the Help section. What do you think?

denschub:
"Forum" feels way to generic to me since it's very ambiguous term... I'd ++ @SuperTux88 here for a added description (that, jftr, should be translatable)

Flaburgan:
Any suggestion then?

Ravenbird:
What about two 'Diaspora Discussion' or 'Discussion and Support'?

denschub:
"Support and project discussions" feels good to me.

Flaburgan:
So I changed the text in the footer and added the link in the help section.

SuperTux88:
I think I would just drop the link in the footer (there are already enough links there, maybe even drop the wiki-link?).

I saw you replaced the tag-link for "#feature", maybe it makes sense to replace the the "#question" link with the support category?

Flaburgan:
Wow @SuperTux88 I missed your feedback sorry it toke that long to have an answer.

>I think I would just drop the link in the footer (there are already enough links there, maybe even drop the wiki-link?).

If we want to clean-up the footer a bit, I would have put the "diasporafoundation.org" link to the "powered by diaspora*" text, removed the wiki and maybe moved the link to the source code to the help or section section (and also the changelog, it keep being 404 on d-fr, I guess it's the way I rebase the branch?), but I feel like the discourse link should stay. Do you want me to do these changes in this PR?

>I saw you replaced the tag-link for "#feature", maybe it makes sense to replace the the "#question" link with the support category?

Hm, I changed #feature because as you saw with the recent "needz groups" discussion this is not productive, but I was more hesitant about #question. I'd love to see users helping each other without us. I agree we can move that to discourse but I'd like to keep the "community doing the support" thing.

SuperTux88:
> I would have put the "diasporafoundation.org" link to the "powered by diaspora*" text

👍 

> removed the wiki
👍 (it's already in the sidebar, i think that's enough)

> maybe moved the link to the source code to the help section 

No, the link to the source code is the version number, and I think the version number should stay. That is a very useful information.

> (and also the changelog, it keep being 404 on d-fr, I guess it's the way I rebase the branch?)

It tries to link to the commit you're currently running, but if you don't have that pushed in a repo that is connected (forked) with the original repo, then this doesn't work (I push my branch to github and then pull on the server this merged branch, but if you merge/rebase on the server, then you have a commit that doesn't exist anywhere else). But you can set a working URL [in the config](https://github.com/diaspora/diaspora/blob/develop/config/diaspora.yml.example#L536-L539) when the auto-detection doesn't work.

I don't know, I don't mind if this link is in the footer or in the sidebar, so we can move this to the sidebar if you like to.

> but I feel like the discourse link should stay.

I still think the discourse link shouldn't be in the footer, but it's not a strong opinion, what do others think?

But I would change the link in the sidebar to point to the start page, not only "support forum".

> Hm, I changed #feature because as you saw with the recent "needz groups" discussion this is not productive, but I was more hesitant about #question. I'd love to see users helping each other without us. I agree we can move that to discourse but I'd like to keep the "community doing the support" thing.

A "needz groups" discussion in discourse wouldn't be productive either ;) But to link #question to discourse too was just an idea, we can keep it as it is now.

Flaburgan:
Alright so wiki link removed, link to foundation website moved to "powered by diaspora" and added to the left nav bar. I kept the link to the root of discourse at the moment. It looks great imo.

skyschub (approved):
I like it. Let's stop bikeshedding here and merge this, we can always change stuff later if we feel like we need to!

denschub:
> I still think the discourse link shouldn't be in the footer

I have no opinion really. Let's release this and see if people notice it, and if we should move it somewhere else.

goobertron:
> Let's stop bikeshedding here

OK, I've not yet really come to terms with 'dogfooding' – now 'bikeshedding'? The only thinks that makes me think of are either smoking or snogging behind the bike sheds at school.

I am definitely getting too old... :P

SuperTux88:
Merged as ddda28102cd553b136a15654b4910bde3a9e8e96 and ea9e50085df7ff6a57c5abaca2be1a855d9de59b

Thank you

denschub:
> I am definitely getting too old... :P

I'm sorry, software engineer slang, I guess. ;) In this case, "bikeshedding"s origins are in [Parkinson's law of triviality](https://en.wikipedia.org/wiki/Law_of_triviality), where he describes discussions about the color and materials used for a bike shed while building a nuclear power plant.

We are not quite building a nuclear power plant here (although, arguably, diaspora might be just as dangerous), but having intense discussions about a link that may not even be perceived by a majority of users feels indeed like discussing on small, trivial details.

However - I find the idea of the entire diaspora team snogging behind a bike shed a bit amusing. And creepy. But amusing nonetheless.

goobertron:
Hehe. Thanks for the explanation – that makes sense now. The nearest phrase I know is 'rearranging the deck chairs on the Titanic', although that has perhaps a certain feel of looming catastrophe about it!

We might not be building a nuclear power plants, but I'm sure we've gone into meltdown a few times...

PR #7447, opened 2017-05-10
Add Nodeinfo 2.0 support

### LINKED ISSUE #7421: Add podmin account name to the statistics collected
If someone has set the `configuration.admins.account` setting in `config/diaspora.yml`, it would be useful to include this in the statistics output. This wouldn't result in any privacy leak that I'm aware of, because it would be opt-in (only if a podmin has enabled stats collection and if they have set a podmin account), and it would only give away the account name of the nominated podmin account (such as podmin@pod.example.com) rather than an email address or anything like that.

What think you?

SuperTux88:
I added support for the new NodeInfo 2.0. I reused the existing `version_10_hash` where 2.0 is still the same as 1.0. Only `protocols` are different now, because there is no separate `inbound`/`outbound` anymore.

I also added the `adminAccount` to the `metadata` if it is set in the diaspora.yml, this fixes #7421.

### DIFF
--- a/app/presenters/node_info_presenter.rb
+++ b/app/presenters/node_info_presenter.rb
@@ -26,8 +26,7 @@ class NodeInfoPresenter
 
   def add_static_data(doc)
     doc.software.name = "diaspora"
-    doc.protocols.inbound << "diaspora"
-    doc.protocols.outbound << "diaspora"
+    doc.protocols.protocols << "diaspora"
   end
 
   def add_user_counts(doc)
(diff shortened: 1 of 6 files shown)


skyschub (approved):
We discussed this long enough, so let's get this merged. :)

denschub:
Merged as e1777067693f8365d1d5d2f9d5155d70cc46746c 7934c1e969729686bcc964948bac3a381273ae03.

PR #7448, opened 2017-05-10
Don't raise when the public key of a person is "broken"

SuperTux88:
Breaking a public key of a person can be used to "block" receiving posts from this person on the pod. So we should handle that case better and not just trigger many retries for something that will fail again.

### DIFF
--- a/spec/models/person_spec.rb
+++ b/spec/models/person_spec.rb
@@ -560,6 +560,19 @@ describe Person, :type => :model do
     end
   end
 
+  describe "#public_key" do
+    it "returns the public key for the person" do
+      key = @person.public_key
+      expect(key).to be_a(OpenSSL::PKey::RSA)
+      expect(key.to_s).to eq(@person.serialized_public_key)
+    end
+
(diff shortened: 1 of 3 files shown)


spec/models/person_spec.rb
```diff
@@ -560,6 +560,19 @@
...
 
+  describe "#public_key" do
+    it "returns the public key for the person" do
+      key = @person.public_key
+      expect(key).to be_a(OpenSSL::PKey::RSA)
+      expect(key.to_s).to eq(@person.serialized_public_key)
+    end
+
+    it "handles broken keys and returns nil" do
+      @person.update_column(:serialized_public_key, "broken")
```
diaspora-code-review: Avoid using `update_column` because it skips validations.

skyschub (approved):
Looks super fine to me, thanks for the quick PR! <3

SuperTux88:
Thanks for the fast review @denschub, merged as 02b4d3f347b3150f3e2fa2906df8bd10fb3b5ccc.

Issue #7363, opened 2017-03-13
NoMethodError when sending a Welcome Message if the admin account is not found


hfroese:
Enabling welcome message causes error 500 to new registered user.
Not getting_started page but error 500.

Flaburgan:
Please check the `production.log` and post the stacktrace here.

hfroese:
OK - here we go:

```
[2017-03-13T17:36:00] INFO  PID-3056 TID-12996840 Rails: Started POST "/users" for 84.141.162.14 at 2017-03-13 17:36:00 +0100
[2017-03-13T17:36:00] INFO  PID-3056 TID-12996840 ActionController::Base: Processing by RegistrationsController#create as HTML
[2017-03-13T17:36:00] INFO  PID-3056 TID-12996840 ActionController::Base:   Parameters: {"utf8"=>"✓", "authenticity_token"=>"tt8aLmO/dDht8R2ZpCwkJezYINEtBXAcyVMTSA9Y/TtKMDpIHx4s9R1n6p4rJGnmduXFJPpSVNdNfe24WNFZLA==", "user"=>{"email"=>"testaccount@diaspora.zone", "username"=>"testaccount", "password"=>"[FILTERED]", "password_confirmation"=>"[FILTERED]"}, "commit"=>"Konto erstellen"}
[2017-03-13T17:36:02] INFO  PID-3020 TID-97685100 Sidekiq: Workers::DeferredDispatch JID-b1efbd4a8621c570d7884e89: start
[2017-03-13T17:36:02] INFO  PID-3056 TID-12996840 ActionController::Base: Completed 500 Internal Server Error in 1708ms (ActiveRecord: 51.2ms)
[2017-03-13T17:36:02] FATAL PID-3056 TID-12996840 Rails: 
NoMethodError (undefined method `person' for nil:NilClass):
  app/models/user.rb:435:in `send_welcome_message'
  app/controllers/registrations_controller.rb:17:in `create'
  lib/rack/internet_explorer_version.rb:34:in `call'
```

denschub:
You don't have an admin account set up, which is why there is no sender for the welcome message. [Please promote an account](https://wiki.diasporafoundation.org/FAQ_for_pod_maintainers#What_are_roles_and_how_do_I_use_them.3F_.2F_Make_yourself_an_admin_or_assign_moderators) to the admin role and the error will be gone.

SuperTux88:
[...]

PR #7452, opened 2017-05-11
Fix issue #7363

mahonrig:
Simple fix for issue #7363  - server crash if admin set in diaspora.yml is not a valid user and the welcome message is enabled.

Not sure how to write a test for this issue or if one is even needed.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -432,6 +432,7 @@ class User < ActiveRecord::Base
     return unless AppConfig.settings.welcome_message.enabled? && AppConfig.admins.account?
     sender_username = AppConfig.admins.account.get
     sender = User.find_by(username: sender_username)
+    return if sender.nil?
     conversation = sender.build_conversation(
       participant_ids: [sender.person.id, person.id],
       subject:         AppConfig.settings.welcome_message.subject.get,
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -875,6 +875,12 @@ describe User, :type => :model do
         user.send_welcome_message
         expect(user.conversations.count).to eq 0
       end
+
+      it "should send no welcome message if podmin is invalid" do
+        AppConfig.admins.account = "invalid"
+        user.send_welcome_message
+        expect(user.conversations.count).to eq 0
+      end
     end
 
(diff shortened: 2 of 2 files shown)


SuperTux88:
Thanks for your contribution.

> Not sure how to write a test for this issue or if one is even needed.

You can write a similar test like [here](https://github.com/diaspora/diaspora/blob/fa71af71c17911e4ce7a013ee1cc153cc8959d3a/spec/models/user_spec.rb#L873-L877), but with setting an invalid username: `AppConfig.admins.account = "invalid"`.

app/models/user.rb
```diff
@@ -432,6 +432,8 @@ def send_welcome_message
     return unless AppConfig.settings.welcome_message.enabled? && AppConfig.admins.account?
     sender_username = AppConfig.admins.account.get
     sender = User.find_by(username: sender_username)
+    # Check if admin account is no longer valid user per Issue #7363
```
SuperTux88: I don't think we need a comment for that, the check should be pretty self-explaining.

SuperTux88:
Can you please squash the commits to a single one and also write a more informative commit message like "Fix welcome message when user is not found" or so, so you don't need to open the issue to see what's done in this commit :) You can add "Fixes #7363" in the second line, so the commit is still linked with the issue.

mahonrig:
For future reference, is it preferred to keep pull requests down to one commit, or to just squash them before merge after any recommended changes have been made?

SuperTux88 (approved):

SuperTux88:
It is preferred to create as many commits as needed. So for example, don't create an extra commit when fixing a typo from a previous commit. But it is also preferred to create separate commits when doing 2 separate things, where it's useful to have separate commit messages (to describe what you have done in which commit). Or create separate commits when you do first a refactoring and then fix something or add a feature (so it's easier to review when refactorings are in it's own commit). But for this PR it's OK to have the fix and the test in one commit. 👍

SuperTux88:
Merged as 942e090b3a8d5c117f7b3d84385794c7ffc39556

Thank you!

Flaburgan:
Thank you and congratulation for your first pull request!

PR #7453, opened 2017-05-15
Use guid with some of links

cmrd-senya:
Use guid instead of id at permalink and in single post view

This changes links from `/posts/:id` to `/posts/:guid` where the links are not clicked very often. Id search is faster than guid search so we can't change it everywhere, but these links are not very useful for clicking, but can be used for easier guid look up.

P.S. Might need some tests to be fixed

### DIFF
--- a/app/assets/templates/single-post-viewer/single-post-content_tpl.jst.hbs
+++ b/app/assets/templates/single-post-viewer/single-post-content_tpl.jst.hbs
@@ -39,11 +39,11 @@
             {{/if}}
             <span class="post-time">
               {{#if root}}
-                <a href="/posts/{{root.id}}">
+                <a href="/posts/{{root.guid}}">
                   <time datetime="{{root.created_at}}" title="{{localTime root.created_at}}" />
                 </a>
               {{else}}
-                <a href="/posts/{{id}}">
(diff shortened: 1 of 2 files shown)


app/assets/templates/single-post-viewer/single-post-content_tpl.jst.hbs
```diff
@@ -39,11 +39,11 @@
             {{/if}}
             <span class="post-time">
               {{#if root}}
-                <a href="/posts/{{root.id}}">
+                <a href="/posts/{{root.guid}}">
```
SuperTux88: Should we change this in line 94 too?

cmrd-senya: Yes, we should use `guid` on the line 94.

Now I think that for this line (42)  we can leave `root.id` instead of `root.guid`, since it's more likely to be used directly to get the original post, so it may affect performance.

SuperTux88 (approved):

SuperTux88:
Merged as 6abd0509f911c0f3c0b15fbb6ea33c6462e2a89b

Thank you! 👍

Issue #7195, opened 2016-11-17
Save new email address without confirmation when mail is disabled in config


svbergerem:
When a user changes their email address on the settings page we will send a confirmation mail and save the new email address after it has been confirmed. If the podmin didn't enable `mail` in the pod's config we won't send a confirmation mail. Thus we should save the new email address without any confirmation.

zenara:
This is solved right? I have tested this and works fine. I didn't get any confirmation mails for changing my email address.

PR #7455, opened 2017-05-15
Save email without confirmation when podmail is disabled.

mahonrig:
Fix and test for issue #7195 

Not entirely sure this is optimal - I kept the first user save so that the validations on the unconfirmed email could run, and then checked podmail status before setting email and saving again.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -211,15 +211,23 @@ class UsersController < ApplicationController
   end
 
   def change_email(user_data)
-    @user.unconfirmed_email = user_data[:email]
-    if @user.save
-      @user.send_confirm_email
-      if @user.unconfirmed_email
+    if AppConfig.mail.enable?
+      @user.unconfirmed_email = user_data[:email]
+      if @user.save
+        @user.send_confirm_email
         flash.now[:notice] = t("users.update.unconfirmed_email_changed")
+      else
+        @user.reload # match user object with the database
+        flash.now[:error] = t("users.update.unconfirmed_email_not_changed")
       end
     else
-      @user.reload # match user object with the database
-      flash.now[:error] = t("users.update.unconfirmed_email_not_changed")
+      @user.email = user_data[:email]
+      if @user.save
+        flash.now[:notice] = t("users.update.settings_updated")
+      else
+        @user.reload
+        flash.now[:error] = t("users.update.unconfirmed_email_not_changed")
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -203,6 +203,13 @@
         expect(Workers::Mail::ConfirmEmail).to receive(:perform_async).with(@user.id).once
         put(:update, :id => @user.id, :user => { :email => "my@newemail.com"})
       end
+
+      it "saves unconfirmed_email when podmail is disabled" do
+        AppConfig.mail.enable = false
+        put(:update, :id => @user.id, :user => { :email => "my@newemail.com"})
```
diaspora-code-review: - Use the new Ruby 1.9 hash syntax.
- Space inside { detected.

svbergerem: Please fix this code style issue. The hash syntax should be 

```ruby
put(:update, id: @user.id, user: {...})
```

You can check the code style locally as described here: https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review

mahonrig: Want me to fix the rest of the tests in that file? There's a number of them using the older syntax (I had just copied one of the existing tests and modified)

svbergerem: For bigger pull requests it's a good idea to keep unrelated changes separate. This makes the changes more readable and makes the review easier.

In this case it's just a small PR, so do whatever you like. :)

app/controllers/users_controller.rb
```diff
@@ -213,9 +213,19 @@ def change_language(user_data)
...
-      @user.send_confirm_email
-      if @user.unconfirmed_email
-        flash.now[:notice] = t("users.update.unconfirmed_email_changed")
+      if AppConfig.mail.enable?
+        @user.send_confirm_email
+        if @user.unconfirmed_email
+          flash.now[:notice] = t("users.update.unconfirmed_email_changed")
+        end
+      else
+        @user.email = @user.unconfirmed_email
```
SuperTux88: We don't need to save the email to `@user.unconfirmed_email` if mail is disabled on that pod. We can save it directly to `@user.email` then.

SuperTux88 (changes requested):

app/controllers/users_controller.rb
mahonrig: OK, it had looked like there was some useful validations that only ran on the unconfirmed_email field, such as the duplicate email check.

I'll get it changed.

SuperTux88: The `email` has also a unique index on it, because we need to validate duplicates when somebody registers with an already used mail-address. If there are other checks only on `unconfirmed_email`, we should probably add them to `email` too, because `unconfirmed_email` is only to store the new mail-address temporarily until it is confirmed, and we don't need that when we cant confirm it.

SuperTux88 (approved):

SuperTux88:
Merged as 5739f6760594b3f153c3d2d11cd81f5cd87a36e8

Thank you 👍

PR #7465, opened 2017-05-24
Cleanup some translations

SuperTux88:
While doing some translations, I found some unused keys. I also fixed some double-spaces.

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -55,14 +55,10 @@ en:
           attributes:
             person_id:
               taken: "must be unique among this user’s contacts."
-        request:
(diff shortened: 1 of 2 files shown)


goobertron:
Nice catch. I can 'review' the removal of double spaces as correct, but I'm not able to check for missing keys, sorry. Hope someone else sees this soon.

skyschub (approved):

SuperTux88:
Merged as 0e13a316f6922eb88507fc68cfa2fc742b5e7378

Thanks @goobertron and @denschub for the reviews 👍 

(Normally I wouldn't add that to the changelog, so it's only to not look that empty ;) )

PR #7473, opened 2017-06-14
Warn user if he leaves profile editing page

mahonrig:
Added cuke test to orphaned pull request #6589 

Still learning cucumber but this seems to work correctly...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -112,6 +112,7 @@ source "https://rails-assets.org" do
 
   gem "rails-assets-autosize",                            "3.0.20"
   gem "rails-assets-blueimp-gallery",                     "2.25.0"
+  gem "rails-assets-jquery.are-you-sure",                 "1.9.0"
   gem "rails-assets-jquery-placeholder",                  "2.3.1"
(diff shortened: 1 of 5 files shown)


Flaburgan:
We already warn the user when I quit a page with the publisher filled. Why do you need a new library to do that?

mahonrig:
The only feedback on this old pull request had been to ask for a test, so I thought I'd try my hand at that. Didn't write the rest of it.
 
But as for the new library it makes it much easier to attach a handler to all fields of the form and warn only if they've changed. I haven't checked to see how it's done with the publisher, but since it's only one field I think the solution there is different. 

This [stack overflow answer](https://stackoverflow.com/questions/7317273/warn-user-before-leaving-web-page-with-unsaved-changes#7317311) goes a bit in depth on the problem and ends up recommending the AreYouSure library if jquery is already being used.

SuperTux88:
I had the same question as @Flaburgan at the beginning, but I think you're right. In this case the library makes sense, while the publisher is only a single field. So thanks for finishing the orphaned PR.

Merged as df1139234f5e43edf093f401660c8b627478d161

PR #7477, opened 2017-06-30
Add a Contributing file

Flaburgan:
As [suggested on discourse](https://discourse.diasporafoundation.org/t/learn-to-code-review/677/6)

Pinging @goobertron for the english verifications ;)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,42 @@
+# Contributing to diaspora\*
+
+First of all: thank you very much for helping us out!
+
+## Things you need to know before contributing
+
+If you want to get in touch with other diaspora\* developers, [check our wiki][how-we-communicate] for information on how we communicate. Feel free to ask if you have any questions!
+
+Everyone interacting with our code, issue trackers, chat rooms, mailing lists, the wiki, and the discourse forum is expected to follow the [diaspora\* code of conduct][code-of-conduct].
+
+## Report a security issue
+
+Found a security issue? Please disclose it responsibly. We have a team of developers listening to [security@diasporafoundation.org][sec-mail]. The PGP fingerprint is [AB0D AB02 0FC5 D398 03AB 3CE1 6F70 243F 27AD 886A][pgp].
+
+## Contributing to translations
+
+We use [WebTranslateIt][webtranslateit] to manage translations of the app interface. Please read [our wiki page][translation-wiki] to find out more about this. If your language is featured on WebTranslateIt, please do **not** open a pull request to update translations. If it is not already featured on WebTranslateIt, please read the wiki article above to find out how to proceed.
+
+## Contributing to the code
+
+**Before opening a pull request** please read [how to contribute][contribute]. Doing things the right way from the start will save us time and mean that your contribution can be integrated quicker!
+- Follow our [git workflow][git-workflow] method. In particular, *do not* open a pull request from the `master` or the `develop` branch.
+- Follow our [styleguide][styleguide] and run pronto, our syntax analyzer, **locally before opening a pull request**. See [our wiki][pronto] for information on how to do this.
+- [Write tests][testing-workflow].
+- Use meaningful commit-messages and split larger tasks, e.g. refactoring, into separate commits. This makes the review process much easier.
+
+## Other ways to contribute
+
+You don’t know code? No worries, there are plenty other ways to help the diaspora* project! Please find out how you can help [on our wiki][other-ways].
+
+[code-of-conduct]: https://github.com/diaspora/diaspora/blob/develop/CODE_OF_CONDUCT.md
+[how-we-communicate]: https://wiki.diasporafoundation.org/How_we_communicate
+[pgp]: https://pgp.mit.edu/pks/lookup?op=get&search=0x6F70243F27AD886A
+[sec-mail]: mailto:security@diasporafoundation.org
+[webtranslateit]: https://webtranslateit.com/en/projects/3020-Diaspora
+[translation-wiki]: https://wiki.diasporafoundation.org/Contribute_translations
+[contribute]: https://wiki.diasporafoundation.org/Getting_started_with_contributing
+[git-workflow]: https://wiki.diasporafoundation.org/Git_Workflow
+[styleguide]: https://wiki.diasporafoundation.org/Styleguide
+[pronto]: https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review
+[testing-workflow]: https://wiki.diasporafoundation.org/Testing_Workflow
+[other-ways]: https://wiki.diasporafoundation.org/Other_ways_to_contribute


goobertron:
Oh, thank you Fla for doing this! I had been meaning to, but am a bit shaky on the technical side. I'll have a proper look at the English tomorrow (just a couple of tiny changes, I think).

SuperTux88 (approved):
The technical side looks good, but I am a bit shaky on the english side ;)

goobertron:
Here are my suggestions: https://github.com/goobertron/diaspora/commit/bba75b761823d4f4e86ee7f0e83e2acb3a7b8d26

Basically:
- a little bit of clean-up to the English,
- a sentence about what to do if your language is not on WebTranslateIt,
- add a link to /Getting_started_with_contributing on the wiki.

Hope that helps! And thanks again for doing this.

CONTRIBUTING.md
```diff
@@ -14,26 +14,27 @@ Found a security issue? Please disclose it responsibly. We have a team of develo
 
 ## Contributing to translations
 
-We use [WebTranslateIt][webtranslateit] to manage translation. Please read [our wiki page][translation-wiki] to know more about it. Except if your language isn't on webtranslateit, please do **not** open a pull request to update translations.
+We use [WebTranslateIt][webtranslateit] to manage translations of the app interface. Please read [our wiki page][translation-wiki] to find out more about this. If your language is on WebTranslateIt, please do **not** open a pull request to update translations. If you want to contribute to a language that is not featured on WebTranslateIt, please contact our developers to find out how to proceed.
```
Flaburgan: >If you want to contribute to a language that is not featured on WebTranslateIt, please contact our developers to find out how to proceed

Not too sure about that, I'd like to redirect to the wiki page, but we definitely need to update its content. @jhass you were the one who managed translations, what's the current recommendations to contribute to translations? Do you have some time to update https://wiki.diasporafoundation.org/Contribute_translations ?

Flaburgan:
Thank you for your feedback @goobertron! I integrated your commit. I will probably squash it before merging this but still, it's there :p

CONTRIBUTING.md
goobertron: I added that because I \*think\* that is how things have been done in the past for a new language. But I'm by no means sure. I suppose it would have been better if I'd have asked Jonne and @denschub rather than just adding that sentence to my PR! Hopefully they can confirm for you what is the proper procedure.

Flaburgan: We currently have to tell contributors (who want to add a new language) what to do because the wiki page is outdated. Instead of telling people come on IRC, hoping that someone who knows will be available, we should write his answer on the wiki.

denschub: The list of "Languages handled outside of WebTranslateIt" in the wiki is quite accurate.

goobertron: I was thinking also about how to handle a language that is not yet part of D* at all. I don't think we have any instructions for that. Let's add a section to the wiki article about what to do for a completely new language, and then the final sentence in the snippet above can be changed to:

> If you want to contribute to a language that is not featured on WebTranslateIt, please read the wiki article above to find out how to proceed.

Flaburgan:
Alright, corrected, and commits squashed. Should be ready to go and I'll dig into the wiki page.

Flaburgan:
I reorganized https://wiki.diasporafoundation.org/Contribute_translations please review. It feels way clearer for me now.

goobertron:
That looks a lot better, thanks. I don't fully understand how this works, so excuse me if this is wrong, but in the paragraph where you say 'Copy the en files (listed below), choose the correct language code (see above) and rename every occurrence of "en" in the files names and in the code.' – while this is often correct for an entirely new language, isn't it best to copy the files of the locale on which your new language is based, and to which strings will fall-back if there is no translation specified in the new locale? E.g. for de-AT, you'd copy the de files and just leave in the strings that will be different from the standard de locale. I think!

I might now finally get around to completing a British English translation – been meaning to do this for ages.

Flaburgan:
This is a fair point. The problem can be, we never know the state of the other languages than english (where all the keys are for sure). I don't know the opinion of others on that point, what do we want to recommend?
Also, adding a new language doesn't mean it has to be updated with github, it can also be integrated in WebTranslateIt. I need help from others, possibly @jhass to modify the page about that.
But we're going out of topic, let's focus on this PR here. There are discussion pages on the wiki if needed.

SuperTux88:
About new languages: When you want to create a new language that is managed in WebTranslateIt (which should be the default), you don't need (and shouldn't) copy any files. A manager on WebTranslateIt needs to create the language there, and that files will be created when pulling from WebTranslateIt. You only need to manage the files by hand, when the language isn't managed with WebTranslateIt.

SuperTux88:
@goobertron can you approve this PR when the English is OK?

goobertron:
I've just one small change to request: to change the last sentence of line 17 to:

> If it is not already featured on WebTranslateIt, please read the wiki article above to find out how to proceed.

I think it's worth mentioning WebTranslateIt again here to make the meaning absolutely clear.

Everything else is fine.

Flaburgan:
Done ;)

Flaburgan:
>About new languages: When you want to create a new language that is managed in WebTranslateIt (which should be the default), you don't need (and shouldn't) copy any files. A manager on WebTranslateIt needs to create the language there, and that files will be created when pulling from WebTranslateIt. You only need to manage the files by hand, when the language isn't managed with WebTranslateIt.

But is there possible to fork this language from a based one on WebTranslateIt? I mean the de, de-AT example is accurate.

denschub:
There is no such thing as "forking" a language. There is no need to copy an entire language. If you want to create a new language based on an existing one, you have to change the `fallbacks` array in [`local_settings.yml`](https://github.com/diaspora/diaspora/blob/d62d79bc293c89c6439ce4ec2414a17d5dc5b11e/config/locale_settings.yml#L77), that's all. Mass-duplicating segments is discouraged and bad.

Flaburgan:
That's exactly the kind of feedback I need as I don't know how things work. Could you please update the wiki page?

SuperTux88:
@goobertron after [this](https://discourse.diasporafoundation.org/t/i-want-to-start-contributing-to-diaspora-development-where-should-i-start/817/9) discussion on discourse, I saw that this PR is still open. Is everything OK from your side with this? We should still update the wiki, but that is out of scope of this PR. So if everything is OK, I'd like to merge this.

goobertron:
To save me reading through the whole file again, then if nothing apart from my suggested edit has been changed since my last comment, my comment that 'Everything else is fine.' still stands.

SuperTux88:
I don't have the commit-ID from 17 days ago anymore, so I can't check that. @Flaburgan can you confirm that you only changed the suggested edit?

Flaburgan:
Yep, the suggestion about WebTranslateIt. Nothing else than the translation paragraph changed.

goobertron:
Then it's still good to go, as far as I'm concerned!

SuperTux88:
Merged, thanks @Flaburgan and @goobertron 👍

PR #7478, opened 2017-07-04
Increase poltergeist timeout

Flaburgan:
Testing if this solves [the activity failing test](https://travis-ci.org/diaspora/diaspora/jobs/235463906#L589) as recommended there, because we [have](https://travis-ci.org/diaspora/diaspora/jobs/244192324#L612) [it](https://travis-ci.org/diaspora/diaspora/jobs/244192325#L612) [a](https://travis-ci.org/diaspora/diaspora/jobs/244192328#L589) [lot](https://travis-ci.org/diaspora/diaspora/jobs/244192329#L589).

### DIFF
--- a/features/desktop/activity_stream.feature
+++ b/features/desktop/activity_stream.feature
@@ -6,11 +6,11 @@ Feature: The activity stream
       | Bob Jones   | bob@bob.bob       |
       | Alice Smith | alice@alice.alice |
     And a user with email "bob@bob.bob" is connected with "alice@alice.alice"
-    When "alice@alice.alice" has posted a status message with a photo
+    And "alice@alice.alice" has posted a status message with a photo
 
   Scenario: delete a comment
-    When "bob@bob.bob" has commented "is that a poodle?" on "Look at this dog"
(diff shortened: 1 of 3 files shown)


Flaburgan:
Only 3 jobs failed and it was not related to the activity test. I relaunched those, they are still failing but it's still unrelated. I launch them again a last time, but if my activity is still green then I guess we're safe to say this PR improved the situation :)

SuperTux88:
Hmm, it seems that the first web-action is too slow, I just wonder how long this helps and when we need to increase it more. And 60 seconds was already increased (and it's really long in my eyes, default is 30 seconds). But it works, so ¯\_(ツ)_/¯

Uplifted to 0.6.7.0 (hopefully it helps to get a green master after the release) and merged as 59c343c0c60b7116a2294cf9f216e24b06b06a5e

Thanks ❤️

Flaburgan:
Yeah, I guess the clean solution should be to understand why this step is taking that long. I tried, but it was over my skills here, or I needed to spend more time and I didn't want to :p

PR #7295, opened 2017-01-25
Add admin pages to the mobile version

SansPseudoFix:
I was starting to code the report view, but I figured out I have to code all admin pages, so…
This is a early-PR because I don't want to code every pages and finally understand it won't be merge or wanted.

So far, reports page, user-search and user entry pages are mobile friendly (doesn't mean totally finished though). I also add admin menu to the drawer.

Here some screenshots:

drawer
report
usersearch

What do you think? Do I continue or not?

Next:

- [x] dashboard
- [x] weekly user stats
- [x] pod stats
- [x] pod network?
- [x] ~~sidekiq monitor?~~ already mobile friendly
- [x] refactoring/pronto

Edit: and, of course, I forgot my "damit, I forgot pronto" commit… Sorry.

PR #7479, opened 2017-07-05
Add access to the report page on mobile for moderators

Flaburgan:
Follow up of #7295 where I forgot about the moderator role. This PR add a link in the mobile drawer to allow moderators to access the report page.
I also added a test to be sure that those links are not displayed to normal users.

@SuperTux88 I know you branched 0.6.7.0 tonight, as you're the one who dispatches commits everywhere, you decide if this can be included in it or not ;) (The review should be very easy).

### DIFF
--- a/app/views/layouts/_drawer.mobile.haml
+++ b/app/views/layouts/_drawer.mobile.haml
@@ -52,5 +52,12 @@
                 = t("admins.admin_bar.report")
             %li= link_to t("admins.admin_bar.pod_network"), admin_pods_path
             %li= link_to t("admins.admin_bar.sidekiq_monitor"), sidekiq_path
+      - elsif current_user.moderator?
+        %li
+          %a{href: report_index_path}
+            - if unreviewed_reports_count > 0
+              .pull-right.badge
+                = unreviewed_reports_count
+            = t("admins.admin_bar.report")
       %li= link_to t("layouts.application.toggle"), toggle_mobile_path
       %li= link_to t("layouts.header.logout"), destroy_user_session_path, method: :delete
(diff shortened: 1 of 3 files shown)


app/views/layouts/_drawer.mobile.haml
```diff
@@ -35,6 +35,13 @@
           = person_image_tag(current_user, size: :thumb_small)
       %li= link_to t("_contacts"), contacts_path
       %li= link_to t("layouts.header.settings"), edit_user_path
+      - if current_user.moderator? && !current_user.admin?
```
SuperTux88: Can you do that the same way [as it is done on desktop](https://github.com/diaspora/diaspora/blob/526ebcfa13a799cae6bfbcc822c958ebecfa4d2b/app/assets/templates/header_tpl.jst.hbs#L102-L106)? So put it as `- elsif current_user.moderator?` below the `- if current_user.admin?` block, then you don't need `&& !current_user.admin?`

SuperTux88 (changes requested):
Only a small change

app/views/layouts/_drawer.mobile.haml
Flaburgan: Sure! Done. I wasn't satisfied by this solution but didn't think about that. I was surprised to discover that the `current_user.moderator?` returns `true` for admins.

SuperTux88 (approved):

SuperTux88:
Uplifted to 0.6.7.0 and merged as 2dc142bd70797bd00beebc6a5bccbf35d4234d4b

Thank you! 👍

PR #7480, opened 2017-07-05
Fix height too high on mobile SPV, closes #7402

### LINKED ISSUE #7402: Spaceing between comment-box and counters on mobile SPV is too high
See

screen shot 2017-03-28 at 07 00 58

Which looks pretty broken. There is a min-height on the (empty) comments container in [`app/assets/stylesheets/mobile/mobile.scss:58`](https://github.com/diaspora/diaspora/blob/ec6a6d51385ee7e6ef672c1e0d4d3ed5691f1ae0/app/assets/stylesheets/mobile/mobile.scss#L58) which we maybe can remove.

Flaburgan:
Looks like the min-height was not needed even for the  `.stream-element` elems.

### DIFF
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -55,7 +55,6 @@ textarea { resize: vertical; }
   overflow: auto;
   position: relative;
   text-align: left;
-  min-height: 34px;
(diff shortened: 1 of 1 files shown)


SuperTux88 (approved):

SuperTux88:
Merged, thank you 👍

PR #7483, opened 2017-07-07
Always link comment count text on mobile

### LINKED ISSUE #7393: Always link comment count text on mobile
diaspora

As stated in my other [issue](https://github.com/diaspora/diaspora/issues/7388) users on my pod have problems commenting posts on mobile version.

When there are no comments the "0 comments" text is not linked to open the comment section and write a comment, while it is when there are comments. It should always be linked and behave the same way.

Another problem is, that the icons on the right for comment and like do not look like they are linked to an action, but just indicators for the counters. This is enhanced by the fact that the comment count information is redundant (text "0 comments" beside comment icon 0).

Instead of writing comments they end up writing new posts, as the write icon in the menu bar is more appealing.

Is there a way to make the icons more attractive? On desktop I would expect them to have a hover effect if they are clickable, but that does not work on touch screen devices.

Flaburgan:
Closes #7393

### DIFF
--- a/app/helpers/mobile_helper.rb
+++ b/app/helpers/mobile_helper.rb
@@ -43,15 +43,12 @@ module MobileHelper
       entypo_class = "entypo-chevron-down"
     end
 
-    if post.comments_count > 0
-      link_to "#{t('admins.stats.comments', count: post.comments_count)}<i class='#{entypo_class}'></i>".html_safe,
-              post_comments_path(post, format: "mobile"),
-              class: "show-comments #{klass}"
-    else
-      html = "<span class='show-comments'>"
-      html << t("admins.stats.comments", count: post.comments_count)
-      html << "</span>"
-    end
+    link_to safe_join([
+                        t("admins.stats.comments", count: post.comments_count),
+                        content_tag(:i, nil, class: entypo_class)
+                      ]),
+            post_comments_path(post, format: "mobile"),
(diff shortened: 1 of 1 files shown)


app/helpers/mobile_helper.rb
```diff
@@ -43,15 +43,9 @@ def show_comments_link(post, klass="")
...
-    if post.comments_count > 0
-      link_to "#{t('admins.stats.comments', count: post.comments_count)}<i class='#{entypo_class}'></i>".html_safe,
-              post_comments_path(post, format: "mobile"),
-              class: "show-comments #{klass}"
-    else
-      html = "<span class='show-comments'>"
-      html << t("admins.stats.comments", count: post.comments_count)
-      html << "</span>"
-    end
+    link_to "#{t('admins.stats.comments', count: post.comments_count)}<i class='#{entypo_class}'></i>".html_safe,
```
diaspora-code-review: Tagging a string as html safe may be a security risk, prefer `safe_join` or other Rails tag helpers instead.

Flaburgan: Looks like it works, thank you. Pronto made me write a weird indent but eh, if that's the ruby style.

SuperTux88 (approved):

SuperTux88:
Merged, thanks ❤️

Issue #6559, opened 2015-11-22
Ergonomy suggestion


EricC-59:
Diaspora's robot sends e-mails with headers such as:

```
From: "podmin@podname.org (diaspora*)" <no-reply@podname.org>
To: "me@podname.org" <me@example.com>
```

… or:

```
From: “Firstname Lastname (diaspora*)” <no-reply@podname.org>
To: “Myfirstname Mylastname” <me@example.com>
```

Given the fact that (too) many mail user agents hide e-mail addresses, the users may see this:

> From: podmin@podname.org (diaspora*)
> To: me@podname.org

This can be disturbing for some users, because it seems that they get an e-mail on me@podname.org… which is not an e-mail address, nor podmin@podname.org!

Worse: those users can see that:

> From: Firstname Lastname (diaspora*)
> To: Myfirstname Mylastname

In this case, it seems logical to use the "reply" button to reply to "Mr Lastname"...  which, of course, results in an “Undelivered Mail Returned to Sender” (no-reply@podname.org).

Here is my suggestion:
- Do not directly use the sender's name in the From: field. Instead, use a text like "Diaspora\* (on behalf of Firstname Lastname)"
- **Never** use Diaspora\* identifiers in an e-mail address description. Using nothing is less disturbing.

See with my two examples:

```
From: "podmin@podname.org (diaspora*)" <no-reply@podname.org>
To: "me@podname.org" <me@example.com>
```

becomes

```
From: "Diaspora*” <no-reply@podname.org>
To: <me@example.com>
```

and the end-user sees:

> From: Diaspora*
> To: me@example.com

and

```
From: “Firstname Lastname (diaspora*)” <no-reply@podname.org>
To: “Myfirstname Mylastname” <me@example.com>
```

becomes

```
From: "Diaspora* (on behalf of Firstname Lastname)” <no-reply@podname.org>
To: "Myfirstname Mylastname” <me@example.com>
```

and the end-user sees:

> From: Diaspora\* (on behalf of Firstname Lastname)
> To: "Myfirstname Mylastname”

More newbie-friendly, isn't it?

Note that my suggestion comes from a real life case, where I tried to convince a Facebook user to join me on Diaspora*. With this e-mail issue, he told me:

> I don't understand anything to your diaspora thing. It's too complicated. (..) I even couldn't reply to _your_ e-mail from diaspora. I did reply and send, and the e-mail came back with "undelivered mail returned to sender"
##

Flaburgan:
[...]

PR #7484, opened 2017-07-07
Include first/last name in invitation emails body

### LINKED ISSUE #7126: Include first/last name in invitation emails
Invitation emails should look like:

> Hello!
> 
> You have been invited to join diaspora\* by Bob Surname (bob99@pod.geraspora.de)!
> (...)

Currently they look like:

> Hello!
> 
> You have been invited to join diaspora\* by bob99@pod.geraspora.de!
> (...)

Flaburgan:
Closes #7126

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -22,7 +22,7 @@ class Person < ActiveRecord::Base
   end
 
   has_one :profile, dependent: :destroy
-  delegate :last_name, :image_url, :tag_string, :bio, :location,
+  delegate :last_name, :full_name, :image_url, :tag_string, :bio, :location,
            :gender, :birthday, :formatted_birthday, :tags, :searchable,
            :public_details?, to: :profile
   accepts_nested_attributes_for :profile
--- a/app/views/notifier/invite.markerb
+++ b/app/views/notifier/invite.markerb
@@ -1,1 +1,8 @@
-<%= t('.message', :invite_url => invite_code_url(@invitation_code), :diasporafoundation_url => 'https://diasporafoundation.org/', :diaspora_id => @inviter.try(:diaspora_handle)) %>
+<%= t('.message',
+  invite_url: invite_code_url(@invitation_code),
+  diasporafoundation_url: 'https://diasporafoundation.org/',
+  user: @inviter.try(:diaspora_handle) == @inviter.try(:name) ?
+    @inviter.try(:diaspora_handle) :
+    @inviter.try(:name) + " (" + @inviter.try(:diaspora_handle) + ")",
+  diaspora_id: @inviter.try(:diaspora_handle))
+%>
--- a/app/views/notifier/invite.markerb
+++ b/app/views/notifier/invite.markerb
@@ -1,8 +1,6 @@
 <%= t('.message',
   invite_url: invite_code_url(@invitation_code),
   diasporafoundation_url: 'https://diasporafoundation.org/',
-  user: @inviter.try(:diaspora_handle) == @inviter.try(:name) ?
-    @inviter.try(:diaspora_handle) :
-    @inviter.try(:name) + " (" + @inviter.try(:diaspora_handle) + ")",
+  user: @inviter.try(:full_name).empty? ? @inviter.try(:diaspora_handle) : "#{@inviter.name} (#{@inviter.diaspora_handle})",
   diaspora_id: @inviter.try(:diaspora_handle))
 %>
--- a/features/step_definitions/user_steps.rb
+++ b/features/step_definitions/user_steps.rb
@@ -168,10 +168,14 @@ Then /^I should have (\d+) email delivery$/ do |n|
   ActionMailer::Base.deliveries.length.should == n.to_i
 end
 
-Then /^I should not see "([^\"]*)" in the last sent email$/ do |text|
+Then /^I should( not)? see "([^\"]*)" in the last sent email$/ do |negate, text|
   email_text = Devise.mailer.deliveries.first.body.to_s
   email_text = Devise.mailer.deliveries.first.html_part.body.raw_source if email_text.blank?
-  email_text.should_not match(text)
+  if negate
+    expect(email_text).to_not have_content(text)
+  else
+    expect(email_text).to have_content(text)
+  end
 end
 
 When /^"([^\"]+)" has posted a (public )?status message with a photo$/ do |email, public_status|
--- a/Changelog.md
+++ b/Changelog.md
@@ -36,6 +36,7 @@ If so, please delete it since it will prevent the federation from working proper
 * Render mentions as links in comments [#7327](https://github.com/diaspora/diaspora/pull/7327)
 * Add support for mentions in comments to the front-end [#7386](https://github.com/diaspora/diaspora/pull/7386)
 * Support direct links to comments on mobile [#7508](https://github.com/diaspora/diaspora/pull/7508)
+* Add inviter first and last name in the invitation e-mail [#7484](https://github.com/diaspora/diaspora/pull/7484)
 
 # 0.6.8.0
 
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -44,9 +44,9 @@ class Notifier < ActionMailer::Base
     @invitation_code = invitation_code
 
     I18n.with_locale(locale) do
-      mail_opts = {:to => email, :from => AppConfig.mail.sender_address,
-                 :subject => I18n.t('notifier.invited_you', :name => @inviter.name),
-                 :host => AppConfig.pod_uri.host}
+      mail_opts = {to: email, from: "\"#{AppConfig.settings.pod_name}\" <#{AppConfig.mail.sender_address}>",
+                 subject: I18n.t("notifier.invited_you", name: @inviter.name),
+                 host: AppConfig.pod_uri.host}
 
       mail(mail_opts) do |format|
         format.text { render :layout => nil }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -46,7 +46,7 @@ class User < ActiveRecord::Base
 
   delegate :guid, :public_key, :posts, :photos, :owns?, :image_url,
            :diaspora_handle, :name, :atom_url, :profile_url, :profile, :url,
-           :first_name, :last_name, :gender, :participations, to: :person
+           :first_name, :last_name, :full_name, :gender, :participations, to: :person
   delegate :id, :guid, to: :person, prefix: true
 
   has_many :aspects, -> { order('order_id ASC') }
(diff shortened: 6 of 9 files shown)


SuperTux88:
I think the name should only be added if the user has set a name, if not it would just be the diaspora ID twice.

Flaburgan:
Corrected.

app/views/notifier/invite.markerb
```diff
@@ -1 +1,8 @@
-<%= t('.message', :invite_url => invite_code_url(@invitation_code), :diasporafoundation_url => 'https://diasporafoundation.org/', :diaspora_id => @inviter.try(:diaspora_handle)) %>
+<%= t('.message',
+  invite_url: invite_code_url(@invitation_code),
+  diasporafoundation_url: 'https://diasporafoundation.org/',
+  user: @inviter.try(:diaspora_handle) == @inviter.try(:name) ?
```
SuperTux88: Sorry, but this is really unreadable. Also it feels wrong to first fallback to the diaspora ID and then ignore it again. It would be better to handle it directly. As I wrote in https://github.com/diaspora/diaspora/issues/6559#issuecomment-314927738 we should probably create a method, that returns `nil` when there is no name.

SuperTux88: Sorry, but this is really unreadable. Also it feels wrong to first fallback to the diaspora ID and then ignore it again. It would be better to handle it directly. As I wrote in https://github.com/diaspora/diaspora/issues/6559#issuecomment-314927738 we should probably create a method, that returns `nil` when there is no name.

app/views/notifier/invite.markerb
```diff
@@ -1 +1,8 @@
...
+    @inviter.try(:diaspora_handle) :
+    @inviter.try(:name) + " (" + @inviter.try(:diaspora_handle) + ")",
```
SuperTux88: This should be `"#{@inviter.try(:name)} (#{@inviter.try(:diaspora_handle)})"`

SuperTux88 (changes requested):

app/views/notifier/invite.markerb
SuperTux88: After thinking about this, we should probably rename the existing `.name` to `.display_name` or something like this. And then we could create a new `.name` that returns `nil`. That would also be a good preparation for #3648.

app/views/notifier/invite.markerb
fwoelm: How about combining your proposal with a method #name_with_handle or #pretty_name or #humanize_handle that returns:

~~~
# Return the user's named (if set) and their handle 
def name_with_handle
  if self.name.present?
    return "#{self.name} (#{self.handle})"     # User (userhandle@domain.tld)
  else
    return "#{self.handle}"                    # userhandle@domain.tld
  end
end
~~~

SuperTux88: I don't think we need the name with the diaspora ID anywhere else.

fwoelm: Fair enough :grin:

app/views/notifier/invite.markerb
Flaburgan: Indeed, but that's a big refactor (291 results in 99 file for .name) and I don't want to do it in this PR. I'll create a new method which will return nil. How should I call it at the moment?

SuperTux88: Ok, you're right, and changing all of these probably isn't a good idea anyway. So how about this:
```ruby
@inviter.try(:full_name).empty? ? @inviter.try(:diaspora_handle) : "#{@inviter.name} (#{@inviter.diaspora_handle})"
```

That would work for now and I'll do the refactoring when I do #3648

Flaburgan: So no new "nil" method? It could also be used in https://github.com/diaspora/diaspora/issues/6559#issuecomment-314927738

app/controllers/conversations_controller.rb
```diff
@@ -117,9 +117,9 @@ def new
 
   def contacts_data
     current_user.contacts.mutual.joins(person: :profile)
-      .pluck(*%w(contacts.id profiles.first_name profiles.last_name people.diaspora_handle))
-      .map {|contact_id, *name_attrs|
-        {value: contact_id, name: ERB::Util.h(Person.name_from_attrs(*name_attrs)) }
-      }
+                .pluck(*%w(contacts.id profiles.full_name people.diaspora_handle))
```
diaspora-code-review: - Pass array contents as separate arguments.
- `%w`-literals should be delimited by `[` and `]`.

app/controllers/conversations_controller.rb
```diff
@@ -117,9 +117,9 @@ def new
...
+                .map {|contact_id, *name_attrs|
+                  {value: contact_id, name: ERB::Util.h(Person.name_from_attrs(*name_attrs)) }
```
diaspora-code-review: Space inside } detected.

app/models/person.rb
```diff
@@ -237,11 +237,15 @@ def name(opts = {})
     if self.profile.nil?
       fix_profile
     end
-    @name ||= Person.name_from_attrs(self.profile.first_name, self.profile.last_name, self.diaspora_handle)
+    @name ||= Person.name_from_attrs(self.profile.full_name, self.diaspora_handle)
```
diaspora-code-review: Redundant `self` detected.

app/views/notifier/invite.markerb
Flaburgan: I just added the full_name method but hey, here comes a new fallback: first_name falls back to the part before the @ of the diaspora id when first_name is empty :p I'll correct that.

app/controllers/conversations_controller.rb
SuperTux88: Careful: `profile.full_name` currently contains the full name in lowercase, and can't be used for display, it's only used for search.

app/views/notifier/invite.markerb
```diff
@@ -1 +1,8 @@
-<%= t('.message', :invite_url => invite_code_url(@invitation_code), :diasporafoundation_url => 'https://diasporafoundation.org/', :diaspora_id => @inviter.try(:diaspora_handle)) %>
+<%= t('.message',
+  invite_url: invite_code_url(@invitation_code),
+  diasporafoundation_url: 'https://diasporafoundation.org/',
+  user: @inviter.try(:full_name).empty? ?
+    @inviter.try(:diaspora_handle) :
+    "#{@inviter.full_name} (#{@inviter.diaspora_handle})",
```
SuperTux88: Use `#{@inviter.name}` here, `full_name` is lowercase.

app/models/person.rb
```diff
@@ -237,11 +237,15 @@ def name(opts = {})
...
   end
 
-  def self.name_from_attrs(first_name, last_name, diaspora_handle)
-    first_name.blank? && last_name.blank? ? diaspora_handle : "#{first_name.to_s.strip} #{last_name.to_s.strip}".strip
+  def full_name
```
SuperTux88: You created this method in the `Person` model, while there already is a `full_name` attribute in the `Profile` model. I think this is really confusing, because you confused yourself and used `profile.full_name` where you probably wanted to use `person.full_name`.

app/models/person.rb
```diff
@@ -237,11 +237,15 @@ def name(opts = {})
...
+    "#{first_name.to_s.strip} #{last_name.to_s.strip}".strip
+  end
+
+  def self.name_from_attrs(full_name, diaspora_handle)
+    full_name.blank? ? diaspora_handle : full_name
```
SuperTux88: That would execute `full_name` twice (when there is a name).

SuperTux88 (changes requested):
Again, I would just use `full_name` now for the check, and don't add a second `full_name` which is more confusing than it helps. It would be removed again by the refactoring for #3648 anyway.

So just use `@inviter.try(:full_name).empty?` for the check if there is a name (which you already do in the latest diff) and `#{@inviter.name}` for now (instead of `#{@inviter.full_name}`).

Everything else should be refactored when fixing #3648

app/models/person.rb
Flaburgan: I didn't see the full_name method in Profile, sorry. `@inviter.try(:full_name)` returns nil in `invite.markerb` so I thought you wanted me to create the full_name method. I'll revert this commit, but I suppose I have something more to do to be able to access `full_name` in the e-mail body.

SuperTux88: Oh, is `@inviter` the person (sorry, didn't check that)? Then it would be `@inviter.profile.full_name.empty?` (or you can delegate it in the person model).

SuperTux88: Delegate is probably better (https://github.com/diaspora/diaspora/blob/develop/app/models/person.rb#L25), because you would need `@invite.try(:profile).try(:full_name).empty?` otherwise.

SuperTux88 (dismissed):

SuperTux88:
@Flaburgan about some tests, you can add your tests here: https://github.com/diaspora/diaspora/blob/develop/spec/mailers/notifier_spec.rb

app/views/notifier/invite.markerb
```diff
@@ -1 +1,6 @@
-<%= t('.message', :invite_url => invite_code_url(@invitation_code), :diasporafoundation_url => 'https://diasporafoundation.org/', :diaspora_id => @inviter.try(:diaspora_handle)) %>
+<%= t('.message',
+  invite_url: invite_code_url(@invitation_code),
+  diasporafoundation_url: 'https://diasporafoundation.org/',
+  user: @inviter.try(:full_name).empty? ? @inviter.try(:diaspora_handle) : "#{@inviter.name} (#{@inviter.diaspora_handle})",
```
dantetekanem: If Diaspora supports 2.3.x, why not use the safe operator `@inviter&.full_name.empty?` instead?

SuperTux88: Because we currently still support 2.1. We probably change this soon (for the next major version), but `try()` works just as good here.

dantetekanem: @SuperTux88 awesome :) Thanks for clarify.

Flaburgan:
Alright I added a quick test only to see if the first and last name are in the emails.

features/step_definitions/user_steps.rb
```diff
@@ -168,10 +168,16 @@
   ActionMailer::Base.deliveries.length.should == n.to_i
 end
 
+Then /^I should see "([^\"]*)" in the last sent email$/ do |text|
```
SuperTux88: While this check doesn't hurt, I think it's better to use a regex for `( not)?` and combine it with the existing check below, see [here](https://github.com/diaspora/diaspora/blob/33ad411bbd0d48652d161a60627657f75d467292/features/step_definitions/notifications_steps.rb#L9-L18) for an example with a regex for "not".

Flaburgan:
Alright, remark fixed and rspec test written ;)

SuperTux88 (approved):
❤️

PR #7486, opened 2017-07-12
Fix failing mention modal test

Flaburgan:
Let's see if it changes something.

### DIFF
--- a/features/step_definitions/custom_web_steps.rb
+++ b/features/step_definitions/custom_web_steps.rb
@@ -78,6 +78,7 @@ end
 
 And /^I want to mention (?:him|her) from the profile$/ do
   find('#mention_button').click
+  expect(find("#publisher")).to be_visible
   within('#mentionModal') do
     click_publisher
   end


SuperTux88:
Still fails :(

And `And I want to mention her from the profile` isn't the step that fails, it's `And I append "I am eating a yogurt" to the publisher in the mention modal` with

`expected to find field "status_message[text]" with value "@{alice@localhost:9887} I am eating a yogurt" but there were no matches. Also found "", which matched the selector but not all filters. (RSpec::Expectations::ExpectationNotMetError)`

fwoelm:
I think I might have a fix for you. Let me confirm and I'll report back tomorrow :)

fwoelm:
Good news: I've fixed the (falsely) failing features concerning mentioning.

Bad news: A bunch of other random failures remain and it's beyond me why. I have a feeling it has to do with poor performance of some of the feature tests. The error rate on Travis appears to be higher than the error rate on my own device, which *could* be due to lower CPU on Travis builds that cause some of the low Capybara methods to time out in Travis when they pass on my computer.

One approach worth testing would be to run all four (2x ruby, 2x db) cucumber tests 5 times on Travis (for a total of 20 builds) with `Capybara.max_wait_time` set to 10 and tracking which errors remain. Errors that disappear could *probably* be attributed to poor performance.

I'll create a PR for the mentioning features in a few minutes.

SuperTux88:
Superseeded by #7491

PR #7487, opened 2017-07-13
Fix: Check for asset precompilation when starting server

FinnWoelm:
I set up my public/assets folder to by symlinked (working with Capistrano) and the server script falsely reported that my assets were not compiled. Adding a trailing slash to public/assets fixes the problem.

Since having the trailing slash makes the check work for all cases (someone is using a regular directory or a symlink), I thought it would not hurt to make this minor update.

Adding [ci skip] since it seems redundant to run tests for this :smiley: 

---

**Edit**: Uhmmmm. so much for skipping the CI tests... :confused:  Guess, it should have gone in the commit, not the PR. Sorry! :man_shrugging: 

**Edit 2:** Rebased as #7488

### DIFF
--- a/script/server
+++ b/script/server
@@ -150,7 +150,7 @@ then
 fi
 
 # Check if assets are precompiled
-if [ "$RAILS_ENV" = "production" -a -z "$(find public/assets -maxdepth 1 -name 'main-*.js' -print -quit 2>/dev/null)" ]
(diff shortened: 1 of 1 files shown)


fwoelm:
I'm going to rebase and add [ci skip] to the commit this time

SuperTux88:
Thanks again for your contribution.

For the next time: You don't need to create a new PR, can just update the existing PR.

But in this case (when travis already ran), it doesn't make a difference if you add `[ci skip]` or not, because it won't run again.

PR #7488, opened 2017-07-13
Fix: Check for asset precompilation when starting server

FinnWoelm:
I set up my public/assets folder to by symlinked (working with Capistrano) and the server script falsely reported that my assets were not compiled. Adding a trailing slash to public/assets fixes the problem.

Since having the trailing slash makes the check work for all cases (someone is using a regular directory or a symlink), I thought it would not hurt to make this minor update.

Adding [ci skip] since it seems redundant to run tests for this :smiley:

### DIFF
--- a/script/server
+++ b/script/server
@@ -150,7 +150,7 @@ then
 fi
 
 # Check if assets are precompiled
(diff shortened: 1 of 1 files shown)


SuperTux88 (approved):
Thank you 👍

PR #7491, opened 2017-07-14
Fix: Use Capybara#fill_in for appending to publisher

FinnWoelm:
Capybara's `native#send_key` function is slow when it is passed a string longer than just a few characters. This often results in timeout issues and Capybara (falsely) reporting features as failing.

To fix this, we use the faster function` #fill_in`. This does not trigger JavaScript events on the input, so we manually trigger them after `#fill_in` by just sending a single key. This can be any key but since we
do not want to modify the text in the input, non-text keys should be used. For a list of non-text keys, see
http://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FElement%3Asend_keys

There is an alternative to the above:
1. Use `#fill_in` to enter all text except for the last character:  
      
       fill_in ..., with: "#{status_message_text} #{txt[0..-2]}"
      
1. And then use `#send_key` to send that last character:

       find("#status_message_text").native.send_key(txt.last)

At the moment, both approaches work equally well. The second approach is documented here just in case it becomes relevant in the future.

---

Okay, Travis, don't disappoint me now! :grin:

### DIFF
--- a/features/support/publishing_cuke_helpers.rb
+++ b/features/support/publishing_cuke_helpers.rb
@@ -5,8 +5,9 @@ module PublishingCukeHelpers
 
   def append_to_publisher(txt)
     status_message_text = find("#status_message_text").value
-    find("#status_message_text").native.send_key(" #{txt}")
-    expect(page).to have_field("status_message[text]", with: "#{status_message_text} #{txt}")
+    fill_in id: "status_message_text", with: "#{status_message_text} #{txt}"
+    # trigger JavaScript event listeners
+    find("#status_message_text").native.send_key(:end)
   end
 
   def upload_file_with_publisher(path)


fwoelm:
Builds are still failing but they are not related to mentioning.

I see a lot of the same failures re-occurring. That gives me some hope. Here are the failures from this CI test -- I've encountered them in the CI tests on my fork, too:

- **features/desktop/comments.feature:45**

      When I make a show page comment "I think that’s a cat"        # features/step_definitions/comment_steps.rb:14

      Unable to find button "Comment" (Capybara::ElementNotFound)

- **features/desktop/signs_up.feature:56**

      And I submit the publisher                                                       # features/step_definitions/posts_steps.rb:84

      expected "mention_popup publisher row" to include "closed" (RSpec::Expectations::ExpectationNotMetError)

- **features/desktop/notifications.feature:113**

     Then I should see a ".read .unread-toggle"                                  # features/step_definitions/custom_web_steps.rb:177

      One or more errors were raised in the Javascript code on the page. If you don't care about these errors, you can ignore them by setting js_errors: false in your Poltergeist configuration (see documentation for details).

      TypeError: null is not an object (evaluating 'that.$element.off')

SuperTux88:
I triggered the cucumber jobs on travis a second time.

fwoelm:
I've also run them about three times on Travis (via my fork) and five times locally. No more failed mentions :)

SuperTux88:
Yes, looks much better 👍 

This time only one failed cuke:

```
  Scenario: Application is authorized                                                  # features/desktop/oidc_auth_code_flow.feature:19
...
    And I parse the auth code and create a request to the token endpoint               # features/step_definitions/auth_code_steps.rb:22
      expected `nil.present?` to return true, got false (RSpec::Expectations::ExpectationNotMetError)
```

fwoelm:
I also ran the specific feature 30 times (using a ruby script):

    cucumber features/desktop/mentions_from_profile_page.feature

**Before fix:** (using native.send_key)
Total Time: 1001.7561831319908
Successes: 21
Failures:  9

**After fix:** (combining fill_in and native.send_key)
Total Time: 719.7683899190015
Successes: 30
Failures:  0

[Here is the script I used](https://gist.github.com/fwoelm/e6f0a65d5e0f580fbba41437aa640181).

fwoelm:
Any insights on those remaining failures? False failures are frustrating me... :[

SuperTux88:
> Any insights on those remaining failures? False failures are frustrating me... :[

Not yet, but we have #7373 to track them (but I think there are also some new since the list was created, I'll try to update the list later). And yes, that's really frustrating, so if you have any idea about how to debug and fix them, you are more than welcome to help here.

SuperTux88 (approved):
Looks good, thanks for fixing that! ❤️

Flaburgan:
Thank you for dealing with that!

PR #7495, opened 2017-07-17
Issue #6559: Changed default mail FROM header and tests

Rete2:
Hey
so here's our first pull request. We had some remarks and questions regarding our solution. Here we go:

Code:
- We changed the default_headers method in the `app/mailers/notification_mailers/base.rb` file to defer between the case where sender has a First and/or Last name and the case where the sender has no First and Last name, but just a diaspora handle.

We have decided the following about the headers:
- Sender present/Firsname Lastname exists: We opted for the `Diaspora* (Firstname Lastname)`. "On behalf of" seemed quite formal to us. But of course we are open to changing it again.
- Sender present/Firstname Lastname doesn’t exist: `Diaspora*`.
We chose the above mentioned solutions as they are short, clear and does. 
- No sender present: We decided to keep this case without `Diaspora*` so that you see directly which email it comes from (is this in most cases a no-reply email?). We address this option even though the issue doesn’t refer to that case.

Other options, we discussed were:
a. when Firstname and/or Lastname is present: 
	`Diaspora* (from Firstname Lastname) <no-reply@example.org>`, 
	`Diaspora* Notifications (from Firstname Lastname) <no-reply@example.org`>, 
	`Diaspora* (on behalf of Firstname Lastname) <no-reply@example.org>`.

b. when Firstname and/or Lastname is missing: 
	`<no-reply@example.org>`, 
	`Diaspora* Notifications <no-reply@example.org>`.

c. when no sender is present:
	`Diaspora* <no-reply@example.org>`

We need to keep in mind that any additional string in the header will require translations. 

Something else we encountered regarding the default sender_address: if the pod admin configured an email address in the `defaults.yml`, a reply will go that email. This can be quite confusing because the user will think they write the sender but actually it goes to the admin. (could be a privacy issue too)

Tests:
- we changed the already existing tests regarding the FROM header in the `spec/notifier_spec.rb`
- we wrote tests for the case when there is no sender: 
	- reports (`report_spec.rb` line 44)
	- confirm_email (`spec/notifier_spec.rb` line 481)
	- csrf_token_fail (`spec/notifier_spec.rb` line 502)
- we wrote a test for the case when there is a sender but no First and/or Last name (`notifier_spec.rb `line 547). We placed the test for now in the describe 'base' block at the bottom of the file. We were wondering whether we should write this test for all notification types or keep it at the bottom in the describe 'base' block?

Another tests issue:
We realized that not for all notification types there are tests regarding the FROM: header. Do they need to be added in each describe “notification type” block. Now it seems not to be a bit inconsitent. (The following notification types have so far no test regarding FROM: reshared, liked, mentioned, mentioned_in_comment, started_sharing). We didn’t want to include this in this pull request, because we feel it’s beyond the scope of the 6559 issue. Should we create a new issue to discuss this?

As this is our first contribution, we hope we didn't break any conventions. Let us know, we're happy to learn. :)

### DIFF
--- a/app/mailers/notification_mailers/base.rb
+++ b/app/mailers/notification_mailers/base.rb
@@ -38,8 +38,9 @@ module NotificationMailers
         host: "#{AppConfig.pod_uri.host}",
         to:   name_and_address(@recipient.name, @recipient.email)
       }
-
-      headers[:from] = "\"#{@sender.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>" if @sender.present?
+      return headers if @sender.blank?
+      sender_in_header = @sender.profile.full_name.empty? ? @sender.username : @sender.name
+      headers[:from] = "\"Diaspora* (#{sender_in_header})\" <#{AppConfig.mail.sender_address}>"
 
       headers
     end
--- a/app/mailers/notification_mailers/base.rb
+++ b/app/mailers/notification_mailers/base.rb
@@ -34,13 +34,13 @@ module NotificationMailers
 
     def default_headers
       headers = {
-        from: AppConfig.mail.sender_address.get,
+        from: "\"#{AppConfig.settings.pod_name}\" <#{AppConfig.mail.sender_address}>",
         host: "#{AppConfig.pod_uri.host}",
         to:   name_and_address(@recipient.name, @recipient.email)
       }
       return headers if @sender.blank?
       sender_in_header = @sender.profile.full_name.empty? ? @sender.username : @sender.name
-      headers[:from] = "\"Diaspora* (#{sender_in_header})\" <#{AppConfig.mail.sender_address}>"
+      headers[:from] = "\"#{AppConfig.settings.pod_name} (#{sender_in_header})\" <#{AppConfig.mail.sender_address}>"
 
       headers
     end
--- a/app/mailers/report_mailer.rb
+++ b/app/mailers/report_mailer.rb
@@ -1,5 +1,5 @@
 class ReportMailer < ActionMailer::Base
-  default from: AppConfig.mail.sender_address
+  default from: "\"#{AppConfig.settings.pod_name}\" <#{AppConfig.mail.sender_address}>"
 
   def self.new_report(report_id)
     report = Report.find_by_id(report_id)
--- a/spec/mailers/notifier_spec.rb
+++ b/spec/mailers/notifier_spec.rb
@@ -246,7 +246,7 @@ describe Notifier, type: :mailer do
     end
 
     it "FROM: contains the sender's name" do
-      expect(@mail["From"].to_s).to eq("\"#{@cnv.author.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>")
+      expect(@mail["From"].to_s).to eq("\"Diaspora* (#{@cnv.author.name})\" <#{AppConfig.mail.sender_address}>")
     end
 
     it "should use a generic subject" do
@@ -290,7 +290,7 @@ describe Notifier, type: :mailer do
       end
 
       it "FROM: contains the sender's name" do
-        expect(comment_mail["From"].to_s).to eq("\"#{eve.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>")
+        expect(comment_mail["From"].to_s).to eq("\"Diaspora* (#{eve.name})\" <#{AppConfig.mail.sender_address}>")
       end
 
       it "SUBJECT: has a snippet of the post contents, without markdown and without newlines" do
@@ -331,7 +331,7 @@ describe Notifier, type: :mailer do
       end
 
       it "FROM: has the name of person commenting as the sender" do
-        expect(comment_mail["From"].to_s).to eq("\"#{eve.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>")
+        expect(comment_mail["From"].to_s).to eq("\"Diaspora* (#{eve.name})\" <#{AppConfig.mail.sender_address}>")
       end
 
       it "SUBJECT: has a snippet of the post contents, without markdown and without newlines" do
@@ -386,7 +386,7 @@ describe Notifier, type: :mailer do
         end
 
         it "FROM: contains the sender's name" do
-          expect(mail["From"].to_s).to eq("\"#{bob.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>")
+          expect(mail["From"].to_s).to eq("\"Diaspora* (#{bob.name})\" <#{AppConfig.mail.sender_address}>")
         end
 
         it "SUBJECT: does not show the limited post" do
@@ -411,7 +411,7 @@ describe Notifier, type: :mailer do
         end
 
         it "FROM: contains the sender's name" do
-          expect(mail["From"].to_s).to eq("\"#{bob.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>")
+          expect(mail["From"].to_s).to eq("\"Diaspora* (#{bob.name})\" <#{AppConfig.mail.sender_address}>")
         end
 
         it "SUBJECT: does not show the limited post" do
@@ -442,7 +442,7 @@ describe Notifier, type: :mailer do
       end
 
       it "FROM: contains the sender's name" do
-        expect(mail["From"].to_s).to eq("\"#{bob.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>")
+        expect(mail["From"].to_s).to eq("\"Diaspora* (#{bob.name})\" <#{AppConfig.mail.sender_address}>")
       end
 
       it "SUBJECT: does not show the limited post" do
@@ -478,7 +478,11 @@ describe Notifier, type: :mailer do
       expect(@confirm_email.to).to eq([bob.unconfirmed_email])
     end
 
-    it "has the unconfirmed emil in the subject" do
+    it "FROM: header should be the default sender address" do
+      expect(@confirm_email["From"].to_s).to eq(AppConfig.mail.sender_address.to_s)
+    end
+
+    it "has the unconfirmed email in the subject" do
       expect(@confirm_email.subject).to include(bob.unconfirmed_email)
     end
 
@@ -500,6 +504,10 @@ describe Notifier, type: :mailer do
 
     it "goes to the right person" do
       expect(email.to).to eq([alice.email])
+    end
+
+    it "FROM: header should be the default sender address" do
+      expect(email["From"].to_s).to eq(AppConfig.mail.sender_address.to_s)
     end
 
     it "has the correct subject" do
@@ -534,6 +542,12 @@ describe Notifier, type: :mailer do
       expect {
         Notifier.send_notification("started_sharing", bob.id, person.id)
       }.to_not raise_error
+    end
+
+    it "FROM: header should be 'Diaspora* (username)' when there is no first and last name" do
+      bob.person.profile.update_attributes(first_name: "", last_name: "")
+      mail = Notifier.send_notification("started_sharing", alice.id, bob.person.id)
+      expect(mail["From"].to_s).to eq("\"Diaspora* (#{bob.person.username})\" <#{AppConfig.mail.sender_address}>")
     end
   end
 end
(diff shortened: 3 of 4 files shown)


app/mailers/notification_mailers/base.rb
```diff
@@ -39,8 +39,13 @@ def default_headers
         to:   name_and_address(@recipient.name, @recipient.email)
       }
 
-      headers[:from] = "\"#{@sender.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>" if @sender.present?
-
+      if @sender.present?
+        if @sender.profile.full_name.empty?
```
diaspora-code-review: Use the return of the conditional for variable assignment and comparison.

spec/mailers/notifier_spec.rb
```diff
@@ -478,7 +478,11 @@
       expect(@confirm_email.to).to eq([bob.unconfirmed_email])
     end
 
-    it "has the unconfirmed emil in the subject" do
+    it "FROM: header should be the default sender address" do
+      expect(@confirm_email["From"].to_s).to eq("#{AppConfig.mail.sender_address}")
```
diaspora-code-review: Prefer `to_s` over string interpolation.

spec/mailers/notifier_spec.rb
```diff
@@ -502,6 +506,10 @@
       expect(email.to).to eq([alice.email])
     end
 
+    it "FROM: header should be the default sender address" do
+      expect(email["From"].to_s).to eq("#{AppConfig.mail.sender_address}")
```
diaspora-code-review: Prefer `to_s` over string interpolation.

spec/mailers/notifier_spec.rb
```diff
@@ -535,5 +543,12 @@
         Notifier.send_notification("started_sharing", bob.id, person.id)
       }.to_not raise_error
     end
+
+    it "FROM: header should be 'Diaspora*' when there is no first and last name" do
+      bob.person.profile.update_attributes(:first_name => "", :last_name => "")
```
diaspora-code-review: Use the new Ruby 1.9 hash syntax.

spec/mailers/notifier_spec.rb
```diff
@@ -535,5 +543,12 @@
...
+      mail = Notifier.send_notification("started_sharing", alice.id, bob.person.id)
+      expect(mail["From"].to_s).to eq("Diaspora* <#{AppConfig.mail.sender_address}>")
+    end
+
```
diaspora-code-review: Extra empty line detected at block body end.

spec/mailers/report_spec.rb
```diff
@@ -41,6 +41,11 @@
       expect(ActionMailer::Base.deliveries[1].to[0]).to include(@user2.email)
     end
 
+    it "FROM: header should be the default sender address" do
+      ReportMailer.new_report(@post_report.id).each(&:deliver_now)
+      expect(ReportMailer.default[:from].to_s).to eq("#{AppConfig.mail.sender_address}")
```
diaspora-code-review: Prefer `to_s` over string interpolation.

Flaburgan:
>We need to keep in mind that any additional string in the header will require translations.

That's fine, translations shoudn't stop you to improve user experience.

>if the pod admin configured an email address in the defaults.yml, a reply will go that email. This can be quite confusing because the user will think they write the sender but actually it goes to the admin. (could be a privacy issue too)

Don't worry about privacy here, the podmin already has access to every pod users data. However I'm surprised that the no-reply address can be changed, are you sure about this? This doesn't look like the expected behavior. We use the podmin emails in the terms and in the left side column "Got a problem", but only there as far as I know.

Flaburgan:
Thank you for your work and welcome aboard!

I have a suggestion: I think we should use the pod name instead of a generic "Diaspora*". It would be useful if a user has an account on different pods, and some users doesn't even know they are using diaspora*.

SuperTux88:
+1 for using `AppConfig.settings.pod_name` instead of "Diaspora*". And I would also change the case when there is no sender, to `"#{AppConfig.settings.pod_name} <no-reply@example.org>"`.

And I think we should still fallback to the diaspora ID when there is no real name. Because I sometimes use my mailclient to search a notification from someone for a post, and then it's useful when there is something in the from field. What do others think about this? Do we really need to remove the ID from the from field (because it looks like an email address)? Or would `pod* (on behalf of username@example.org) <no-reply@example.org>` be OK?

Also: As @Flaburgan already wrote: Translation is no problem, it only means you don't hardcode the string. You extract it to the english translation file (so others can translate it).

@Flaburgan You can (and must) set the sender address [here](https://github.com/diaspora/diaspora/blob/7c37a625934430fa60c8b620673a9c53b74b2b91/config/diaspora.yml.example#L623-L624), because the mail-server you use, must be allowed to send mails from that address. Also using hardcoded `no-reply@something` doesn't really add any advantages. Because when you reply to a `no-reply` address, a podmin can still receive it. And a podmin could also reject mails sent to `notifications@something`.

netagonen:
@Flaburgan and @SuperTux88  - thanks for the reply. 
1. We are not using diaspora handles because this is what the issue is about. It is specifically suggested in the issue: 

> Never use Diaspora* identifiers in an e-mail address description. Using nothing is less disturbing

For reasoning you can take a look at [the issue](https://github.com/diaspora/diaspora/issues/6559), but basically it is too confusing for non-expert diaspora users when the FROM  header looks like an email.

2. > And I think we should still fallback to the diaspora ID when there is no real name. 

In order to avoid confusion but still have some identification parameters, we could use diaspora username instead of the whole handle.      

3. > "I think we should use the pod name instead of a generic "Diaspora*". It would be useful if a user has an account on different pods, and some users doesn't even know they are using diaspora*".

 From my personal experience as a new user, I didn't remember my pod name for a while, so it would still be confusing to get an email where just the pod name is in the header.

Rete2:
Thanks for your reply also from my side. Neta already wrote on the main discussion points.
Just one more thing about the sender_address in the diaspora.yml line 624: 
```
## Sender address used in mail sent by Diaspora.
    #sender_address: 'no-reply@example.org'
```

> Don't worry about privacy here, the podmin already has access to every pod users data. 

The privacy issue we were thinking of was more, that someone gets a notification from someone on diaspora and the person (someone not familiar with diaspora and what notifications are) thinks they can reply to the email and if it doesn't show the email address it is sending to and the podmin has set an actual email address, they will receive a personal email that this person has written thinking it goes to the user.

Regarding the issue with no First and Lastname: I really like the idea of using the username instead.

SuperTux88:
> In order to avoid confusion but still have some identification parameters, we could use diaspora username instead of the whole handle.

Yes, that would be enough to differentiate notifications from different users who haven't set a name.

> I didn't remember my pod name for a while, so it would still be confusing to get an email where just the pod name is in the header.

I don't think that's a problem, at least you start remembering the pod name better.

> they will receive a personal email that this person has written thinking it goes to the user.

I already received some of these emails (but not many, it's more a once or twice a year thing), and at least I could then answer these users and tell them, that they need to answer on diaspora, so the answer isn't lost and the user don't know why they don't get a reply.

But this email address must match the outgoing mailserver the podmin uses, so it must be settable. And then it's the decision of the podmin what to set there.

app/mailers/notification_mailers/base.rb
```diff
@@ -38,9 +38,12 @@ def default_headers
         host: "#{AppConfig.pod_uri.host}",
         to:   name_and_address(@recipient.name, @recipient.email)
       }
-
-      headers[:from] = "\"#{@sender.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>" if @sender.present?
-
+      return headers unless @sender.present?
```
diaspora-code-review: Use `if @sender.blank?` instead of `unless @sender.present?`.

app/mailers/notification_mailers/base.rb
```diff
@@ -38,9 +38,12 @@ def default_headers
...
+      headers[:from] = if @sender.profile.full_name.empty?
+        "\"Diaspora* (#{@sender.username})\" <#{AppConfig.mail.sender_address}>"
```
diaspora-code-review: Use 2 (not -15) spaces for indentation.

app/mailers/notification_mailers/base.rb
```diff
@@ -38,9 +38,12 @@ def default_headers
...
+      else
```
diaspora-code-review: Align `else` with `if`.

app/mailers/notification_mailers/base.rb
```diff
@@ -38,9 +38,12 @@ def default_headers
...
+        "\"Diaspora* (#{@sender.name})\" <#{AppConfig.mail.sender_address}>"
+      end
```
diaspora-code-review: `end` at 46, 6 is not aligned with `if` at 42, 23.

SuperTux88:
Did you know you can run codestyle checks locally before pushing? See https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review

spec/mailers/notifier_spec.rb
```diff
@@ -478,7 +478,11 @@
       expect(@confirm_email.to).to eq([bob.unconfirmed_email])
     end
 
-    it "has the unconfirmed emil in the subject" do
+    it "FROM: header should be the default sender address" do
+      expect(@confirm_email["From"].to_s).to eq(AppConfig.mail.sender_address.to_s)
```
SuperTux88: You don't need the `.to_s` the second time:
```ruby
expect(@confirm_email["From"].to_s).to eq(AppConfig.mail.sender_address)
```

netagonen: We did this because the checks before "complained" about this. but we could change it if you prefer

Rete2: Actually now we changed it and removed the `.to_s`. Sorry for the confusion...

spec/mailers/notifier_spec.rb
```diff
@@ -246,7 +246,8 @@
     end
 
     it "FROM: contains the sender's name" do
-      expect(@mail["From"].to_s).to eq("\"#{@cnv.author.name} (diaspora*)\" <#{AppConfig.mail.sender_address}>")
+      pod_name = AppConfig.settings.pod_name
```
SuperTux88: You can write `let(:pod_name) { AppConfig.settings.pod_name }` once at the top of this file (below the `let(:person) ...` line), and then you don't need this 6 times (in every test).

SuperTux88:
Can you please change the default from header when there is no sender to `"\"#{AppConfig.settings.pod_name}\" <#{AppConfig.mail.sender_address}>"`. Everything else looks good to me.

app/mailers/report_mailer.rb
```diff
@@ -1,5 +1,5 @@
 class ReportMailer < ActionMailer::Base
-  default from: AppConfig.mail.sender_address
+  default from: "#{AppConfig.settings.pod_name} <#{AppConfig.mail.sender_address}>"
```
SuperTux88: Here are the quotes missing, should be `"\"#{AppConfig.settings.pod_name}\" <#{AppConfig.mail.sender_address}>"`.

SuperTux88 (changes requested):
Only found one thing, looks good otherwise. Can you please squash some commits (961eecb + 98d5c17 + b043d79 and a3849d3 + a482904 + 481fe2d and also squash the fix of my comment to `Add pod name to FROM: header for no sender case`), then it should be ready to merge.

SuperTux88:
Merged, thanks @Rete2 and @netagonen 🎉 

Two small things for future PRs:
* When squashing, please remove something like `fix comments on pull request` from the commit message. That isn't useful when reading the commit message in the future. So please write the commit message so, that somebody reading them in the future knows what's done here and maybe also why. (I edited the commit messages for this PR for you, I hope that's OK)
* Please write a small comment when you think you are ready for another review after fixing requested changes. Even when I receive notifications when you push new commits (not when you squash or edit existing commits), I don't know if you are finished or if you are still working on something and there maybe comes another commit. So a little ping can help speeding things up.

And thanks again for working on diaspora ❤️

Flaburgan:
Congratulation for your first PR merged! 🎉

netagonen:
Hey @SuperTux88. Thanks for your answer and for the advice regarding our contributions. We're always happy to learn, especially as we are new to this whole process as you could tell :) We will definitely try from now on to follow those conventions.  
Regarding the first point: of course it's ok that you edited the message so it makes more sense for the future. 
Obviously, we're happy that you merged it and starting to look at the next issue

netagonen:
Thnks @Flaburgan :)

PR #7501, opened 2017-07-23
Send public profiles publicly

### LINKED ISSUE #3164: When a user with a lot of contacts updates their profile, bad things happen...
Whenever we update DiasporaHQ, the server has a heart attack. Max says there's probably something really dumb in our profile update action. Help please.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/362-when-a-user-with-a-lot-of-contacts-updates-their-profile-bad-things-happen?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SuperTux88:
This will help big public profiles (for example hq@pod.diaspora.software), and it's public anyway, so no need to encrypt them.

fixes #3164

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -40,6 +40,10 @@ class Profile < ActiveRecord::Base
     Person.joins(:contacts).where(contacts: {user_id: person.owner_id})
   end
 
+  def public?
+    public_details?
+  end
+
   def diaspora_handle
     #get the parent diaspora handle, unless we want to access a profile without a person
     (self.person) ? self.person.diaspora_handle : self[:diaspora_handle]
--- a/lib/diaspora/federation/entities.rb
+++ b/lib/diaspora/federation/entities.rb
@@ -160,7 +160,8 @@ module Diaspora
           location:         profile.location,
           searchable:       profile.searchable,
           nsfw:             profile.nsfw,
-          tag_string:       profile.tag_string
+          tag_string:       profile.tag_string,
+          public:           profile.public_details
         )
       end
 
--- a/lib/diaspora/federation/receive.rb
+++ b/lib/diaspora/federation/receive.rb
@@ -122,7 +122,8 @@ module Diaspora
             location:         entity.location,
             searchable:       entity.searchable,
             nsfw:             entity.nsfw,
-            tag_string:       entity.tag_string
+            tag_string:       entity.tag_string,
+            public_details:   entity.public
           )
         end
       end
--- a/spec/lib/diaspora/federation/entities_spec.rb
+++ b/spec/lib/diaspora/federation/entities_spec.rb
@@ -188,6 +188,7 @@ describe Diaspora::Federation::Entities do
       expect(federation_entity.searchable).to eq(diaspora_entity.searchable)
       expect(federation_entity.nsfw).to eq(diaspora_entity.nsfw)
       expect(federation_entity.tag_string.split(" ")).to match_array(diaspora_entity.tag_string.split(" "))
+      expect(federation_entity.public).to eq(diaspora_entity.public_details)
     end
 
     it "builds a reshare" do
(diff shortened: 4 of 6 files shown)


skyschub (approved):
Approved as soon as travis is green.

SuperTux88:
Travis is greenish, merged. Thanks for the review @denschub 🍪

Flaburgan:
Can you please explain what you did here exactly? I guess previously the public profile was sent to each contact and is now sent once per pod? But why isn't there any old code removed then? (Just trying to understand).

denschub:
> But why isn't there any old code removed then?

Because private profiles still have to be sent like private posts.

denschub:
I should add more context.

The entire send logic is shared between all models anyway, this is not related to only profile updates and not related to this PR, we check if `public?` is true when sending entities, that's all. This PR merely adds the possibility for profile updates to be handled like public posts, for example.

To answer your question: Yes, this does boil down distributing public profiles only once per pod, but that's independent of what we actually federate. As soon as a entity is public, it gets handled by [`Diaspora::Federation::Dispatcher::Public`](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/federation/dispatcher/public.rb), which indeed does make sure we only post public stuff only once per pod. However, this is not the interesting part here.

The actual interesting change here is the way we create the list of recipients. If Alice shares a private entity, we generate our target list based on the people Alice has in her aspects. If Alice shares a public entity, however, we generate the target list based on the list of people who have Alice in their aspects. This means a pod receives Alice's public entities even if she does not share with anyone from that pod, as long as there is someone on that pod sharing with her.

SuperTux88:
@denschub thanks for the explanation. But the last point isn't completely true. We always send the profile [to all contacts](https://github.com/diaspora/diaspora/blob/f9476409c8d8627b91d58dc036285d0c1fee8f7f/app/models/profile.rb#L39-L41). But I didn't change anything here, because we already have the problem of outdated profiles, I don't want to make that problem worse.

I thought about that problem many times and already know how I want to fix that:

* public details + searchable:
  * send to all known pods publicly
* public details + not searchable:
  * send to all contacts publicly
* private details + searchable:
  * send to all contacts the person shares with privately
  * send public profile (without private parts) to all other known pods publicly
* private details + not searchable:
  * send to all contacts the person shares with privately
  * send public profile to all other contacts publicly

That solves the problem of outdated profiles and also solves the problem, that we leak some private information. The only problem is, that I don't have a clean way to implement that yet (because we need to send the same object twice in different ways with different content to different recipients), but I'll probably give this another try soon (I want to have this in 0.7).

denschub:
oh, yeah, I missed that `subscribers` override. good catch!

zotlabs:
Could you please clarify which details are public and which are private? I do not see any distinction made in the protocol documentation. We're currently only sending profile messages privately so there is no leakage, but in order to mirror this ability I sort of need to know which fields are off limits in a public context.

SuperTux88:
The name, avatar and tags are the base profile, and are always public. `searchable` and `nsfw` are metadata and also "public" but not directly visible to the user.

In diaspora the users can choose if their extended profile (`bio`, `birthday`, `gender` and `location` is public or private, that's what the `public` flag indicates. If it's `true` all information are visibile to all users on the receiving pod, if it's `false` the extended profile is only visible to the contacts. If everything is public, you can send the profile once per pod to the public endpoint (that's what this PR does).

I think you're coming from diaspora/diaspora_federation#79? Once this is merged and updated in diaspora you can also send profiles with the `public` flag set to `false` via the public route (if you don't have contacts on the target pod), if you only send the base profile parts. That is to fix the problem of outdated profiles (wrong names and avatar) on pods which once fetched the profile via WebFinger but aren't connected to the person, so they currently don't receive updates.

And thanks for the question, I'll add information about base and extended profile to the documentation.

PR #7502, opened 2017-07-23
Cache assets-cache for cucumber on travis

SuperTux88:
Let's see what travis thinks about this.

(Since [this](https://travis-ci.org/diaspora/diaspora/jobs/256488043#L1175) still fails, and I don't want to increase the global timeout any higher only for the first login, let's see if we can if we can fix this differently)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ cache:
   bundler: true
   directories:
     - app/assets/images
+    - tmp/cache/assets/test
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


SuperTux88:
OK, that finally seems to work. What do others think about this? It's a hack, but I think it's better than increasing the global timeout everywhere (and we don't know how big we need to make it, because 80 seconds still wasn't enough, see #7478).

I'm also not sure if I should keep the debug `puts` (because it's displayed under `Background:` of the first scenario, which looks weird), but I think it helps us monitoring travis. So I think we should at least keep it now and maybe remove it later (when everything works stable).

SuperTux88:
OK, I finally found the reason why it takes that long the first time. When I delete `tmp/cache/assets/test` locally, I can reproduce this. So let's add that folder to the cached directories on travis, then we shouldn't need the workaround hack.

SuperTux88:
Aaaaaaand it works \o/ 🎉

skyschub (approved):
❤️ 😻 🍪 🍰

SuperTux88:
Merged, thanks for the review @denschub 👍 

And also linking it with #7373.

PR #7504, opened 2017-07-25
Update jasmine

svbergerem:
A lot of tests were failing with the latest jasmine version. Some of them because of gon being undefined (solution: set it in `spec/javascripts/jasmine_helpers/SpecHelper.js`) and others because of `app.currentUser` being undefined (solution: initialize `app` again after the `app.js` tests in `spec/javascripts/app/app_spec.js`).


SuperTux88 (approved):
Thanks for fixing that and welcome back ❤️

PR #7510, opened 2017-07-31
Fix bookmarklet for sprockets 3

SuperTux88:
This fixes:
```
rake aborted!
NoMethodError: undefined method `[]' for nil:NilClass
/srv/diaspora/lib/bookmarklet_renderer.rb:9:in `source_name'
/srv/diaspora/lib/bookmarklet_renderer.rb:22:in `compile'
/srv/diaspora/lib/tasks/assets.rake:10:in `block (2 levels) in <top (required)>'
/srv/diaspora/lib/tasks/assets.rake:38:in `block (2 levels) in <top (required)>'
Tasks: TOP => assets:uglify_bookmarklet
```
in production environment.

`Rails.application.assets` is only available when `config.assets.compile` is true (which is false in production). So the old way with a separate rake task doesn't work in production. But we can get the filename of the precompiled file from `Rails.application.assets_manifest.assets`.

(The second commit is only a small refactoring after I found `Rails.application.assets_manifest.assets`)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -72,6 +72,7 @@ module Diaspora
       jquery_ujs.js
       main.js
       jsxc.js
+      bookmarklet.js
       mobile/bookmarklet.js
       mobile/mobile.js
       templates.js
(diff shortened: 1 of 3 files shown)


skyschub (approved):

SuperTux88:
Merged, thanks for the review @denschub 🍪

PR #7512, opened 2017-08-03
Cleanup .gitignore and add note about .well-known to the changelog

SuperTux88:
I noticed that some pods still have a static `/.well-known/host-meta` from the past. That wasn't a problem, because the old file had the same content as the generated version from the federation lib. But with the new federation-lib 0.2.0 the content changed, and with the old static file webfinger doesn't work anymore. So I added a note to the changelog, so people need to check if they have an old cached `public/.well-known` and if so, delete it.

I also cleaned up some other old unused lines from the `.gitignore`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,11 @@
 This release recommends using Ruby 2.4, while retaining Ruby 2.3 as an officially supported version.
 Ruby 2.1 is no longer officially supported.
 
+## Delete public/.well-known/
+
+Before upgrading, please check if your `public/` folder contains a hidden `.well-known/` folder.
+If so, please delete it since it will prevent the federation from working properly.
+
 ## Refactor
 
 * Make the mention syntax more flexible [#7305](https://github.com/diaspora/diaspora/pull/7305)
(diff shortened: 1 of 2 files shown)


skyschub (approved):

Changelog.md
```diff
@@ -5,6 +5,11 @@
 This release recommends using Ruby 2.4, while retaining Ruby 2.3 as an officially supported version.
 Ruby 2.1 is no longer officially supported.
 
+## Delete public/.well-known/
+
+Before upgrading, please check if your `publlic/` folder contains a hidden `.well-known` folder.
```
cmrd-senya: typo: "publ**l**ic"

skyschub (changes requested):
Oh noez! See senyas comment!

Changelog.md
denschub: And this, ladies and gentleman, is the reason why I should not approve sentences I wrote myself!

SuperTux88: Fixed, thanks 👍 (RubyMine has a spellchecker ... but since it's in `...` it didn't check it ;) )

skyschub (approved):

Flaburgan:
I do have a `.well-known` folder in `public/` but that's where I have the acme-challenge for Let's Encrypt. We should probably be more precise about this folder and what's in which could block the federation, to tell everyone to delete doesn't seem careful.

PR #7514, opened 2017-08-07
Upgrade to rails 5.1

SuperTux88:
This PR still depends on 2 unreleased gems for rails 5.1 compatibility.

- [x] benlangfeld/mobile-fu#65 (used `mobile_fu` for now)
- [x] jhass/js_image_paths#4
- [x] jhass/js_image_paths#5
- [x] Solution for `schema.rb` that works with postgres and mysql on travis (remove it)
- [x] Fix conflict between `bootstrap-sass` and `rails-assets-bootstrap` (The workaround with `_boostrap` works and is OK for now)
- [x] Fix mobile cukes on travis
- [x] Fix migrations (auto increment for ID missing or IDs bigints problem)

ToDos after merging:

- [x] File a follow-up to merge old migrations we no longer need.
- [x] Adjust installation docs to use `db:migrate` instead of `db:schema:load`


cmrd-senya:
Huge job! :+1:

SuperTux88:
And my workaround in [config/initializers/assets.rb](https://github.com/SuperTux88/diaspora/blob/33f06f97d61a5515bbdc6c4794514d8bafa5b6a1/config/initializers/assets.rb#L16-L23) doesn't work on travis (works locally in development mode).

I pushed another workaround-commit. When importing `_bootstrap` with underscore, it works ... but pronto doesn't like the underscore ...

 If anybody has an idea on how to solve this (in a cleaner way), I'm open for help ...

SuperTux88:
And the updated `schema.rb` doesn't work with postgres ... :/ Maybe we should switch to postgres and maybe the postgres-version of the schema.rb works with mysql?

Anyway ... I go to bed and try that at some other time.

SuperTux88:
Oh, and mobile cukes broke with timeouts ... that worked locally. Need to check what's the problem there. Maybe some more caches that are there locally but not on travis?

denschub:
All migrations are dead, since they need explicit Rails version information now:

```
Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class CreateSchema < ActiveRecord::Migration[4.2]
```

SuperTux88:
The problem with cucumber is weird, for example [here](https://travis-ci.org/diaspora/diaspora/jobs/261802337#L3940) ... everything worked, until one cuke failed and every following cuke failed with timeout!?

denschub:
Pushed some commits. I fixed the migrations to be working with Rails 5.1, making sure that you can run all migrations from an empty database.

As for the `schema.rb`, this is tricky. Basically, according to https://github.com/rails/rails/issues/26209, it is *not a bug* that the `schema.rb` is not compatible across databases, but expected behavior. While it's usually good to have it checked in, it makes no sense if you share the schema between databases.

So I decided to remove it. For initial setup, users have to run migrations, but I'll file a follow-up for merging old migrations so we can reduce the number a bit. Also, as a side effect, travis now tests our migrations across databases, so that's cool. We have to change our documentation for installing, but that's doable.

I'll fix any issues with the migrations coming up here (for example, MySQL just failed for some reason)

SuperTux88:
I reverted https://github.com/diaspora/diaspora/commit/d511f3cb1707341f068b119308549048fdaa84a8 and replaced it with https://github.com/diaspora/diaspora/commit/49df3e6f5b8a06d0b15cec7763b8ae2ca005df6a ... that fixed the ID bigint/auto-increment problem.

skyschub (approved):
Approval for all of @SuperTux88's commits, he will review mine!

SuperTux88:
I reviewed @denschub commits (thanks for the help ❤️) and approve the (can't approve my own PR)

I'll rebase this once again and use `mobile_fu` release for now. If they release a new version of `mobile-fu` before we release 0.7.0.0, we can switch that back.

denschub:
👍

PR #7516, opened 2017-08-08
Move back to top to the right to avoid misclicks

Flaburgan:
I have users complaining about clicking on back-to-top instead of the "comment" button.
This misclick is easy on 1367px screens:
capture du 2017-08-08 13-04-12

So I moved it to the right, at the same margin than it has with the bottom:
screen shot 2017-08-08 at 12 55 11

The only possible problem I see is if it overlaps the chat. But I don't manage to activate the chat and I still think it is unsupported at the moment, so not a blocker to that imo. If someone wants to test though, he's welcome.

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -71,7 +71,7 @@ pre { word-wrap: break-word; }
   line-height: 50px;
   opacity: 0;
   position: fixed;
-  right: 54px;
+  right: 20px;
   transition: opacity ease 400ms;
   width: 50px;
   z-index: 49;


Candidesk8:
i have this problem.
pod framasphere
screen size 1024px
webbrowser firefox
thank your for your attention

SuperTux88 (approved):

SuperTux88:
> The only possible problem I see is if it overlaps the chat.

The chat overwrites this, so this PR is only an improvement for pods with disabled chat.

SuperTux88:
Merged, thank you ❤️

PR #7522, opened 2017-08-10
Call autosize when changing input via markdown editor

### LINKED ISSUE #7521: New publisher not extending in comment when inserting markdown
A small issue regarding #7482, when the publisher is "full" and you continue to tape text, its height increases. However, this does not work when you add content using the markdown buttons (like inserting a link or an image). The height stays the same and a scrollbar appears. If you enter a character with the keyboard then the height extends accordingly. The post publisher works correctly.

I don't think that this is a release blocker but I wanted to report it.

svbergerem:
Fixes #7521


SuperTux88 (approved):
Thanks ❤️

PR #7523, opened 2017-08-11
Improve adding posts to stream

svbergerem:
* Use id as fallback when comparing posts
* Remove no posts info when adding posts to the stream


SuperTux88 (approved):
Thanks ❤️ ❤️ ❤️ ❤️ ❤️

Flaburgan:
Can I please have more context here, to get what's going on?

svbergerem:
Two posts with the same created at date could appear in arbitrary order in the stream. This made some tests fail in the Rails5 branch, where it is important to know the order of the posts before loading the stream.

PR #7525, opened 2017-08-11
Hide 'Sign in' header link on the connection page

Flaburgan:
This pull request removes the "Sign in" link in the header on the connection page (`/users/sign_in`) for both mobile and desktop version of diaspora*. I find this link confusing especially on mobile because it could give the user the feeling it will send the connection form when it's only a link to the page where you already are. So I think it can be removed.


SuperTux88 (approved):

SuperTux88:
Merged, thanks 👍

PR #7526, opened 2017-08-12
Use config.eager_load_paths instead of config.autoload_once_paths

SuperTux88:
See http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#autoloading-is-disabled-after-booting-in-the-production-environment


SuperTux88:
OK, this doesn't work in `development` now :(

skyschub (approved):
Fine by me if travis is green.

PR #7530, opened 2017-08-13
add jquery are you sure to comment form

### LINKED ISSUE #7173: No "Stay or Leave Page" prompt when writing a comment
Hey folks,

I ran into this by accident.
I noticed that when I comment on a post, I don't get a prompt in Firefox for "Staying or Leaving" the page like I do when I write a post if I click a link or use the Back button.

Does this pertain to anyone else?

ivantedja:
Right now I'm not adding `confirm on page` after dismissing the dialog (on `cuke`)
It seem that, right now it doesn't support:

```
When I am on "alice@alice.alice"'s page
Then I should be on "alice@alice.alice"'s page
```

Fix #7173

### DIFF
--- a/app/assets/javascripts/app/views/comment_stream_view.js
+++ b/app/assets/javascripts/app/views/comment_stream_view.js
@@ -37,6 +37,8 @@ app.views.CommentStream = app.views.Base.extend({
       }.bind(this),
       onFocus: this.openForm.bind(this)
     });
+
+    this.$("form").areYouSure();
   },
 
   presenter: function(){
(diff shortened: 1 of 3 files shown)


ivantedja:
@svbergerem sorry for the delay
can you give me some insight about making the jasmine test?
- is it checking the existence of the jquery.areyousure
- or checking the alert is triggered?

svbergerem:
I was thinking about a test that checks that `areYouSure` is being called when calling `postRenderTemplate`. You can find a similar test here:

https://github.com/diaspora/diaspora/blob/11372035b89c32dcf47f58e8ea1a9b63a58cca0d/spec/javascripts/app/views_spec.js#L153-L158

svbergerem (approved):

svbergerem:
Thanks! 👍

PR #7533, opened 2017-08-13
Add missing indexes from #7234

SuperTux88:
I added the indexes reviewed by @CSammy from https://github.com/diaspora/diaspora/pull/7234#issuecomment-267187943

I also refactored the `AccountDeleter` to work with the person, that can be used in #6750 too.

I didn't add the unique index for the `guid` in the `participations` table. We already have an index there and we have a unique index for the `target` which should be enough. I planned to remove the `guid` column anyway soon, because we don't really need it, so there is no need to change the existing index to a unique index there.

### DIFF
--- a/lib/account_deleter.rb
+++ b/lib/account_deleter.rb
@@ -106,6 +106,6 @@ class AccountDeleter
   end
 
   def mark_account_deletion_complete
-    AccountDeletion.where(:diaspora_handle => self.person.diaspora_handle).where(:person_id => self.person.id).update_all(["completed_at = ?", Time.now])
+    AccountDeletion.find_by(person: person)&.update_attributes(completed_at: Time.now.utc)
   end
 end
--- a/lib/account_deleter.rb
+++ b/lib/account_deleter.rb
@@ -17,9 +17,9 @@ class AccountDeleter
 
   attr_accessor :person, :user
 
-  def initialize(diaspora_handle)
-    self.person = Person.where(:diaspora_handle => diaspora_handle).first
-    self.user = self.person.owner
+  def initialize(person)
+    self.person = person
+    self.user = person.owner
   end
 
   def perform!
@@ -34,7 +34,6 @@ class AccountDeleter
         #user deletion methods
         remove_share_visibilities_on_contacts_posts
         delete_standard_user_associations
-        disconnect_contacts
         tombstone_user
       end
 
@@ -44,17 +43,17 @@ class AccountDeleter
 
   #user deletions
   def normal_ar_user_associates_to_delete
-    %i(tag_followings services aspects user_preferences
-       notifications blocks authorizations o_auth_applications pairwise_pseudonymous_identifiers)
+    %i[tag_followings services aspects user_preferences
+       notifications blocks authorizations o_auth_applications pairwise_pseudonymous_identifiers]
   end
 
   def special_ar_user_associations
-    %i(person profile contacts auto_follow_back_aspect)
+    %i[person profile contacts auto_follow_back_aspect]
   end
 
   def ignored_ar_user_associations
-    %i(followed_tags invited_by contact_people aspect_memberships
-       ignored_people share_visibilities conversation_visibilities conversations reports)
+    %i[followed_tags invited_by contact_people aspect_memberships
+       ignored_people share_visibilities conversation_visibilities conversations reports]
   end
 
   def delete_standard_user_associations
@@ -67,10 +66,6 @@ class AccountDeleter
     normal_ar_person_associates_to_delete.each do |asso|
       self.person.send(asso).destroy_all
     end
-  end
-
-  def disconnect_contacts
-    user.contacts.destroy_all
   end
 
   # Currently this would get deleted due to the db foreign key constrainsts,
@@ -97,7 +92,7 @@ class AccountDeleter
   end
 
   def normal_ar_person_associates_to_delete
-    %i(posts photos mentions participations roles)
+    %i[posts photos mentions participations roles]
   end
 
   def ignored_or_special_ar_person_associations
--- a/spec/integration/account_deletion_spec.rb
+++ b/spec/integration/account_deletion_spec.rb
@@ -1,11 +1,12 @@
 describe "deleteing account", type: :request do
   def account_removal_method
-    AccountDeleter.new(subject.diaspora_handle).perform!
+    AccountDeleter.new(person).perform!
     subject.reload
   end
 
   context "of local user" do
     subject(:user) { FactoryGirl.create(:user_with_aspect) }
+    let(:person) { user.person }
 
     before do
       DataGenerator.create(subject, :generic_user_data)
@@ -29,9 +30,7 @@ describe "deleteing account", type: :request do
         }.to(eq([true] * user.send(:clearable_fields).count)))
     end
 
-    it_behaves_like "it removes the person associations" do
-      subject(:person) { user.person }
-    end
+    it_behaves_like "it removes the person associations"
   end
 
   context "of remote person" do
--- a/app/models/account_deletion.rb
+++ b/app/models/account_deletion.rb
@@ -18,7 +18,7 @@ class AccountDeletion < ApplicationRecord
 
   def perform!
     Diaspora::Federation::Dispatcher.build(person.owner, self).dispatch if person.local?
-    AccountDeleter.new(diaspora_handle).perform!
+    AccountDeleter.new(person).perform!
   end
 
   def subscribers
--- a/app/models/account_deletion.rb
+++ b/app/models/account_deletion.rb
@@ -5,23 +5,15 @@
 class AccountDeletion < ApplicationRecord
   include Diaspora::Federated::Base
 
-  scope :uncompleted, -> { where('completed_at is null') }
+  scope :uncompleted, -> { where("completed_at is null") }
 
   belongs_to :person
-  after_commit :queue_delete_account, :on => :create
+  after_commit :queue_delete_account, on: :create
 
-  def person=(person)
-    self[:diaspora_handle] = person.diaspora_handle
-    self[:person_id] = person.id
-  end
-
-  def diaspora_handle=(diaspora_handle)
-    self[:diaspora_handle] = diaspora_handle
-    self[:person_id] ||= Person.find_by_diaspora_handle(diaspora_handle).id
-  end
+  delegate :diaspora_handle, to: :person
 
   def queue_delete_account
-    Workers::DeleteAccount.perform_async(self.id)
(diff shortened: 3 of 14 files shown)


lib/account_deleter.rb
```diff
@@ -69,10 +68,6 @@ def delete_standard_person_associations
     end
   end
 
-  def disconnect_contacts
```
cmrd-senya: Why did you remove this?

cmrd-senya:
Your refactorings conflict with the changes from #6750. What do you think we should merge first?

lib/account_deleter.rb
SuperTux88: See commit-message: Also remove `disconnect_contacts` methods, because contacts are already removed with aspects memberships in `before_destroy`.

I had a problem after changing this to use `person` instead of `diaspora_handle`, because it tried to do this twice, but the second time it failed, because `contact.user` was nil.

SuperTux88:
> Your refactorings conflict with the changes from #6750. What do you think we should merge first?

I know, that's why I already mentioned it in the PR description. Since you changed `AccountDeleter` in #6750 to handle persons too, you can use the change here, and you don't need [this if](https://github.com/cmrd-senya/diaspora/blob/5dbc1b0adf9f1f593024173dd59de9b4da1708fa/lib/account_deleter.rb#L21-L25) anymore there. So I think it's cleaner to merge this first, and the conflicts should be easy to resolve (I only removed `diaspora_handle`, and since you don't rely on that, that shouldn't be a problem for your PR).

lib/account_deleter.rb
SuperTux88: And the tests ensures that `contacts` are empty after the `AccountDeleter` run, so that shouldn't be a problem when something changes in the future.

SuperTux88:
Oh, I broke the tests last minute, because pronto told me so :( I'll fix that ...

cmrd-senya:
Usage of `person` in `AccountDeleter` is what I originally was trying to do in #6783 (which I later included to #6726).

@denschub disapproved this change back then: https://github.com/diaspora/diaspora/pull/6783#r60526640

SuperTux88:
Hmm, I don't see why we shouldn't use `person` here, we already have it and can continue using it. It doesn't make sense to query an entity again.

cmrd-senya:
> I know, that's why I already mentioned it in the PR description. Since you changed AccountDeleter in #6750 to handle persons too, you can use the change here, and you don't need this if anymore there. So I think it's cleaner to merge this first, and the conflicts should be easy to resolve (I only removed diaspora_handle, and since you don't rely on that, that shouldn't be a problem for your PR).

Okay, I need to review this PR then to check it doesn't affect anything that I use in migration.

#6750 also changes some associations in objects and in `account_deleter` (method `normal_ar_person_associates_to_delete` and friends) so that's not only `diaspora_handle`. I should check this I think.

cmrd-senya:
> Hmm, I don't see why we shouldn't use person here, we already have it and can continue using it. It doesn't make sense to query an entity again.

That's what I thought either, but I changed it to this `if` approach just because of this comment.

denschub:
> Hmm, I don't see why we shouldn't use person here, we already have it and can continue using it.

If we already have inside the current worker, that's totally fine, but passing around objects between worker causes them to get serialized to redis, which is a bit of a bummer. However, it does not seem like we end up with that here, so I may made a brainfart in the comment back then.

SuperTux88:
> Okay, I need to review this PR then to check it doesn't affect anything that I use in migration.

I checked while editing this (but that was before your latest push), and you extacted the `close_user` to a new method, but that is an easy change (you need to remove `disconnect_contacts` there too) and can be resolved.

I didn't change anything `normal_ar_person_associates_to_delete`, but I changed from `%i()` to `%i[]`, because it was really hard to read because rubymine adds curly yellow lines when there is a rubocop error, and I don't see if there is a `_` or a ` ` ;) I but I can revert that back if you like.

SuperTux88:
Also, [tests fixed](https://jenkins.coding4.coffee/job/diaspora-2.4-other/884/) (travis is still busy with your PR) :)

skyschub (approved):
Take this as approved once travis is green.

spec/integration/account_deletion_spec.rb
```diff
@@ -1,19 +1,21 @@
 describe "deleteing account", type: :request do
   def account_removal_method
-    AccountDeleter.new(subject.diaspora_handle).perform!
+    AccountDeleter.new(subject).perform!
     subject.reload
   end
 
   context "of local user" do
-    subject(:user) { FactoryGirl.create(:user_with_aspect) }
```
cmrd-senya: I'd rahter not change this part here since it will be hard to rebase. Maybe it's better to add this change on top of my PR instead?

cmrd-senya: (this applies to most of the changes in this file, maybe excluding line 3)

cmrd-senya (approved):

lib/account_deleter.rb
cmrd-senya: Actually `before_destroy` block in `AspectMembership` doesn't delete contacts if it's `sharing: true`. So it's not the same thing as `disconnect_contacts` in `AccountDeleter`. I have no idea why it worked for you, but it now fails with contacts not empty for me.

Here is my pry session output that shows what happens to a contact actually:
```
From: /home/senya/source/diaspora/app/models/aspect_membership.rb @ line 14 :

     9:   has_one :user, :through => :contact
    10:   has_one :person, :through => :contact
    11: 
    12:   before_destroy do
    13:     binding.pry
 => 14:     if self.contact && self.contact.aspects.size == 1
    15:       self.user.disconnect(self.contact)
    16:     end
    17:     true
    18:   end
    19: 

[2] pry(#<AspectMembership>)> self.contact.aspects.size
=> 1
[4] pry(#<AspectMembership>)> user.contacts
=> [#<Contact:0x0000000dfeff08
  id: 1105,
  user_id: 1139,
  person_id: 4295,
  created_at: Mon, 14 Aug 2017 01:56:40 UTC +00:00,
  updated_at: Mon, 14 Aug 2017 01:56:40 UTC +00:00,
  sharing: true,
  receiving: true>]
[5] pry(#<AspectMembership>)> next

From: /home/senya/source/diaspora/app/models/aspect_membership.rb @ line 15 :

    10:   has_one :person, :through => :contact
    11: 
    12:   before_destroy do
    13:     binding.pry
    14:     if self.contact && self.contact.aspects.size == 1
 => 15:       self.user.disconnect(self.contact)
    16:     end
    17:     true
    18:   end
    19: 
    20:   def as_json(opts={})

[5] pry(#<AspectMembership>)> next

From: /home/senya/source/diaspora/app/models/aspect_membership.rb @ line 17 :

    12:   before_destroy do
    13:     binding.pry
    14:     if self.contact && self.contact.aspects.size == 1
    15:       self.user.disconnect(self.contact)
    16:     end
 => 17:     true
    18:   end
    19: 
    20:   def as_json(opts={})
    21:     {
    22:       :id => self.id,

[6] pry(#<AspectMembership>)> user.contacts.reload
=> [#<Contact:0x0000000f50cfa8
  id: 1105,
  user_id: 1139,
  person_id: 4295,
  created_at: Mon, 14 Aug 2017 01:56:40 UTC +00:00,
  updated_at: Mon, 14 Aug 2017 01:57:25 UTC +00:00,
  sharing: true,
  receiving: false>]
```

cmrd-senya: Don't you mind me reverting the change with `disconnect_contacts` in the account deleter?

SuperTux88: You are right ... when the other person still shares with you, the contact doesn't get deleted ... So the data generator probably only creates one-way contacts? because user.contacts were empty after reload.

SuperTux88: No, one half of the contacts are deleted via `delete_contacts_of_me` and the second half is deleted with deleting aspects in `delete_standard_user_associations`. And when I wanted to delete it a third time in `disconnect_contacts` it fails with:
```
NoMethodError: undefined method `disconnect' for nil:NilClass
./app/models/aspect_membership.rb:14:in `block in <class:AspectMembership>'
...
./lib/account_deleter.rb:73:in `disconnect_contacts'
./lib/account_deleter.rb:37:in `block in perform!'
...
```

When I do `user.contacts.reload.destroy_all` in `disconnect_contacts` it works, but there are no contacts left.

cmrd-senya: Ah, I see, I missed that before, but I made a little change in `DataGenerator` for #6750 which actually created a double-sided contact: https://github.com/diaspora/diaspora/pull/6750/files#diff-4af80ea1ce63ed0a143b1b026bdb7eec

When I made methods for `DataGenerator` that are used in the account deletion spec I followed the spec that was there before. This spec didn't create any double-sided contacts. There are methods in `DataGenerator` which can make a double-sided contact but I didn't call them for the account deletion spec, only for exporter. But with the change in #6750 it is now different.

cmrd-senya: > No, one half of the contacts are deleted via delete_contacts_of_me and the second half is deleted with deleting aspects in delete_standard_user_associations

No, `delete_contacts_of_me` are contacts of local people who have us in their aspects. The thing that isn't deleted are contacts where `user` is us, and `person` is someone else, while `delete_contacts_of_me` deletes contacts where we're a person. So that's a different thing.

SuperTux88: You are right, when I add your change from the DataGenerator, it fails without `disconnect_contacts` ... We probably need better tests, because without the additional double-sided contact, it fails without `.reload`. I created #7538 to fix that, that should now cover both problems. 

I need to sleep now, because I'm tired and I need to work tomorrow, but I'll have a closer look at this tomorrow, to improve the tests. Or if you have the time to improve the tests in your PR, you can also do it.

@denschub: can you merge #7538 as soon as travis is green?

cmrd-senya: > We probably need better tests

What tests exactly? Some more tests for this method in `spec/lib/account_deleter_spec.rb`?

cmrd-senya: @SuperTux88, good night and thank you for your work!

SuperTux88: In this case: tests with different combinations of contacts. Without `disconnect_contacts` it fails when you have double sided contacts, for me it failed without `.reload`. So something that covers that.

But I just tested, when I remove the `.reload` it now also fails with the updated DataGenerator, so it's maybe already enough. I just want to make sure that we don't remove the `.reload`  or `disconnect_contacts` and something is broken again without a failing test.

But I'll have a closer look at this again tomorrow. #7538 should fix what I broke and as it looks it also makes the tests failing if I break something again, so for now that's good enough ;) 😴

PR #7534, opened 2017-08-13
added tag for supporting apps prior to Chrome M39

### LINKED ISSUE #7485: Enable "Mobile Web App Capable"
Chrome/Chromium for Android let you add web pages to home screen and if they are marked as web app the address bar is hidden.

Diaspora should work well as web app in both mobile and desktop versions.

It should be enough to simply add the following in the header:

```
<meta name="mobile-web-app-capable" content="yes">
```

For more info: https://developer.chrome.com/multidevice/android/installtohomescreen

oneturkmen:
Hi all,

This fixes #7485.

Added support for web apps prior to Chrome ver. M39 (with help of @alxlg and @SuperTux88) with the following (HAML) meta tag:

```%meta{:content => "yes", :name => "mobile-web-app-capable"}/```

Hope this solves the issue.

### DIFF
--- a/app/views/application/_head.haml
+++ b/app/views/application/_head.haml
@@ -4,6 +4,8 @@
 %meta{charset: "utf-8"}/
 = content_for?(:meta_data) ? yield(:meta_data) : metas_tags
 
+%meta{content: "yes", name: "mobile-web-app-capable"}/
+
 / favicon
 / For Apple devices
 %link{rel: "apple-touch-icon", href: image_path("apple-touch-icon.png")}
@@ -26,3 +28,4 @@
 = current_user_atom_tag
 
 = yield(:head)
+

### CODE REVIEW

File: app/views/application/_head.haml [hunk 1]
```diff
@@ -4,6 +4,8 @@
 %meta{charset: "utf-8"}/
 = content_for?(:meta_data) ? yield(:meta_data) : metas_tags
 
+%meta{:content => "yes", :name => "mobile-web-app-capable"}/
```
SuperTux88: Our style-check is currently down, but you can run it locally with `pronto run -c develop`:
```
app/views/application/_head.haml:7 W: Use the new Ruby 1.9 hash syntax.
```
So the line should look like this:
```
%meta{content: "yes", name: "mobile-web-app-capable"}/
```

SuperTux88:
Looks good 👍 can you please squash and force-push your commits?

oneturkmen:
Just squashed two commit into one and force-pushed them. Hope that I have done everything correctly.

SuperTux88 (approved):
Thanks 👍

PR #7535, opened 2017-08-13
Reset stuck exports and handle errors

SuperTux88:
When anything goes wrong while exporting, the export stays stuck, and the user needs to contact the podmin and the podmin needs to reset the boolean in the database.

This fixes that problem and also adds a migration to unstuck all users who currently has a broken export from the past. So everybody would be able to create a new export with the updated export format.

I expect that while upgrading there are no running exports. But if there are, a user could start a second export, if the user is really fast ... but I think that's not a big problem. Worst case it fails and the user can try again.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -297,18 +297,22 @@ class User < ApplicationRecord
   mount_uploader :export, ExportedUser
 
   def queue_export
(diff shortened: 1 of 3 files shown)


skyschub (approved):
\o/

PR #7543, opened 2017-08-16
Try to fix invitations from stream and aspect navigation cukes

SuperTux88:
These are the two most annoying randomly failing tests on travis, probably 80-90% of red builds are because of these two.

If this improves travis, I want to uplift that to the frozen release branch. Since this only changes tests and no code, that should be OK and hopefully we get green release builds then :)

#7373

### DIFF
--- a/features/desktop/invitations.feature
+++ b/features/desktop/invitations.feature
@@ -52,7 +52,8 @@ Feature: Invitations
 
   Scenario: sends an invitation from the stream
     When I sign in as "alice@alice.alice"
-    And I press the first "a.invitations-link" within "#no_contacts"
+    Then I should see "There are no posts to display here yet." within ".no-posts-info"
+    When I press the first "a.invitations-link" within "#no_contacts"
     Then I should see "Invite someone to join diaspora*!" within "#invitationsModalLabel"
     And I fill in the following:
       | email_inviter_emails         | alex@example.com    |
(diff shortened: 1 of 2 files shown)


SuperTux88:
First run was green 🎉 , retriggering all cucumber jobs, and see what happens :)

SuperTux88:
OK, travis didn't restart the jobs, so after waiting some minutes I amended the last commit and forcepushed to trigger a completely new build, and then travis also restarted the old build ;) I canceled the rspec jobs from the second push, no need to run them again, but we have now 2 times 4 cucumber builds again in the pipeline. :)

SuperTux88:
I just had 3 times 4 successful cucumber jobs, so 12 successful builds with this 🎉

features/desktop/invitations.feature
```diff
@@ -52,6 +52,7 @@ Feature: Invitations
 
   Scenario: sends an invitation from the stream
     When I sign in as "alice@alice.alice"
+    Then I should see "There are no posts to display here yet." within ".no-posts-info"
     And I press the first "a.invitations-link" within "#no_contacts"
```
svbergerem: > When I press the first ...

SuperTux88: Done :)

svbergerem (approved):
👍

SuperTux88:
Merged as b49b33675fc611b0d57606300b531d050b1335a9 and cdc3084990c59c9b29ce00b71706f680673fa909

Thank you for reviewing @svbergerem 👍

PR #7546, opened 2017-08-17
Fix mentioned_spec

SuperTux88:
Match parameter with `match_array` to fix order mismatch.

#7373


svbergerem (approved):

SuperTux88:
Merged as 17949ffb4ebc83380514601a9c1026894b55e20d

Thanks for the review @svbergerem ❤️

PR #7550, opened 2017-08-17
issue #4888: show user's vote in poll

netagonen:
Hi! 

Just wanted to let you know that we didn't fix the following checks for the reasons stated below them:

```
app/assets/javascripts/app/views/poll_view.js:36 W: Identifier 'original_post_link' is not in camel case.
app/assets/javascripts/app/views/poll_view.js:37 W: Identifier 'poll_answers' is not in camel case.
```
Here we are following exactly the code that is already there and it seems like it is properly formatted.

```
app/models/poll.rb:28 W: Use the new Ruby 1.9 hash syntax.
app/models/poll.rb:28 W: Redundant `self` detected.
```
These are not changes that we have made.

CC @Rete2

### DIFF
--- a/app/assets/javascripts/app/views/poll_view.js
+++ b/app/assets/javascripts/app/views/poll_view.js
@@ -17,15 +17,22 @@ app.views.Poll = app.views.Base.extend({
     var isReshare = (this.model.get('post_type') === 'Reshare');
     var showForm = defaultPresenter.loggedIn &&
                    !isReshare &&
-                   !this.model.get('already_participated_in_poll');
+                   !this.model.get("poll_participation_answer_id");
     var originalPostLink = isReshare && this.model.get('root') ?
       '<a href="/posts/' + this.model.get('root').id + '" class="root_post_link">' + Diaspora.I18n.t('poll.original_post') + '</a>' :
       '';
+    var answerGiven = this.model.get("poll_participation_answer_id");
+
+    if (defaultPresenter.poll && defaultPresenter.poll.poll_answers) {
+      defaultPresenter.poll.poll_answers.forEach(function(answer) {
+        _.extend(answer, {isCurrentUserVote: answerGiven ? answer.id === answerGiven : false});
+      });
+    }
 
     return _.extend(defaultPresenter, {
       show_form: showForm,
       is_reshare: isReshare,
-      original_post_link: originalPostLink
+      originalPostLink: originalPostLink
     });
   },
 
--- a/app/assets/stylesheets/poll.scss
+++ b/app/assets/stylesheets/poll.scss
@@ -26,6 +26,10 @@
       height: 100%;
       text-align: left;
     }
+
+    .users-vote {
+      background-color: $brand-primary;
+    }
   }
 
   .submit[disabled] {
--- a/app/assets/templates/poll_tpl.jst.hbs
+++ b/app/assets/templates/poll_tpl.jst.hbs
@@ -13,13 +13,24 @@
       {{#poll.poll_answers}}
         <div class="result-row">
           <input type="radio" name="vote" value="{{id}}"/>
-          <label>{{answer}}</label>
+          {{#if isCurrentUserVote}}
+            <label>
+              {{answer}}
+              <span class="label label-primary">{{t "poll.your_vote"}}</span>
+            </label>
+          {{else}}
+            <label>{{answer}}</label>
+          {{/if}}
           <div class="poll-result pull-right">
             <span class="percentage"></span>
             ({{t "poll.answer_count" count=vote_count}})
           </div>
           <div class="poll_progress_bar_wrapper progress">
-            <div class="poll_progress_bar bar" data-answerid="{{id}}" style="height: 100%"></div>
+            {{#if isCurrentUserVote}}
+              <div class="poll_progress_bar bar users-vote" data-answerid="{{id}}" style="height: 100%"></div>
+            {{else}}
+              <div class="poll_progress_bar bar" data-answerid="{{id}}" style="height: 100%"></div>
+            {{/if}}
           </div>
         </div>
       {{/poll.poll_answers}}
@@ -38,7 +49,7 @@
 
       {{#if is_reshare }}
         <div class="poll_footer">
-          {{{t "poll.go_to_original_post" original_post_link=original_post_link}}}
+          {{{t "poll.go_to_original_post" original_post_link=originalPostLink}}}
         </div>
       {{/if}}
     </div>
--- a/app/presenters/post_presenter.rb
+++ b/app/presenters/post_presenter.rb
@@ -67,7 +67,7 @@ class PostPresenter < BasePresenter
       title:                        title,
       location:                     @post.post_location,
       poll:                         @post.poll,
-      already_participated_in_poll: already_participated_in_poll,
+      poll_participation_answer_id: poll_participation_answer_id,
       participation:                participate?,
       interactions:                 build_interactions_json
     }
@@ -121,10 +121,8 @@ class PostPresenter < BasePresenter
     @post.reshare_for(current_user).try(:as_api_response, :backbone)
   end
 
-  def already_participated_in_poll
-    if @post.poll && user_signed_in?
-      @post.poll.already_participated?(current_user)
-    end
+  def poll_participation_answer_id
+    @post.poll&.participation_answer(current_user)&.poll_answer_id if user_signed_in?
   end
 
   def participate?
--- a/config/locales/javascript/javascript.en.yml
+++ b/config/locales/javascript/javascript.en.yml
@@ -344,3 +344,4 @@ en:
         other: "<%=count%> votes"
       show_result: "Show result"
       close_result: "Hide result"
+      your_vote: "Your vote"
--- a/spec/javascripts/app/views/poll_view_spec.js
+++ b/spec/javascripts/app/views/poll_view_spec.js
@@ -65,7 +65,7 @@ describe("app.views.Poll", function(){
       expect(this.view.$('form').length).toBe(1);
     });
     it("hides vote form when user voted before", function(){
-      this.view.model.attributes.already_participated_in_poll = true;
+      this.view.model.set("poll_participation_answer_id", this.view.poll.poll_answers[0].id);
       this.view.render();
       expect(this.view.$('form').length).toBe(0);
     });
@@ -73,6 +73,27 @@ describe("app.views.Poll", function(){
       logout();
       this.view.render();
       expect(this.view.$('form').length).toBe(0);
+    });
+  });
+
+  describe("answer given", function() {
+    it("adds 'users-vote' class to progress bar for the option the user voted for", function() {
+      var answer = this.view.poll.poll_answers[0];
+      this.view.model.set("poll_participation_answer_id", answer.id);
+      expect(this.view.$(".poll_progress_bar.users-vote").length).toBe(1);
+    });
+
+    it("doesn't add 'users-vote' class to progress bar of the options the user didn't vote for", function() {
+      var answer1 = this.view.poll.poll_answers[0];
+      var answer2 = this.view.poll.poll_answers[1];
+      this.view.model.set("poll_participation_answer_id", answer1.id);
+      expect(this.view.$(".poll_progress_bar[data-answerid='" + answer2.id + "']").hasClass("users-vote")).toBe(false);
+    });
+
+    it("adds label next to the answer the user voted for", function() {
+      var answer = this.view.poll.poll_answers[0];
+      this.view.model.set("poll_participation_answer_id", answer.id);
+      expect(this.view.$(".label.label-primary").length).toBe(1);
     });
   });
 });
--- a/spec/javascripts/jasmine_helpers/factory.js
+++ b/spec/javascripts/jasmine_helpers/factory.js
@@ -211,9 +211,9 @@ var factory = {
   },
 
   postWithPoll :  function(overrides) {
-    var defaultAttrs = _.extend(factory.postAttrs(),  {"author" : this.author()});
-    defaultAttrs = _.extend(defaultAttrs,  {"already_participated_in_poll" : false});
-    defaultAttrs = _.extend(defaultAttrs,  {"poll" : factory.poll()});
+    var defaultAttrs = _.extend(factory.postAttrs(), {"author": this.author()});
+    defaultAttrs = _.extend(defaultAttrs, {"poll_participation_answer_id": null});
+    defaultAttrs = _.extend(defaultAttrs, {"poll": factory.poll()});
     return new app.models.Post(_.extend(defaultAttrs, overrides));
   },
 
--- a/spec/presenters/post_presenter_spec.rb
+++ b/spec/presenters/post_presenter_spec.rb
@@ -135,6 +135,19 @@ describe PostPresenter do
       presenter = PostPresenter.new(status_message_with_poll)
       expect(presenter.as_json).to be_a(Hash)
     end
+
+    it "returns the answer id of the current user's poll participation" do
+      presenter = PostPresenter.new(status_message_with_poll, alice)
+      poll_answer = status_message_with_poll.poll.poll_answers.first
+      poll_participation = status_message_with_poll.poll.poll_participations
+      poll_participation = poll_participation.create(poll_answer: poll_answer, author: alice.person)
+      expect(presenter.as_json[:poll_participation_answer_id]).to eql(poll_participation.poll_answer_id)
+    end
+
+    it "returns nil if the user did not participate in a poll" do
+      presenter = PostPresenter.new(status_message_with_poll, alice)
+      expect(presenter.as_json[:poll_participation_answer_id]).to eql(nil)
+    end
   end
 
   describe "#tags" do
--- a/app/assets/stylesheets/color_themes/_color_theme_override.scss
+++ b/app/assets/stylesheets/color_themes/_color_theme_override.scss
@@ -6,7 +6,5 @@ body {
 
   .left-navbar .hoverable:hover { background-color: $main-color-essence; }
 
-  .poll_form .progress .bar { background-color: $main-color-dark; }
-
   .badge { background-color: $brand-primary; }
 }
--- a/app/assets/stylesheets/color_themes/_color_theme_override_dark.scss
+++ b/app/assets/stylesheets/color_themes/_color_theme_override_dark.scss
@@ -22,6 +22,7 @@ body {
   .poll_form .progress {
     background-color: $gray-dark;
     .bar { background-color: $gray-light; }
+    .users-vote { background-color: $brand-primary; }
   }
 
   .stream-element .collapsible {
--- a/app/models/poll.rb
+++ b/app/models/poll.rb
@@ -21,19 +21,19 @@ class Poll < ApplicationRecord
 
   def as_json(options={})
     {
-      :poll_id => self.id,
-      :post_id => self.status_message.id,
-      :question => self.question,
-      :poll_answers => self.poll_answers,
-      :participation_count => self.participation_count,
+      poll_id:             id,
+      post_id:             status_message.id,
+      question:            question,
+      poll_answers:        poll_answers,
+      participation_count: participation_count
     }
+  end
+
+  def participation_answer(user)
+    poll_participations.find_by(author_id: user.person.id)
   end
 
   def participation_count
     poll_answers.sum("vote_count")
-  end
-
-  def already_participated?(user)
-    poll_participations.where(:author_id => user.person.id).present?
   end
 end
--- a/spec/models/poll_spec.rb
+++ b/spec/models/poll_spec.rb
@@ -20,4 +20,23 @@ describe Poll, type: :model do
       expect(poll.errors).to have_key(:question)
     end
   end
+
+  describe "poll_participation" do
+    it "should return the answer object after a user voted in a poll" do
+      answer = poll.poll_answers.build(answer: "1")
+      answer.poll = poll
+      poll.poll_answers.build(answer: "2").poll = poll
+      poll.save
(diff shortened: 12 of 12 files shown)


svbergerem:
Please fix (some of) the code style errors. We still have a lot of old code that doesn't follow our styleguides and we agreed on changing the code to follow the styleguides whenever we touch it.

When fixing the camel case issues you would also have to change the identifiers in the poll template. If you don't want to do that, then I'd say it's okay to leave the code like it is.

The ruby code style issues in `app/models/poll.rb` should be fixed.

You will also have to add a fix for the dark theme (in `vim app/assets/stylesheets/color_themes/_color_theme_override_dark.scss`).

poll dark theme

netagonen:
> When fixing the camel case issues you would also have to change the identifiers in the poll template. If you don't want to do that, then I'd say it's okay to leave the code like it is.

We changed the camel case also in the template.

> The ruby code style issues in app/models/poll.rb should be fixed.

We changed lines 24-28 the the new hash syntax and not only line 28. 
Regarding the `Redundant `self``: what does that mean?

SuperTux88:
> Regarding the `Redundant 'self'`: what does that mean?

That means, that you don't need `self` there, you can simply remove it.

netagonen:
Hi, we fixed the style errors. BUT we are still working on adding a label next to the progress-bar, so it's not ready for a full review.

SuperTux88:
By the way: You can add `[WIP]` (work in progress) to a PR in the title, so it's visible to us, that you are still working on it (but you want to ask a question or need some feedback about something). I added that now, you can remove it (and write a comment, because title changes don't send notifications) when you think you are finished and it's ready for a full review (only the person who opened the PR can edit the title).

netagonen:
We are now ready for a full review :)
We only got this error that we are not sure how to fix, can you give us a hint how to?
`.eslintrc: F: Parsing error: Unexpected token }`

app/assets/javascripts/app/views/poll_view.js
```diff
@@ -21,11 +21,21 @@ app.views.Poll = app.views.Base.extend({
     var originalPostLink = isReshare && this.model.get('root') ?
       '<a href="/posts/' + this.model.get('root').id + '" class="root_post_link">' + Diaspora.I18n.t('poll.original_post') + '</a>' :
       '';
+    var answerGiven = this.model.get("poll_participation_answer_id");
+    var currentUser = defaultPresenter.current_user;
+
+    if (defaultPresenter.poll && defaultPresenter.poll.poll_answers) {
+      var pollAnswers = defaultPresenter.poll.poll_answers;
+      pollAnswers = _.map(pollAnswers, function(answer) {
+        return _.extend(answer, {isCurrentUserVote: answer.id === answerGiven, currentUser});
```
svbergerem: This line is causing the syntax error. It should be `currentUser: currentUser`.

netagonen: Thanks! fixed! now it's completely ready

app/assets/javascripts/app/views/poll_view.js
```diff
@@ -21,11 +21,21 @@ app.views.Poll = app.views.Base.extend({
...
+        return _.extend(answer, {isCurrentUserVote: answer.id === answerGiven, currentUser: currentUser});
+      });
+    }
 
     return _.extend(defaultPresenter, {
       show_form: showForm,
       is_reshare: isReshare,
-      original_post_link: originalPostLink
+      originalPostLink: originalPostLink,
+      pollAnswers: pollAnswers
```
svbergerem: You don't use it in the poll template, so you can just remove it.

app/assets/javascripts/app/views/poll_view.js
```diff
@@ -21,11 +21,21 @@ app.views.Poll = app.views.Base.extend({
...
       '<a href="/posts/' + this.model.get('root').id + '" class="root_post_link">' + Diaspora.I18n.t('poll.original_post') + '</a>' :
       '';
+    var answerGiven = this.model.get("poll_participation_answer_id");
+    var currentUser = defaultPresenter.current_user;
+
+    if (defaultPresenter.poll && defaultPresenter.poll.poll_answers) {
+      var pollAnswers = defaultPresenter.poll.poll_answers;
+      pollAnswers = _.map(pollAnswers, function(answer) {
+        return _.extend(answer, {isCurrentUserVote: answer.id === answerGiven, currentUser: currentUser});
+      });
```
svbergerem: The only purpose of these four lines is changing `defaultPresenter.poll.poll_answers` and I'd like to make that more explicit. Maybe something like this:

```JavaScript
defaultPresenter.poll.poll_answers.forEach(function (answer) {
  // set answer.isCurrentUserVote and answer.currentUser
});
```

app/assets/templates/poll_tpl.jst.hbs
```diff
@@ -13,13 +13,26 @@
...
+            {{#if isCurrentUserVote}}
+              <label>
+                {{answer}}
+                  {{#linkToAuthor currentUser}}
+                    {{{personImage this 'small' 'micro'}}}
+                  {{/linkToAuthor}}
+              </label>
+            {{else}}
+              <label>{{answer}}</label>
+            {{/if}}
```
svbergerem: The indentation here is a bit weird. `{{#if isCurrentUserVote}}` should be indented like `<input type="radio" name="vote" value="{{id}}"/>` and both `{{answer}}` and `{{#linkToAuthor currentUser}}` like `<label>`.

app/assets/templates/poll_tpl.jst.hbs
```diff
@@ -13,13 +13,26 @@
       {{#poll.poll_answers}}
         <div class="result-row">
           <input type="radio" name="vote" value="{{id}}"/>
-          <label>{{answer}}</label>
+            {{#if isCurrentUserVote}}
+              <label>
+                {{answer}}
+                  {{#linkToAuthor currentUser}}
+                    {{{personImage this 'small' 'micro'}}}
+                  {{/linkToAuthor}}
```
svbergerem: I'm not sure if we should add a link to our own profile or just a simple avatar without the link.

spec/javascripts/app/views/poll_view_spec.js
```diff
@@ -75,4 +75,32 @@ describe("app.views.Poll", function(){
...
+      this.view.model.set("poll_participation_answer_id", answer.id);
+      expect(this.view.$(".poll_progress_bar.users-vote").length).toBe(1);
+    });
+
+    it("doesn't add 'users-vote' class to progress bar of the options the user didn't vote for", function() {
+      var answer1 = this.view.poll.poll_answers[0];
+      var answer2 = this.view.poll.poll_answers[1];
+      this.view.model.set("poll_participation_answer_id", answer1.id);
+      var progressBarAnswer2 = ".poll_progress_bar[data-answerid='" + answer2.id + "']";
+      expect(this.view.$(progressBarAnswer2).hasClass("users-vote")).toBe(false);
```
svbergerem: Using a variable for the selector looks unnecessary to me. You could replace `progressBarAnswer2` by the value set before and still follow the maximum line length rule. (120 chars)

spec/javascripts/app/views/poll_view_spec.js
```diff
@@ -75,4 +75,32 @@ describe("app.views.Poll", function(){
...
+      var answer = this.view.poll.poll_answers[0];
+      this.view.model.set("poll_participation_answer_id", answer.id);
+      expect(this.view.$(".micro.avatar.img-responsive.center-block").length).toBe(1);
+    });
+
+    it("shows the 'brand-primary' color on the option the user voted for", function() {
+      var answer = this.view.poll.poll_answers[0];
+      this.view.model.set("poll_participation_answer_id", answer.id);
+      expect(this.view.$(".poll_progress_bar.users-vote").css("background-color")).toBe("rgb(0, 136, 230)");
+    });
```
svbergerem: You already tested before that the `users-vote` class is added, so here you only check the CSS rule for `.users-vote`. I'd say we can remove this test, since it should only test things our JavaScript code does. @denschub Any other opinion?

svbergerem (changes requested):
These are mostly small remarks. The new color for the progress bars looks good, but there is something about the avatar next to the label that I don't like. Unfortunately I'm currently unable to describe the reason for that. I tried to add some margin to the left of the avatar, but I'm still not satisfied. Maybe someone else has a good idea? I'll add a screenshot of the current solution below.

svbergerem:
Here is a screenshot of the current design:

poll view avatar

SuperTux88:
Hmm, I see what you mean, the avatar doesn't really fit there. While I somehow like it, it also looks strange. 

Also: When a user doesn't have an avatar it would simply display the default fallback avatar, which isn't clear then. So maybe just a text label with "your vote" or something like this is better?

Rete2:
We put a text label next to the answer now. For some reason I can´t post a screenshot here right now. I will try that later.
But one question about translations:
We added a translation in `config/locales/javascript/javascript.en.yml` and use it then in the template for the text in the label. Is this the right thing to do or do we need to add it in a different file?

SuperTux88:
> We added a translation in `config/locales/javascript/javascript.en.yml` and use it then in the template for the text in the label. Is this the right thing to do or do we need to add it in a different file?

Yes, that's the correct file.

spec/javascripts/app/views/poll_view_spec.js
netagonen: We removed the test

Rete2:
We´re still working on your comments, so no need to look at the latest commit yet.

netagonen:
We made the corrections :) so it's ready on our part

goobertron:
Sorry, I'm coming late to these latest changes.

I think the avatar is unnecessary: the blue colour of the bar indicates which option the user has voted for; and, as it is merely a reminder of which option they themselves voted for, I don't think it's necessary to reinforce this with their own avatar or a text label.

The avatar also doesn't really fit well in amongst the bars. It it a bit of visual 'noise'. While I haven't seen a screenshot of the text label there, I think this would have the same effect.

I think it's best to keep the design as simple as possible by just altering the bar colour.

netagonen:
At the moment we switched the avatar to a label. Originally we wanted this extra indication for the user's vote because a change of the progress bar color could be interpreted also as "the most popular vote" when it is actually the most popular vote. But if the majority here thinks it's unnecessary, we can drop this change.

SuperTux88:
Here is a screenshot:


I think the label is OK (better than the avatar). I think it's cool to show what the color means. When the colored bar is also the winning one, I think it's maybe confusing if the color is because it's winning or because it's your vote. So I think a indicator would be nice. An alternative would be to add a tooltip when hovering the colored bar and show "your vote" there.

config/locales/javascript/javascript.en.yml
```diff
@@ -344,3 +344,4 @@ en:
         other: "<%=count%> votes"
       show_result: "Show result"
       close_result: "Hide result"
+      vote_label: "Your vote"
```
SuperTux88: I think `your_vote` would be a better key than `vote_label`.

app/assets/javascripts/app/views/poll_view.js
```diff
@@ -21,11 +21,19 @@ app.views.Poll = app.views.Base.extend({
     var originalPostLink = isReshare && this.model.get('root') ?
       '<a href="/posts/' + this.model.get('root').id + '" class="root_post_link">' + Diaspora.I18n.t('poll.original_post') + '</a>' :
       '';
+    var answerGiven = this.model.get("poll_participation_answer_id");
+    var currentUser = defaultPresenter.current_user;
+
+    if (defaultPresenter.poll && defaultPresenter.poll.poll_answers) {
+      defaultPresenter.poll.poll_answers.forEach(function(answer) {
+        return _.extend(answer, {isCurrentUserVote: answer.id === answerGiven, currentUser: currentUser});
```
svbergerem: You don't need the `return` here. Just write `_.extend(...);`.

app/assets/templates/poll_tpl.jst.hbs
```diff
@@ -13,15 +13,26 @@
...
           <input type="radio" name="vote" value="{{id}}"/>
-          <label>{{answer}}</label>
-          <div class="poll-result pull-right">
-            <span class="percentage"></span>
-            ({{t "poll.answer_count" count=vote_count}})
-          </div>
-          <div class="poll_progress_bar_wrapper progress">
+          {{#if isCurrentUserVote}}
+            <label>
+            {{answer}}
```
svbergerem: I'm sorry if I made a mistake in my comment before (I'm unable to check right now) but here are two whitespaces missing for indentation.

Flaburgan:
There is some work to do to deal with long labels: https://github.com/diaspora/diaspora/issues/7573 this is not part of this PR, but if we add a label "Your vote" at the end, it will be even more common.

config/locales/javascript/javascript.en.yml
svbergerem: I agree.

svbergerem (changes requested):

goobertron:
Thanks for that screenshot. The label looks a lot better than I imagined; I thought it would be the user's name (which can be very long).

If everyone else wants a label, I'm happy for this to be used.

app/assets/templates/poll_tpl.jst.hbs
```diff
@@ -13,15 +13,26 @@
...
+        </div>
+        <div class="poll_progress_bar_wrapper progress">
+          {{#if isCurrentUserVote}}
+            <div class="poll_progress_bar bar users-vote" data-answerid="{{id}}" style="height: 100%"></div>
+          {{else}}
             <div class="poll_progress_bar bar" data-answerid="{{id}}" style="height: 100%"></div>
-          </div>
+          {{/if}}
         </div>
+      </div>
```
svbergerem: Mhh, it looks like the indentation in this whole block needs some work.

svbergerem:
Let's start with the label. We can still change it in future commits if we think that there is a better solution, but for now I think it's ok.

app/assets/javascripts/app/views/poll_view.js
```diff
@@ -21,11 +21,19 @@ app.views.Poll = app.views.Base.extend({
     var originalPostLink = isReshare && this.model.get('root') ?
       '<a href="/posts/' + this.model.get('root').id + '" class="root_post_link">' + Diaspora.I18n.t('poll.original_post') + '</a>' :
       '';
+    var answerGiven = this.model.get("poll_participation_answer_id");
+    var currentUser = defaultPresenter.current_user;
```
svbergerem: You don't use `currentUser` anymore in the template, so you can remove it here.

app/assets/javascripts/app/views/poll_view.js
```diff
@@ -21,11 +21,19 @@ app.views.Poll = app.views.Base.extend({
...
+
+    if (defaultPresenter.poll && defaultPresenter.poll.poll_answers) {
+      defaultPresenter.poll.poll_answers.forEach(function(answer) {
+        _.extend(answer, {isCurrentUserVote: answer.id === answerGiven, currentUser: currentUser});
```
svbergerem: And you can remove `currentUser` here as well.

app/assets/stylesheets/poll.scss
```diff
@@ -46,6 +50,10 @@
     font-weight: normal;
     margin-bottom: 5px;
     vertical-align: middle;
+
+    .center-block {
+      display: inline;
+    }
```
svbergerem: I think this rule is not needed anymore.

svbergerem (changes requested):
Just some cleanup after the latest changes. Everything else in the front-end looks good to me.

svbergerem (approved):
The front-end code looks good to me. 👍

app/presenters/post_presenter.rb
```diff
@@ -47,6 +47,10 @@ def page_title
     post_page_title @post
   end
 
+  def poll_participation_answer_id
+    @post.poll.participation_answer(current_user).poll_answer_id if already_participated_in_poll
```
SuperTux88: You can combine this with `already_participated_in_poll`, so this line looks like this:
```ruby
@post.poll&.participation_answer(current_user)&.poll_answer_id if user_signed_in?
```
and then also use `poll_participation_answer_id` to decide "if already participated" in the `poll_view.js`.

Then you can remove `already_participated_in_poll` from `post_presenter.rb` and remove `already_participated?` from `poll.rb`.

This has two reasons:
1. less code
2. less database queries

with the old way you queried the participation 3 times from the database:
1. for the `if already_participated_in_poll` in this line
2. for the `poll.participation_answer(current_user)` here
3. again `already_participated_in_poll` to add it to `non_directly_retrieved_attributes`.

With the new line you do only one query, because you can decide "if already participated" if this method returns nil or the id.

spec/presenters/post_presenter_spec.rb
```diff
@@ -135,6 +135,14 @@
...
       expect(presenter.as_json).to be_a(Hash)
     end
+
+    it "returns the answer id of the current user's poll participation" do
+      presenter = PostPresenter.new(status_message_with_poll, alice)
+      poll_answer = status_message_with_poll.poll.poll_answers.first
+      poll_participation = status_message_with_poll.poll.poll_participations
+      poll_participation = poll_participation.create(poll_answer: poll_answer, author: alice.person)
+      expect(presenter.poll_participation_answer_id).to eql(poll_participation.poll_answer_id)
+    end
```
SuperTux88: When you do the change in `post_presenter_spec.rb` this would also need a negative test to test it returns nil when the user didn't participate yet.

SuperTux88 (changes requested):

spec/presenters/post_presenter_spec.rb
Rete2: We did those changes and hope they are ok. Had to touch some other files now, but it was only because of the change of the method.

app/presenters/post_presenter.rb
```diff
@@ -47,6 +47,10 @@ def page_title
     post_page_title @post
   end
 
+  def poll_participation_answer_id
```
SuperTux88: The only reason you made this method public is because you need it in the tests. But you shouldn't change code only for tests. Can you please move this method down in the private scope (for example where `already_participated_in_poll` was).

spec/presenters/post_presenter_spec.rb
```diff
@@ -135,6 +135,19 @@
       presenter = PostPresenter.new(status_message_with_poll)
       expect(presenter.as_json).to be_a(Hash)
     end
+
+    it "returns the answer id of the current user's poll participation" do
+      presenter = PostPresenter.new(status_message_with_poll, alice)
+      poll_answer = status_message_with_poll.poll.poll_answers.first
+      poll_participation = status_message_with_poll.poll.poll_participations
+      poll_participation = poll_participation.create(poll_answer: poll_answer, author: alice.person)
+      expect(presenter.poll_participation_answer_id).to eql(poll_participation.poll_answer_id)
```
SuperTux88: You can replace this with:
```ruby
expect(presenter.to_json[:poll_participation_answer_id]).to eql(poll_participation.poll_answer_id)
```

spec/presenters/post_presenter_spec.rb
```diff
@@ -135,6 +135,19 @@
...
+    end
+
+    it "returns nil if the user did not participate in a poll" do
+      presenter = PostPresenter.new(status_message_with_poll, alice)
+      expect(presenter.poll_participation_answer_id).to eql(nil)
```
SuperTux88: Same as above.

SuperTux88 (changes requested):
Looks good, only one small thing.

@svbergerem can you have a quick look at the frontend-change again (replace `already_participated_in_poll` with `poll_participation_answer_id`)? But I think it's OK.

spec/javascripts/jasmine_helpers/factory.js
```diff
@@ -211,9 +211,9 @@ var factory = {
   },
 
   postWithPoll :  function(overrides) {
-    var defaultAttrs = _.extend(factory.postAttrs(),  {"author" : this.author()});
-    defaultAttrs = _.extend(defaultAttrs,  {"already_participated_in_poll" : false});
-    defaultAttrs = _.extend(defaultAttrs,  {"poll" : factory.poll()});
+    var defaultAttrs = _.extend(factory.postAttrs(), {"author": this.author()});
+    defaultAttrs = _.extend(defaultAttrs, {"poll_participation_answer_id": false});
```
svbergerem: This should be `null` instead of `false`, because that's what the server will return.

spec/javascripts/app/views/poll_view_spec.js
```diff
@@ -65,7 +65,7 @@ describe("app.views.Poll", function(){
       expect(this.view.$('form').length).toBe(1);
     });
     it("hides vote form when user voted before", function(){
-      this.view.model.attributes.already_participated_in_poll = true;
+      this.view.model.attributes.poll_participation_answer_id = this.view.poll.poll_answers[0].id;
```
svbergerem: You can replace this by
```javascript
this.view.model.set("poll_participation_answer_id", this.view.poll.poll_answers[0].id);
```
to make pronto happy.

svbergerem (changes requested):
Two minor remarks, everything else looks good.

spec/presenters/post_presenter_spec.rb
netagonen: And thanks for all the explanations!

svbergerem:
I just found another small bug in the post preview while reviewing #7579:

poll preview

In the preview both the answer ids and the `poll_participation_answer_id` are undefined. We should only set `isCurrentUserVote` to `true` if the user gave an answer.

svbergerem (approved):
The front-end code looks good to me.

SuperTux88 (approved):
Backend looks good now.

SuperTux88:
Merged as b556ad5211825cbc7fb711e6d0d3ca4e51666316

Thank you @Rete2 and @netagonen 🍪

PR #7556, opened 2017-08-19
Change order of fields in fill_in_new_user_form

SuperTux88:
The `user_email` is the first field anyway, and login form as a `user_username` field too, so it maybe happens that we already fill in the `user_username` filed there before page load finished.


svbergerem (approved):

SuperTux88:
Merged as 4ef0c19ce30dcb54166d23e793cff68e220b8db9

Thanks for the review @svbergerem 🍪

PR #7558, opened 2017-08-19
Fix local migration run without old private key

cmrd-senya:
It was possible to run migration locally without providing old private
key. This way migration was performed but not dispatched, which obviously
leads to desynchronization of the federation state so let's validate sender
before performing any actual actions.

### DIFF
--- a/app/models/account_migration.rb
+++ b/app/models/account_migration.rb
@@ -26,6 +26,7 @@ class AccountMigration < ApplicationRecord
   # executes a migration plan according to this AccountMigration object
   def perform!
     raise "already performed" if performed?
+    validate_sender if locally_initiated?
 
     ActiveRecord::Base.transaction do
       account_deleter.tombstone_person_and_profile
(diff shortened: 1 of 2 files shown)


app/models/account_migration.rb
```diff
@@ -113,6 +114,10 @@ def ephemeral_sender
     EphemeralUser.new(old_person.diaspora_handle, old_private_key)
   end
 
+  def validate_sender
+    raise if sender.nil?
```
SuperTux88: This `raise` looks weird. I was wondering why there isn't a message and then after looking twice I realized that `ephemeral_sender` will raise if sender couldn't be created. So I think just calling `sender` once and add a little comment that explains that calling `sender` will raise if needed is better I think.

cmrd-senya:
Updated

SuperTux88 (approved):
👍 🍪

SuperTux88:
Merged as 9ee9dbe9695a0fb74b17511b620e7f382650a9da

Thank you 🍪

PR #7565, opened 2017-08-21
use asset manifest instead of single entries in config.assets.precompile

Fensterbank:
This PR adds a `manifest.js` in which files or directories can be linked.
It has been introduced in Sprockets 3 and it's the preferred way for Sprockets 3.x and 4.x.
See also here: https://github.com/rails/sprockets/blob/master/UPGRADING.md#manifestjs

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,0 +1,13 @@
+//= link contact-list.js
+//= link ie.js
+//= link jquery3.js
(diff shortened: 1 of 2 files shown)


skyschub (approved):
Thanks! 🍪

SuperTux88:
Fixed the merge conflict and merged as 398235ee6f0ed23b4e95ffa7228025d87237600c

Thank you 🍪

PR #7566, opened 2017-08-21
Set fog_provider and load carrierwave/storage/fog if S3 is enabled

### LINKED ISSUE #7563: S3 support is broken with 0.7.0.0
For more details see: https://discourse.diasporafoundation.org/t/error-when-migrating-to-0-7/868

I have an idea what's going wrong, `CarrierWave::Storage::Fog` is not loaded automatically anymore, so maybe this:
```ruby
config.fog_provider = "fog/aws"
require "carrierwave/storage/fog"
```
in the [initializers/carrierwave.rb](https://github.com/diaspora/diaspora/blob/ac147cc9f438f0d7a374b27a155d8e1fc4952df7/config/initializers/carrierwave.rb#L9) (before `config.storage = :fog`) already helps? But I have no S3 and I can't test it.

The bigger problem with this is: Nobody of the developers uses S3 (as far as I know) and can maintain and test S3 support. So it can break anytime again. Should we simply remove S3 support or does anybody want to take care of it?

SuperTux88:
Fixes #7563

### DIFF
--- a/config/initializers/carrierwave.rb
+++ b/config/initializers/carrierwave.rb
@@ -6,6 +6,8 @@
 ENV['SSL_CERT_FILE'] = AppConfig.environment.certificate_authorities.get
 CarrierWave.configure do |config|
   if !Rails.env.test? && AppConfig.environment.s3.enable?
+    config.fog_provider = "fog/aws"
+    require "carrierwave/storage/fog"
     config.storage = :fog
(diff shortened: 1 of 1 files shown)


skyschub (approved):

SuperTux88:
Merged as 5ea7b8ee134f8b52e9485bc439f3a38e714adb1c

Thanks @denschub for the review and thanks to @trek-e and @AutumnFox for testing this.

PR #7569, opened 2017-08-22
add rails/webpacker to manage application-like JavaScript

Fensterbank:
This PR adds [rails/webpacker](https://github.com/rails/webpacker) to manage application-like JavaScript in diaspora.
It was [announced](https://twitter.com/dhh/status/808348184481124352) and published with Rails 5.1 and is the recommended way to use ES2015+, yarn and other modern JavaScript development technologies in Rails.

While gem `sprockets-es6` tried to add ES6 support (but seems very inactive) and `sprockets 4` is in an unchanged beta release since last year, `webpacker` seems to be a very active repository and coexists with the asset pipeline without side effects.

Since the compiled code in `/public/packs` is not committed and compiled during precompile process, developer and podmins will need Node.js 6.4.0+ and Yarn 0.20.1+ which could be a bigger change.

**Why this huge change?**
As a web developer I would love to improve the frontend and implement cool new features.
Since I mostly work with modern web technologies like ES2015+ and React during my daytime job, I don't see a future in the way, JavaScript files and modules are handled in diaspora today.

With this PR, we will get the full power of webpack, npm packages and dynamic imports and we will open the door to a better and more efficient way of JavaScript frontend development in diaspora.

PS: I'm not sure of how to get pronto to accept the new syntax in several config files. Adding `"ecmaVersion": 7` to `.eslintrc` did not solved the issue. Help is appreciated. 🙂


denschub:
Please use npm instead of yarn.

Fensterbank:
> Please use npm instead of yarn.

webpacker ships with yarn.
Creating a new Rails 5.1 app with the `rails new myapp --webpack` parameter ships yarn.
I don't think it's something, we can change here.

> I have serious doubts about whether this is the time to actually do this or not. I see no reason of splitting our frontend off right now.

It's not about splitting the frontend and it's not, that I want to build a React app right into our UI or something.
It's about to have a way to create and use independent JavaScript modules, exports and the power of ES2015+.
The whole point of `webpacker` is to separate the way JavaScript modules are handled and compiled because the asset pipeline in Rails was never designed to handle modules. Even ES6 in `Sprockets 4` would never allow to do that.

Example: Why should Diaspora.ProfilePhotoUploader be available on every page of diaspora? This doen't feel modern and as a module compiled with webpacker it would only be available on `profile/edit`.

> This certainly is the kind of change that need that would have benefited from discussions before implementing it.

The implementation wasn't quite hard, so I presented you a solution we can discuss.

SuperTux88:
I would like to prevent any additional dependencies here. We already have nodejs as dependency (I don't know which version, debian comes with 4.8 and it works), but I don't know where to get yarn, there is no package in debian or ubuntu sources. Our installation guide is already complicated enough, I don't want to make it even more complicated. The only thing that needs to be installed additionally is currently ruby (with rvm), because diaspora is written in ruby. Adding more dependencies only makes it unnecessary complicated. When it works with what we already have, then that's fine to me (I don't understand what all these dependencies do and why they are needed, I only see that it currently perfectly works without them).

denschub:
> Creating a new Rails 5.1 app with the rails new myapp --webpack parameter ships yarn.
I don't think it's something, we can change here.

No, webpacker does **not** ship yarn. Yarn is a prerequisite of webpacker and not included in their distribution. We already depend on Node, so `npm` is available, so the additional step of running `npm i -g yarn` (or, `npm i`, since you commited the `package.json`) is both unnecessary and wrong. I do not want to get involved in us supporting issues with global npm packages - nor do I want to be involved in helping people having issues when their local node bundle fails. In any way, this is something that adds quite a few steps for installing diaspora*, both for development and production, and quite honestly, I see no reason why we should do this.

> The whole point of webpacker is to separate the way JavaScript modules are handled and compiled because the asset pipeline in Rails was never designed to handle modules. Even ES6 in Sprockets 4 would never allow to do that.

No. The point of webpacker is to ship webpack in a Rails application. The only "integration" between the rails app and the frontend stuff there is is basically webpacker using the Rails assets pipeline to discover files to build. That's it. Webpacker literally uses webpack to call *their own* build chain for frontend files, and that includes you commiting `package.json` for node into our repository, having the podmin (or webpacker) running `npm i` to install *their own* `babel`, *their own* `coffee-script`, *their own* `scss` compiler, ... effectively duplicating our entire frontend build chain. Heck, per default, they even use `webpack-dev-server` to distribute assets, in addition to the worker already running.

Sorry, but no.

Yes, hot module reload for the frontend would be nice, and actually using webpack to ship our frontend would be nice as well, but not as long as the frontend is dependent on backend components. having that chain set up makes sense if and only if your frontend is actually independent, and that includes all html assets as well. This step would be right if you magically would be able to rewrite everything in react, for example, where the backend only serves JSON. If the projects are split, we don't need the asset pipeline in rails (and hey, that's exactly what the Rails 5 API apps are doing), but we are not there yet.

> It's about to have a way to create and use independent JavaScript modules, exports and the power of ES2015+.

And if you can't do that by using the asset pipeline, then you cannot do that. I am not willing to ignore all the issues I mentioned above just for the sake of having a "nicer" (and even that is questionable) code structure for one feature that would use it (the one you are planning on working on). Sorry.

---

In addition, let me add something:

> Example: Why should Diaspora.ProfilePhotoUploader be available on every page of diaspora? This doen't feel modern and as a module compiled with webpacker it would only be available on profile/edit.

You are wrong, sadly. Even with webpacker, all modules are shipped everywhere. They might not be available in the scope the `profile/edit` page lives in, since requirejs throws each site on its own scope with dependency injection, but they are still shipped to the browser. And actually, that's a good thing, because this means that all dependencies are bundled together in *one file* that does not change between page loads, thus being cachable by browsers. The only gain you get is a smaller memory footprint *in theory*, since modules are not in the scope of the `window` object. However, *in practice*, this doesn't even matter, since the modules are in scope for requirejs anyway, and scope assignments are based by reference, not by value. So let's call the memory savings as... negligible.

Fensterbank:
> In any way, this is something that adds quite a few steps for installing diaspora*, both for development and production, and quite honestly, I see no reason why we should do this.

That was indeed my biggest doubt.

> If the projects are split, we don't need the asset pipeline in rails (and hey, that's exactly what the Rails 5 API apps are doing), but we are not there yet.

I'm looking forward to that day... \o/

> I am not willing to ignore all the issues I mentioned above just for the sake of having a "nicer" (and even that is questionable) code structure for one feature that would use it (the one you are planning on working on)

Hm, in my opinion it should not be a discussion point for what features it will be used.
If I would have created this PR with five new features using it, it would be not in the scope of this single PR. So I just created a PR which opens the door, the features using it should be developed and merged in other PRs.

Of course I can implement my features in old jquery polluted object syntax like I did ten years ago, but it's not fun. And if UI development in diaspora is not fun, then there are less developers wanting implement cool new stuff.

But anyway, thanks for your detailed answers and opinions in this discussion.
Since I did not really expect this PR to be accepted and it wasn't big work it's no problem for me. :)

Flaburgan:
@Fensterbank I know it's not shiny new, but the front-end is using backbone.js and we only support IE11, so if you want to write JS without jquery, you're welcome to do so. As @denschub said, we do want to update our stack. Now is just not the time to do that change. But please, don't be stopped by that if you want to contribute!

PR #7570, opened 2017-08-22
Switch back to mobile-fu

SuperTux88:
There is a new release with the old name now, so we can switch back to it.


svbergerem (approved):

svbergerem:
Merged as b530ed0be08d9d7f6bfabeb1dc2b18a172f5015d. Thanks!

PR #7575, opened 2017-08-23
Fix mixed username and timestamp with LTR/RTL scripts

### LINKED ISSUE #7424: User name and timestamp intermixed with LTR/RTL scripts
I just noticed that an RTL character in a user name which ends in an RTL character, that character appears to the right of the hyphen and the date part of the timestamp. See image.

_rtl

The majority of the user name, which I have blurred out, is in LTR; just the final character, which appears after a space, is in RTL (Arabic).

This doesn't seem to fit with the other open RTL issues I could find. Possibly related to #3529, which was recently closed.

svbergerem:
Fixes #7424

### DIFF
--- a/app/assets/stylesheets/stream_element.scss
+++ b/app/assets/stylesheets/stream_element.scss
@@ -90,6 +90,7 @@
     .author-name {
       font-weight: bold;
       margin-bottom: 4px;
+      unicode-bidi: isolate;
     }
     .feedback {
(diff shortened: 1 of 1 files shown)


skyschub (approved):

SuperTux88:
Merged as ce40ec128f5f7c9938d2075b06da190b0e8208a7

Thanks you 🍪

PR #7576, opened 2017-08-24
Remove tiff support from photos

manuelVo:
Currently the list of supported extensions for photos in the frontend and the list in the backend are inconsistent. The fontend allows `tiff` and `tif` files while the backend does not (not everywhere at least).

I propose removing `tiff` and `tif` files from the lists of supported extensions for photos since the browser support for them is quite bad (none of the GIMP generated `tiff` files i tested could be displayed in either Chrome or Firefox).


SuperTux88 (approved):
Yes, lets remove tiff 👍

SuperTux88:
Merged as 0f771dbb107045d78c3f9e01efcc20c232b3ff04

Thank you 🍪

PR #7578, opened 2017-08-24
Remove root_guid from reshares when post is deleted

SuperTux88:
And cleanup old reshares with deleted root posts

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -33,6 +33,10 @@ class Post < ApplicationRecord
     self.touch(:interacted_at)
   end
 
+  before_destroy do
+    reshares.update_all(root_guid: nil) # rubocop:disable Rails/SkipsModelValidations
+  end
+
   #scopes
   scope :includes_for_a_stream, -> {
     includes(:o_embed_cache,
--- a/db/migrate/20170824202628_cleanup_root_guids_from_reshares.rb
+++ b/db/migrate/20170824202628_cleanup_root_guids_from_reshares.rb
@@ -1,0 +1,9 @@
+class CleanupRootGuidsFromReshares < ActiveRecord::Migration[5.1]
+  def up
+    # rubocop:disable Rails/SkipsModelValidations
+    Reshare.joins("LEFT OUTER JOIN posts as root ON root.guid = posts.root_guid")
+           .where("root.id is NULL AND posts.root_guid is NOT NULL")
+           .update_all(root_guid: nil)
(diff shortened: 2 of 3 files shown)


db/migrate/20170824202628_cleanup_root_guids_from_reshares.rb
```diff
@@ -0,0 +1,13 @@
...
+  def up
+    # rubocop:disable Rails/SkipsModelValidations
+    Reshare.joins("LEFT OUTER JOIN posts as root ON root.guid = posts.root_guid")
+           .where("root.id is NULL AND posts.root_guid is NOT NULL")
+           .update_all(root_guid: nil)
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
cmrd-senya: Do we really need this `IrreversibleMigration` here? This will block migrations from roll back. But it feels safe to roll back here by doing nothing: in this case we will stay with `root_guid=nil` for deleted roots which is fine for pre-migration state as well. So why fail?

SuperTux88: You are right, I removed it, thanks 👍

jhass (commented):
I think there can be a slight optimization.

app/models/post.rb
```diff
@@ -33,6 +33,10 @@ class Post < ApplicationRecord
     self.touch(:interacted_at)
   end
 
+  before_destroy do
+    reshares.each {|reshare| reshare.update_attributes(root_guid: nil) }
```
jhass: This ought to be possible with a single update statement, if reshares is a relation just skipping the each should do it.

SuperTux88: No, I could use `update_all` (like in the migration):
```ruby
reshares.update_all(root_guid: nil)
```
but pronto doesn't like this: ``app/models/post.rb:37 W: Avoid using `update_all` because it skips validations.``

Should I disable pronto on this line too?

jhass: I think so, it's a innocuous enough statement, and doing an N+1 query here just feels worse to me.

SuperTux88: You're right, after thinking more about it, it makes much more sense that way. I updated it.

jhass (approved):

SuperTux88:
Merged as b040f85790c61a4155a044f3f1fa198e49a96925

Thanks for the review @jhass 🍪

PR #7580, opened 2017-08-25
Merge migrations from before 0.6.0.0 to 0000_create_schema.rb

SuperTux88:
Follow-up for #7514

Merge all migrations from before 0.6.0.0 to the initial migration. This has no effect for existing pods, but makes travis and installation of new pods a bit faster. It's still possible to upgrade from 0.6.x to 0.7.x


skyschub (approved):
❤️ 🍪

SuperTux88:
Merged as e08440d7ba28fca9e6eb53d785faade7ce10b8a0

Issue #7577, opened 2017-08-24
Frontend hangs when submitting photo with extension that is not supported by the backend


manuelVo:
When submitting an photo which has an extension that isn't listed in the list of supported extensions of the backend (but allowed by the frontend) no error is reported by the backend. Instead the frontend is stuck on a filled progress bar which won't disappear.

Steps to reproduce:
1. Add extension `txt` to `allowedExtensions` in `app/assets/javascripts/app/views/publisher/uploader_view.js`
2. Upload a `.txt` file as a photo
3. No error will show up after uploading. The progress bar will stay on the screen forever.

pipcet:
This is a combination of several minor issues:
1. We don't respond with JSON to the failed image upload request.
2. We don't deal properly with failed image uploads in the frontend.
3. Fine Uploader doesn't deal with non-JSON responses that have okay HTTP status codes.

Here's a "fix" for (1) and (2); at least it should highlight the source locations where the problems are. It removes the permanent message entirely, and it'd be nicer to give it a close button instead, and it might misrespond to queries that should legitimately get an HTML response: https://github.com/diaspora/diaspora/compare/develop...pipcet:photo-upload?expand=1 (I've had to change a few minor things to get a test installation going and modify uploader_view.js, so let me know if this doesn't work).

Is it worth polishing this or should we just be waiting for an HTML5-canvas based new image uploader?

tclaus:
Should be closed with https://github.com/diaspora/diaspora/pull/8205
(frontend only allows image types now)

PR #7581, opened 2017-08-25
Profile photo upload with cropping function

### LINKED ISSUE #5700: Let the user crop avatar
This is an idea I've had in mind since I've seen avatar resizing mess on mobile.
I think it would be a cool idea to let the user crop its avatar when he uploads it.

I found a [jQuery plugin](https://github.com/fengyuanchen/cropper) that does it (maybe there's others). My main concern would be performance on low configurations.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8834453-let-the-user-crop-avatar?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Fensterbank:
This PR adds a cropping function to the profile photo upload.
It's a refactoring of the old profile photo uploader using ES2015.

1. User selects a photo to set as profile picture.
2. Picture is loaded into browser and displayed in a cropper component.
3. User selected the section he wants to use as profile picture.
4. Confirm selection with the accept-button.
5. Picture will be uploaded and saved as profile picture.

**Two problems are getting solved:**

- New profile pictures always get an aspect ratio of 1:1 so we don't have weird scaled or cropped pictures in the stream
- It's no problem to use a 30 MB camera photo (if your browser handles it) as profile picture, because the cropped image uploaded to server would be resized and won't exceed 4 MB.

grafik

**Notice:** This PR does not affect the mobile page. The old class `ProfilePhotoUploader` has been moved to mobile folders and will be included in the mobile page.

Fixes #5700


Flaburgan:
This fixes #5700. Does it also apply the changes described in https://github.com/diaspora/diaspora/issues/4997?

Fensterbank:
No, sine I did not change anything on the client side logic, chunk uploading described in #4997 is not implemented.
But with this change the resize option of fineuploader is used, so If the cropped section of the image exceeds a maximum of 1600px it will be resized.

It's very unlikely or impossible for JPG images of this size to exceed the 4 MB limit, so I don't see the very need for chunk upload.

PS: To let users take their (cropped and resized) camera photos as profile pictures, I had to remove the size limit option in fineuploader.
Therefore it is currently possible (when using PNG) to create a profile picture exceeding the limit which could cause webserver errors on upload. It's not possible to set the file size limit after cropping and resizing.
I'm not sure, how to properly deal with it... Why do we have to allow PNGs as profile picture? ;)

Flaburgan:
Users mainly upload images in posts, not for their avatars, that's why #4997 is really important.

denschub:
> Users mainly upload images in posts, not for their avatars, that's why #4997 is really important.

Totally agree, but this PR is about adding a cropping thingy for profile pictures, so probably unrelated. :)

app/assets/javascripts/helpers/profile_photo_uploader.es6
```diff
@@ -0,0 +1,250 @@
...
+    this.createButton = (icon, onClick) =>
+      ($(`<button class="btn btn-default" type="button"><i class="entypo-${icon}"></i></button>`)
+        .on("click", onClick));
+
+    /**
+     * Shows a message using flash messages or alert for mobile.
+     * @param {string} type - The type of the message, e.g. "error" or "success".
+     * @param text - The text to display.
+     */
+    this.showMessage = (type, text) => { return app.flashMessages ? app.flashMessages[type](text) : alert(text); };
```
denschub: I guess using implicit returns here is fine:

```js
this.showMessage = (type, text) => app.flashMessages ? app.flashMessages[type](text) : alert(text);
```

app/assets/javascripts/helpers/profile_photo_uploader.es6
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  /**
+   * Initializes the fine uploader component
+   */
+  initFineUploader() {
+    this.fineUploader = new qq.FineUploaderBasic({
+      element: this.fileInput,
+      validation: {
+        allowedExtensions: ["jpg", "jpeg", "png"]
```
denschub: Do we have any easy way of passing the list of valid extensions from the backend to the frontend? Thinking about ways to elegantly workaround #7577.

app/assets/javascripts/helpers/profile_photo_uploader.es6
```diff
@@ -0,0 +1,250 @@
...
+
+      messages: {
+        typeError: Diaspora.I18n.t("photo_uploader.invalid_ext"),
+        sizeError: Diaspora.I18n.t("photo_uploader.size_error"),
+        emptyError: Diaspora.I18n.t("photo_uploader.empty")
+      },
+
+      callbacks: {
+        onProgress:
+          (id, fileName, loaded, total) => (this.info.innerText = `${fileName} ${Math.round(loaded / total * 100)}%`),
```
denschub: This looks weird. I'd remove the implicit return and have the function body in its own line, just to make it look a bit nicer.

Flaburgan:
Well as long as 

>PS: To let users take their (cropped and resized) camera photos as profile pictures, I had to remove the size limit option in fineuploader.

only apply to avatar upload then yes, it's unrelated. However it looks like even for that PNG can be bigger than 4MB, so maybe #4997 has to be done before this PR? Or maybe we can set a max size way under 1600px, that feels really big for an avatar. 800px should be enough imo.

app/assets/javascripts/helpers/profile_photo_uploader.es6
```diff
@@ -0,0 +1,250 @@
...
+    const file = this.fileInput.querySelector("input").files[0];
+
+    // ensure browser's file reader support
+    if (FileReader && file) {
+      try {
+        const fileReader = new FileReader();
+        fileReader.onload = () => this.initCropper(fileReader.result);
+        fileReader.readAsDataURL(file);
+      } catch (e) {
+        this.setLoading(false);
```
denschub: Hm, when could `readAsDataURL` throw? Is this something that could occur during normal use? I cannot think of any scenario where it might.

So maybe, catching isn't the best idea here, since we'll lose any way of debugging stuff if uploading an image fails from the user perspective. Maybe `finally` would be more suitable to ensure the loading animation is disabled, but have the error be actually thrown.

app/assets/javascripts/helpers/profile_photo_uploader.es6
```diff
@@ -0,0 +1,250 @@
...
+   */
+  initCropper(imageData) {
+    // cache the current picture source if the user cancels
+    this.previousPicture = this.picture.getAttribute("src");
+
+    this.mimeType = imageData.split(";base64")[0].substring(5);
+
+    this.picture.onload = () => {
+      // set the preferred size style of the cropper based on picture orientation
+      this.picture.setAttribute("style", (this.picture.naturalHeight > this.picture.naturalWidth ?
```
denschub: I wouldn't mind having this written out a bit more verbose to avoid line breaks inside ternaries.

app/assets/javascripts/helpers/profile_photo_uploader.es6
```diff
@@ -0,0 +1,250 @@
...
+      .append(this.controls.rotateLeft).append(this.controls.rotateRight));
+
+    // preview images in the middle
+    this.controlRow.append($("<div class='previews'>")
+      .append("<div class='preview-lg'>")
+      .append("<div class='preview-md'>")
+      .append("<div class='preview-sm'>"));
+
+    // main buttons on the right
+    this.controlRow.append($("<div class='btn-group buttons-main' role='group'>")
```
denschub: `.append` takes an array!

app/assets/javascripts/helpers/profile_photo_uploader.es6
```diff
@@ -0,0 +1,250 @@
...
+   * @param {number} id - The current file's id.
+   * @param {string} fileName - The current file's name.
+   * @param {object} responseJSON - The server's json response.
+   */
+  onUploadCompleted(id, fileName, responseJSON) {
+    this.setLoading(false);
+    this.fileInput.classList.remove("hidden");
+    this.info.innerText = Diaspora.I18n.t("photo_uploader.completed", {"file": fileName});
+
+    if (responseJSON.data !== undefined) {
```
denschub: What if uploading failed for server-side reasons?

skyschub (changes requested):
I did a first pass by just looking at the source, have to do more testing.

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: Was my initial idea, but the active eslint rule [no-confusing-arrow](https://eslint.org/docs/rules/no-confusing-arrow) doesn't accept it.

denschub: I... totally knew that! Disregard, then.

Fensterbank: on the other hand: I don't like this rule and see no confusion here. We could deactivate it or at least set `allowParens` to `true`. The implicit return looks so much nicer.

denschub: I like, and understand the rule. However, with `allowParens`, it's fine by me. @svbergerem, any opinion?

denschub:
@Flaburgan this only affects profile image uploads!

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: do we have a list of accepted extensions? do we want to save them in a global variable or something?

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: i just pushed an update and fixed this the way you suggested.

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: you're right, `readAsDataURL` cannot throw here and the existence of `FileReader` is checked, so no need for `try catch`. I just pushed an update and fixed this.

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: I just pushed an update and improved this.

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: I… totally knew it! :)

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: I just pushed an update and enabled `allowParens`. Looks good for me.

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: In this case the following error is shown in the UI. The UI is changed back to initial state (same behavior as when image cropping process is cancelled by the user).
upload-error

app/assets/javascripts/helpers/profile_photo_uploader.es6
Fensterbank: At the current state we have multiple places where the allowed extensions are defined.
I don't know how to bring the list of atteched types from backend to frontend (without refactoring many things).
In addition we accept gif in posts while we only accept jpg and png for profile pictures. So also making a single frontend and/or a single backend setting doesn't make sense here.
Needs more thinking in the future…

Fensterbank:
@denschub I took care of the requested changes and, if necessary, commented on the subject.
Could you please review again or notify me if there are any further steps to be taken?

denschub:
Thanks for putting the changes to the uploader in place, as we've discussed off-GitHub. Looks great now.

So, let's talk about this PR. I have reviewed the sources to my best ability, although I do not have 100% confidence that we do not introduce any side effects here. My ability to judge that in the current frontend is... limited, mainly caused by some design issues in the frontend code. We already started discussions about future plans on Discourse, so let's not make this off-topic here.

Luckly, I don't need 100% confidence, as we have the ability to test pull requests in the `develop` branch, which gives it time to show any potential problems we have. Also, this gives us a little bit more time to check if the es6 bootstrapping has some sideeffects we only discover in production.

Sooo....

PR #7589, opened 2017-08-27
Prevent users from zooming in IE Mobile

### LINKED ISSUE #7499: Top bar shall not resize when zooming on Edge Mobile
User Story:
As a diaspora* user I don’t want that the top bar resizes when I zoom into the page because the bar begins to covers the page.

User Acceptance Test:
Given I use Edge Mobile AND I visit a diaspora web client AND I zoom into the page THEN the top bar is NOT resized.

svbergerem:
Fixes #7499

According to https://msdn.microsoft.com/en-us/library/dn629259(v=vs.85).aspx, “the minimum-scale, maximum-scale, and initial-scale properties are currently unsupported for Internet Explorer for Windows Phone“.

For further information on the viewport properties see https://drafts.csswg.org/css-device-adapt/#viewport-meta

### DIFF
--- a/app/views/layouts/application.mobile.haml
+++ b/app/views/layouts/application.mobile.haml
@@ -12,7 +12,7 @@
     = render "head"
     = include_color_theme "mobile"
 
-    %meta{name: "viewport", content: "width=device-width, minimum-scale=1 maximum-scale=1"}/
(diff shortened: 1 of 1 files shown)


skyschub (approved):

SuperTux88:
Merged as cd09c75c50d6b4b344df80a5f44f3f69ef316346

Thanks @svbergerem 🍪 and thanks to @denschub for the review 🍪

PR #7590, opened 2017-08-27
Fix root guid validation

### LINKED ISSUE #7587: "Unprocessable Entity": Can't comment or like a deleted reshared post anymore
[This is a reshared post which got deleted](https://nerdpol.ch/posts/390822906d2f01357131423d69c21395), so as usual the message "The original post has been deleted by the author" (or whatever the english phrase is) is shown.

Usually i write a comment to tell the person that their reshared post has been deleted, so they can decide if they want to delete their reshare or not. 👍

But at this post, i neither can like nor comment. When i click the like button it says "Like didn't work" (or similar) and when i want to send a comment, it shows a JSON object as an error message, telling that this entity can't be processed. That person does not have me on their ignore list. Check the screenshot: https://i.imgur.com/C2UXZot.png

SuperTux88:
Follow-up for #7578

Fixes #7587

### DIFF
--- a/spec/models/reshare_spec.rb
+++ b/spec/models/reshare_spec.rb
@@ -3,15 +3,41 @@ describe Reshare, type: :model do
     expect(FactoryGirl.build(:reshare)).to be_valid
   end
 
-  it "requires root" do
-    reshare = FactoryGirl.build(:reshare, root: nil)
-    expect(reshare).not_to be_valid
-  end
+  context "validation" do
+    it "requires root" do
+      reshare = FactoryGirl.build(:reshare, root: nil)
+      expect(reshare).not_to be_valid
+    end
 
-  it "require public root" do
-    reshare = FactoryGirl.build(:reshare, root: FactoryGirl.create(:status_message, public: false))
-    expect(reshare).not_to be_valid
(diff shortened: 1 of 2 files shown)


skyschub (approved):

spec/models/reshare_spec.rb
```diff
@@ -3,15 +3,46 @@
     expect(FactoryGirl.build(:reshare)).to be_valid
   end
 
-  it "requires root" do
-    reshare = FactoryGirl.build(:reshare, root: nil)
-    expect(reshare).not_to be_valid
-  end
+  context "validation" do
+    it "requires root when the author is local" do
```
cmrd-senya: Why can't there be local reshare with no root? What if root post was deleted and reshare author is local?

SuperTux88: This is only validated on create, so it's fine to have the reshare without root later, but a local user can't create a new reshare without root. We maybe need to change that later for the import and validate that somewhere else, but for now that's the current state.

SuperTux88:
Merged as 04035c3eb7e8f85ce5ebbf6b34fd9a4cb57b527a and 163ffdb19b134d54d446ec0a255994b9cf0f628b

denschub:
Btw, thanks for the impressive turnover time! ;)

PR #7593, opened 2017-08-27
Replace where...first with find_by

### LINKED ISSUE #7539: Inefficient api where.first
In post.rb, we can use find_by to replace where.first which will not do the extra order by the primary key.

svbergerem:
Fixes #7539

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -45,7 +45,7 @@ class Comment < ApplicationRecord
 
   after_destroy do
     self.parent.update_comments_counter
-    participation = author.participations.where(target_id: post.id).first
+    participation = author.participations.find_by(target_id: post.id)
     participation.unparticipate! if participation.present?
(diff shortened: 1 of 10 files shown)


denschub:
Besides somewhat ruining `git blame` by touching all these queries, have you found any actual performance gain? Because I tested them, and query times were exactly the same, so I'm a bit hesitant of merging that stuff just because(tm).

svbergerem:
No, I didn't test this. With this PR we omit the unnecessary `ORDER BY` and I think that `find_by` is better readable than `where(...).first`.

denschub:
Okay!

SuperTux88 (approved):
👍 🍪

SuperTux88:
Merged as 9d2763089d2796deb5a5d2388f0a92d13447761b

Thanks 🍪

PR #7595, opened 2017-08-27
Enable frozen string literals

SuperTux88:
This is supported since ruby 2.3 and will be the default with ruby 3.0, so it's a good thing to be prepared ;)

There is one pronto warning that I didn't fix:
```
app/models/acts_as_taggable_on-tag.rb:1 W: The name of this source file (`acts_as_taggable_on-tag.rb`) should use snake_case.
```


SuperTux88:
I marked this with the 0.7.1.0 milestone. Theoretically this could cause regressions when there are string manipulations in parts not covered by tests. But we are still 4 weeks away from freeze, so we hopefully find regressions in this time (if there are any).

skyschub (approved):
I'm unsure, since I usually don't like adding something to all files...

However, frozen strings will be default in Ruby 3, and enabling them helps us testing if we break something. Also, Pronto will make sure we don't miss the comment in new files, so it's probably okay...

SuperTux88:
Merged as ... many commits ... thanks @denschub for the review 🍪

PR #7596, opened 2017-08-27
Remove unused scripts

SuperTux88:
They were added a long time ago and they talk about rails 3 ... we have rails 5 now and scripts in the bin folder, so I think we can remove them.


skyschub (approved):

SuperTux88:
Merged as 1d2394571c48a0da5c785a75e1870502c9adf55d

Thank you @denschub for the review 🍪

PR #7597, opened 2017-08-27
Remove rails_admin_histories table

SuperTux88:
This was a leftover from rails_admin which was removed in #7440.

Not sure if we should merge this to 0.7.1.0 or 0.8.0.0, but since we already have migrations for 0.7.1.0 and removing a table should be really fast, I think we can add this as drive-by for 0.7.1.0.

### DIFF
--- a/db/migrate/20170827222357_remove_rails_admin_histories.rb
+++ b/db/migrate/20170827222357_remove_rails_admin_histories.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class RemoveRailsAdminHistories < ActiveRecord::Migration[5.1]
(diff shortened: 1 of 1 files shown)


db/migrate/20170827222357_remove_rails_admin_histories.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
```
denschub: ;D

skyschub (approved):

SuperTux88:
Merged as 4c967903fed684dacb9d14ecbfc6f06158e3df5c

PR #7598, opened 2017-08-27
Remove "string" column from "ppid" table

SuperTux88:
This was added by accident in 4e18f3849dcc5d36bda93e68b597b869735d5517.

Same as for #7597:

Not sure if we should merge this to 0.7.1.0 or 0.8.0.0, but since we already have migrations for 0.7.1.0 and this table probably is still empty, so removing a column should be really fast. I think we can add this as drive-by for 0.7.1.0.


skyschub (approved):

SuperTux88:
Merged as 028199a6f352b9b20bd72b3f514241cb0fcd1e6b

PR #7606, opened 2017-09-03
Don't truncate ar_internal_metadata after cucumber scenarios

SuperTux88:
This is an internal active record table and shouldn't be touched. When this is truncated you need to fix it again by running `bin/rails db:environment:set RAILS_ENV=test`.

This is a temporary fix until there is a new version of
database_cleaner. See DatabaseCleaner/database_cleaner#445.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -51,6 +51,9 @@ Capybara.default_max_wait_time = 30
 # of your scenarios, as this makes it hard to discover errors in your application.
 ActionController::Base.allow_rescue = false
 
+# TODO: Temporary fix for rails 5, remove the next line after a new version of database_cleaner is released.
+# See https://github.com/DatabaseCleaner/database_cleaner/issues/445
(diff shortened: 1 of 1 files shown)


features/support/env.rb
```diff
@@ -51,6 +51,7 @@
 # of your scenarios, as this makes it hard to discover errors in your application.
 ActionController::Base.allow_rescue = false
 
+Cucumber::Rails::Database.javascript_strategy = :truncation, {except: %w[ar_internal_metadata]}
```
svbergerem: Could you add a comment above for this temporary fix with a link to the issue / pull request in the database cleaner repo?

SuperTux88: Sure, done :)

svbergerem (approved):

svbergerem:
Merged as 45e714776ebcbb13d6f40f67d2a6414820e362f0. Thank you! :)

PR #7614, opened 2017-09-14
Fix database issues for export

SuperTux88:
I fixed some database issues also blocking the profile export. So the PR contains DB-migrations, but since it's for the export I think it's important that they are in the minor release.

Times (with a local copy of a new backup of my production DB):
```
== 20170914202650 CleanupInvalidLikes: migrated (28.7232s) ====================
== 20170914212336 CleanupInvalidPolls: migrated (2.9790s) =====================
```

### DIFF
--- a/app/models/poll.rb
+++ b/app/models/poll.rb
@@ -3,8 +3,8 @@ class Poll < ApplicationRecord
   include Diaspora::Fields::Guid
 
   belongs_to :status_message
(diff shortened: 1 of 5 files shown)


skyschub (approved):

SuperTux88:
Merged as 655fe2a9129a70d6add9462f816fb55c274b31db and 00296ffda51aee1d572219a40fe59f64c521f562

Thank you for the review @denschub 🍪

PR #7620, opened 2017-09-17
Delete invalid oEmbed caches with binary titles

SuperTux88:
There are a few old oEmbed caches which have the title saved in binary (because they contain Chinese characters). This fails with `ActionView::Template::Error ("\xE5" from ASCII-8BIT to UTF-8)`. Since I found only very old OEmbed caches with this problem (newest from 2012), I think we can just remove these. When I create a new oEmbed cache for the same URL it creates it without `!binary`.

`== 20170917163640 CleanupInvalidOEmbedCaches: migrated (14.2081s) =============`


skyschub (approved):

SuperTux88:
Merged as b8fb4b62515b8bdcfb1052df7d51afef0ba35966

Thank you for the review @denschub 🍪

PR #7622, opened 2017-09-17
Don't change the filename of the export everytime the user is touched

SuperTux88:
Otherwise the filename is reseted when the user tries to download it, because `last_seen` time is touched ... and the user can't download the archive because the new filename doesn't exist anymore ...


skyschub (approved):

SuperTux88:
Merged as 95fbd3fc66bd9a01099c4b4777921ac065cc2191

PR #7624, opened 2017-09-22
Create birthday notification and mailer

Rete2:
There is one rubocop complaint left, which refers to the set_headers:
```
app/mailers/notification_mailers/contacts_birthday.rb:7 W: Do not prefix writer method names with `set_`.
```

It needs to stay like this because this method is called in the base.rb for all notification mailers.

### DIFF
--- a/app/mailers/notification_mailers/contacts_birthday.rb
+++ b/app/mailers/notification_mailers/contacts_birthday.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module NotificationMailers
+  class ContactsBirthday < NotificationMailers::Base
+    attr_accessor :person
+
+    def set_headers(person_id)
+      @person = Person.find(person_id)
+      @headers[:subject] = I18n.t("notifier.contacts_birthday.subject", name: @person.name)
+    end
+  end
+end
--- a/app/models/notifications/contacts_birthday.rb
+++ b/app/models/notifications/contacts_birthday.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Notifications
+  class ContactsBirthday < Notification
+    def mail_job
+      Workers::Mail::ContactsBirthday
+    end
+
+    def popup_translation_key
+      "notifications.contacts_birthday"
+    end
+
+    def self.notify(contact, _recipient_user_ids)
+      recipient = contact.user
+      actor = contact.person
+      create_notification(recipient, actor, actor).try(:email_the_user, actor, actor)
+    end
+  end
+end
--- a/app/views/notifications/index.html.haml
+++ b/app/views/notifications/index.html.haml
@@ -30,6 +30,8 @@
                   %i.entypo-reshare
                 - when "started_sharing"
                   %i.entypo-add-user
+                - when "contacts_birthday"
+                  %i.entypo-calendar
               = t("." + key)
 
     .col-md-9.stream.notifications
--- a/app/views/notifier/contacts_birthday.markerb
+++ b/app/views/notifier/contacts_birthday.markerb
@@ -1,0 +1,6 @@
+
+<%= t(".birthday", name: @notification.person.name) %>
+
+[<%= t(".view_profile", name: @notification.person.name) %>][1]
+
+[1] <%= local_or_remote_person_path(@notification.person, absolute: true) %>
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -642,6 +642,10 @@ en:
       zero: "%{actors} have reshared your post %{post_link}."
       one: "%{actors} has reshared your post %{post_link}."
       other: "%{actors} have reshared your post %{post_link}."
+    contacts_birthday:
+      zero: "%{actors} have their birthday on %{date}."
+      one: "%{actors} has their birthday on %{date}."
+      other: "%{actors} have their birthday on %{date}."
     also_commented_deleted:
       zero: "%{actors} commented on a deleted post."
       one: "%{actors} commented on a deleted post."
@@ -676,6 +680,7 @@ en:
       mentioned_in_comment: "Mentioned in comment"
       reshared: "Reshared"
       started_sharing: "Started sharing"
+      contacts_birthday: "Birthday"
       no_notifications: "You don't have any notifications yet."
       and_others:
         zero: "and nobody else"
@@ -720,6 +725,10 @@ en:
     reshared:
         reshared: "%{name} reshared your post"
         view_post: "View post >"
+    contacts_birthday:
+        subject: "%{name} has their birthday today"
+        birthday: "%{name} has their birthday today. Wish them 'Happy Birthday'!"
+        view_profile: "View %{name}’s profile"
     confirm_email:
         subject: "Please activate your new email address %{unconfirmed_email}"
         click_link: "To activate your new email address %{unconfirmed_email}, please follow this link:"
@@ -1213,6 +1222,7 @@ en:
       comment_on_post: "someone comments on your post"
       also_commented: "someone comments on a post you’ve commented on"
       private_message: "you receive a private message"
+      birthday: "someone has their birthday"
       download_export: "Download my profile"
       request_export: "Request my profile data"
       request_export_update: "Refresh my profile data"
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -104,7 +104,8 @@ class NotificationsController < ApplicationController
       "mentioned"            => "Notifications::MentionedInPost",
       "mentioned_in_comment" => "Notifications::MentionedInComment",
       "reshared"             => "Notifications::Reshared",
-      "started_sharing"      => "Notifications::StartedSharing"
+      "started_sharing"      => "Notifications::StartedSharing",
+      "contacts_birthday"    => "Notifications::ContactsBirthday"
     }
   end
   helper_method :types
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -16,6 +16,8 @@ module NotificationsHelper
       elsif %w(Notifications::CommentOnPost Notifications::AlsoCommented Notifications::Reshared Notifications::Liked)
             .include?(note.type)
         opts.merge!(opts_for_post(note.linked_object))
+      elsif note.is_a?(Notifications::ContactsBirthday)
+        opts.merge!(opts_for_birthday(note.linked_object))
       end
     end
     translation(target_type, opts)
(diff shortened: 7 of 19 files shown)


app/mailers/notification_mailers/contacts_birthday.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module NotificationMailers
+  class ContactsBirthday < NotificationMailers::Base
+    attr_accessor :contact
+
+    def set_headers(contact_id)
```
diaspora-code-review: Do not prefix writer method names with `set_`.

Rete2: As stated above, we think this should stay. By the way, this PR is ready for review :)

app/models/notifications/contacts_birthday.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def popup_translation_key
+      "notifications.contacts_birthday"
+    end
+
+    def self.notify(contact, _recipient_user_ids)
+      target = contact.person.profile
+      recipient = contact.user
+      actor = contact.person
+      create_notification(recipient, target, actor).try(:email_the_user, target, actor)
```
SuperTux88: You use the `contact_id` in the `ContactsBirthday` mailer, but you use the `target` (which is the `profile`) here for `email_the_user`. This doesn't work (unless the profile id and contact id are the same by coincidence). But since you don't need the contact in the mail (you only use `contact.person` there), you can use the `person` here directly and select the Person instead of the Contact in the mailer.

I also don't know why you used the `profile` as target here? Why don't you just use the `person` for everything?

SuperTux88 (changes requested):
Sorry for the delay, here is some feedback about the PR. Thanks for your work 👍

config/locales/diaspora/en.yml
```diff
@@ -627,6 +627,10 @@ en:
       zero: "%{actors} have reshared your post %{post_link}."
       one: "%{actors} has reshared your post %{post_link}."
       other: "%{actors} have reshared your post %{post_link}."
+    contacts_birthday:
+      zero: "%{actors} have their birthday on %{date}."
+      one: "%{actors} has their birthday on %{date}."
+      other: "%{actors} have their birthday on %{date}."
```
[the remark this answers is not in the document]
netagonen: So should we leave it like this and open a new issue about combining several birthday notifications together?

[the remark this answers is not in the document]
SuperTux88: Yes, that sounds like a good idea. 👍

app/views/notifications/index.html.haml
```diff
@@ -30,6 +30,8 @@
                   %i.entypo-reshare
                 - when "started_sharing"
                   %i.entypo-add-user
+                - when "contacts_birthday"
+                  %i.entypo-heart
```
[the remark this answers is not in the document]
netagonen: We also wanted to use the cake in the first place, but it didn't work. So we used the calendar now. Thanks for the suggestion.

app/models/notifications/contacts_birthday.rb
netagonen: we are completely  confused about this...

We looked at this again for a while now, but it seems to be working. We are calling `Notifications::ContactsBirthday.notify(contact, [])` in the console when the `contact.id` and `profile.id` are different and it seems to create the notification and the email with the correct header:

screenshot from 2017-09-28 14-27-38

We use contact in the mail in order to have access to the person that has the birthday (since this email doesn't really have a sender, we don't use the `@sender` variable defined in the `base::initialize`)

We used profile as a target because that object has the birthday, and it's cosistent with other notifications targets where the target is the thing the notification is about (like `like`, `mention`). We realize now that we might not have to pass the profile but when we removed it doesn't work and we get this error:

```
ActiveRecord::AssociationTypeMismatch: User(#47106434963060) expected, got #<User id: 3, username: "neta", serialized_private_key: "-----BEGIN RSA PRIVATE KEY-----\nMIIJKQIBAAKCAgEA4q...", getting_started: true, disable_mail: false, language: "en", email: "neta@iamneta.com", created_at: "2017-09-28 12:08:17", updated_at: "2017-09-28 12:46:28", invited_by_id: nil, authentication_token: "htxYVY2uw2wgUZyscwUHvrrR9_75Dw", confirm_email_token: nil, show_community_spotlight_in_stream: true, auto_follow_back: false, auto_follow_back_aspect_id: nil, hidden_shareables: {}, last_seen: "2017-09-28 12:46:28", remove_after: nil, export: "neta_diaspora_data_YbACeaC1WcASvUy_XrBL6Q.json.gz", exported_at: nil, exporting: false, strip_exif: true, exported_photos_file: nil, exported_photos_at: nil, exporting_photos: false, color_theme: "original", post_default_public: false> which is an instance of User(#47106419171140)
```

Now we are even more confused... any suggestions to how to figure this out?

SuperTux88: > We used profile as a target because that object has the birthday, and it's consistent with other notifications targets where the target is the thing the notification is about (e.g. `like`, `mention`). We realize now that we don't have to pass the profile so instead we removed the target variable completely and passed the actor twice (which is the person), hope that's better.

The person is better, because even when the profile has the birthday attribute, the person has birthday, not the profile ;) Also we link everywhere to the person and never to the profile directly. So using `actor` twice is perfect here.

> We are calling `Notifications::ContactsBirthday.notify(contact, [])` in the console when the `contact.id` and `person.id` are different and it seems to create the notification and the email with the correct header

This:
```ruby
try(:email_the_user, target, actor)
```

goes here:

https://github.com/diaspora/diaspora/blob/82b51fe2be2e831a65901525d46b479326fb5e7b/app/models/notification.rb#L17-L19

So it calls `target.id` and target is the profile (or when you use `actor` twice here, it's the person).

This then goes here:

https://github.com/diaspora/diaspora/blob/82b51fe2be2e831a65901525d46b479326fb5e7b/app/models/user.rb#L343-L349

Which creates a `Mail::ContactsBirthday` job with the `args`. The first two args are `recipient_id` and `actor.id` (sender) from above, and the last argument is the `profile.id`.

This then goes here:

https://github.com/diaspora/diaspora/blob/82b51fe2be2e831a65901525d46b479326fb5e7b/app/workers/mail/notifier_base.rb#L8-L10

which calls

https://github.com/diaspora/diaspora/blob/82b51fe2be2e831a65901525d46b479326fb5e7b/app/mailers/notifier.rb#L58-L59

which then calls with the three arguments (`recipient_id`, `sender_id` and the last still `profile.id`). So in `args` remains only the last argument now.

https://github.com/diaspora/diaspora/blob/82b51fe2be2e831a65901525d46b479326fb5e7b/app/mailers/notification_mailers/base.rb#L13-L23

which finally calls

https://github.com/diaspora/diaspora/blob/82b51fe2be2e831a65901525d46b479326fb5e7b/app/mailers/notification_mailers/contacts_birthday.rb#L7-L10

with the profile id.

So if it works, only because the contact ID is the same as the profile ID, which was also the case on my local test pod (it's likely to happen on small test pods, when you have only two users the profile have ID 1 and 2 and the contacts too), after I stopped (completely in both directions) and started sharing again the contact had a new ID and the whole job failed.

But when you use the person here now (the `actor` twice, so it is `person.id`), the it's simply, you can just change the mailer to:

```ruby
    attr_accessor :person

    def set_headers(person_id)
      @person = Person.find(person_id)
      @headers[:subject] = I18n.t("notifier.contacts_birthday.subject", name: @person.name)
    end
```

and replace `@notification.contact.person` with `@notification.person` in the `contacts_birthday.markerb`, so it even gets simpler with the `person` instead of the `contact`.

I hope I could explains what happens here and it's clearer now what is/was the problem. Since I don't see what you already changed I can only explain it with the old state of this PR.

netagonen: Hi @SuperTux88 , thanks! We did follow this path but you made it much clearer now. I have made the changes and will push them soon.

config/locales/diaspora/en.yml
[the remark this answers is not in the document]
netagonen: Probably best after this will be merged, right?

netagonen:
Anything else needed to be fixed?

app/views/notifier/contacts_birthday.markerb
```diff
@@ -0,0 +1,6 @@
+
+<%= @notification.person.name %> <%= t('.birthday') %>
+
+[<%= t('.view_profile', :name => @notification.person.name) %>][1]
```
SuperTux88: I just saw that pronto doesn't check this file, but can you use double quotes and ruby 1.9 hash syntax here too? (in the whole file)

SuperTux88 (changes requested):
Sorry, I was really busy the last few weeks, and I used the little time I had for diaspora to fix the last bugs for 0.7.1.0 so we can finally release it.

Anyway, thanks for your changes, your PR looks almost ready. But travis is red and I found one little thing that we should change to make translations work better with different languages (and a little code style issue you can fix as drive-by when you change the translation).

netagonen:
Hi!

No problem, we will work on that soon.

config/locales/diaspora/en.yml
```diff
@@ -705,6 +710,10 @@ en:
     reshared:
         reshared: "%{name} reshared your post"
         view_post: "View post >"
+    contacts_birthday:
+        subject: "%{name} has their birthday today"
+        birthday: "has their birthday today. Wish them 'Happy Birthday'!"
```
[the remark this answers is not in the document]
netagonen: Just to make sure- do you mean in line 715?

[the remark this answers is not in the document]
SuperTux88: Yes, it should be `"%{name} has their birthday ..." like the line above, because maybe for some language the "name" is in the middle of the sentence.

netagonen:
Well, I pushed the changes, hope everything is ok now.

SuperTux88 (commented):
Thanks for updating the PR. I found one little thing that wasn't changed completely ... but since travis didn't trigger (I don't know why), I'll do a force-push to your branch to trigger travis, and while force-pushing anyway I'll fix that little leftover myself, so hopefully it will be ready to merge (when travis is green).

And you included a database backup in your commit, which I'll remove too, you can find it [here](https://github.com/diaspora/diaspora/blob/6cf133c994a20e855fe2249557a7000d912cb249/diaspora_dev.backup) if you still need it.

SuperTux88:
Ok, pronto ran with 10 minutes delay after the force-push, but still no travis after another 10 minutes of waiting ... I have no idea why travis doesn't start (it doesn't even show up [here](https://travis-ci.org/diaspora/diaspora/pull_requests)). Maybe @denschub can have a look why travis doesn't trigger in the admin settings (I don't have access there)?

netagonen:
Thanks!  Anyway, the  database backup in my commit were a mistake...

SuperTux88 (approved):
I did another force push now and this time travis triggered 🎉 and was green 💚 so I'll merge this soon, thanks for your work 🍪

SuperTux88:
Merged as fc33a2ac5dd494c8d1f59aa18baea1549b6a4d3d

Thanks again 🍪

Flaburgan:
Congratulations for your work @netagonen and @Rete2! A very useful feature just landed thanks to you!

netagonen:
YAY
thank you!!!

Issue #7615, opened 2017-09-14
Optimize profile export


SuperTux88:
After fixing diaspora/diaspora_federation#73, diaspora/diaspora_federation#75 and #7614 I was finally able to export my profile on my local computer. It took ~50 minutes and **10.8GB** RAM for sidekiq. While I don't see the time as a problem (it's an old profile, it takes some time), I think the RAM-usage is a problem. At least on my production pod I won't be able to export my profile with that RAM-usage, because the pod would be really fast out-of-memory. For my local test I restarted sidekiq before the export and the export was the only task it was doing during the 50 minutes. A production pod would also need to federate and process other task during that time. And I think that my pod will not be the only one having problems with this, especially other smaller pods which exists for some time (so they have a longer history).

The final export was 28.4MB small (gz) and after unzipping it was 105.3MB. So the export doesn't need 10.8GB. Maybe we can split the export in smaller parts and concat at the end, so we don't need to hold everything in memory?

But we need to do something before the release of 0.7.1.0 (that's why I added the `blocking-release` label), either we fix it or we should disable the export until we fixed it, otherwise this could DOS some pods.

cmrd-senya:
I tried running profile export from rails console with my test profile from a test database using these commands:
```ruby
u = User.find_by(username: "senya")
e = Diaspora::Exporter.new(u).execute
```

I used [ruby-mass](https://github.com/archan937/ruby-mass) gem to analyze objects lifetime. What I can see is that after running export all of the `ActiveRecord` objects that was used by the serializers are still alive:

```ruby
2.4.0 :010 > Mass.index ActiveRecord::Base
[...]

Issue #2999, opened 2012-03-13
Users should be able to like comments again


ghost:
I haven't found an issue for that ;)
##

qm3ster:
Heart requires no translation, and in Russian 'Like' is much dumber than +1 (Which people use in normal conversation sometimes.)
Also requesting this feature, convinced it should not cause bloat. :sweet_potato:

Flaburgan:
@qm3ster as said by @jaywink https://github.com/diaspora/diaspora/issues/4920 is a blocker for this issue. Hopefully https://github.com/diaspora/diaspora/pull/6586 will be enough to close it.

cmrd-senya:
How about removing `author_signature` for relayables which are not on the home pod of the author of a parent post? It'll be the best we can do to reduce the DB bloat.

@SuperTux88, doesn't your refactoring includes this change by any chance?

SuperTux88:
Yes, I'm working on removing unused signatures

cmrd-senya:
Great, then fixing this one should be a piece of cake

Hypolite:
+1 for this feature. I like being able to notice someone I liked their comment without having to add a comment myself if I don't have anything meaningful to add to the conversation.

Flaburgan:
Nice news, the new 0.6 version now released drastically reduced the database size. This feature could be activated again pretty easily I guess (I didn't look at the code at the moment). Does anyone here want to work on it or can I add the "help wanted" label on this one?

cmrd-senya:
Looks attractive. I don't know whether I can dedicate enough time to that. I'm definetely interested to participate, but not right now.

SuperTux88:
It is only a frontend thing, the backend is still functional and likes on comments from friendica are stored and relayed.

SuperTux88:
At least I think it still works, I haven't tested it, but I have likes for comments from friendica in my DB. I don't know how relaying works, because the parent is the comment, maybe this needs improvements on the federation.

Timoses:
+1!

Flaburgan:
@SuperTux88 any chance you can check the current state of the federation regarding this? It would allow a possible contributor to know if he can solve it, a simple front-end or a modification of the federation are two very different tasks :)

SuperTux88:
The federation already works in some cases (relaying is the tricky part here), and it should be easy to make it fully functional again. I already planned to fix this (because friendica already have likes on comments), but didn't have the time yet.

[...]

PR #7625, opened 2017-09-23
Upgrade federation gem, diaspora:// links, RFC 7033 Webfinger, relay likes for comments

### LINKED ISSUE #5963: Implement RFC 7033 Webfinger

SuperTux88:
First: This PR contains multiple changes, the main reason for this is that all of these rely on the new version of the `diaspora_federation` gem, and we can't merge this until #7615 is fixed, because the new federation gem fixes bugs that are currently preventing a OOM. The different changes are in different commits, so I hope it's OK.

The biggest/main change is support for diaspora:// links. I created a `references` table (create table migrations are fast, so it shouldn't be a problem for a minor release), where all valid references are stored. We can use this in the future to:
* list all posts that reference another post
* notify the author of the referenced post
* embed referenced posts (without parsing the text every time)
* display a different link (which for example only opens a popup or something like this), when the viewing person can't see the referenced post (for example when it is private), so we don't link to a 404 page.

But that are all plans for the future, here I add only basic support for forward compatibility (diaspora:// links are converted to clickable links and not displayed as dead links), so the `references` table is not used yet, but I think it's still good to already store the information when receiving posts with diaspora:// links.

I first tried to handle diaspora:// links with linkify-it in the desktop view, but sadly linkify-it only handles plain links, not when they are embedded in markdown (`[foo](diaspora://...)` didn't work), so we probably need to develop a plugin similar to the mention plugin to handle diaspora:// links in the preview. But since this PR only adds support to receive and display diaspora:// links, this is something that can be handled in the future.

Fixes #5963, adds forward compatibility for #2999,  #2884 and diaspora/diaspora_federation#75


skyschub (approved):

SuperTux88:
Merged as many commits, thanks @denschub for the review 🍪

PR #7627, opened 2017-09-25
Optimize memory usage on profile export

### LINKED ISSUE #7615: Optimize profile export

cmrd-senya:
fix #7615 

see commit descriptions for details


spec/support/fixture_builder.rb
```diff
@@ -1,47 +1,49 @@
 # frozen_string_literal: true
 
-FixtureBuilder.configure do |fbuilder|
+def create_basic_users
```
diaspora-code-review: Assignment Branch Condition size for create_basic_users is too high. [38.28/20]

SuperTux88 (approved):
I tested this again with my profile, and it used 1.2GB RAM and 35 minutes. This is still not perfect and can hopefully be improved in the future, but I think it's enough for the average profile on most pods.

SuperTux88:
Merged as c6ed850a8532049bb6a65233795334be1c9952f4, ea57fc5d25c60c3e11efd704a928b1b80f89e1c5, 9f0b74ebbb82dff16f19334c179eb9eb1abc1f3b and 265a7ee2530999f17cdec5c0f09934b776dad5f3

Thanks for the work @cmrd-senya 🍪

PR #7629, opened 2017-09-28
Limit the number of parallel exports that are allowed to run

SuperTux88:
As I suggested in https://github.com/diaspora/diaspora/issues/7615#issuecomment-330380452, I added a check to limit the number of parallel export jobs. Because the export still takes a lot of memory for a bigger profile (it's better now, but there is still room for more optimization), so it could be a problem when there are many exports running in parallel. I added a setting to make it configurable, I set the default to only 1, to protect small servers, big pods (with more users and more resources) probably need to increase this, because otherwise the exports could queue up.

### DIFF
--- a/app/workers/export_user.rb
+++ b/app/workers/export_user.rb
@@ -4,12 +4,25 @@
 #   licensed under the Affero General Public License version 3 or later.  See
 #   the COPYRIGHT file.
 
-
 module Workers
   class ExportUser < Base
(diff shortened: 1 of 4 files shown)


skyschub (approved):

SuperTux88:
Merged as fd36517dee51d9d78796d81242f91a72f65165b2

PR #7630, opened 2017-09-30
Delete invalid diaspora IDs from friendica

SuperTux88:
Friendica allows to be installed in a subfolder, but diaspora doesn't support this. In the past receive worked (send never worked), today diaspora is more strict and blocks these entities completely. But because of that we have now still some entities from the past in the database which can fail the export. The best solution is just to delete the invalid people (the can't be recreated with the current validations), because otherwise people could always create new invalid data.

I didn't create migration models for this migration, because otherwise I would have needed to copy all models (and even add some hacks for polymorphic relations, because migration models are namespaced) ...

On a newer pod this migration will just do nothing and exit.

```
== 20170928233609 CleanupInvalidDiasporaIds: migrated (13.2132s) ==============
```

### DIFF
--- a/app/models/aspect_membership.rb
+++ b/app/models/aspect_membership.rb
@@ -5,7 +5,6 @@
 #   the COPYRIGHT file.
 
 class AspectMembership < ApplicationRecord
-
   belongs_to :aspect
   belongs_to :contact
   has_one :user, :through => :contact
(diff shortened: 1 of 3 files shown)


spixi:
This is a good idea, because usernames with slash could severe issues. e. g. in the atom feed link.

skyschub (approved):

SuperTux88:
Merged as 32067246df724b7f6455f3ef95f747391c48d043

PR #7632, opened 2017-10-01
Handle bugs related to missing user DB entries

### LINKED ISSUE #7631: AccountDeletion raises exception on performing
Currently, `AccountDeletions` seem to be broken. I have several unperformed deletions, and when triggering them manually, they fail with:

```
NoMethodError: undefined method `disconnect' for nil:NilClass
	from app/models/aspect_membership.rb:16:in `block in <class:AspectMembership>'
	from lib/account_deleter.rb:100:in `delete_contacts_of_me'
	from lib/account_deleter.rb:32:in `block in perform!'
	from lib/account_deleter.rb:28:in `perform!'
	from app/models/account_deletion.rb:23:in `perform!'
	from (irb):6
```

Can't figure out when this regressed, but we should probably have a look at this...

cmrd-senya:
fixes #7631

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -90,6 +90,10 @@ class User < ApplicationRecord
 
   after_save :remove_invalid_unconfirmed_emails
 
+  before_destroy do
+    raise "Never destroy users!"
+  end
+
   def self.all_sharing_with_person(person)
     User.joins(:contacts).where(:contacts => {:person_id => person.id})
   end
--- a/db/migrate/20171012202650_cleanup_invalid_contacts.rb
+++ b/db/migrate/20171012202650_cleanup_invalid_contacts.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+class CleanupInvalidContacts < ActiveRecord::Migration[5.1]
+  class Contact < ApplicationRecord
+    belongs_to :user
+    belongs_to :person
+
+    has_many :aspect_memberships, dependent: :delete_all
+
+    before_destroy :destroy_notifications
+
+    def destroy_notifications
+      Notification.where(
+        target_type:  "Person",
+        target_id:    person_id,
+        recipient_id: user_id,
+        type:         "Notifications::StartedSharing"
+      ).destroy_all
+    end
+  end
+
+  class User < ApplicationRecord
+  end
+
+  class Person < ApplicationRecord
+    belongs_to :owner, class_name: "User", optional: true
+  end
+
+  class Notification < ApplicationRecord
+    self.inheritance_column = nil
+    has_many :notification_actors, dependent: :delete_all
+  end
+
+  class NotificationActor < ApplicationRecord
+  end
+
+  class AspectMembership < ApplicationRecord
+  end
+
+  def up
+    Contact.left_outer_joins(:user).where("users.id is NULL").destroy_all
+    Contact.left_outer_joins(person: :owner).where("people.owner_id is NOT NULL").where("users.id is NULL").destroy_all
+  end
+end
--- a/app/models/aspect_membership.rb
+++ b/app/models/aspect_membership.rb
@@ -12,9 +12,7 @@ class AspectMembership < ApplicationRecord
   has_one :person, :through => :contact
 
   before_destroy do
-    if self.contact && self.contact.aspects.size == 1
-      self.user.disconnect(self.contact)
-    end
+    user&.disconnect(contact) if contact&.aspects&.size == 1
     true
   end
 
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -6,7 +6,7 @@
 #
 class Notification < ApplicationRecord
   belongs_to :recipient, class_name: "User"
-  has_many :notification_actors, dependent: :destroy
+  has_many :notification_actors, dependent: :delete_all
   has_many :actors, class_name: "Person", through: :notification_actors, source: :person
   belongs_to :target, polymorphic: true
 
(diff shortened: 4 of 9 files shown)


cmrd-senya:
According to travis, the "improperly deleted users" can affect the account migration as well.

cmrd-senya:
Looks like `User#destroy` was used as account deletion (close) method in the past. I wonder if we actually need a migration which removes the dangling contacts.

app/models/user.rb
```diff
@@ -90,6 +90,10 @@ class User < ApplicationRecord
 
   after_save :remove_invalid_unconfirmed_emails
 
+  before_destroy do
+    raise "Never destroy users!"
```
SuperTux88: +1 for this one, so podmins don't try to destroy the user via rails console.

SuperTux88 (commented):
> I wonder if we actually need a migration which removes the dangling contacts.

Yes, we should add a migration for that. And with a migration I'm not sure if we need the check anymore, because the user should always be there in the future ... but since the code is simpler now than before I think the check doesn't hurt ;)

SuperTux88 (changes requested):
@cmrd-senya can you add a migration to remove contacts with missing users? You can also remove contacts from other users with local people of missing users, then you can remove the commit "Drop contacts where respective user entry is missing".

cmrd-senya:
Updated

SuperTux88 (approved):
(wrong button ... and I can't delete reviews ...)

SuperTux88 (changes requested):
Can you please add migration models to the migration? Otherwise it looks good 👍 

(Turns out I have some missing users from 2011 too)

cmrd-senya:
Updated

db/migrate/20171012202650_cleanup_invalid_contacts.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class CleanupInvalidContacts < ActiveRecord::Migration[5.1]
+  class Contact < ApplicationRecord
+    belongs_to :user
+    belongs_to :person
+
+    has_many :aspect_memberships, dependent: :destroy
+    has_many :aspects, through: :aspect_memberships
```
SuperTux88: This line isn't needed

db/migrate/20171012202650_cleanup_invalid_contacts.rb
```diff
@@ -0,0 +1,41 @@
...
+        target_type:  "Person",
+        target_id:    person_id,
+        recipient_id: user_id,
+        type:         "Notifications::StartedSharing"
+      ).destroy_all
+    end
+  end
+
+  class User < ApplicationRecord
+    has_one :person, inverse_of: :owner, foreign_key: :owner_id
```
SuperTux88: Since you join from contacts via person to user, you should add the person model with the user relation. This line can be removed then, because it isn't needed.

db/migrate/20171012202650_cleanup_invalid_contacts.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  class Notification < ApplicationRecord
+    has_many :notification_actors, dependent: :destroy
```
SuperTux88: Since `NotificationActor` doesn't do anything special on destroy, you can use `dependent: :delete_all` here, which is faster, because it doesn't instantiate the model (we could change that in the real model too).

SuperTux88 (changes requested):
Only some small things :)

db/migrate/20171012202650_cleanup_invalid_contacts.rb
cmrd-senya: Maybe we can use `:delete_all` in `Contact` `:aspect_memberships` relation in migration model as well?

SuperTux88: Sure, that makes sense 👍

cmrd-senya:
Updated

SuperTux88 (changes requested):
Just tested it locally with my database and got an error:

```
Invalid single-table inheritance type: Notifications::StartedSharing is not a subclass of CleanupInvalidContacts::Notification
```

So it needs to also have this as a model:

```ruby
  module Notifications
    class StartedSharing < Notification
    end
  end
```

SuperTux88:
OK, that doesn't work too :(

SuperTux88:
OK, I did a different solution and pushed it directly to the PR:

```diff
diff --git a/db/migrate/20171012202650_cleanup_invalid_contacts.rb b/db/migrate/20171012202650_cleanup_invalid_contacts.rb
index c0b84c4b9..709724723 100644
--- a/db/migrate/20171012202650_cleanup_invalid_contacts.rb
+++ b/db/migrate/20171012202650_cleanup_invalid_contacts.rb
@@ -27,6 +27,7 @@ class CleanupInvalidContacts < ActiveRecord::Migration[5.1]
   end

   class Notification < ApplicationRecord
+    self.inheritance_column = nil
     has_many :notification_actors, dependent: :delete_all
   end
```
Polymorphic associations and single table inheritance always are bad with migration models :(

skyschub (approved):
Only approving @SuperTux88's changes, assuming the rest is reviewed by him.

SuperTux88 (approved):

SuperTux88:
Yes, everything else looks good, so merged as 2e9c6f46dc6c01a63123445da876f23780b3398d, ad025850cece51672257929ab9e54b93f7fa70aa, 973e9d98c60226d216ecdd5874311e7b03e1445d and f212b51f7f83f63cb59dff45f38f82480bd7bb14

Thank you @cmrd-senya for the work 🍪 and thanks to @denschub for the quick review of the last fix 🍪

PR #7637, opened 2017-10-10
Cleanup relayables where the signature is missing

SuperTux88:
When we should have the signature but don't have it, the user data export fails. There are a few comments from back in 2011 where the signature is missing.

Also some podmins maybe messed with signatures in their database (#4920), which would also break the exports now.

Running with a copy of my production DB (and nothing deleted which should be the case with a clean DB):
`== 20171009232054 CleanupRelayablesWithoutSignature: migrated (8.3341s) =======`

### DIFF
--- a/db/migrate/20171009232054_cleanup_relayables_without_signature.rb
+++ b/db/migrate/20171009232054_cleanup_relayables_without_signature.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+class CleanupRelayablesWithoutSignature < ActiveRecord::Migration[5.1]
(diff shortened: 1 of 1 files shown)


denschub:
> Also some podmins maybe messed with signatures in their database (#4920), which would also break the exports now.

Since we explicitly said that this is a horrible idea and should not be done, I'm fine with dropping invalid relayables now.

skyschub (approved):

SuperTux88:
Merged as 2711b9dc831d8c58bc9ba7c1bd917e6e16aa107e

Thanks you for the review @denschub 🍪

PR #7639, opened 2017-10-12
Reduce account deletion memory usage

SuperTux88:
This massively reduces the memory usage for account deletions (which could cause OOMs when deleting many big accounts in parallel). With this PR the sidekiq process uses only 180MB ram after deleting a rss-bot account with 17k posts (after start sidekiq uses 150MB, so this is only 30MB more than after start). Without this PR the same account deletion used 1.8GB ram.

I also did some cleanups in the `AccountDeleter` and added a check for duplicate account deletions.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,13 @@
 
 # 0.7.1.0
 
+## Ensure account deletions are run
+
+There were some issues causing accounts deletions to not properly perform in some cases, see
+[#7631](https://github.com/diaspora/diaspora/issues/7631) and [#7639](https://github.com/diaspora/diaspora/pull/7639).
+To ensure these are reexecuted properly, please run `RAILS_ENV=production bin/rake migrations:run_account_deletions`
+after you've upgraded.
+
 ## Refactor
 * Remove title from profile photo upload button [#7551](https://github.com/diaspora/diaspora/pull/7551)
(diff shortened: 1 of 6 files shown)


goobertron:
Whooooooooooo! 

Another fantastic performance improvement. Thanks!

denschub:
I wonder if we should create a short rake task to run old account deletions, so people can re-run them. What do you think?

skyschub (changes requested):
As discussed on a private channel, let's add a rake task. Nobody will probably see it, and we can only put it into the 0.8.0.0 upgrade instructions, but it does not cause any harm if we ship it earlier.

Other than that, code looks fine.

SuperTux88:
I added the rake task, but this also depends on #7632 to fix another account deletion related bug.

skyschub (approved):

SuperTux88:
Merged as 2bd9c663c509f4f5d1ccae0e20fa13e220510627, 6d5647ec112d3fd0c93d7f217dc2f6beed9fde04, b920ddbff55dd0facd54da884943d16e568860f2, f704f305727feeb9fa528de6a08ded56d40674d2, f4902421eaa50dea98dc5046848f95c959cb3c88 and 4e6d92ce63c96d577cef04e338bed8a17372b635

PR #7642, opened 2017-10-17
Do not autocomplete change password and close account password input fields

### LINKED ISSUE #7633: Password field for `close account' should have autocomplete="off"
https://github.com/diaspora/diaspora/blob/1fdc17d96cf5190a9e9ac069510c1985608a7a58/app/views/users/_close_account_modal.haml#L37

The password field in the `close account' form should have autocomplete="off".

Flaburgan:
Fixes #7633 

This adds autocomplete="off" to the input fields. This is how [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password#Allowing_autocomplete) indicates it should be done but weirdly it didn't work for me: Firefox is still autocompleting the input and Chromium never was... Can anyone else please test?

### DIFF
--- a/app/views/users/_close_account_modal.haml
+++ b/app/views/users/_close_account_modal.haml
@@ -34,7 +34,8 @@
           = f.error_messages
           .form-group
             = f.label :close_account_password, t("users.edit.current_password"), for: :close_account_password
-            = f.password_field :current_password, id: :close_account_password, class: "form-control"
+            = f.password_field :current_password, id: :close_account_password, class: "form-control",
+              autocomplete: "current-password"
         .modal-footer
           .btn.btn-default{type: "button", data: {dismiss: "modal"}, aria: {hidden: "true"}}
             = t("cancel")
--- a/app/views/users/_edit.haml
+++ b/app/views/users/_edit.haml
@@ -39,18 +39,18 @@
             = f.label :current_password, t(".current_password"), class: "col-sm-6 control-label"
             .col-sm-6
               = f.password_field :current_password, placeholder: t(".current_password_expl"),
-                class: "form-control"
+                class: "form-control", autocomplete: "current-password"
           .form-group
             = f.label :password, t(".new_password"), class: "col-sm-6 control-label"
             .col-sm-6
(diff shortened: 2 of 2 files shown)


SuperTux88:
The MDN sounds like it's a bad idea to do this (which could also explain while it isn't working?):

> Note that some software ignores this value, since it's typically harmful to users' ability to maintain safe password practices.

I agree with this, when users want to use a password manager to use a secure password, we shouldn't try to block that (it's the same as we shouldn't create rules like "the password should contain x and y, but shouldn't contain z"). So for the change password form, we should probably use `current-password` and `new-password` instead of `off`.

For the "close account"-scenario I understand that we want the user to input something and not just click the button. But another possible solution could be to add a checkbox that the user needs to check, or a second input field where the user needs to input "I'm sure, close the account!" (this isn't a replacement for the password, it would be an additional thing, because we can't know if the `autocomplete="off"` is ignored or not).

Flaburgan:
Hm for the close account, the idea to me was to avoid to see the account erased by someone who has access to the account without knowing the password (found a lost phone connected, for example).

SuperTux88:
Yes, that's why we should keep the password. But when somebody founds a lost phone with an unlocked password manager that's not our problem. Because for example when you use the internal password manager of firefox or chrome (without master password and without lockscreen), then you can just show the password in the settings and don't need the autocompletion. And the damage from stealing the passwords is probably bigger than a closed diaspora account (and nothing we can protect people from).

So yes, the password is important there, but when people want to use password managers, it's their job to protect their password managers, it's not our job (to try) to block them from using the password manager.

Flaburgan:
PR updated according the discussion on IRC and [the spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls:-the-autocomplete-attribute:autofill-expectation-mantle-3)

SuperTux88:
Merged as b156c2f1a343854180549706484f543776301510 (I changed the commit-message myself, because the old commit-message wasn't true anymore. But I need to edit the commit anyway when adding the changelog, so I hope that's OK.)

Thank you. 🍪

PR #7646, opened 2017-10-18
Add some progress indicators to the account deletions rake task

SuperTux88:
@Flaburgan asked if the rake task is stuck, because it doesn't output anything until it's finished (which can take a while when there are multiple accounts or big accounts needed to be deleted). So I thought it's maybe a good idea to print some stuff.

### DIFF
--- a/lib/tasks/migrations.rake
+++ b/lib/tasks/migrations.rake
@@ -63,11 +63,22 @@ namespace :migrations do
   desc "Run uncompleted account deletions"
   task run_account_deletions: :environment do
     if AccountDeletion.uncompleted.count > 0
-      puts "Running account deletions.."
-      AccountDeletion.uncompleted.find_each(&:perform!)
+      puts "Running account deletions..."
+      AccountDeletion.uncompleted.find_each do |account_deletion|
+        print "Deleting #{account_deletion.person.diaspora_handle} ..."
+        progress = Thread.new {
+          loop {
+            sleep 10
+            print "."
+          }
+        }
+        account_deletion.perform!
+        progress.kill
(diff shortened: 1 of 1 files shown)


Flaburgan:
Thanks! Can you please also add an indication in the changelog that it can take a while, and maybe should be run in tmux?

SuperTux88:
It doesn't need to be run in tmux, when it stops or is canceled (Ctrl-C) it just continues where it's stopped after restart. The account deletion is only marked as completed when it's completed.

lib/tasks/migrations.rake
```diff
@@ -63,11 +63,22 @@ namespace :migrations do
...
+          }
+        }
+        account_deletion.perform!
+        progress.kill
+        puts " Done"
+      end
       puts "OK."
     else
-      puts "No acccount deletions to run."
+      puts "No account deletions to run."
```
denschub: But... the hidden phun...!

skyschub (approved):

SuperTux88:
Merged as c9423bfdd203bf8f127dd7bfa13d67a0b96d6337

@Flaburgan what I wanted to say was, that you don't need tmux, but if you have an unstable connection you maybe use tmux for the upgrade anyway and then you can continue using tmux. But if something happens you can just restart it and it continues where it's stopped, so nothing bad happens :)

lib/tasks/migrations.rake
SuperTux88: I copy pasted that from an older rake task ... and while adding the progress indicators the spell-checker of my IDE told me that there is something wrong here ;)

PR #7568, opened 2017-08-22
Order notifications by updated_at instead of created_at

svbergerem:
**before**
order by created_at

**after**
bildschirmfoto von 2017-08-22 16-34-28

PR #7648, opened 2017-10-20
Group notifications by updated_at

### LINKED ISSUE #7647: Incorrect date displayed in notifications list
Bit of an odd one this: I found dates out of order in the notifications list on pod.diaspora.software.

out-of-order

On investigation, some of the timestamps don't match the date listed. The four notifications marked **13 October** are in fact listed in order (albeit the first two under the wrong displayed date): from the top, their timestamps are:
- 15/10/27, 17:36:14
- 15/10/27, 04:52:06
- 13/10/27, 15:49:38
- 13/10/27, 15:44:00

The one listed below as **15 October** is in fact out of order; its timestamp is:
- 15/10/27, 05:13:50

Any ideas why this might have happened? Pinging podmin @Zauberstuhl in case it's a configuration issue.

SuperTux88:
The notifications need to be grouped by the same date as they are sorted, otherwise the date used for the group doesn't match all timestamps in the group and also the groups are sorted by the wrong date.

This fixes #7647, a regression of #7568.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -40,7 +40,7 @@ class NotificationsController < ApplicationController
 
       pager.replace(result)
     end
-    @group_days = @notifications.group_by{|note| note.created_at.strftime('%Y-%m-%d')}
+    @group_days = @notifications.group_by {|note| note.updated_at.strftime("%Y-%m-%d") }
 
     @unread_notification_count = current_user.unread_notifications.count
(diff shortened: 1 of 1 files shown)


skyschub (approved):

SuperTux88:
Merged as 0f688b9d7317f06b22915ef4d968190f00500ae4

Thanks @denschub for the fast review 🍪

PR #7654, opened 2017-10-28
Fix compatibility with Bundler 1.6

### LINKED ISSUE #7653: Incompatibility with Bundler 1.6
When Bundler 1.6 gets released, we most likely have an issue. We depend on `Bundler::settings#with` in our application initializer here:

https://github.com/diaspora/diaspora/blob/635139919775f8eed8aa8dd845a0575fe01d3706/config/application.rb#L9

But this function got removed in https://github.com/bundler/bundler/commit/8ccf49c3b4a086f1e0cad2130dc30c04304175b7, so we throw a no method error for people who *accidentally* upgraded to 1.6. Given the PR behind it, https://github.com/bundler/bundler/pull/5811, this may or may not be an accident, but since we use internal APIs which were never documented, this is pretty much our fault and not Bundler's. :)

We should be able to replace the function call with `Bundler.settings[:with]`. Maybe, there is an even nicer option.

SuperTux88:
Stop using `Bundler.settings.with`, because it will be removed from Bundler 1.6.

Also, as described in #7653, we could use `Bundler.settings[:with]`, but that would be internal API again, so it probably breaks again in the future. That's why I added a `BundlerHelper` module to parse the required optional group from our config files, without the use of any internal Bundler API. That only breaks when we change something in our configs, but then we can change the `BundlerHelper` too, and we're not surprised any more by any new Bundler versions.

The `BundlerHelper` is also used for the new `script/configure_bundler` script to generate the correct config needed for `bin/bundle install`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,13 +24,11 @@ branches:
     - 'develop'
 
 before_install:
-  - gem install bundler
+  - script/ci/prepare.sh
   - mkdir travis-phantomjs
   - wget http://cifiles.diasporafoundation.org/phantomjs-2.1.1-linux-x86_64.tar.bz2 -O $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2
   - tar -xvf $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2 -C $PWD/travis-phantomjs
   - export PATH=$PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH
-
-bundler_args: "--deployment --without development --with mysql postgresql --jobs 3 --retry 3"
 
 script: "./script/ci/build.sh"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
   - tar -xvf $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2 -C $PWD/travis-phantomjs
   - export PATH=$PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64/bin:$PATH
 
-script: "./script/ci/build.sh"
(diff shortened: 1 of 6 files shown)


jhass:
Now I start to wonder, can we maybe just list all our optional groups explicitly and bundler would just ignore the not installed ones? I never tried!

SuperTux88:
It looks like that works with the current bundler version and it doesn't fail when you require a group that isn't installed. However it looks very dirty and I think that is something that can break again with a future bundler version when they start to validate that required groups should be installed. So I think it's better to only require the needed groups where we are sure they are installed.

SuperTux88:
I think it's ready for review. I also used the new `script/configure_bundler` script on travis, so we see it works.

As a drive-by I removed the `script/ci/build.sh` and call `bin/rake` directly from the `.travis.yml`, because we don't need `xvfb` anymore with poltergeist.

After this is released we should update our installation and upgrade guides to use the new script, because as described in https://github.com/diaspora/diaspora/issues/7653#issuecomment-340204893 the current way with `--with ...` won't work anymore with Bundler 2.0.

SuperTux88:
Bundler just [released 1.16.0](https://github.com/bundler/bundler/releases/tag/v1.16.0) ... so from now on everybody updating diaspora will have problems :/ Should we hotfix this PR like we did it for [0.6.0.1](https://github.com/diaspora/diaspora/releases/tag/v0.6.0.1)?

skyschub (approved):

SuperTux88:
Merged as 7c4648030f4256c8e405db470c0f3d77cd225add, 4e267bb1efdd0a295c3b6909f74cf9c60b6856a4, a14115119c643c00f19366be94251ded1be6ec32 and 63fcc9c1bcfe5a5033344358fff13ab803f724a7, released as [0.7.1.1]()

Thanks @denschub for the research and review 🍪

PR #7655, opened 2017-10-30
Fix multiple photos upload progress bar

### LINKED ISSUE #6721: Upload progress-bar of multiple files working incorrectly
progress-bar which shows the status of upload seems to go backwards while uploading multiple images
whereas there is no issue while uploading a single file
on diaspora version Code 0.5.6.3-pad20bb05
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31365871-upload-progress-bar-of-multiple-files-working-incorrectly?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Flaburgan:
Fixes #6721 

Here is a quick [screencast](https://diaspora-fr.org/progress-upload-bug.ogv) showing the bug in the develop branch and then the bug fixed in that branch.

### DIFF
--- a/app/assets/javascripts/app/views/publisher/uploader_view.js
+++ b/app/assets/javascripts/app/views/publisher/uploader_view.js
@@ -7,108 +7,92 @@
 app.views.PublisherUploader = Backbone.View.extend({
   initialize: function(opts) {
     this.publisher = opts.publisher;
-
-    this.info = $("<div id=\"fileInfo\" />");
-    this.publisher.wrapperEl.before(this.info);
     this.publisher.photozoneEl.on("click", ".x", _.bind(this._removePhoto, this));
 
     // Initialize the PostPhotoUploader and subscribe its events
     this.uploader = new Diaspora.PostPhotoUploader(this.el);
-
     this.uploader.onUploadStarted = _.bind(this.uploadStartedHandler, this);
     this.uploader.onProgress = _.bind(this.progressHandler, this);
     this.uploader.onUploadCompleted = _.bind(this.uploadCompleteHandler, this);
   },
 
   // add photo placeholders to the publisher to indicate an upload in progress
-  _addPhotoPlaceholder: function() {
+  _addPhotoPlaceholder: function(id) {
     var publisher = this.publisher;
     publisher.setButtonsEnabled(false);
 
     publisher.wrapperEl.addClass("with_attachments");
     publisher.photozoneEl.append(
-      "<li class=\"publisher_photo loading\" style=\"position:relative;\">" +
+      "<li id=\"upload-" + id + "\"class=\"publisher_photo loading\" style=\"position:relative;\">" +
       "  <div class=\"progress\">" +
       "    <div class=\"progress-bar progress-bar-striped active\" role=\"progressbar\"></div>"+
       "  </div>" +
-      "  <img src=\"\"+Handlebars.helpers.imageUrl(\"ajax-loader2.gif\")+\"\" class=\"ajax-loader\" alt=\"\" />"+
+      "  <div class=\"spinner\"></div>" +
       "</li>"
     );
   },
 
-  uploadStartedHandler: function() {
+  uploadStartedHandler: function(id) {
     this.$el.addClass("loading");
-    this._addPhotoPlaceholder();
+    this._addPhotoPlaceholder(id);
   },
 
-  progressHandler: function(fileName, progress) {
-    this.info.text(fileName + " " + progress + "%").fadeTo(200, 1);
+  progressHandler: function(id, fileName, progress) {
     this.publisher.photozoneEl
-      .find("li.loading").first().find(".progress-bar")
+      .find("li.loading#upload-" + id + " .progress-bar")
       .width(progress + "%");
   },
 
-  uploadCompleteHandler: function(_id, fileName, response) {
+  uploadCompleteHandler: function(id, fileName, response) {
     if (response.success){
-      this.info.text(Diaspora.I18n.t("photo_uploader.completed", {file: fileName})).fadeTo(2000, 0);
-
-      var id  = response.data.photo.id,
+      var photoId = response.data.photo.id,
           image = response.data.photo.unprocessed_image;
 
-      this._addFinishedPhoto(id, image);
+      this._addFinishedPhoto(id, photoId, image);
       this.trigger("change");
     } else {
-      this._cancelPhotoUpload();
-      this.trigger("change");
-      this.info.text(Diaspora.I18n.t("photo_uploader.error", {file: fileName}));
+      this._cancelPhotoUpload(id);
       this.publisher.wrapperEl.find("#photodropzone_container").first().after(
         "<div id=\"upload_error\">" +
-        Diaspora.I18n.t("photo_uploader.error", {file: fileName}) +
+          Diaspora.I18n.t("photo_uploader.error", {file: fileName}) +
         "</div>"
       );
+      this.trigger("change");
     }
   },
 
   // replace the first photo placeholder with the finished uploaded image and
   // add the id to the publishers form
-  _addFinishedPhoto: function(id, image) {
+  _addFinishedPhoto: function(id, photoId, image) {
     var publisher = this.publisher;
 
     // add form input element
     publisher.$(".content_creation form").append(
-      "<input type=\"hidden\", value=\""+id+"\" name=\"photos[]\" />"
+      "<input type=\"hidden\", value=\"" + photoId + "\" name=\"photos[]\" />"
     );
     // replace placeholder
-    var placeholder = publisher.photozoneEl.find("li.loading").first();
+    var placeholder = publisher.photozoneEl.find("li.loading#upload-" + id);
     placeholder
-      .removeClass("loading")
       .prepend(
-        "<div class=\"x\"></div>"+
-        "<div class=\"circle\"></div>"
-       )
-      .find("img").attr(
-      {
-        "src": image.thumb_medium.url,
-        "data-small": image.thumb_small.url,
-        "data-scaled": image.scaled_full.url,
-        "data-id": id
-      }).removeClass("ajax-loader");
-    placeholder
-      .find("div.progress").remove();
+        "<div class=\"x\"></div>" +
+        "<div class=\"circle\"></div>" +
+        "<img src=\"" + image.url + "\" data-id=\"" + photoId + "\" alt=\"\" class=\"hidden\" />"
+      ).removeClass("loading");
+    placeholder.find("div.progress").remove();
+    placeholder.find("img").on("load", function(ev) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/app/views/publisher/uploader_view.js
```diff
@@ -57,23 +57,23 @@ app.views.PublisherUploader = Backbone.View.extend({
...
   // add photo placeholders to the publisher to indicate an upload in progress
-  _addPhotoPlaceholder: function() {
+  _addPhotoPlaceholder: function(id) {
     var publisher = this.publisher;
     publisher.setButtonsEnabled(false);
 
     publisher.wrapperEl.addClass("with_attachments");
     publisher.photozoneEl.append(
-      "<li class=\"publisher_photo loading\" style=\"position:relative;\">" +
+      "<li id=\"upload-" + id + "\" class=\"publisher_photo loading\" style=\"position:relative;\">" +
```
Flaburgan: This id comes from fine-uploader and is a number, but please confirm it can't come from the user and there is no injection here as I don't know how fine-uploader works internally.

SuperTux88 (changes requested):
While testing this I found some things:

1. When the second image finishes first, it still replaces the first progress-bar (while the second progress-bar is stuck at 100% while the first file still uploads). See related code for this:

   https://github.com/diaspora/diaspora/blob/c4a84b5d019e51845b89c57fb64e6f78d75a1647/app/assets/javascripts/app/views/publisher/uploader_view.js#L117

2. (unrelated to your change) This is broken:

   https://github.com/diaspora/diaspora/blob/c4a84b5d019e51845b89c57fb64e6f78d75a1647/app/assets/javascripts/app/views/publisher/uploader_view.js#L80

   and creates the following html:

   ``

   This should render the ajax-loader2.gif after the upload is finished while loading the uploaded image as preview. But the ajax-loader2.gif has a white background, which looks really ugly with the dark theme.
  
   So while you're at it, you maybe can simply change this to a `src=""` (without any ajax-loader2.gif), which would work/look like it does currently (no loader is displayed), but without broken html. We can then add a loader when we have one with transparent background in the future.

3. Can you place add some tests for this. The current tests are here and use `null` as `id` (because it wasn't used until now):

   https://github.com/diaspora/diaspora/blob/77296354852e9ee50b9a44659f68ad8ca728903f/spec/javascripts/app/views/publisher_view_spec.js#L516-L661

app/assets/javascripts/app/views/publisher/uploader_view.js
SuperTux88: The `id` is the index in the array, so it's safe to use it:

https://github.com/FineUploader/fine-uploader/blob/5.13.0/client/js/upload-data.js#L73-L79

Flaburgan:
@SuperTux88 I finally pushed my changes, sorry to have been that long.
- first point should be okay (I did that code a long time ago, I don't remember the exact state)
- second point: I preferred to load the ajax loader at the moment. I'll try to find an image which is fine with the dark theme soon.
- third point: I didn't write tests at the moment, I'll see if I have time to do so today.

Flaburgan:
@SuperTux88 updated, please tell me if that's fine for you.

spec/javascripts/app/views/publisher_view_spec.js
```diff
@@ -575,12 +578,10 @@ describe("app.views.Publisher", function() {
...
         beforeEach(function() {
-          $('#photodropzone').html('<li class="publisher_photo loading"><img src="" /></li>');
-
-          this.uploader.onComplete(null, 'test.jpg', {
+          this.uploader.onComplete(0, "test.jpg", {
             data: { photo: {
-              id: '987',
-              unprocessed_image: { url: 'test.jpg' }
+              id: "987",
+              unprocessed_image: {url: "test.jpg"}
```
diaspora-code-review: Identifier 'unprocessed_image' is not in camel case.

spec/javascripts/app/views/publisher_view_spec.js
```diff
@@ -609,14 +610,42 @@ describe("app.views.Publisher", function() {
...
-      context('unsuccessful completion', function() {
+      context("successful completion with two images", function() {
         beforeEach(function() {
-          $('#photodropzone').html('<li class="publisher_photo loading"><img src="" /></li>');
+          this.uploader.onSubmit(1);
+          this.uploader.onComplete(1, "test2.jpg", {
+            data: {
+              photo: {
+                id: "989",
+                unprocessed_image: {url: "test2.jpg"}
```
diaspora-code-review: Identifier 'unprocessed_image' is not in camel case.

spec/javascripts/app/views/publisher_view_spec.js
```diff
@@ -609,14 +610,42 @@ describe("app.views.Publisher", function() {
...
+      });
+
+      context("unsuccessful completion", function() {
+        beforeEach(function() {
+          this.uploader.onComplete(0, "test.jpg", {
             data: { photo: {
-              id: '987',
-              unprocessed_image: { url: 'test.jpg' }
+              id: "987",
+              unprocessed_image: {url: "test.jpg"}
```
diaspora-code-review: Identifier 'unprocessed_image' is not in camel case.

app/assets/javascripts/app/views/publisher/uploader_view.js
```diff
@@ -86,10 +85,10 @@ app.views.PublisherUploader = Backbone.View.extend({
     if (response.success){
       this.info.text(Diaspora.I18n.t("photo_uploader.completed", {file: fileName})).fadeTo(2000, 0);
 
-      var id  = response.data.photo.id,
+      var photoId = response.data.photo.id,
           url = response.data.photo.unprocessed_image.url;
 
-      this._addFinishedPhoto(id, url);
+      this._addFinishedPhoto(_id, photoId, url);
```
SuperTux88: The `_` at the beginning indicates that the parameter is unused in the function. Since you use it here now, you should remove the `_`.

SuperTux88 (changes requested):
I found one thing, see below.

And as already mentioned the loader should be replaced with something that looks better (the current `ajax-loader2.gif` looks pixelated and doesn't work well with the dark theme) or as I already wrote, you could just remove it for now, since it doesn't work at the moment, this wouldn't be a regression.

denschub:
> And as already mentioned the loader should be replaced with something that looks better (the current `ajax-loader2.gif` looks pixelated and doesn't work well with the dark theme)

For the record, we have `<div class="spinner"></div>`, which only needs explicit `width`/`height` in the context of use and everything will be fine!

spec/javascripts/app/views/publisher_view_spec.js
```diff
@@ -609,14 +610,41 @@ describe("app.views.Publisher", function() {
...
```
diaspora-code-review: Identifier 'unprocessed_image' is not in camel case.

SuperTux88 (changes requested):
OK, now the broken spinner is gone, but the new spinner is not visible either. It's because you already remove it as soon as the image is uploaded. The spinner should be shown after the image is uploaded while downloading the uploaded image again from the server (at least that was where the spinner was shown in the past). So it should be removed only after the uploaded image has finished downloading again.

Flaburgan:
Okay, code fixed and rebased, ready to be reviewed.

app/assets/javascripts/app/views/publisher/uploader_view.js
```diff
@@ -71,41 +64,39 @@ app.views.PublisherUploader = Backbone.View.extend({
...
+        "<div class=\"circle\"></div>" +
+        "<img src=\"" + image.url + "\" data-id=\"" + photoId + "\" alt=\"\" class=\"hidden\" />"
+      )
+      .removeClass("loading");
+    placeholder.find("div.progress").remove();
+    placeholder.find("img").on("load", function(ev) {
+      $(ev.target).removeClass("hidden");
+      placeholder.find(".spinner").remove();
+    });
+    this.$el.removeClass("loading");
```
SuperTux88: This line shouldn't be here, because the `loading` class on the uploader-button should only be removed after all uploads are finished, which is done below (I'm not sure if this class is used for anything at the moment, but if it's used, then it should only be removed after all uploads are done).

SuperTux88 (changes requested):
It works now as it should, only a little nitpicking.

app/assets/javascripts/app/views/publisher/uploader_view.js
Flaburgan: Nice catch, I forgot it, it's duplicate it's also on the line below. Corrected.

Flaburgan: Actually, it should have been here :p it's not on the uploader-button, it's on the `li` containing the image. I'm putting it back.

SuperTux88: No, `this.$el` is the upload button, the `loading` class on the `li` element is removed here:

https://github.com/diaspora/diaspora/blob/522629ce40c29b4a23b03d7da45bc15bb93576c8/app/assets/javascripts/app/views/publisher/uploader_view.js#L82

Flaburgan: Hm this is weird because without that line [this test](https://travis-ci.org/diaspora/diaspora/jobs/393373631#L1877) is failing.

SuperTux88: But with this line the test is failing too.

I had a closer look at this now and the tests are probably broken since you fixed the first point of my first review (https://github.com/diaspora/diaspora/pull/7655#pullrequestreview-83994653), because since then the `onComplete` callback also uses the `id` to select the `li` element:

https://github.com/diaspora/diaspora/blob/522629ce40c29b4a23b03d7da45bc15bb93576c8/app/assets/javascripts/app/views/publisher/uploader_view.js#L75

And the `id="upload-x"` is missing here:

https://github.com/diaspora/diaspora/blob/522629ce40c29b4a23b03d7da45bc15bb93576c8/spec/javascripts/app/views/publisher_view_spec.js#L579

and you have `null` as `id` parameter here:

https://github.com/diaspora/diaspora/blob/522629ce40c29b4a23b03d7da45bc15bb93576c8/spec/javascripts/app/views/publisher_view_spec.js#L582

That's why it can't remove the `loading` class (and do everything else) from the `li`, because it can't select it. Hopefully that helps you with fixing the tests.

SuperTux88 (approved):

SuperTux88:
Merged to develop only, because there were too many conflicts (mostly from #7734).

Thanks for fixing this! :cookie:

PR #7656, opened 2017-10-30
Fix deformed avatar in hovercards

### LINKED ISSUE #6039: Hovercard avatar deformed
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/20130873-hovercard-avatar-deformed?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Flaburgan:
So, from the [different options](https://github.com/diaspora/diaspora/issues/6039#issuecomment-339945467), I went the safest way: keep the avatar to 70x70 and truncate the image as done for the `small` avatar everywhere else. The "truncation" is not really one: I simply set the image as a background one and center center in a 70x70 container.

Fixes #6039

### DIFF
--- a/app/assets/stylesheets/hovercard.scss
+++ b/app/assets/stylesheets/hovercard.scss
@@ -12,16 +12,10 @@
   font-size: small;
 
   .avatar {
-    position: relative;
     float: left;
-    height: 70px !important;
-    width: 70px !important;
-    top: 0 !important;
-
-    margin: {
-      right: 10px;
-      left: 0;
-    }
+    height: 70px;
+    object-fit: cover;
+    width: 70px;
   }
 
   $image_width: 80px; /* including margin */


SuperTux88 (changes requested):
Thanks for the PR, while this fixed the deformed avatars, it breaks some other things:

* It doesn't have rounded corners anymore.
* The error handling doesn't work anymore. When the avatar doesn't exist anymore or couldn't be loaded, it just displays an empty box instead of the default avatar:

https://github.com/diaspora/diaspora/blob/1c1af74e9f3cfc5f54ba07ecec78ccc1a6379308/app/assets/javascripts/app/views.js#L158-L160

Flaburgan:
I don't reproduce your problem with Firefox and Chromium. The avatar is rounded and the default avatar is here. It was not correctly displayed so I added `background-position: cover;`. On which page and with which browser did you test this code?

SuperTux88:
> The avatar is rounded

Ok, that was "my fault", I didn't look close enough ... it is indeed rounded ... but I didn't see it, because I uploaded a full-hd (1920x1080) Image, which was then resized to 100x63, which is too small to see the corners ... so it's just another problem that was visible there.

Also, when uploading a perfectly squared avatar, you see the opposite problem, the avatar is displayed too big (because you use the 100x100 image as background). Here you see an example with the small avatar beside it, so you can see how it should look:



It's even worse with a user from friendica, where the avatar is 175x175:



> and the default avatar is here.

No, when the avatar isn't available anymore (deleted or down), there is just a blank space. Here is a screenshot of it, and you see that the fallback still works with the avatar beside it, where the linked code above sets the default avatar when the linked avatar isn't available, but it doesn't work for the background avatar:



> On which page and with which browser did you test this code?

Tested with both firefox and chromium, at stream and single-post-view.

Flaburgan:
>No, when the avatar isn't available anymore (deleted or down)

Oh, that was the use case. I tried with a user who never uploaded an avatar. I'm going to look at that, thank you.

Flaburgan:
>Also, when uploading a perfectly squared avatar, you see the opposite problem, the avatar is displayed too big

Aaaaaand this one is my fault, I test CSS directly in Firefox because the page is slow to load in dev mode, and I wrote `background-position: cover;` when it should have been `background-size: cover;`. My bad, this is now fixed and the avatar should display correctly. Now let me check the default avatar issue.

Flaburgan:
Alright so forget about this CSS approach with background image, to have the avatar fallback it was way easier to keep the HTML one. I had to do some research but I discovered exactly what we wanted: the [object-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) property. Way cleaner code, isn't it? I just had to add this line. (I still cleaned the `!important` mention as I hate it and it looks like they were useless with Firefox and Chrome. Let's test in `develop` if that's really true).

For information, the browser support of `object-fit` is Chrome 31, Edge 16 and Firefox 36. No IE, but the consequence is a deformed avatar like the current situation so we should be good 👍

SuperTux88 (approved):
Tested and works as expected 👍 

> For information, the browser support of object-fit is Chrome 31, Edge 16 and Firefox 36. No IE, but the consequence is a deformed avatar like the current situation so we should be good 👍

Agree 👍

SuperTux88:
Merged as 53603917eba5cb91bd824304ff2e6ad3b2e44150

Thank you 🍪

PR #7657, opened 2017-10-31
Replace factory_girl with factory_bot

emrox:
because factory_girl is renamed and will be deprecated
see https://robots.thoughtbot.com/factory_bot

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -287,7 +287,7 @@ group :test do
 
   # General helpers
 
-  gem "factory_girl_rails", "4.8.0"
+  gem "factory_bot_rails",  "4.8.2"
   gem "shoulda-matchers",   "3.1.2"
   gem "timecop",            "0.9.1"
   gem "webmock",            "3.0.1", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -205,10 +205,10 @@ GEM
       state_machines
       thor
     fabrication (2.16.3)
-    factory_girl (4.8.0)
+    factory_bot (4.8.2)
       activesupport (>= 3.0.0)
-    factory_girl_rails (4.8.0)
-      factory_girl (~> 4.8.0)
+    factory_bot_rails (4.8.2)
+      factory_bot (~> 4.8.2)
       railties (>= 3.0.0)
     faraday (0.11.0)
       multipart-post (>= 1.2, < 3)
@@ -788,7 +788,7 @@ DEPENDENCIES
   diaspora_federation-test (= 0.2.2)
   entypo-rails (= 3.0.0)
   eye (= 0.9.2)
-  factory_girl_rails (= 4.8.0)
+  factory_bot_rails (= 4.8.2)
   faraday (= 0.11.0)
   faraday-cookie_jar (= 0.0.6)
   faraday_middleware (= 0.12.2)
--- a/features/step_definitions/message_steps.rb
+++ b/features/step_definitions/message_steps.rb
@@ -3,7 +3,7 @@
 Then /^I should see the "(.*)" message$/ do |message|
   text = case message
            when "alice is excited"
-             @alice ||= FactoryGirl.create(:user, :username => "Alice")
+             @alice ||= FactoryBot.create(:user, :username => "Alice")
              I18n.translate('invitation_codes.excited', :name => @alice.name)
            when "welcome to diaspora"
              I18n.translate('users.getting_started.well_hello_there')
--- a/features/step_definitions/oidc_common_steps.rb
+++ b/features/step_definitions/oidc_common_steps.rb
@@ -1,14 +1,14 @@
 # frozen_string_literal: true
 
 Given /^a client with a provided picture exists for user "([^\"]*)"$/ do |email|
-  app = FactoryGirl.create(:o_auth_application, logo_uri: "/assets/user/default.png")
+  app = FactoryBot.create(:o_auth_application, logo_uri: "/assets/user/default.png")
   user = User.find_by(email: email)
-  FactoryGirl.create(:auth_with_read, user: user, o_auth_application: app)
+  FactoryBot.create(:auth_with_read, user: user, o_auth_application: app)
 end
 
 Given /^a client exists for user "([^\"]*)"$/ do |email|
   user = User.find_by(email: email)
-  FactoryGirl.create(:auth_with_read, user: user)
+  FactoryBot.create(:auth_with_read, user: user)
 end
 
 When /^I register a new client$/ do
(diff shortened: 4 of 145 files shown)


emrox:
Any ideas about the rubocop and travis errors?

denschub:
Thanks for the PR!

Travis had some technical issues, I retriggered the jobs and they're green now. As for rubocop, well, you changed a lot of lines so you retriggered validation for quite some old files. However, our bot got banned in the process of trying to comment those,since apparently, it was a bit too much. Check [here](https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review) for guidance on how to run the style checks locally.

emrox:
Thanks for feedback and for restarting Travis.

I ran pronto and there are 227 warnings. And my opinion is it's the reason you also mentioned: these are warnings because I've touched all these old files. Any suggestions how to solve this? In my opinion correcting these warnings is out of scope for this MR.

Some possible solutions:
- just ignore these warnings. 
- correct these first, then apply the patch. Could need some time and it's would result in some just style related PRs
- since pronto just uses rubocop: generate a `.rubocop_todo.yml` and include this within the project (`rubocop --auto-gen-config`)

I would recommend the last option since it documents current TODOs regarding code style and is also applied for future PRs touching old lines of code. (I would create a separate PRs if this solution is a way to go for you)

SuperTux88:
Normally we fix style issues of lines when we touch the line, and since you touched really many lines here, that's a lot of open issues.

However, I'm not sure if it's really worth to change all the lines here and replace `FactoryGirl` with `FactoryBot` everywhere, because of the same reason we don't fix all open style issues at once, and only fix them when we need to touch the line anyway (that's why the pronto bot only checks changed lines). Because it messes with the git history, and when you want to know why a the code on a lines does what it does, it's harder to find the last commit where the actual logic changed, when there are a lot of "fixed style" commits. Sometimes we still do it as a drive-by and fix one or two errors (for example when there are only two errors left in the file I'm editing, I cleanup the whole file, or for example I fix the hash syntax for the whole method and not only one line), but this PR changes 145 files and 900 lines ... for "nothing".

So how about just adding `FactoryGirl = FactoryBot` to the `spec_helper.rb` file and don't touch any specs for now. So we can upgrade FactoryGirl and aren't stuck with an old version and we can start using `FactoryBot` for new code/specs and we can replace old code with `FactoryGirl` the same as we do it with codestyle: only when we touch the line anyway.

emrox:
could also be a possible solution. One downside would be all contributors and reviewers have to keep it in mind and use `FactoryBot` in future since I don't know if there is a easy way to include it into linting. If it's ok for you I create another PR with this solution later.

emrox:
Just another thought: maybe it's worth it to consider just using `create(:factory)` or `build(:factory)` in future specs. This would make the creation/building of factories independent of the used library in the future

SuperTux88:
> could also be a possible solution. One downside would be all contributors and reviewers have to keep it in mind and use `FactoryBot` in future.

I don't think that it's a problem for active contributors and reviewers (because we are only a few reviewers anyway). New contributors probably don't know it.

> since I don't know if there is a easy way to include it into linting.

Adding a custom cop for rubocop would be nice, but I don't think it's possible :( So we would need to tell it to contributors manually, but new contributors usually need some talking anyway, so I think that isn't a problem.

BTW: I would already replace `FactoryGirl` with `FactoryBot` in some central files (for example `factories.rb`), but not touch normal specs.

denschub:
> So how about just adding FactoryGirl = FactoryBot to the spec_helper.rb file and don't touch any specs for now.

There was a reason why the gem was renamed - and I agree with that. I'm fine with replacing the names without fixing style issues, however. That wouldn't really cause issues with our git history since the commit would be meaningful in that context and to be fair, those lines aren't really complex anyway. I'm also kinda against adding a simple alias.

Anyway, not going to fight over that.

SuperTux88:
> There was a reason why the gem was renamed - and I agree with that.

I disagree.

The reason was because some people were unhappy with the name, but it's only a name and that's no reason to change everything. They even fucked up the release of the rename, then deleted a already released version (which makes the fuckup even worse, because it broke builds of people already upgraded) and it only adds work for every project using it ... The old name worked just as well as the new name, it doesn't change anything important.

However, it's not about if we agree or disagree with the rename, because they didn't ask us, and it's already renamed ...

> I'm fine with replacing the names without fixing style issues, however.

When you really think we need to rename it everywhere, then that's my most preferred solution, because everything else would be worse. But please check if it doesn't add any new style issues (indention for example).

Adding an alias would just be a solution without the need to change everything, because as said, there is no reason to change it, it just works as it is.

svbergerem:
I agree with @denschub. They changed the name for a reason and they also provide proper [upgrade instructions](https://github.com/thoughtbot/factory_bot/blob/4-9-0-stable/UPGRADE_FROM_FACTORY_GIRL.md). Let's use the new name.

emrox:
I've tried to change indentations within this PR whenever there was something looking 'wrong'. So the PR is not just a pure replacing but also reviewing surrounding lines. But please review first if you consider merging this. Maybe I missed some cases.

Since this is my first PR here I would stick with your decision.

SuperTux88:
The upgrade instructions just search and replace it, they don't pay attention to the new shorter name. I don't know how @emrox did the changes, but you should at least check if the rename doesn't add new style issues (as written: for example when the next line needs to be indented less now).

emrox:
Some examples:

- I kept the ugly indention here since it was already this way: https://github.com/diaspora/diaspora/pull/7657/files#diff-904a4107d89f81a525345d37804626a9R4
- one example for changing other indentations as well: https://github.com/diaspora/diaspora/pull/7657/files#diff-b709bad60390eff76d8e893b0e31911dR76

General issues I stumbled upon to consider for future refactorings:
- everything already existing style related (hash assignment syntax, indention, ...)
- use shorthand method like `create(:factory)` over `FactoryBot.create(:factory)`
- use `create_list(:factory, x_times)` instead of `x.times { create(:factory) }` (f.e. see https://github.com/diaspora/diaspora/pull/7657/files#diff-c753f123c775922ea5db350239ce1b38R34)

SuperTux88:
> everything already existing style related (hash assignment syntax, indention, ...)

As said: we want to keep the cops active, so we can fix them when we touch the line anyway. But we don't want to fix everything without the need of it (see for example https://github.com/diaspora/diaspora/pull/6157#issuecomment-118315342)

> use shorthand method like `create(:factory)` over `FactoryBot.create(:factory)`

I think that's something we should decide now. Because we then don't need to change everything again when we decide it later. And we don't need to change everything again when they rename it again... And as a side effect it's shorter (so less problems with long lines). 👍

I don't see any disadvantages?

But it doesn't matter ... I just think when we already change everything we can maybe change it to something useful ...

denschub:
+1 for using something that abstracts away the underlying framework altogether. Also +1 in ignoring all cop warnings here, replacing the name and changing obvious indentation issues is more than enough. :)

emrox:
I'm not longer interested in supporting this MR since the only issue on the end was the rubocop raising warnings for code which already was not aligned to your rules. Fixing these would take more time and resources I can currently spend and also are not related to the PR itself.

I recommend you to spend some time for refactoring - even just for the sake of refactoring because in the end it makes the code better maintainable and may attracts other developers to your project.

PR #7662, opened 2017-11-12
change twitter limit characters from 140 to 280

skorotkiewicz:


SuperTux88:
You also need to fix the "message size limits"-tests in `spec/models/services/twitter_spec.rb` to have longer messages now, because they are now broken on travis.

Flaburgan:
@modInfo do you have time to finish this? I'd like it to be merged.

PR #7684, opened 2017-12-28
 Rearrange userdata-export section and cleanup old archives

### LINKED ISSUE #5727: Add warning for old profile data
With the new system to download profile data, user has to refresh its data before downloading them. Even if there is a date indication, it could be great to add a kind of warning message to indicate for instance if the last refresh is older than 2 weeks in order to prevent download of old data.

I don't know if this message should appear after an arbitrary time or if it is possible to know if a user has been active (e.g. add a comment, publish a post) since the last profile data refresh. I would prefer the second solution since data can already be up-to-date after two weeks if you didn't do anything.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9185340-add-warning-for-old-profile-data?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Janakas:
I changed the position of the export/download buttons to make it clear what belongs together.
I also replaced the timestamps with the age of the archive.

I added a cronjob to remove old archives, this saves space on the servers. The data is older than 2 weeks anyway.

Existing archives are also removed when closing an account.

This fixes #5727.

### DIFF
--- a/app/assets/stylesheets/settings.scss
+++ b/app/assets/stylesheets/settings.scss
@@ -36,3 +36,7 @@
   max-width: 100%;
   min-width: 100%;
 }
+
+.account-data h6 {
+  color: $text-grey;
+}
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -539,6 +539,8 @@ class User < ApplicationRecord
      :post_default_public].each do |field|
       self[field] = false
     end
+    self.remove_export = true
+    self.remove_exported_photos_file = true
     self[:disable_mail] = true
     self[:strip_exif] = true
     self[:email] = "deletedaccount_#{self[:id]}@example.org"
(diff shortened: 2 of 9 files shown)


skyschub (approved):

denschub:
Merged as 0347507b2030e17889f90c810c3e224487beb9ef d4e5d13a8fb9185be1a678fdc2e185573a7b6ee9 and 497c647983c740544f6209925fa3b77fa2b76286, thank you very much for your work! 🍪

PR #7690, opened 2018-01-03
Fix invite link on /contacts

Janakas:
I fixed the invite link that is displayed on the contacts page if you don't have any contacts.

### DIFF
--- a/app/views/contacts/index.html.haml
+++ b/app/views/contacts/index.html.haml
@@ -27,7 +27,7 @@
                 != t(".no_contacts_message",
                      community_spotlight: link_to(t(".community_spotlight"), community_spotlight_path))
             %p
-              .btn.btn-link{"data-toggle" => "modal"}
+              .btn.btn-link.invitations-button{"data-toggle" => "modal"}
                 = t("invitations.new.invite_someone_to_join")
 
       #paginate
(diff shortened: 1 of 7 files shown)


app/views/contacts/index.html.haml
```diff
@@ -27,7 +27,7 @@
                 != t(".no_contacts_message",
                      community_spotlight: link_to(t(".community_spotlight"), community_spotlight_path))
             %p
-              .btn.btn-link{"data-toggle" => "modal"}
+              .btn.btn-link#invitations-button{"data-toggle" => "modal"}
```
SuperTux88: This adds the `#invitations-button` ID a second time to the page, because it is also in the sidebar (where you probably copied it from):

https://github.com/diaspora/diaspora/blob/bb2826842fc43733ff39d501c44ebb5542684c9f/app/views/contacts/_sidebar.html.haml#L14

And an ID should be unique on a page and only appear once.

But this ID can easily be replaced with a class `.invitations-button` which also works multiple times per page. The critical place (why it was broken on this page) is here:

https://github.com/diaspora/diaspora/blob/9fa032e4c91710df072291a26b5344429fd69d65/app/assets/javascripts/app/pages/contacts.js#L12

But it also appears at some other places in the sourcecode, so just use your search function :)

SuperTux88 (changes requested):
Thanks for your pull request, nice catch! 🍪 There is only one "little" thing (that will make the pull request much bigger than just one line ;) )

SuperTux88 (approved):

SuperTux88:
Merged as d150a6b2af1ecc857cd3565d332722849836ba45

Thanks for the fix 🍪

PR #7698, opened 2018-01-20
Make mobile bookmarklet work the same way as with desktop UI

SuperTux88:
This fixes that it fails when you call it without notes parameter.


skyschub (approved):

denschub:
Merged, thanks. :)

PR #7702, opened 2018-01-25
Fix mention in #newhere message when invited by another person

### LINKED ISSUE #7701: Mention in first message is broken
When new users registered using an invite link, their first message looks like "Hi everyone, I'm #newHere, I'm interested in #tags, thanks for the invite @userWhoGaveTheInvitationLink"

This last mention is broken.

SuperTux88:
fixes #7701

### DIFF
--- a/lib/stream/multi.rb
+++ b/lib/stream/multi.rb
@@ -46,7 +46,7 @@ class Stream::Multi < Stream::Base
 
     if inviter = self.user.invited_by.try(:person)
       prefill << I18n.t("shared.publisher.new_user_prefill.invited_by")
-      prefill << "@{#{inviter.name} ; #{inviter.diaspora_handle}}!"
+      prefill << "@{#{inviter.diaspora_handle}}!"
     end
 
     prefill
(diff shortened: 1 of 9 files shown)


lib/stream/multi.rb
```diff
@@ -46,7 +46,7 @@ def publisher_prefill
 
     if inviter = self.user.invited_by.try(:person)
       prefill << I18n.t("shared.publisher.new_user_prefill.invited_by")
-      prefill << "@{#{inviter.name} ; #{inviter.diaspora_handle}}!"
+      prefill << "@{#{inviter.diaspora_handle}}!"
```
Flaburgan: I thought this line would be enough to fix the bug. Is the usage of gon necessary or is it a refactor for a cleaner code?

SuperTux88: No, this change is needed to make gon work (`addPersonToMentions` only works with this syntax).

The main fix was in the [publisher.rb](
https://github.com/diaspora/diaspora/pull/7702/files#diff-a51abdb338000222bc972a44538e8b6bL15), because the `.plain_text` changed this prefill message to plain text and replaced the mention syntax with the plain text version (which then obviously doesn't work anymore, because it's only plain text and no mention).

But that only would fix it in the final post. The gon part is to make the preview work.

skyschub (approved):

denschub:
Merged, thanks. 🍪

PR #7705, opened 2018-01-27
Bump diaspora federation

SuperTux88:
Update `diaspora_federation` gem to [0.2.3](https://github.com/diaspora/diaspora_federation/releases/tag/v0.2.3) and do some related changes:

* Remove participants limits for conversations (second part of diaspora/diaspora_federation#91)
* Start sending the `blocking` flag, and increase retries for `Contact` entity (second part of diaspora/diaspora_federation#80)
 (Receiving the `blocking` flag would need a place to store it, so this needs a migration, but we can add this later in develop-branch and benefit from older releases already sending it)

### DIFF
--- a/spec/workers/send_base_spec.rb
+++ b/spec/workers/send_base_spec.rb
@@ -8,13 +8,13 @@ describe Workers::SendBase do
   end
 
   it "increases the interval for each retry" do
-    expect(Workers::SendBase.new.send(:seconds_to_delay, 2)).to be >= 625
-    expect(Workers::SendBase.new.send(:seconds_to_delay, 3)).to be >= 1_296
-    expect(Workers::SendBase.new.send(:seconds_to_delay, 4)).to be >= 2_401
-    expect(Workers::SendBase.new.send(:seconds_to_delay, 5)).to be >= 4_096
-    expect(Workers::SendBase.new.send(:seconds_to_delay, 6)).to be >= 6_561
-    expect(Workers::SendBase.new.send(:seconds_to_delay, 7)).to be >= 10_000
-    expect(Workers::SendBase.new.send(:seconds_to_delay, 8)).to be >= 14_641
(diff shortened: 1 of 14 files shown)


spec/workers/send_base_spec.rb
```diff
@@ -16,5 +16,15 @@
     expect(Workers::SendBase.new.send(:seconds_to_delay, 7)).to be >= 10_000
     expect(Workers::SendBase.new.send(:seconds_to_delay, 8)).to be >= 14_641
     expect(Workers::SendBase.new.send(:seconds_to_delay, 9)).to be >= 20_736
+    expect(Workers::SendBase.new.send(:seconds_to_delay, 10)).to be >= 28_561
```
denschub: Is there any reason not to write this entire test as

```ruby
(2..19).each do |count|
  expect(Workers::SendBase.new.send(:seconds_to_delay, count)).to be >= ((count + 3)**4)
end
```

?

I really dislike this repeating source :/

SuperTux88: The reason was: I didn't want to test the formula with the same formula, so when there is an error in the formula at both places (because when you change it, you would just copy paste it to the test), we don't see it, while with explicit numbers, it is visible that the result is how we want it to be.

But I can change this, if you want. Maybe I can do it as you suggested, but then test 2, 9 and 19 with explicit numbers, so we still see that we have the expected increase. What do you think?

denschub: Sounds good to me! Maybe a little comment why we test these cases "manually" wouldn't harm either. :)

SuperTux88: Updated, please review again :)

skyschub (approved):

denschub:
Thanks for the follow-ups, merged.

Issue #6806, opened 2016-04-19
Our federation model produces dangling data


cmrd-senya:
I wasn't sure of the best place to put it. The question I want to raise is quite fundamental for the federation, but too technical to put it on loomio, so I think github fits better. And it's actually an issue. I'm also not aware of any previous discussions on that topic, but it is well-known for sure for those who work on the federation.

So, we use push model for the most of our interactions. However we use pull model for some. This includes profiles fetching and posts fetching now and will include more in future, like comments fetching and if we ever implement the message editing feature it'll be affected as well.

Our push model implementation now supposes that users are exchanging messages only when they are sharing. It doesn't matter if the sharing is mutual or not. But it must be there. Updates, including post comments, retractions, profile updates and profile deletions are delivered only to those pods where a user have contacts.

On the other hand, we use pull model that allows us to fetch profile or post without having a sharing relashionship set up. This leads to the problem, that if there is no one on our pod who shares with some person whose content we have fetched, then we **can't trust** this content. We will definetely miss any updates on this content (user deletion #6791, retraction #6436, comments). This also happens if there was sharing but it was removed then.

I guess this applies to the public posts only, since we don't use the pull model for private posts. The issue wasn't noticed too much by users because most of the time we have some kind of sharing set up, and also most of the users use big pods so there is at least someone on your pod who shares with the other person, so you get updates through his subscribtion anyway. But if the federation growth, it'll become worse. Also if #4491 is fixed it'll make the thing worse. Also this will make the account migration less reliable.

We shouldn't allow this to happen.

Here is a solution I think of. First, if there ever was sharing between Alice from podA and Bob from podB, and they stop sharing, both pods must still send updates on user profiles, even if there are no one left who share between the pods. Second, the same stands for posts. If a post was federated, all the updates must be delivered anyway, including retractions, comments, in future - editions, again even if Alice and Bob don't share anymore and there is no one between podA and podB who shares. Third, if there was no user sharing between podA and podB, and podA fetches some profiles and posts from podB, then podA must also send some feedback address to podB, so podB knows that it must send update information somewhere. Thus it'll be podA's responsibility to set a correct address to receive updates. For podB podA then becomes just another subscriber along with ordinary users.

@SuperTux88, since you are the main federation contributor at the moment, I hope to get feedback from you. Maybe you already have idea on solution for that or maybe you're going to fix it? You don't want me to participate to your federation work, but if it is the part you aren't working on, then maybe I can work on it?

[...]

PR #7708, opened 2018-01-30
Send participation after receiving a public post

### LINKED ISSUE #6436: Published post stay on stream of another pod accounts after deletion contacts
Given
Account1 on pod1
Account2 on pod2
Accounts is linked between each other in some aspect
Public posts of both accounts are shown in streams of each accounts

Steps
1. Sign in account1
2. Add public post in account1
3. Sign in account2 - post from step2 is shown
4. Delete contacts between accounts in both accounts
5. Delete post from step2
6. Check in account2 visibility of post from step2 

Expected result
Deleted post from step 2 is not shown in both accounts

Actual result
Deleted post from step 2 stay be shown in account2

SuperTux88:
I'm currently working on a PR to improve the situation described in #6806 and #6436. The main work is bigger (and not finished yet) and includes big migrations, so it's something for 0.8.0.0. But I thought it would be useful if older pods already start sending participations, so we can make use of them later. So I thought lets make this commit a last minute PR for 0.7.3.0 ;)

It already improves the situation only with this commit, because it lets the owner of the post know, that this pod is interested in updates about this post. This improves for incomplete threads of a post (but it will be better after the second part is done), and it also deletes a post on all pods that received it (so this fixes #6436, at least for public posts).

The sending user is only used to verify that the participation was sent from this pod, but lets use an admin/podmin account if available.

### DIFF
--- a/lib/diaspora/federation/receive.rb
+++ b/lib/diaspora/federation/receive.rb
@@ -153,7 +153,7 @@ module Diaspora
             guid:       entity.guid,
             created_at: entity.created_at,
             root_guid:  entity.root_guid
-          )
+          ).tap {|reshare| send_participation_for(reshare) }
         end
       end
 
@@ -193,6 +193,8 @@ module Diaspora
             status_message.photos = save_or_load_photos(entity.photos)
 
             status_message.save!
+
+            send_participation_for(status_message)
           end
         end
       end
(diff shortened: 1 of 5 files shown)


Flaburgan:
There's not enough love in the ❤️ smiley to show how grateful I am! Thank you for your work!

goobertron:
This looks like being a fix for something I asked a number of years ago, 'Is there any way of doing this?', so a huge thank you and a heart from me too!

denschub:
I... am confused. Wouldn't that mean the admin/first user would get notifications for... everything?

SuperTux88:
No, because the `participation` is only temporary for sending it, it isn't persisted in the database, so no notifications for the user. The user is really only needed because we need a valid key from any user of that pod.

denschub:
Ah, thanks.

skyschub (approved):

denschub:
Thanks for the extra test case and the explanation 🍪 - Merged.

PR #7715, opened 2018-02-16
Fix followed tags posts in stream for posts with multiple tags

### LINKED ISSUE #4503: followed tags don't work anymore?

SuperTux88:
When there were posts with many followed tags they were returned multiple times, resulting in less than 15 unique posts. That resulted in some posts to be missed in the stream.

Fixes #4503

### DIFF
--- a/app/models/status_message.rb
+++ b/app/models/status_message.rb
@@ -46,7 +46,7 @@ class StatusMessage < Post
   end
 
   def self.public_tag_stream(tag_ids)
-    all_public.tag_stream(tag_ids)
+    all_public.select("DISTINCT #{table_name}.*").tag_stream(tag_ids)
   end
 
   def self.tag_stream(tag_ids)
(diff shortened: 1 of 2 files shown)


denschub:
Merged as beeb073dbe0d95b0dd38c91f8b129fc0039bd3a0.

PR #7717, opened 2018-02-17
Don't retry for PublicKeyNotFound

SuperTux88:
We only return `nil` when we know the person, but don't know the key or the key is invalid, so it doesn't make sense to retry in this case. When the person isn't known and can't be fetched we raise a `DiscoveryError` which will be retried.

Also the errors were moved to the `Signable` module in the last release.


denschub:
Merged as f0aca56c8f0cbb8ffd1f47e712bd29cb6a982553.

PR #7718, opened 2018-02-18
Bump diaspora federation

SuperTux88:
This fixes the ``NoMethodError: undefined method `optional_props' for WebFinger:Module`` bug for discovery from socialhome and hubzille (or others with missing optional fields in webfinger).

I also added the `edited_at` properties to profile messages, so we can use this later when receiving profiles to make sure we don't overwrite newer data with older data.

Since `edited_at` could theoretically also be present in comments now, we need to store it in `additional_data` to keep the signatures valid.


denschub:
Merged as 602a1dcd5501c4370a247b01ab7b3eb67d13baca 0b4b3515a500f864e520d216700e85ce18fb3fea 86d52a0da3013fbf023e18d531b7312584baf0a5.

PR #7724, opened 2018-02-26
Profile tags order

SuperTux88:
I already saw [this spec](https://travis-ci.org/diaspora/diaspora/jobs/346058334#L1703) fail multiple times, and I now debugged why this happens. The tag order in a profile depends on the order in which the `tags` (not `taggings`) were created on the pod, so when for some reason `#two` is created first in the database, it's always the first tag in the list. This isn't a big problem (since we don't offer a way to sort the profile tags), but I also don't know if it's intended that the profile tags are different on each pod.

However `acts-as-taggable-on` has the feature to keep the order, and I think it's useful to enable it for profile tags. That way people can order their tags by removing them and add them in the order they like, and maybe we can add a feature in the future to order them in a simpler way. And I didn't find any other way to make this test not randomly fail. :(

I also refactored the shared taggable tests to use `let` instead of `@object`, that makes the usage a big simpler and I didn't want to add the hack with `before` a second time like it was in [`comment_spec.rb`](https://github.com/diaspora/diaspora/pull/7724/files#diff-33939bbe6ae0e01f344f9254b1c86bb0L125).

### DIFF
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -11,7 +11,7 @@ class Profile < ApplicationRecord
   include Diaspora::Taggable
 
   attr_accessor :tag_string
-  acts_as_taggable_on :tags
+  acts_as_ordered_taggable
   extract_tags_from :tag_string
   validates :tag_list, :length => { :maximum => 5 }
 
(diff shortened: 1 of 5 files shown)


denschub:
Merged as 4e5ee36675246038009ba231b1a4101dbb36a59e 4cd8de532740079ad8541ce460cc7add0462ebf3.

Didn't add it to the changelog, since technically, nothing changes: tag order was hard to determine before, and still is.

Issue #7686, opened 2017-12-30
User setting: enable/disable notifications in-app as well as by email


goobertron:
I thought I had opened an issue about this several years ago, but I can't find it. (Apologies if there is already an open issue that I've missed.)

I would really like to be able to enable/disable notifications in the app (notifications drop-down and notifications page) as well as by email.

Two use-cases: for someone like me who isn't interested when someone 'likes' their post; and a user with a huge number of contacts (for example, Diaspora HQ) gets bombarded with notifications they might not need (e.g. the new birthday notifications).

I can't see a down-side to making in-app notifications configurable, just like email notifications are. Is there one?

Here's a mock-up of what it could look like:

![notif_prefs](https://user-images.githubusercontent.com/2530168/34451058-519530fa-ed12-11e7-918c-d9c1257b5781.png)

The 'in-app' and 'by email' labels could be given as tool-tips instead of as part of the page, in which case 'in-app' could be lengthened to 'in diaspora\*'.

I hope something like this is possible!

SuperTux88:
I planned to do this (regardless of if there is an issue or not), but didn't have the time yet. :+1:

goobertron:
Great, thanks!

PR #7725, opened 2018-02-26
Add app notifications

Radicalchaos-zz:
This PR will add a symmetrical set of options for in-app notifications, as is currently available for email.

- [x] HAML form
- [ ] User settings controller
- [ ] Control in-app notifications conditionally

### DIFF
--- a/app/views/users/_edit.haml
+++ b/app/views/users/_edit.haml
@@ -117,9 +117,9 @@
 
     = render partial: "post_default"
 
-    - if AppConfig.mail.enable?
-      .row
-        .col-md-12
+    .row
+      - if AppConfig.mail.enable?
+        .col-md-6
           %h3
             = t(".receive_email_notifications")
           = form_for "user", url: edit_user_path, html: {method: :put} do |f|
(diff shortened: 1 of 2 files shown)


app/views/users/_edit.haml
```diff
@@ -176,9 +176,66 @@
                   = type.check_box :contacts_birthday, {checked: email_prefs["contacts_birthday"]}, false, true
                   = t(".birthday")
                 .small-horizontal-spacer
+      .col-md-6
+        %h3
+          = t(".receive_app_notifications")
+        = form_for "user", url: edit_user_path, html: {method: :put} do |f|
+          = f.fields_for :email_preferences do |type|
+            #email_prefs
```
diaspora-code-review: - Do not repeat id "#email_prefs" on the page
- `id` attribute must be in lisp-case

denschub:
Thanks for working on that, @Radicalchaos! Being able to turn off notifications completely for different kinds of events would be super cool.

Please be aware that our current system for storing preferences is... not the nicest, and it may not even be easily possible to add preferences for notification types, as our current email notifications are currently implemented via the [`UserPreference`](https://github.com/diaspora/diaspora/blob/198dddd69e9264f1fefda1f49964a665aabc6a2e/app/models/user_preference.rb#L3) table, which isn't really designed to handle anything else than emails. If you already worked out a solution, cool. If not, feel free to [drop by our usual communication channels](https://wiki.diasporafoundation.org/How_we_communicate) so we can discuss about potential solutions. :)

(linking this PR to the matching feature request, which is #7686)

denschub:
Hey @Radicalchaos! Do you plan on continuing this work? Is there anything we can do to assist you here?

denschub:
After a team-internal discussion, let's go ahead and close this PR. There is a lot of work left to do, especially since this PR is a front-end only PR for now. The backend portion is the more complicated stuff, and the frontend has to be adjusted either way, so closing this won't get rid of much.

If someone is interested in working on this, feel free to pick up this PR and base your own work on the work being done here. We will eventually work on this feature, because we agree that this is important and we really want to have it ourselves. :)

PR #7726, opened 2018-02-26
Add /camo/ to robots.txt

SuperTux88:
I almost forgot to PR this. I added this a while ago on my pod after some (porn gif) /camo/ urls suddenly got popular on google and the traffic exploded. Camo only proxies images hosted somewhere else, so it doesn't make sense to add the proxied versions to search engines. It only creates traffic for camo when /camo/* urls are in search results.

Here is an image of the traffic from before, after some "magic SEO" happened and then after I added this to the robots.txt and some days delay (crawlers first need to find the new robots.txt) it was even lower than before.

image

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -8,3 +8,4 @@
 User-Agent: *
 Disallow: /people/
 Disallow: /u/
+Disallow: /camo/


denschub:
... welp ...

denschub:
Merged as f883c6ede2d7acf61a2320ce27e316e866da50e4.

Issue #6475, opened 2015-10-09
[feature] add support for LDAP authentication


pravi:
disroot.org integrated email, xmpp and owncloud with single sign on, but diaspora needs a separate account. If we add LDAP as a supported authentication method, it would be useful to setups like disroot.org which already has a userbase which they want to integrate with diaspora.

svbergerem:
We closed the same feature request about two years ago. #4748

jhass:
Closing as duplicate. If you want to see this in main, find a committed and active maintainer for it.

pravi:
This is not helpful. I'm trying to find support through bounty source and you can't insist on finding a supporter before even looking for one.

jhass:
I can, we had to rip out tons of stuff already because nobody maintained it, I'm rather careful about adding new things that will easily remain unmaintained.

jhass:
The forks that implemented it are all outdated and broken, there's clearly nobody willing to commit to maintaining it.

jhass:
Oh and Bountysource might motivate yet another integration, but how likely is it that that contributor will also maintain it?

pravi:
If it is used in community infra like disroot.org, they'd want to keep it working.

jhass:
That might generate bugreports, but will there be somebody to do the fixes?

pravi:
We can make maintenance a criteria for accepting contributions on it.

jhass:
Looking forward to the pull request including such a commitment  then. I see no reason to track an issue for it.

pravi:
Status of bounty source is linked to status of this issue. Keeping it closed will give a wrong message to any effort to seek support.

jhass:
As explained I think a BountySource for this would only encourage a one time contribution, not a long term maintenance commitment.

JosephKiranBabu:
@pravi @jhass We're trying to integrate diaspora into the FreedomBox project. FreedomBox uses LDAP for most applications. 
Would you be willing to merge LDAP integration into diaspora master if the FreedomBox team maintains it?

Flaburgan:
I would say yes but we also need @SuperTux88 and @denschub opinion here I guess.

denschub:
Basically, see this comment: https://github.com/diaspora/diaspora/issues/6475#issuecomment-146867235

SuperTux88:
Basically, what @denschub says :)

[...]

PR #7728, opened 2018-03-02
pam authentication

devkral:
Adds pam authentication to diaspora. Based on my patches to mastodon.
For people requiring ldap authentication:
I wrote a package for pam which extracts the email from ldap:
https://github.com/devkral/pam_email_extractor

Not tested yet => WIP.


app/views/users/_edit.haml
```diff
@@ -32,28 +32,31 @@
...
-              class: "col-sm-6 control-label"
-            .col-sm-6
-              = f.password_field :password_confirmation, placeholder: t(".character_minimum_expl"),
-                class: "form-control", autocomplete: "new-password"
-
-          .clearfix
-            = f.submit t(".change_password"), class: "btn btn-primary pull-right", name: "change_password"
+        - if !use_pam? || resource.encrypted_password.present? || !resource.is_pam_account?
+          = form_for @user, url: edit_user_path, html: {method: :put, class: "form-horizontal"} do |f|
+            - if (@user.errors.keys & %i(password password_confirmation current_password)).present?
```
diaspora-code-review: `%i`-literals should be delimited by `[` and `]`.

app/views/users/_edit.haml
```diff
@@ -32,28 +32,31 @@
...
+              = f.label :password_confirmation, t("registrations.new.password_confirmation"),
+                class: "col-sm-6 control-label"
+              .col-sm-6
+                = f.password_field :password_confirmation, placeholder: t(".character_minimum_expl"),
+                  class: "form-control", autocomplete: "new-password"
+
+            .clearfix
+              = f.submit t(".change_password"), class: "btn btn-primary pull-right", name: "change_password"
+        -else
+          = t('.pam_account')
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/views/users/_edit.haml
```diff
@@ -32,28 +32,31 @@
...
+            .form-group
+              = f.label :password_confirmation, t("registrations.new.password_confirmation"),
+                class: "col-sm-6 control-label"
+              .col-sm-6
+                = f.password_field :password_confirmation, placeholder: t(".character_minimum_expl"),
+                  class: "form-control", autocomplete: "new-password"
+
+            .clearfix
+              = f.submit t(".change_password"), class: "btn btn-primary pull-right", name: "change_password"
+        -else
```
diaspora-code-review: The - symbol should have one space separating it from code

app/models/user.rb
```diff
@@ -418,6 +421,54 @@ def setup(opts)
     self
   end
 
+  def is_pam_account?
```
diaspora-code-review: Rename `is_pam_account?` to `pam_account?`.

app/models/user.rb
```diff
@@ -418,6 +421,54 @@ def setup(opts)
...
+    return false unless Devise.pam_authentication
+    name = pam_get_name
+    return false unless name
+    Rpam2.account(Devise.pam_default_service, name).present?
+  end
+
+  def no_pam_conflict
+    errors.add(:username, 'username conflicts with pam account') if pam_conflict?
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/user.rb
```diff
@@ -418,6 +421,54 @@ def setup(opts)
...
+  end
+
+  def pam_conflict(_attributes)
+    # block pam logins on users with pw
+    nil
+  end
+
+  def pam_setup(attributes)
+    args = {}
+    args['username'] = self[::Devise.usernamefield]
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/user.rb
```diff
@@ -418,6 +421,54 @@ def setup(opts)
...
+    args['email'] = Rpam2.getenv(get_service, self[::Devise.usernamefield], attributes[:password], "email", false)
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/user.rb
```diff
@@ -418,6 +421,54 @@ def setup(opts)
...
+    args['email'] = "#{self[::Devise.usernamefield]}@#{find_pam_suffix}" if args['email'].nil?
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/user.rb
```diff
@@ -418,6 +421,54 @@ def setup(opts)
...
+    opts = ActionController::Parameters.new(args)
+    setup(opts.permit(:username, :email))
+  end
+
+  def find_pam_suffix
+    "#{AppConfig.bare_pod_uri}"
```
diaspora-code-review: Prefer `to_s` over string interpolation.

app/models/user.rb
```diff
@@ -418,6 +421,54 @@ def setup(opts)
...
+  end
+
+  def password_required?
+    return !is_pam_account? if Devise.pam_authentication
+    super
+  end
+
+  def send_reset_password_instructions
```
diaspora-code-review: Method `User#send_reset_password_instructions` is defined at both app/models/user.rb:173 and app/models/user.rb:462.

config/initializers/devise.rb
```diff
@@ -12,6 +12,19 @@ def to_mobile
   end
 end
 
+module Devise
+  mattr_accessor :pam_authentication
+  @@pam_authentication = false
```
diaspora-code-review: Replace class var @@pam_authentication with a class instance var.

config/initializers/devise.rb
```diff
@@ -12,6 +12,19 @@ def to_mobile
...
+  mattr_accessor :pam_controlled_service
+  @@pam_controlled_service = nil
```
diaspora-code-review: Replace class var @@pam_controlled_service with a class instance var.

config/initializers/devise.rb
```diff
@@ -12,6 +12,19 @@ def to_mobile
...
+
+  class Strategies::PamAuthenticatable
```
diaspora-code-review: Use nested module/class definitions instead of compact style.

config/initializers/devise.rb
```diff
@@ -270,4 +283,18 @@ def to_mobile
   # When using omniauth, Devise cannot automatically set Omniauth path,
   # so you need to do it manually. For the users scope, it would be:
   # config.omniauth_path_prefix = '/my_engine/users/auth'
+
+  # implement pam support analog to mastodon
+  if AppConfig.pam.enable
+    config.pam_authentication     = true
+    config.usernamefield          = 'username'
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

config/initializers/devise.rb
```diff
@@ -270,4 +283,18 @@ def to_mobile
...
+    config.emailfield             = 'email'
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

db/migrate/20180302130741_add_remember_token_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddRememberTokenToUsers < ActiveRecord::Migration[5.1]
```
diaspora-code-review: Missing magic comment `# frozen_string_literal: true`.

config/locales/devise/devise.de_formal.yml
```diff
@@ -57,6 +57,7 @@ de_formal:
     passwords:
       edit:
         change_password: "Mein Passwort ändern"
+        pam_account: "Eine Passwortänderung ist nicht möglich - Die Account Authentifizierung erfolgt durch PAM"
```
jhass: Please do not edit any non-English locale, these changes will simply get lost with the next locale update.

config/initializers/devise.rb
```diff
@@ -270,4 +283,18 @@ def to_mobile
   # When using omniauth, Devise cannot automatically set Omniauth path,
   # so you need to do it manually. For the users scope, it would be:
   # config.omniauth_path_prefix = '/my_engine/users/auth'
+
+  # implement pam support analog to mastodon
```
jhass: I don't think we need a reference to mastodon here :P

jhass:
I have similar worries as with supporting LDAP (directly) here. It seems like no feature that any core contributor would use and thus notice any breakages. So at very least we would need good CI coverage for this, which can be tricky given the nature of this feature.

Flaburgan:
>I have similar worries as with supporting LDAP (directly) here. It seems like no feature that any core contributor would use and thus notice any breakages. So at very least we would need good CI coverage for this, which can be tricky given the nature of this feature.

I think @ldidry (sysadmin of framasoft) would be interested in LDAP support in diaspora for our framasphere.org installation. Can you please confirm?

jhass:
Well I have the same suggestion as for LDAP here, we could have this as a fork that we point to in our documentation and if it turns out that fork actually gets some relevant usage and maintenance we can look at upstreaming.

denschub:
We had the same discussion in https://github.com/diaspora/diaspora/issues/6475, and all those comments are relevant... And I still agree with everything in it. I don't think we want this (that is, these kind of authentication providers) in our codebase, unless there is someone committed to permanently maintaining it.

devkral:
the thing is: pam is an authentication provider provider. Instead of coding ldap and whatever in your codebase you just have pam. It is an old standard for solving exactly the problem you mentioned.
Use cases:
* local user mode
* ldap
* whatever pam authentication you use

ldidry:
> I think @ldidry (sysadmin of framasoft) would be interested in LDAP support in diaspora for our framasphere.org installation. Can you please confirm?

I don't confirm. LDAP is good in Framasoft's case for:
- internal services
- new services where our members does not already have account (it makes more easy for our members to create an account, but if they already have an account… no advantage).

I have no interest in having LDAP authentication on framasphere.org **but I think it would be great on diaspora\*** since it would benefit to some people (enterprises, associations…)

denschub:
> but I think it would be great on diaspora* since it would benefit to some people (enterprises, associations…)

Maybe, yes. Except as of right now, noone of these groups ever actively communicated real interested in any of these use cases. If "enterprises" have looked into diaspora\* previously, it was pretty obvious that it wasn't the right fit, and not because we didn't support LDAP, but because we are designed to federate with other nodes.

> It is an old standard for solving exactly the problem you mentioned

Your PR, so far, introduces two new Gems, which want to be updated eventually. And it adds a lot of code in critical parts of the application (authentication, that is). If that stuff is not maintained, we very easily end up having people no longer being able to sign in at all. And because there is no active contributor willing to constantly maintaining it, there is an actual risk here.

Even if LDAP would be a highly requested feature, without someone able to maintain it in the long run, I don't like it. We have much more requested features like the XMPP chat, for example, which got merged in because someone once took the time to implement it, and now, these parts are laying around in a dark corner, unmaintained, and in a bad state.

I'm not too happy having parts inside a central aspect of our codebase in an unmaintained state, to be quite honest. And this sounds like a discussion we should have on Discourse, potentially.

cmrd-senya:
Is it possible and does it makes sense to always use PAM as a standard way of logging in even for the same local user login scenario we have currently?

devkral:
yes.

config/locales/devise/devise.de_formal.yml
devkral: thx, didn't know

denschub:
> does it makes sense to always use PAM as a standard way of logging in

No, not at all. Using PAM required `libpam`/`libpam-dev` to be installed on all systems to even run `bundle install`. I don't see why we should add this dependency. If anything, PAM should be an optional feature.

Besides that, it is absolutely not a good idea to replace existing auth logic provided by devise itself with something else...

I'll just quote what @jhass said earlier

> It seems like no feature that any core contributor would use and thus notice any breakages. So at very least we would need good CI coverage for this, which can be tricky given the nature of this feature. [...] Well I have the same suggestion as for LDAP here, we could have this as a fork that we point to in our documentation and if it turns out that fork actually gets some relevant usage and maintenance we can look at upstreaming.

devkral:
ok, see mastodon, they have a good solution by using environment. pam is only installed if in environment. If you want to - I can implement it this way

db/migrate/20180302130741_add_remember_token_to_users.rb
devkral: fixed?

devkral:
I have to admit: I should have used ffi. Sadly pam-headers are not standard but the pam library is (sudo)

devkral:
my tests were successful. Hope that we find a solution for the dependency problem.

devkral:
maybe polyfilling is a bad idea: it could lead crashing programs which are hard to debug because nobody assumes polyfills. I will make it configurable with the environment. This is the safest way.

config/initializers/devise.rb
```diff
@@ -270,4 +285,26 @@ def to_mobile
   # When using omniauth, Devise cannot automatically set Omniauth path,
   # so you need to do it manually. For the users scope, it would be:
   # config.omniauth_path_prefix = '/my_engine/users/auth'
+
+  config.pam_authentication = AppConfig.pam.enable?
+  if config.pam_authentication
+    begin
+      gem "devise_pam_authenticatable2"
+    rescue Gem::LoadError
+      puts "PAM not available, install with --with pam"
```
diaspora-code-review: Do not write to stdout. Use Rails's logger if you want to log.

devkral:
fixed the problem with pam: I moved it in an optional group. I hope this solution is satisfying.

devkral:
I added a test mode to my rpam gem and will implement soon some tests so the coverage is better

devkral:
the ALLOW_NOPAM environment variable allows the build of the pam package without pam headers.
It is dangerous (this is why it is an environment var) because "pam" won't work for anything except tests this way.

app/models/user.rb
```diff
@@ -418,6 +422,49 @@ def setup(opts)
...
+
+  def pam_conflict(_attributes)
+    # block pam logins on users with pw
+    nil
+  end
+
+  def pam_setup(attributes)
+    args = {}
+    args['username'] = self[::Devise.usernamefield]
+    args['email'] = Rpam2.getenv(find_pam_service, self[::Devise.usernamefield], attributes[:password], 'email')
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/user.rb
```diff
@@ -418,6 +422,49 @@ def setup(opts)
...
+    args['email'] = "#{self[::Devise.usernamefield]}@#{find_pam_suffix}" if args['email'].nil?
```
diaspora-code-review: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

config/environments/test.rb
```diff
@@ -58,4 +58,7 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # set fake_data for pam, don't do real calls, just use fake data
+  Rpam2.fake_data = { usernames: Set['pam_user1', 'pam_user2'], servicenames: Set['pam_test', 'pam_test_controlled'], password: '123456', env: { email: 'me@example.com'} }  if AppConfig.pam.enable?
```
diaspora-code-review: - Space inside { detected.
- Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
- Line is too long. [197/120]
- Space inside } detected.
- Unnecessary spacing detected.

spec/controllers/sessions_controller_spec.rb
```diff
@@ -28,6 +28,12 @@
       expect(response).to be_redirect
       expect(response.location).to match /^#{stream_url}\??$/
     end
+
+    it "auto create and login pam user" do
+      post :create, params: {user: {remember_me: "0", username: "pam_user1", password: "123456"}}
+      expect(response).to be_redirect
+      expect(response.location).to match /^#{getting_started_url}\??$/
```
diaspora-code-review: Ambiguous regexp literal. Parenthesize the method arguments if it's surely a regexp literal, or add a whitespace to the right of the `/` if it should be a division.

denschub:
At this point, I feel like closing this PR. As been quoted multiple times:

If you have a potential user of this feature, please maintain a fork for now, and as written before, if that turns out to be stable, maintained, and indeed a feature that people would use, we can look at upstreaming.

As of right now, I have multiple issues with this. First, it seems like your own library isn't really mature yet, as you rolled over your own approach multiple times. That, by itself, does not feel stable enough to be embedded in a project that scale, especially since it's directly touching very important parts of our source. Moreover, the amount and nature of tests that are failing on a regular basis here is not increasing my confidence either.

As quoted earlier, this is still true:

> It seems like no feature that any core contributor would use and thus notice any breakages. So at very least we would need good CI coverage for this, which can be tricky given the nature of this feature. [...] Well I have the same suggestion as for LDAP here, we could have this as a fork that we point to in our documentation and if it turns out that fork actually gets some relevant usage and maintenance we can look at upstreaming.

But as of right now, I don't really see a reason to merge this. Can I have a second opinion from @diaspora/core-team please? :)

devkral:
May look for the error causes. Last error is because of some environment stuff broke and not because of my code (require term/ansi).
Also because every ruby project uses its own code style it is hard to find methods which can be overloaded safely without a linter complaining (-> Versioning).
Linter warnings are not Compile errors.

devkral:
And I am not a ruby guru. So I may need some approaches.

config/environments/test.rb
```diff
@@ -58,4 +58,7 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # set fake_data for pam, don't do real calls, just use fake data
+  Rpam2.fake_data = { usernames: Set["pam_user1", "pam_user2"], servicenames: Set["pam_test", "pam_test_controlled"], password: "123456", env: { email: "me@example.com" } }  if AppConfig.pam.enable?
```
diaspora-code-review: Line is too long. [198/120]

config/environments/test.rb
```diff
@@ -58,4 +58,12 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # set fake_data for pam, don't do real calls, just use fake data
+  Rpam2.fake_data = {
```
diaspora-code-review: Favor a normal if-statement over a modifier clause in a multiline statement.

config/environments/test.rb
```diff
@@ -58,4 +58,12 @@
...
+    usernames: Set["pam_user1", "pam_user2"],
```
diaspora-code-review: Align the elements of a hash literal if they span more than one line.

config/environments/test.rb
```diff
@@ -58,4 +58,12 @@
...
+    servicenames: Set["pam_test", "pam_test_controlled"],
+    password: "123456",
```
diaspora-code-review: Align the elements of a hash literal if they span more than one line.

config/environments/test.rb
```diff
@@ -58,4 +58,12 @@
...
+    env: { email: "me@example.com" }
```
diaspora-code-review: - Align the elements of a hash literal if they span more than one line.
- Space inside { detected.
- Space inside } detected.

config/environments/test.rb
```diff
@@ -58,4 +58,15 @@
...
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # set fake_data for pam, don't do real calls, just use fake data
+  if AppConfig.pam.enable?
+    Rpam2.fake_data =
+      {
+        usernames: Set["pam_user1", "pam_user2"],
+        servicenames: Set["pam_test", "pam_test_controlled"],
+        password: "123456",
```
diaspora-code-review: Align the elements of a hash literal if they span more than one line.

config/environments/test.rb
```diff
@@ -58,4 +58,15 @@
...
+        env: { email: "me@example.com" }
```
diaspora-code-review: - Align the elements of a hash literal if they span more than one line.
- Space inside { detected.
- Space inside } detected.

devkral:
There is a dependency bug. I may hit it because I use Bundler.require.

devkral:
I see bundler.require is absolutely bad and the culprit. I have to find a better solution.

devkral:
I tested it against my own repo with travis. I wonder why travis doesn't trigger (caching?).

devkral:
its a heisenbug. I just changed a comment. In my travis build nothing failed.

devkral:
I had this problem before in an unrelated project: I think one test fails depending on the daytime/date. Now it should work because it is not the last day of the month+ not too late

jdkruzr:
Hi, I had a little trouble following this -- does Diaspora* have PAM support for accounts now?

devkral:
with this patch/branch - yes. But it is not integrated yet.

SuperTux88:
> But as of right now, I don't really see a reason to merge this. Can I have a second opinion from @diaspora/core-team please? :)

I agree here, we can keep this as fork and see how it works and how many podmins are using it. But I don't feel good with merging something that big and in this case even critical without anybody maintaining it. We already have an unmaintained chat and S3 support which already broke in 0.7.0.0 because nobody was able to test it.

But in addition to everything what @denschub and @jhass already said, I also see a problem with closing accounts. When you remove an account from LDAP, what happens with the account in diaspora? When you just remove access the person would lose access over all their posts, which is bad in my eyes. But when you automatically close the account in diaspora, this can't be undone, and that isn't clear that closing an account in LDAP has this side-effect (while closing an account in diaspora has a clear warning), which is bad too. Also LDAP doesn't ensure that you don't create an account that already existed in the past. Because of these reasons I think external authentication systems don't work well with diaspora.

devkral:
If the account is deleted in pam, the user can simply not logging in.
But he has the possibility to recover his account: requesting a password reset email, it is just blocked for pam users. As soon as the account has a password it is a normal diaspora account.

devkral:
But anyway this easily tweakable, see pam_conflict? and pam_account?

denschub:
Thanks @SuperTux88.

Based on what has been said by three core maintainers now, I'll close this pull request. As for now, there seems to be a negligible interest in having LDAP in the first place, the suggested solution does neither seem stable nor well tested, and we all have serious concerns about supporting LDAP in the first place.

@devkral, if you feel like it, please continue your work in your own diaspora fork. If it turns out that people actually use LDAP in significant amounts **and** the integration and its dependencies are well maintained, please open a thread on our Discourse and we can have a community-driven discussion about adopting your work and making LDAP part of the core project.

PR #7734, opened 2018-03-07
Resize images on client-side before uploading

### LINKED ISSUE #5621: Resize images on client-side before uploading
Instead of limiting the file size to 4 MB, high resolution images should be automatically resized on the client side before uploading them to the server. People are used to upload images straight from their camera and it's inconvenient to resize them manually.

One (easy) way to do this is described here: https://hacks.mozilla.org/2011/01/how-to-develop-a-html5-image-uploader/
## 

<bountysource-plugin>

---
There is a **[$200 open bounty](https://www.bountysource.com/issues/8261972-resize-images-on-client-side-before-uploading?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Fensterbank:
This pull requests implements client-side resizing of images before uploading them.
While FineUploader provides scaling of selected images, its internal image resize code delegates to the `drawImage` method on the browser's native `CanvasRenderingContext2D` object which means linear interpolation in most browser.

So for quality and performance reasons I added the [Pica-Library](https://github.com/nodeca/pica) which is makeing a much better resizing and uses a Webworker for it.
~~Since we support Internet Explorer 11 and Pica / FineUploader are working with promises here I added [Lie](https://github.com/calvinmetcalf/lie) as a small, performant promise library.
I'm not so sure about this or if there is a better way to just make a Promise polyfill for IE11 without affect the modern browsers.~~

Each image is resized to a max width or max height of 3072 pixels before uploading, so in the future the original images stored on the server will never exceed this size.
In my tests I successfully selected big photos (> 12 MB) to upload. The biggest file sent to the server was 2,4 MiB.

Scaling is disabled when running in a PhantomJS instance because it didn't seem to work at all.

This fixes #5621 and makes #4997 less important.

### DIFF
--- a/spec/javascripts/app/views/publisher_view_spec.js
+++ b/spec/javascripts/app/views/publisher_view_spec.js
@@ -539,36 +539,36 @@ describe("app.views.Publisher", function() {
         this.view = new app.views.Publisher();
 
         // replace the uploader plugin with a dummy object
-        var upload_view = this.view.viewUploader;
+        var uploadView = this.view.viewUploader;
         this.uploader = {
-          onProgress: _.bind(upload_view.progressHandler, upload_view),
-          onSubmit:   _.bind(upload_view.submitHandler, upload_view),
-          onComplete: _.bind(upload_view.uploadCompleteHandler, upload_view)
+          onProgress: _.bind(uploadView.progressHandler, uploadView),
+          onUpload: _.bind(uploadView.uploadStartedHandler, uploadView),
+          onComplete: _.bind(uploadView.uploadCompleteHandler, uploadView)
         };
-        upload_view.uploader = this.uploader;
+        uploadView.uploader = this.uploader;
       });
 
-      context('progress', function() {
-        it('shows progress in percent', function() {
-          this.uploader.onProgress(null, 'test.jpg', 20, 100);
+      context("progress", function() {
+        it("shows progress in percent", function() {
+          this.uploader.onProgress("test.jpg", 20);
 
           var info = this.view.viewUploader.info;
-          expect(info.text()).toContain('test.jpg');
-          expect(info.text()).toContain('20%');
+          expect(info.text()).toContain("test.jpg");
+          expect(info.text()).toContain("20%");
         });
       });
 
-      context('submitting', function() {
+      context("submitting", function() {
         beforeEach(function() {
-          this.uploader.onSubmit(null, 'test.jpg');
+          this.uploader.onUpload(null, "test.jpg");
         });
 
-        it('adds a placeholder', function() {
+        it("adds a placeholder", function() {
           expect(this.view.wrapperEl.attr("class")).toContain("with_attachments");
           expect(this.view.photozoneEl.find("li").length).toBe(1);
         });
 
-        it('disables the publisher buttons', function() {
+        it("disables the publisher buttons", function() {
           expect(this.view.submitEl.prop("disabled")).toBeTruthy();
         });
       });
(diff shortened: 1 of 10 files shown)


Flaburgan:
Good news: we don't especially support IE11. If users are able to upload images without resizing them with IE11, we're good enough. We would even tend to not support it at all.

Fensterbank:
Okay. I'm not sure, if we want to have that it's possible to upload the original image without any limitation.
It could easily happen that huge files are saved on the server which the 4 MB limit prevented.

But yes, only for IE11 we don't need another Promise library. In fact the resizing based on linear interpolation brought with FineUploader is enough. Technically it's not the best quality but in my test it was much faster than resizing in IE11 with Pica, so we just use what FineUploader brings.

spec/javascripts/app/views/publisher_view_spec.js
```diff
@@ -542,7 +542,7 @@ describe("app.views.Publisher", function() {
         var upload_view = this.view.viewUploader;
         this.uploader = {
           onProgress: _.bind(upload_view.progressHandler, upload_view),
-          onSubmit:   _.bind(upload_view.submitHandler, upload_view),
+          onUpload:   _.bind(upload_view.uploadHandler, upload_view),
```
diaspora-code-review: - Extra space before value for key 'onUpload'.
- Identifier 'upload_view' is not in camel case.

spec/javascripts/app/views/publisher_view_spec.js
```diff
@@ -560,7 +560,7 @@ describe("app.views.Publisher", function() {
 
       context('submitting', function() {
         beforeEach(function() {
-          this.uploader.onSubmit(null, 'test.jpg');
+          this.uploader.onUpload(null, 'test.jpg');
```
diaspora-code-review: Strings must use doublequote.

denschub:
> Good news: we don't especially support IE11. If users are able to upload images without resizing them with IE11, we're good enough

I mean... if we only support the latest of the major browser, it's fine by me. And that would be Edge! And since Edge has Promise-support, that's fine. At least for me.

However, I'm not sure about your approach. Is there an option to only resize when the image is larger than 4MB and not touch the image at all if it's below the limit?

Fensterbank:
> However, I'm not sure about your approach. Is there an option to only resize when the image is larger than 4MB and not touch the image at all if it's below the limit?

The options can only be set at initialization. But it seems we could achieve this by calling the [scaleImage](https://docs.fineuploader.com/branch/master/api/methods.html#scaleImage) method if necessary in the submit handler and then trigger the upload.

Since `scaleImage` returns a promise, we will definitely lose any scaling in IE11. But we can give the IE11 users the 4 MB validation as it has been before and the moderm browser will scale if necessary.

I'll take a look at it!

Fensterbank:
As suggested by @denschub the picture is now only scaled if its size is above the 4 MB size limit.
Smaller images will be uploaded untouched at it has been before.

If the browser does not support Promises or canvas scaling the logic is the same as always.
This means an IE11 user trying to upload a huge image will see the file too big error message while the modern browsers will scale.

Unfortunately we have a [bug in the FineUploader library](https://github.com/FineUploader/fine-uploader/issues/1978). Firefox for Android is considered to be a stock browser for which the scaling feature is disabled.
This means uploading pictures directly from the camera of an Android smartphone does currently only work in Chrome for Android.

denschub:
Sorry it took me a bit to review, but it looks good to me. Merged to `develop`, and we'll see if we break things. :)

Issue #7739, opened 2018-03-13
Stream sometimes doesn't load when opened in background tab


SuperTux88:
I observed this occasionally for quite some time now (so it isn't new). When loading a user profile in a background tab (I don't know if this is only for profiles or if it also happens for other streams, tags pages), and switch to that tab later, sometimes the spinner is still spinning, but never finishes. I'm only able to reproduce it when loading in background tab (maybe because background tabs get less resources and so some race condition happens that doesn't happen when going to a profile in the same tab). When I then reload the page the stream always loads.

It sometimes takes many tries to be able to reproduce it, but when the error happens and I open the javascript error console I see the following error:

`Error: The partial status-message could not be found`

(Also while trying to reproduce it, I saw that sometimes the timago-timestamp is missing until the first timeago-update, I don't know if this ythe same issue as when the stream doesn't load at all, or if it is the same issue as #7645 (but there the timeago is in the wrong language, not missing), or if this is a completely different issue).

cmrd-senya:
`status-message` is the only template that is used in another template (reshare) as a partial: [code link](https://github.com/diaspora/diaspora/blob/develop/app/assets/templates/reshare_tpl.jst.hbs#L24).

However partials are have to be registered, and here is the code responsible for that: [code link](https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/helpers/handlebars-partials.js).

I guess that this jQuery document ready handler is just not being executed fast enough so it happens that by the time the partial got to be rendered, the handler has not been fired yet.

So it only happens when the profile page (or other pages) you open contain reshares.

cmrd-senya:
@SuperTux88 could you please test if #7740 solves it for you?

PR #7740, opened 2018-03-16
Include Handlebars templates to main.js

cmrd-senya:
This is needed to properly initialize Handlebars partials without a document ready handler.

ref #7739

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -6,7 +6,6 @@
 //= link bookmarklet.js
 //= link mobile/bookmarklet.js
 //= link mobile/mobile.js
-//= link templates.js
 //= link error_pages.css
 //= link admin.css
 //= link rtl.css
--- a/app/assets/javascripts/app/helpers/handlebars-partials.js
+++ b/app/assets/javascripts/app/helpers/handlebars-partials.js
@@ -1,7 +1,4 @@
 // @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-v3-or-Later
 
-/* we need to wrap this in a document ready to ensure JST is accessible */
-$(function(){
-  Handlebars.registerPartial('status-message', HandlebarsTemplates['status-message_tpl']);
-});
+Handlebars.registerPartial("status-message", HandlebarsTemplates["status-message_tpl"]);
 // @license-end
--- a/app/assets/javascripts/jasmine-load-all.js
+++ b/app/assets/javascripts/jasmine-load-all.js
@@ -1,6 +1,5 @@
 //= require jquery3
 //= require handlebars.runtime
-//= require templates
 //= require main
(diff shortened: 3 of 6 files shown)


SuperTux88:
I don't know how to reproduce this in develop mode, and even in production it sometimes takes many tries until it happens, so it's difficult for me to test this.

But I don't know if this is the "correct" way of fixing this. It somehow makes sense to me, that registering it earlier probably solves the problem, I have no idea if this has other side-effect (especially because the comment probably had a reason and I don't know if this still is needed). Maybe @denschub can have a look at it and say if this is a good solution or if it should be done in a different way.

cmrd-senya:
I reproduced it in development by creating a profile with a reshare and then I opened it in a background tab (with a middle mouse button click).

> especially because the comment probably had a reason and I don't know if this still is needed

This code was originally introduced in 2012, here is the first revision: https://github.com/diaspora/diaspora/blob/5d051fcc5226c855d726c64647d5f04fbc40cf08/public/javascripts/app/helpers/handlebars-partials.js

At this point `handlebars_assets` gem wasn't used as the assets pipeline too. So the comment had a reason back then when there probably was some initializing code at the javascript app which had to be called first.

Now we're using `handlebars_assets` gem: https://github.com/leshill/handlebars_assets

The documentation for the gem explicitly [says](https://github.com/leshill/handlebars_assets#templates-directory) that you should require the templates at the javascript application main file (e.g. `app/assets/javascripts/application.js`). `handlebars_assets` will take care of loading the templates now during assets compilation, so no need to set document ready handlers anymore.

What I have done here is:
1) Fixed requiring templates according with the documentation of  `handlebars_assets`
2) Removed some legacy code

So the fix feels quite "correct" to me.

cmrd-senya:
It is deterministic that we register the partial when the template is already loaded because the `handlebars-partials.js` is required after the templates are required in the `app.js` so I don't think there are any side effects.

SuperTux88:
Thanks for the explanation, I think I understand now how it worked before and what you did here. :+1: 

But now as you include the templates to the `main.js` the templates are included twice, because they are also included in the application template here:

https://github.com/diaspora/diaspora/blob/8e88f4b18370a3877631277569d1d0ef5bbc6612/app/views/layouts/application.html.haml#L11

And they don't need to be precompiled separately anymore when they are included in the `main.js`, so this can be removed too:

https://github.com/diaspora/diaspora/blob/398235ee6f0ed23b4e95ffa7228025d87237600c/app/assets/config/manifest.js#L9

And maybe remove the `templates.js` file completely and include it with `require_tree` as described in the `handlebars_assets`.

cmrd-senya:
Updated

SuperTux88 (approved):
Looks good to me and seems to fix the bug on my develop pod.

SuperTux88:
Merged as f632f8a1f8fef3e7c9303c2ef373d72b929994aa

Thanks for debugging and fixing that :cookie:

PR #7742, opened 2018-03-16
Respond 422 with StatusMessagesController#create when aspect_ids don't contain correct aspects

### LINKED ISSUE #3862: Posts can be send to nonexistent aspect_ids
If you fake a post request and use a nonexistent aspect_id a post still get's created. It doesn't appear in the stream, but it can be accessed via the direct link.
This doesn't seem like a wanted behavior to me. Instead 404 or something like this should be returned.
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/114579-posts-can-be-send-to-nonexistent-aspect_ids?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cmrd-senya:
fixes #3862

### DIFF
--- a/app/controllers/status_messages_controller.rb
+++ b/app/controllers/status_messages_controller.rb
@@ -49,6 +49,8 @@ class StatusMessagesController < ApplicationController
       format.mobile { redirect_to stream_path }
       format.json { render json: PostPresenter.new(status_message, current_user), status: 201 }
     end
+  rescue StatusMessageCreationService::BadAspectsIDs
+    render status: 422, plain: I18n.t("status_messages.bad_aspects")
   rescue StandardError => error
     handle_create_error(error)
   end
--- a/app/services/status_message_creation_service.rb
+++ b/app/services/status_message_creation_service.rb
@@ -72,4 +72,7 @@ class StatusMessageCreationService
                        url:           short_post_url(status_message.guid, host: AppConfig.environment.url),
                        service_types: receiving_services)
   end
+
+  class BadAspectsIDs < RuntimeError
+  end
 end
--- a/app/services/status_message_creation_service.rb
+++ b/app/services/status_message_creation_service.rb
@@ -9,15 +9,16 @@ class StatusMessageCreationService
 
   def create(params)
     build_status_message(params).tap do |status_message|
+      load_aspects(params[:aspect_ids]) unless status_message.public?
       add_attachments(status_message, params)
       status_message.save
-      process(status_message, params[:aspect_ids], params[:services])
+      process(status_message, params[:services])
     end
   end
 
   private
 
-  attr_reader :user
+  attr_reader :user, :aspects
 
   def build_status_message(params)
     public = params[:public] || false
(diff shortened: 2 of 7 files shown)


app/controllers/status_messages_controller.rb
```diff
@@ -49,6 +49,8 @@ def create
       format.mobile { redirect_to stream_path }
       format.json { render json: PostPresenter.new(status_message, current_user), status: 201 }
     end
+  rescue StatusMessageCreationService::BadAspectsIDs
+    render status: 422, plain: "Provided aspects IDs aren't applicable (non-existent or not owned)"
```
SuperTux88: Even if this should probably never happen, but when somebody starts writing a post in one tab and then deletes an aspect in another tab, this could still be a normal scenario (so not trying to modify the request). So I think this error-message should be added to the translations.

app/services/status_message_creation_service.rb
```diff
@@ -54,13 +58,20 @@ def add_photos(status_message, photos)
     end
   end
 
-  def process(status_message, aspect_ids, services)
-    add_to_streams(status_message, aspect_ids) unless status_message.public
+  def validate_aspects
+    raise BadAspectsIDs if aspects.empty?
```
SuperTux88: I would move this inside the `load_aspects_from_param` (`load_aspects`) method, so you automatically validate it after load. I also think it simplifies it with fewer methods. And that way you can also simplify the block at line 12 to a oneliner.

app/services/status_message_creation_service.rb
```diff
@@ -54,13 +58,20 @@ def add_photos(status_message, photos)
...
+  end
+
+  def load_aspects_from_param(aspect_ids)
```
SuperTux88: Since you already parse the `params` outside of this method I would just call this `def load_aspects(aspect_ids)`.

SuperTux88 (changes requested):
Mostly nitpicking :)

app/services/status_message_creation_service.rb
cmrd-senya: Well, to my taste having more little methods that are doing single thing is a simpler design than having fewer methods doing more :)

I would move the call of  `validate_aspects` to `load_aspects` but leave the method itself unless you're strong about the deletion.

SuperTux88: > Well, to my taste having more little methods that are doing single thing is a simpler design

It depends. When the method is as simple and short as this one and is only used once, this line could simple replace the line where this method is called. Because then you would save 4 lines which makes the file shorter which increases overall clarity in my opinion.

When the method is used multiple times and/or is a complex line where the method name can simply describe what it does, then I agree with you.

But I think that's not the case here. This method will never be called twice, because it makes only sense to call it at his one point (after loading the aspects) and when you move this to the `load_aspects` method, that is clear and you could even use `@aspects` directly so it's obvious that this validates the just loaded aspects and doesn't call a magic getter that hopefully was loaded before. So it should **never** be called before the `load_aspects` and it doesn't make sense to call it a second time after it, and I think the simplest way to signal this is to just inline the code to `load_aspects` (and the `load_apsect` is then still simple enough even with this code included).

However, if you still think it's better to have this in a separate method, leave it that way, I won't block on that.

SuperTux88:
@Flaburgan yes, that stops users to post to invalid and no aspects.

app/services/status_message_creation_service.rb
cmrd-senya: Okay, it makes sense to me :+1: 

I'll inline the code.

SuperTux88 (approved):

SuperTux88:
Merged as 0080a2567f279ef320f0c3176b3b0797939a5ac5 and e0e40f73c32c3bf4e118f665915333e6a6952016

Thank you :cookie: 

By the way, it would help a little if you would add the "fixes #xxxx" to the commit-message too, because when pulling the PRs to `next-minor` GitHub doesn't detect this PR as merged, so the PR description doesn't close the issues you fixed, but pushed commits would close them. I currently always add that to the commit-message when I add the changelog entry, so it isn't a big deal for me to add it, but it was just something I wanted to mention :)

PR #7744, opened 2018-03-21
Fixes #7377 - Conversation form on contacts page prefilled with only sharing contacts

### LINKED ISSUE #7377: Conversation form on contacts page prefilled with only sharing contacts
We should filter the prefilled contacts in the conversation form on the contacts page. Right now we add all contacts in an aspect but we should only add the mutual contacts.

<bountysource-plugin>

---
There is a **[$5 open bounty](https://www.bountysource.com/issues/43205258-conversation-form-on-contacts-page-prefilled-with-only-sharing-contacts?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lu-x:


SuperTux88 (changes requested):
First: Thanks for your pull request. :heart:

Can you please also change the test [here](https://github.com/diaspora/diaspora/blob/develop/spec/javascripts/app/pages/contacts_spec.js#L292-L302) so it fails without this patch. You probably need to change the fixture generated [here](https://github.com/diaspora/diaspora/blob/26a9e50ef935628c800f9a21d345057556fa5c31/spec/controllers/jasmine_fixtures/contacts_spec.rb#L22-L25) so it includes a contact which isn't mutual.

kas70:
Sure, thanks for your hints.

kas70:
@SuperTux88 done

spec/javascripts/app/pages/contacts_spec.js
```diff
@@ -297,7 +297,7 @@ describe("app.pages.Contacts", function(){
       expect(app.views.ConversationsForm.prototype.initialize).toHaveBeenCalled();
 
       var prefill = app.views.ConversationsForm.prototype.initialize.calls.mostRecent().args[0].prefill;
-      var contacts = app.contacts.filter(function(contact) { return contact.inAspect(app.aspect.get("id")); });
+      var contacts = app.contacts.filter(function(contact) { return contact.person.get("relationship") === "mutual" && contact.inAspect(app.aspect.get("id")); });
```
SuperTux88: I don't know why @diaspora-code-review didn't complain about this (looks like it is broken, maybe @denschub can have a look at that?), but this line is longer than 120 chars. But you can simply write it in multiple lines, the same as it is in `contacts.js`.

By the way, you can run these styleguide checks locally with `bin/pronto run -c develop`, see here for more information about that: https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review

spec/controllers/jasmine_fixtures/contacts_spec.rb
```diff
@@ -20,6 +20,7 @@
     end
 
     it "generates the aspects_manage_contacts_json fixture", fixture: true do
+      alice.disconnect(User.first.contact_for(bob.person))
```
SuperTux88: `User.first` could fail in the future because `alice` doesn't need to be the `first` user.

But I think it's better to add a third non-mutual user instead of converting one of the existing two contacts a non-mutual. That way we can still check if it still works for multiple mutual users and not just returns one user.

So that would be something like that instead:

```ruby
bob.share_with(FactoryGirl.create(:person), @aspect)
```

SuperTux88 (changes requested):
Thanks for the test. :+1:  It looks already quiet well, only some small stuff.

spec/javascripts/app/pages/contacts_spec.js
diaspora-code-review: Line 300 exceeds the maximum line length of 120.

SuperTux88: @denschub thanks for fixing the bot :cookie:

kas70: Done

spec/controllers/jasmine_fixtures/contacts_spec.rb
kas70: Done.

kas70:
@SuperTux88 Everything should be fixed now. Thank you for your advice!

.byebug_history
```diff
@@ -0,0 +1,48 @@
+c
```
SuperTux88: This file should be removed.

SuperTux88 (changes requested):
The code looks OK now, thanks for the fixes :+1: But you added a `.byebug_history` file now that shouldn't be there, but you can add it to the `.gitignore` if you want.

.byebug_history
kas70: Of course, sorry. Should i add it to gitignore?

SuperTux88 (approved):

.byebug_history
SuperTux88: If you want to use byebug regularly in the future, it probably makes sense to add it (otherwise you need to think about it every time not to add it). You can add it in a separate commit here or create a new PR for it.

kas70: Ok, i'll add a commit.
I'd also like to add awesome_print (https://github.com/awesome-print/awesome_print) if you don't mind.

SuperTux88: Thanks.

I'm not sure about awesome_print, I think we want to keep as little dependencies as possible and for me `pp` (pretty print) was always enough. But I think we can discuss that in the future if we want to add that.

kas70: Didn't know pp. Sounds good :+1:

spec/controllers/jasmine_fixtures/contacts_spec.rb
```diff
@@ -11,6 +11,8 @@
       @aspect = bob.aspects.create(:name => "another aspect")
       bob.share_with alice.person, @aspect
       bob.share_with eve.person, @aspect
+      # not mutual contact
+      bob.share_with(FactoryGirl.create(:person), @aspect)
```
SuperTux88: It looks like adding this to the `before` has side-effects to other tests which use the another fixture generated here (see [travis](https://travis-ci.org/diaspora/diaspora/jobs/356786625)). I think it's probably the easiest to just move this line to the `generates the aspects_manage_contacts_json fixture` again where it was before.

SuperTux88 (changes requested):
Sorry I need to unapprove that once more. There are tests failing now.

spec/controllers/jasmine_fixtures/contacts_spec.rb
kas70: Hmm. Everything is green here: https://travis-ci.org/diaspora/diaspora/builds/356791736?utm_source=github_status&utm_medium=notification

SuperTux88: Yes, it doesn't fail always, but it fails randomly, and it's at a place where you changed something, so I think it's related (because these tests worked stable before).

kas70:
@SuperTux88 Done, if specs are green 😉 🤞

SuperTux88 (approved):

SuperTux88:
Merged as 40fc175d6ed052b52aaabad01fc6da5c65ba2186 and 53b733f0cd89669b0e0ba7d970ac8fd0716478d7.

Thank you :cookie: and congratulations for your first pull request! :tada:

SuperTux88:
Oh, and I almost forgot, don't forget to claim your bounty here: https://www.bountysource.com/issues/43205258-conversation-form-on-contacts-page-prefilled-with-only-sharing-contacts

Issue #4299, opened 2013-07-15
Two factor authentication


raisty:
I think there should be two step verification as well.
Use open ID can be usefull too.
##

jaywink:
I saw a bounty was posted on this issue which I guess means we should at least think about how this could be implemented. AFAIK, two-factor identification usually means SMS. I don't believe that is a possibility in the decentralized environment were hundreds of podmins would have to start signing up to SMS gateways, which probably are not free to use.

Are there any other possibilities outside SMS for two factor authentication which would work in a decentralized environment? If not, I suggest we strongly think about closing this issue from the core so that at least bounties will not be fruitlessly wasted on something that is not possible to implement in a decentralized world in a way that *most* of the podmins will be able to activate the feature.

denschub:
Usually, two-factor authentication means TOTP, not SMS. I'm pretty sure there are devise plugins for TOTP out there, so besides the user preference storing and modifying the UI, this should be one of the easier features.

ABISprotocol:
Hello, there may be a decentralized 2FA out there.  Possibly you could use one of a few different options depending on what you want to do. 
1) something like what has been already mentioned by @denschub - that is TOTB / devise, such as
https://github.com/authy/authy-devise
or
2) BitID for open / decentralized 2FA https://github.com/bitid/bitid

Or, maybe 2FA isn't what's needed.  Per Bruce Schneier, something else may be needed entirely:
https://www.schneier.com/blog/archives/2006/11/fighting_fraudu.html

genma:
CozyCloud has implemented Two Factors authentication (https://forum.cozy.io/t/two-factor-authentication-in-cozy/2568) maybe it could help to choose "technology". My 2 cents.

svbergerem:
Here is a devise plugin for TOTP: https://github.com/tinfoil/devise-two-factor

ABISprotocol:
Thanks @svbergerem - that's a good one.

tchncs:
Mastodon just got tinfoil/devise-two-factor - i really enjoy it.
Would be nice to see it in diaspora* as well.

oliof:
I would like to see 2 factor authentication support for u2f/fido tokens. Sample implementation at https://github.com/TwoFactorAuth/ruby.

tursiops33:
Hi, I'm also interested in this, and a quick look on the Web, found this https://dynalogin.org it supports openid which diaspora already has, and is open source.

denschub:
[...]

Issue #7442, opened 2017-05-03
Add possibility to edit users e-mail address in the admin panel


SuperTux88:
Sometimes users use a wrong e-mail address (or they deleted their mail account) and lost their password. If they can prove that the account belongs to them, then the podmin should be able to change the mail-address without going to the database.

We already display the e-mail address in the results of the user search, so it's probably the best place to add a button to edit it.

As an extra feature, it would also be cool to disable mails for a user, because if they use a wrong address, it creates a lot of bounces. We already have a `disable_mail` flag in the `users` table, so we need just a checkbox to set it.

ngetahun:
Hi is this issue up for grabs?

SuperTux88:
Yes, you can work on it. :+1:

PR #7751, opened 2018-03-26
Two factor authentication

lislis:
Hey there!
This is my attempt to implement 2fa, ticket #4299

I know, there are no tests (:see_no_evil:), all strings are hardcoded, and the UI is far from 'done' but I'm at a point where I need some feedback on the general direction.

Some screenshots:

New link on the settings page, 2fa deactivated
screenshot-2018-3-26 two factor

Scan the code and confirm
screenshot-2018-3-26 two factor 1

Confirmation and backup codes
screenshot-2018-3-26 two factor 2

Settings page, 2fa activated
screenshot-2018-3-26 two factor 3

After sign-in, prompt for token
screenshot-2018-3-26 diaspora - sign in

And that is it so far.

Missing:
- [x] tests!
- [x] put copy from templates into translation files
- [x] fix UI

and anything else that's missing


app/controllers/sessions_controller.rb
```diff
@@ -5,10 +5,60 @@
...
+  def find_user
+    if session[:otp_user_id]
+      User.find(session[:otp_user_id])
+    elsif params[:user][:username]
+      User.find_for_authentication(username: params[:user][:username])
+    end
+  end
+
+  def authenticate_with_2fa
+    u = self.resource = find_user
```
kas70: Using shortcuts is bad style imho.

app/controllers/sessions_controller.rb
```diff
@@ -5,10 +5,60 @@
 #   the COPYRIGHT file.
 
 class SessionsController < Devise::SessionsController
+  before_action :authenticate_with_2fa, only: [:create]
   after_action :reset_authentication_token, only: [:create]
   before_action :reset_authentication_token, only: [:destroy]
 
+  def find_user
+    if session[:otp_user_id]
```
kas70: Maybe: 
```
return User.find(session[:otp_user_id]) if session[:otp_user_id]
User.find_for_authentication(username: params[:user][:username]) if params[:user][:username]

```

app/views/two_factor_authentications/_confirm.haml
```diff
@@ -0,0 +1,22 @@
+.row
+  .col-md-12
+    %h3 Confirm two factor authentication
+    %p Two factor authentication is not fully activated yet
+
+    .row
+      .col-md-6
+        %p Please scan the QR code with a TOTP capable app (eg Google authenticator).
```
Flaburgan: I'd love to not encourage Google usage :stuck_out_tongue: So at least list other apps. I'm not familiar with TOTP, I think [FreeOTP](https://freeotp.github.io/) is one. If someone can check if it is available on F-droid? Or an other trusted app listed there. That would be awesome!

app/views/two_factor_authentications/_confirm.haml
```diff
@@ -0,0 +1,22 @@
...
+        = raw RQRCode::QRCode.new(@qrcode_uri).as_svg(padding: 0, module_size: 5)
+
+      .col-md-6
+        %p In case you can't scan the ocde you can manually enter the secret.
```
Flaburgan: Typo ;)

app/views/users/_edit.haml
```diff
@@ -65,7 +65,7 @@
             = f.submit t(".change_language"), class: "btn btn-primary pull-right"
     %hr
 
-    .row
+    .rowq
```
Flaburgan: Typo

.ruby-version
```diff
@@ -1 +1 @@
-2.4
+2.4.2
```
Flaburgan: Please do not do that. We use RVM which works well without hardfixing the hotfix version.

Gemfile.lock
```diff
@@ -138,7 +141,7 @@ GEM
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
     crass (1.0.3)
-    cucumber (2.4.0)
+    cucumber (2.99.0)
```
Flaburgan: I don't think all those bumps are necessary for your PR, are they?

app/models/user.rb
```diff
@@ -19,7 +19,17 @@ class User < ApplicationRecord
   scope :halfyear_actives, ->(time = Time.now) { logged_in_since(time - 6.month) }
   scope :active, -> { joins(:person).where(people: {closed_account: false}) }
 
-  devise :database_authenticatable, :registerable,
+  attribute :otp_secret
+
+  devise :two_factor_authenticatable,
+         otp_secret_encryption_key: "c0f2d6a8fa2873e40be5768f99b99260a3b0462ad9dd3f3\
+71432751b4e632f6ab4519b97dc7fdf5fa90c4dfdffa3ed0a917cd7a51382bac8b26166a5540bc603"
```
Flaburgan: Not familiar with this but usually secret keys are not available on Github ;)

Flaburgan (changes requested):
First of all, thank you very much for working on this. The flow looks good in my opinion, no global blocking points, so you should be able to continue that way and polish your work.

Two remarks:
- I don't see where the hashing algorithm is set. Can you assure it's a modern one, not SHA1 which was the default some years ago?
- I'm a bit afraid of the support load this new feature can bring. Framasphere moderators are already spending a lot of time with users who lost access to their account (that's why we have feature requests like https://github.com/diaspora/diaspora/issues/7442). The question is, is 2FA added to protect users from a stolen password (if so, then when resetting the password by e-mail, the 2FA code should not be asked) or is it also aimed to protect them if their email gets compromised, like you did here. (In that case, you're in trouble anyway but eh, that's in part why 2FA has been built), At least the podmin should be able to access recovery code easily I guess, as we will probably be asked them a lot...

app/models/user.rb
SuperTux88: I had a look at the 2fa topic already too and compared different possible solution, and I think the `devise-two-factor` @lislis is using here was the one I would have chosen too. However this lib requires a key to encrypt the OTP secret, which I don't know if we really need this. But there is no easy way to disable encryption (no direct config flag).

The solution is the documentation is to use an env-variable for the key:

```ruby
 :otp_secret_encryption_key => ENV['YOUR_ENCRYPTION_KEY_HERE']
```

However this would add more complexity to the installation/upgrade process because every podmin would be needed to set this. So I don't like this solution.

We could add the key to the diaspora.yml, but that wouldn't solve the problem, that every podmin would need to change it (and you can't change it later, you always have to use the same key). Also podmins would need to set a key with the correct length, which again is even more complexity.

A static key would solve this problem, but I agree with you, it looks a bit strange to see this on GitHub ;)

We could try to disable the encryption with setting an `if` option and defining the attribute before the library does it, so something like this maybe works (didn't test it):

```ruby
attr_encrypted :otp_secret, if: false
```

The reason why I'm not sure if the encryption is really needed is: When somebody is already in the database and the encryption protects from reading the OTP secret, the attacker is already in the database and can access all other data and maybe even change stuff. When the attacker is on the machine where diaspora runs, they probably even have access to the key and can decrypt the secret. So the additional protection with the encryption is very low compared with the additional complexity. An attacker couldn't even do anything with the OTP secret, without the password you can't login, and you can't use the OTP secret to login on other pages, because it is a generated value only for diaspora.

A possibility would be to generate a key with the correct size at the first start (similar to the `secret_token` for the session cookie). But then we still need to make sure that podmins copy the key to a new server if they migrate the installation to a new hardware and include it in the backups, otherwise all users with 2fa enabled can't login anymore (while when they forget to migrate the session secret token only all sessions are invalid, and people just need to login again).

So I think no key (no encryption, if possible) or a static key is currently the best solution, but I'm open for other opinions.

SuperTux88:
> * I don't see where the hashing algorithm is set. Can you assure it's a modern one, not SHA1 which was the default some years ago?

It's not hashing, it's encryption, but see https://github.com/diaspora/diaspora/pull/7751#discussion_r177337965 for more information about that topic.

> * I'm a bit afraid of the support load this new feature can bring. Framasphere moderators are already spending a lot of time with users who lost access to their account (that's why we have feature requests like #7442). The question is, is 2FA added to protect users from a stolen password (if so, then when resetting the password by e-mail, the 2FA code should not be asked) or is it also aimed to protect them if their email gets compromised, like you did here. (In that case, you're in trouble anyway but eh, that's in part why 2FA has been built), At least the podmin should be able to access recovery code easily I guess, as we will probably be asked them a lot...

Yes, probably some users will try to ask you to reset their 2fa, but you shouldn't reset it, otherwise 2fa is completely useless. It is optional (so everybody activating it should be aware of this) and there are backup keys (and they are there for a reason). If people lose their 2fa secret and their backup keys, they can create a new account. There is no (official) possibility to reset 2fa if you lost it (there is no "forgot 2fa" similar to the "forgot password", obviously you as a podmin can reset your 2fa if you want, but you shouldn't do that for other users you don't know personally). So you can just create a default answer for everybody asking to reset their 2fa and tell them that it isn't possible and they should create an new account and next time backup their backup keys.

(personally I already see setting another mail-address as critical, when you can't verify if it's the real owner of the account, and the mail-address is the only way to do this. But that's another topic.)

app/models/user.rb
Flaburgan: >However this would add more complexity to the installation/upgrade process because every podmin would be needed to set this. So I don't like this solution.

We already do that for Camo, and simply ask the podmin to add `export` in there `.bashrc`. This looks good enough to me, but I agree the encryption is maybe not needed.

Gemfile.lock
SuperTux88: Updating that many dependencies as drive-by doesn't look like a good idea to me, but it can be that some of them got updated because we didn't update many gems for a long time now. I planned to do a gem-update PR again soon and it will probably create a lot of conflicts in the Gemfile.lock, but you can rebase after that and generate a new Gemfile.lock with your changes, then it should look a lot cleaner.

app/models/user.rb
SuperTux88: But Camo is optional for podmins, while this is optional for the users to activate, but it's not optional for podmins, so every podmin would need to do it when updating or in the future when installing diaspora. Diaspora alone currently doesn't need any env variables (all options are in the `diaspora.yml`), and the `.bashrc` wouldn't be the correct place to add something like this (it would maybe work for some situations, but it's nothing I want to write in the official installation guides).

jhass: Don't we already do the key generation for OpenID Connect too?

app/views/two_factor_authentications/_confirm.haml
lislis: Yeah, totally. If you'd recommend FreeOTP then I can write that. Do you know by chance if there is also an alternative app for iOS?

.ruby-version
lislis: Ah yeah, didn't mean to check that in. I'm using rbenv locally and it didn't like that for some reason :roll_eyes:

Gemfile.lock
lislis: Agreed!

app/models/user.rb
lislis: > Don't we already do the key generation for OpenID Connect too?

If it's already being done somewhere that'd be great! Otherwise I'm fine Tux's suggestion, too.

lislis:
Thanks for the comments so far <3 I'll look into updating the PR this weekend!

.ruby-version
jhass: Yeah, rbenv unfortunately intentionally doesn't support this. You can workaround by symlinking a specific version to 2.4, I think there's some alias plugin which automates it a bit.

Gemfile
```diff
@@ -27,7 +27,9 @@ gem "json-schema", "2.8.0"
 # Authentication
 
 gem "devise", "4.4.1"
+gem "devise-two-factor"
 gem "devise_lastseenable", "0.0.6"
+gem "rqrcode"
```
jhass: Please add specific versions for the dependencies you add like everywhere else in the Gemfile, to keep it consistent :)

app/controllers/sessions_controller.rb
```diff
@@ -5,10 +5,60 @@
...
+  end
+
+  def prompt_for_two_factor(user)
+    session[:otp_user_id] = user.id
+    render :two_factor
+  end
+
+  def create
+    super
+  end
```
jhass: Maybe I'm missing something but I think this definition is superfluous :)

jhass (commented):
On a good way!

app/views/two_factor_authentications/_confirm.haml
denschub: > Do you know by chance if there is also an alternative app for iOS?

Yes, FreeOTP!

app/models/user.rb
denschub: > Don't we already do the key generation for OpenID Connect too?

Well, we kinda do, but that's happening in [`lib/api/openid_connect/id_token_config.rb`](https://github.com/diaspora/diaspora/blob/ae17248b2ecb462d7b7471d3173b27c0cb6c64a8/lib/api/openid_connect/id_token_config.rb) and isn't the nicest way of doing that, to be honest. However, there is a [task to generate the session secret](https://github.com/diaspora/diaspora/blob/ae17248b2ecb462d7b7471d3173b27c0cb6c64a8/lib/tasks/generate_session_secret.rake), and one could easily take that as a template for a new task.

denschub:
> The question is, is 2FA added to protect users from a stolen password (if so, then when resetting the password by e-mail, the 2FA code should not be asked) or is it also aimed to protect them if their email gets compromised

That really doens't matter. If you have forgotten your password, you can reset it, and you can get a new one via email. It doesn't matter if we ask for a 2fa token or not (and in fact, the industry standard is *not* to ask, because it's kinda pointless). In any case, resetting the password does not reset the 2fa token, and it should not. So even if your eMail address is somehow borked and someone manages to reset the password, there is still no way to get into the account, which is the point of 2fa.

As @SuperTux88 already outlined, there is no way to reset/disable 2fa without a valid TOTP, and there shouldn't be one. Now, there is a risk of someone losing their phone (which generates the TOTP) without having a backup, and that's bad. Like most large applications supporting 2fa, we should generate some backup tokens for users to store in a secret place (like a password manager) to be used as a valid token if they lost their generator. The gem used here [has built-in support for this](https://github.com/tinfoil/devise-two-factor/tree/359a31bcf7c8ee30484f8c79acb3f8d5ecde73b3#backup-codes), and we should expose that.

If people enable 2fa, but fail to backup their keys and the recovery key, they'll have a hard (read: impossible) time to get their account back. This is by design. And this is just the way it works here on GitHub as well: [If you cannot use the provided validation schemes, you are goofed](https://help.github.com/articles/recovering-your-account-if-you-lost-your-2fa-credentials/), the support will not reset your account. Neither will Google, fwiw.

Besides me rambling around... @lislis, thanks for working on this!

Gemfile.lock
SuperTux88: By the way, the gems were updated in #7772 which was merged 4 days ago and which is the reason you have now many conflicts in the Gemfile and Gemfile.lock. But you can just rebase and the latest `develop` branch and you should have a much cleaner Gemfile.lock then :)

app/views/two_factor_authentications/_confirm.haml
```diff
@@ -0,0 +1,22 @@
+.row
+  .col-md-12
+    %h3 Confirm two factor authentication
+    %p Two factor authentication is not fully activated yet
+
+    .row
+      .col-md-6
+        %p Please scan the QR code with a TOTP capable app (eg Google authenticator).
+        = raw RQRCode::QRCode.new(@qrcode_uri).as_svg(padding: 0, module_size: 5)
```
diaspora-code-review: - Rails/OutputSafety: Tagging a string as html safe may be a security risk.
- Avoid using instance variables in partials views

app/controllers/sessions_controller.rb
```diff
@@ -5,10 +5,51 @@
 #   the COPYRIGHT file.
 
 class SessionsController < Devise::SessionsController
+  before_action :authenticate_with_2fa, only: [:create]
```
diaspora-code-review: Rails/LexicallyScopedActionFilter: `create` is not explicitly defined on the class.

app/models/user.rb
```diff
@@ -19,7 +19,17 @@ class User < ApplicationRecord
...
+
+  devise :two_factor_backupable,
+         otp_backup_code_length: 16,
```
diaspora-code-review: Layout/AlignHash: Align the elements of a hash literal if they span more than one line.

config/routes.rb
```diff
@@ -118,6 +118,12 @@
     get "getting_started_completed" => :getting_started_completed
   end
 
+  resource :two_factor_authentication, only: %i(show create destroy) do
```
diaspora-code-review: Style/PercentLiteralDelimiters: `%i`-literals should be delimited by `[` and `]`.

app/controllers/sessions_controller.rb
lislis: This one is tricky, `#create` comes from devise but I need to run the before action if a user has 2fa enabled

lislis:
Update: found some time to work on this again.

Here the current state of the UI:

screenshot-2018-5-6 two factor authentication
screenshot-2018-5-7 confirm activation
screenshot-2018-5-7 two factor authentication recovery codes
screenshot-2018-5-7 diaspora - two factor authentication

screenshot-2018-5-7 two factor authentication 1

Flaburgan:
It looks very nice, awesome work! Is there a screen where the user can enter the recovery codes too?

app/views/two_factor_authentications/_confirm.haml
```diff
@@ -0,0 +1,31 @@
...
+  .row
+    .col-md-12
+      %h3= t("two_factor_auth.confirm.title")
+      %p= t("two_factor_auth.confirm.status")
+      .small-horizontal-spacer
+      %h4= t("two_factor_auth.confirm.scan_title")
+  .row
+    .col-md-6
+      %p= t("two_factor_auth.confirm.scan_explanation")
+      = raw RQRCode::QRCode.new(@qrcode_uri).as_svg(padding: 0, module_size: 5)
```
diaspora-code-review: - Rails/OutputSafety: Tagging a string as html safe may be a security risk.
- Avoid using instance variables in partials views

app/views/two_factor_authentications/_deactivate.haml
```diff
@@ -0,0 +1,26 @@
...
+            = f.text_field :code, placeholder: t("two_factor_auth.input_token.placeholder"), class: "form-control"
+            = t("two_factor_auth.recovery.reminder")
+        .clearfix= f.submit t("two_factor_auth.activated.change_button"), class: "btn btn-primary pull-right"
+
+      %hr
+
+      %h4= t("two_factor_auth.recovery.title")
+      %p= t("two_factor_auth.recovery.explanation_short")
+      %p= t("two_factor_auth.recovery.invalidation_notice")
+      %p= link_to t("two_factor_auth.recovery.button"), recovery_codes_two_factor_authentication_path, class: "btn btn-default"
```
diaspora-code-review: Line is too long. [127/120]

app/models/user.rb
```diff
@@ -19,7 +19,15 @@ class User < ApplicationRecord
   scope :halfyear_actives, ->(time = Time.now) { logged_in_since(time - 6.month) }
   scope :active, -> { joins(:person).where(people: {closed_account: false}) }
 
-  devise :database_authenticatable, :registerable,
+  attribute :otp_secret
+
+  devise :two_factor_authenticatable,
+         :two_factor_backupable,
+         otp_secret_encryption_key: Diaspora::Application.config.twofa_encryption_key,
```
diaspora-code-review: Layout/AlignHash: Align the elements of a hash literal if they span more than one line.

app/models/user.rb
```diff
@@ -19,7 +19,15 @@ class User < ApplicationRecord
...
+         otp_backup_code_length: 16,
```
diaspora-code-review: Layout/AlignHash: Align the elements of a hash literal if they span more than one line.

config/initializers/twofa_encryption_key.rb
```diff
@@ -0,0 +1,11 @@
...
+
+# The 2fa encryption key is used to encrypt users' OTP tokens in the database.
+
+# You can regenerate this key by running `rake generate:twofa_key`
+
+# If you change this key after a user has set up 2fa
+# the users' tokens cannot be recovered
+# and they will not be able to log in again!
+
+Diaspora::Application.config.twofa_encryption_key = '052717f82d2676edef29695037fc21b6eb83661fb40ab5d3fe709cc34095ba08'
```
diaspora-code-review: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

lib/tasks/generate_2fa_encription_key.rake
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+namespace :generate do
+  desc "Generates a key for encrypting 2fa tokens"
+  task :twofa_key do
+    path = Rails.root.join("config", "initializers", "twofa_encryption_key.rb")
+    key = SecureRandom.hex(32)
+    File.open(path, "w") do |f|
+      f.write <<EOF
+# The 2fa encryption key is used to encrypt users' OTP tokens in the database.
```
diaspora-code-review: Layout/IndentHeredoc: Use 2 spaces for indentation in a heredoc by using `<<~` instead of `<<`.

lib/tasks/generate_2fa_encription_key.rake
```diff
@@ -0,0 +1,22 @@
...
+
+# You can regenerate this key by running `rake generate:twofa_key`
+
+# If you change this key after a user has set up 2fa
+# the users' tokens cannot be recovered
+# and they will not be able to log in again!
+
+Diaspora::Application.config.twofa_encryption_key = '#{key}'
+EOF
```
diaspora-code-review: Naming/HeredocDelimiterNaming: Use meaningful heredoc delimiters.

lib/tasks/generate_2fa_encription_key.rake
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+namespace :generate do
+  desc "Generates a key for encrypting 2fa tokens"
+  task :twofa_key do
```
jhass: I guess we should automagically call this akin to how we do it for the secret token, also have handling for it for ephemeral filesystems such as heroku's. Have a peek in `lib/configuration_methods.rb` `secret_token` method and its usages.

jhass (changes requested):
Sorry for the delayed review. Looking very good, I think the below are the last things we need here :)

app/controllers/sessions_controller.rb
```diff
@@ -5,10 +5,55 @@
...
   after_action :reset_authentication_token, only: [:create]
   before_action :reset_authentication_token, only: [:destroy]
 
+  def create
+    super
+  end
```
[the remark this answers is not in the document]
lislis: It technically works without this, but then pronto gives me `app/controllers/sessions_controller.rb:8 W: Rails/LexicallyScopedActionFilter: `create` is not explicitly defined on the class.
` :woman_shrugging:

config/initializers/twofa_encryption_key.rb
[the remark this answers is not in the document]
lislis: :+1:

lib/configuration_methods.rb
```diff
@@ -69,6 +69,25 @@ def secret_token
       end
     end
 
+    def twofa_encryption_key
+      if heroku?
+        return ENV['TWOFA_ENCRYPTION_KEY'] if ENV['TWOFA_ENCRYPTION_KEY']
```
diaspora-code-review: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

lib/configuration_methods.rb
```diff
@@ -69,6 +69,25 @@ def secret_token
...
+        warn "FATAL: Running on Heroku with TWOFA_ENCRYPTION_KEY unset"
+        warn "       Run heroku config:add TWOFA_ENCRYPTION_KEY=#{SecureRandom.hex(32)}"
+        Process.exit(1)
```
diaspora-code-review: Rails/Exit: Do not use `exit` in Rails applications.

lib/configuration_methods.rb
```diff
@@ -69,6 +69,25 @@ def secret_token
...
+      else
+        key_file = File.expand_path(
+          '../config/initializers/twofa_encryption_key.rb',
```
diaspora-code-review: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

lib/configuration_methods.rb
```diff
@@ -69,6 +69,25 @@ def secret_token
...
+          File.dirname(__FILE__)
+        )
+        unless File.exist? key_file
```
diaspora-code-review: Style/IfUnlessModifier: Favor modifier `unless` usage when having a single-line body. Another good alternative is the usage of control flow `&&`/`||`.

lib/tasks/generate_2fa_encription_key.rake
```diff
@@ -0,0 +1,24 @@
...
+    path = Rails.root.join("config", "initializers", "twofa_encryption_key.rb")
+    key = SecureRandom.hex(32)
+    File.open(path, "w") do |f|
+      f.write <<~EOF
+      # frozen_string_literal: true
```
diaspora-code-review: Layout/IndentHeredoc: Use 2 spaces for indentation in a heredoc.

lib/tasks/generate_2fa_encription_key.rake
lislis: I was wondering about that :D I basically copied and adjusted the secret_token method. I hope it works like that

jhass (approved):

lislis:
@jhass there is still a problem in Travis where it can't find the `twofa_encryption_key`. How should I deal with that?

app/models/user.rb
jhass: Oh, this needs to be `AppConfig.twofa_encryption_key` then.

lislis: Ah, nice!

lib/configuration_methods.rb
```diff
@@ -52,23 +52,42 @@ def show_service?(service, user)
...
         Process.exit(1)
       else
         token_file = File.expand_path(
-          '../config/initializers/secret_token.rb',
+          "../config/initializers/secret_token.rb",
           File.dirname(__FILE__)
         )
         unless File.exist? token_file
-          `DISABLE_SPRING=1 bin/rake generate:secret_token`
+          "DISABLE_SPRING=1 bin/rake generate:secret_token"
```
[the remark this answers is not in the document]
lislis: ha, I was too hasty with my quote replacement here. I'll call system explicitly too make it clearer

lib/configuration_methods.rb
```diff
@@ -52,23 +52,42 @@ def show_service?(service, user)
...
+        warn "FATAL: Running on Heroku with TWOFA_ENCRYPTION_KEY unset"
+        warn "       Run heroku config:add TWOFA_ENCRYPTION_KEY=#{SecureRandom.hex(32)}"
+        Process.exit(1)
+      else
+        key_file = File.expand_path(
+          "../config/initializers/twofa_encryption_key.rb",
+          File.dirname(__FILE__)
+        )
+        unless File.exist? key_file
+          "DISABLE_SPRING=1 bin/rake generate:twofa_key"
```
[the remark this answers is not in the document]
lislis: :heavy_check_mark:

lislis:
Ah, the cucumber test for resetting passwords is outdated. 
With 2fa coming, you won't be logged in automatically anymore after resetting the password since that would bypass the second factor. Instead after typing your new password, you just have to manually sign in with your new password.
I'll look into updating the cucumber test

Gemfile.lock
```diff
@@ -347,10 +359,10 @@ GEM
     json-schema-rspec (0.0.4)
       json-schema (~> 2.5)
       rspec
-    jsonpath (0.8.11)
+    jsonpath (0.9.5)
       multi_json
       to_regexp (~> 0.2.1)
-    jwt (1.5.6)
+    jwt (2.1.0)
```
jhass: Looks like this update breaks our OpenID Connect implementation.

db/migrate/20180302105225_add_two_factor_backupable_to_user.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTwoFactorBackupableToUser < ActiveRecord::Migration[5.1]
+  def change
+    add_column :users, :otp_backup_codes, :string, array: true
```
jhass: Looks like it has to be a `text` as `string` is too short.

Gemfile.lock
lislis: Do you think that's the cause of the test failure and not what I was writing about?

jhass: I only looked at the rspec failures, your explanation for the cucumber one sounds plausible :)

db/migrate/20180302105225_add_two_factor_backupable_to_user.rb
lislis: cool, I'll do that

Gemfile.lock
lislis: I'm not sure if downgrading is an option. Maybe opening a separate PR to upgrade?

jhass: Depends on how big the fix is, if it's a one or two liner do as you prefer, I'd be fine to include it here too.

config/locales/diaspora/en.yml
```diff
@@ -1308,6 +1308,40 @@ en:
...
+      manual_explanation: "In case you can't scan the QR code automatically you can manually enter the secret."
+      input_title: "Confim with TOTP token"
+      input_explanation: "After scanning or entering the secret, enter the 6 digit code you see and confirm the setup."
+      activate_button: "Confirm and activate"
+    input_token:
+      label: "Two factor token"
+      placeholder: "6 digit two factor token"
+    recovery:
+      title: "Recovery codes"
+      reminder: "Alternatively, you can use a recovery code."
```
Flaburgan: you can use *one of the recovery codes* is more clear IMO

app/views/two_factor_authentications/_activate.haml
```diff
@@ -0,0 +1,10 @@
+.col-md-12
+  .row
+    .col-md-12
+      %h3= t("two_factor_auth.title")
+
```
Flaburgan: I think a small informational sentence here would help novice users understand what this is about. It can probably be kept on other pages too. Something like *"Two factor authentication is a powerful way to ensure you are the only one able to login to your account. It requires you to enter a 6 digits code generated from your phone after you entered your password to prove to diaspora\* that you are the owner. Be careful though: if you lose your phone app **and** the recovery codes created when you activate this feature, access to your diaspora\* account will be blocked forever."*

@goobertron can you please check the wording?

Flaburgan (changes requested):
I only did a "user point of view" test. It works very well but can be made a bit friendlier for users who don't know what 2FA is. We're getting there!

config/locales/diaspora/en.yml
```diff
@@ -1308,6 +1308,40 @@ en:
       email_confirmed: "Email %{email} activated"
       email_not_confirmed: "Email could not be activated. Wrong link?"
 
+  two_factor_auth:
+    title: "Two factor authentication"
+    activated:
```
goobertron: I would use 'two-factor authentication' here and throughout. The hyphen is good style when you have compound adjectives, as in this phrase.

config/locales/diaspora/en.yml
```diff
@@ -1308,6 +1308,40 @@ en:
...
+      change_label: "Activate two factor authentication"
+      change_button: "Activate"
+    confirm:
+      title: "Confirm activation"
+      status: "Two factor authentication is not fully activated yet, you need to confirm activation with a TOTP token"
+      scan_title: "Scan the QR code"
+      scan_explanation: "Please scan the QR code with a TOTP capable app (eg FreeOTP (Android, iOS), SailOTP (SailfishOS))."
+      manual_explanation: "In case you can't scan the QR code automatically you can manually enter the secret."
+      input_title: "Confim with TOTP token"
+      input_explanation: "After scanning or entering the secret, enter the 6 digit code you see and confirm the setup."
```
goobertron: Ditto: hyphen in '6-digit'. My personal preference would be to spell it out as 'six-digit', which I think looks better.

config/locales/diaspora/en.yml
```diff
@@ -1308,6 +1308,40 @@ en:
...
+      change_button: "Deactivate"
+    deactivated:
+      status: "Two factor authentication not activated"
+      change_label: "Activate two factor authentication"
+      change_button: "Activate"
+    confirm:
+      title: "Confirm activation"
+      status: "Two factor authentication is not fully activated yet, you need to confirm activation with a TOTP token"
+      scan_title: "Scan the QR code"
+      scan_explanation: "Please scan the QR code with a TOTP capable app (eg FreeOTP (Android, iOS), SailOTP (SailfishOS))."
```
goobertron: I suggest rephrasing this to avoid the double closing bracket at the end:

> "Please scan the QR code with a TOTP capable app, such as FreeOTP (Android, iOS), SailOTP (SailfishOS)."

config/locales/diaspora/en.yml
```diff
@@ -1308,6 +1308,40 @@ en:
...
+      manual_explanation: "In case you can't scan the QR code automatically you can manually enter the secret."
```
goobertron: I think we try to use typographical quotes and apostrophes in text strings: `can’t`.

app/views/two_factor_authentications/_activate.haml
goobertron: Hi, sorry for the delay. For some reason I didn't receive (or, more likely, didn't notice) a ping for this.

I'm not sure I fully understand how 2FA works. So I don't want to edit without understanding it. Your text suggests that this is just for phones, and requires an app. Is that correct? Once I know, I'll try to suggest text.

goobertron (commented):
Apologies; I didn't mean to start a review, but simply to leave a few comments on the text. I pressed the wrong button! I hope that doesn't cause any difficulties.

app/views/two_factor_authentications/_activate.haml
Flaburgan: If you don't get it with my text above, it's a proof that it's not good enough :stuck_out_tongue: 
Two factors authentication requires the user to enter a code provided by an application when he logs in. The application needs to be set up first, and that's done in the user settings. During that process, recovery codes are generated. They allow the user to log in without the code provided by the app (mainly used if the access to the app is lost). Is that clearer?

goobertron: Thanks. 

@Flaburgan, I don't feel competent right now to decide whether the text is factually correct, so I'll just make a few grammatical/stylistic corrections and leave it to someone who knows more about this to approve the content of the text.

> Two-factor authentication is a powerful way to ensure you are the only one able to sign in to your account. When signing in, you will enter a 6-digit code along with your password to prove your identity. Be careful though: if you lose your phone and the recovery codes created when you activate this feature, access to your diaspora* account will be blocked forever.

As this appears to be about how to sign in with 2FA rather than how to set it up, I have deleted a couple of parts that read as if they concern setting up 2FA. Also, you say 'if you lose your phone app'. As I understand it, this concerns signing in to the diaspora\* web-app itself, rather than any third-party app that a user might have on their phone in order to access the diaspora\* web-app. So I deleted 'app' from that clause in order to avoid that confusion.

I hope this is helpful, and I hope that someone else can confirm whether the facts in the text are correct and clear.

Flaburgan: It's shorter and the meaning is still there, so :+1: 

@lislis will ultimately choose anyway :stuck_out_tongue:

HankG:
> I only did a "user point of view" test. It works very well but can be made a bit friendlier for users who don't know what 2FA is. We're getting there!

Is there a server one can experiment with this on to help test?

Flaburgan:
>Is there a server one can experiment with this on to help test?

Looking how good the code actually is, I'm fine with adding it on top of diaspora-fr.org if @lislis rebase it on the develop branch.

lislis:
Hey y'all sorry for radio silence from my part. I'll make this PR a priority this week and finish the tests and last wording bits and of course the rebase :)

app/views/two_factor_authentications/_confirm.haml
```diff
@@ -0,0 +1,31 @@
...
+
+    .col-md-6
+      %p= t("two_factor_auth.confirm.manual_explanation")
+      %pre.well= current_user.otp_secret.scan(/.{4}/).join(" ")
```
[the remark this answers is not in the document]
lislis: I installed FreeOTP to better understand this and I'm quite confused myself. SailOTP is fine with the code as it is currently displayed, but apparently FreeOTP wants it base32 encoded (?!). I need to look into that

app/views/two_factor_authentications/_activate.haml
lislis: Reads well to me :)

lib/tasks/generate_2fa_encription_key.rake
```diff
@@ -0,0 +1,24 @@
...
+        # The 2fa encryption key is used to encrypt users' OTP tokens in the database.
+
+        # You can regenerate this key by running `rake generate:twofa_key`
+
+        # If you change this key after a user has set up 2fa
+        # the users' tokens cannot be recovered
+        # and they will not be able to log in again!
+
+        Diaspora::Application.config.twofa_encryption_key = "#{key}"
+      EOF
```
diaspora-code-review: Naming/HeredocDelimiterNaming: Use meaningful heredoc delimiters.

app/assets/stylesheets/two-factor-authentication.scss
```diff
@@ -0,0 +1,5 @@
+// Styles around two-factor authentication
+
+.two-factor-qr {
+  background-color: white;
```
diaspora-code-review: - Color `white` should be written in hexadecimal form as `#ffffff`
- Color literals like `white` should only be used in variable declarations; they should be referred to via variable everywhere else.

app/views/two_factor_authentications/_confirm.haml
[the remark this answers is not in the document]
lislis: It seems FreeOTP is sort of dead, I left comments in the tickets addressing the confusing manual UI and the base32 encoding but I think we should recommend another Android client that is more active. People suggested https://github.com/andOTP/andOTP and I think I'd update the copy to that

lislis:
I've stared to map out the steps I need to test this in cucumber tests but I'm super new to that. Any remarks (or useful tutorials) appreciated!

app/views/two_factor_authentications/_confirm.haml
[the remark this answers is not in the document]
Flaburgan: andOTP is available on F-droid so :+1: from me. Scrolling the list of OTP apps there, I also found FreeOTP+ which is "a fork of FreeTOP provided by Red Hat". Maybe you want to try it?

app/assets/stylesheets/two-factor-authentication.scss
```diff
@@ -0,0 +1,5 @@
+// Styles around two-factor authentication
+
+.two-factor-qr {
+  background-color: #ffffff;
```
diaspora-code-review: - Color literals like `#ffffff` should only be used in variable declarations; they should be referred to via variable everywhere else.
- Color `#ffffff` should be written as `#fff`

Flaburgan: We have $white in our variables already ;)

lislis:
Ok, finally made Travis pass :smiley_cat: I removed my train-of-thought cucumber file (for now) because I have no idea how writing the token generation part in there would work (yet). I need to give this some more thought. 
But I'd like to have a test for the user flow of signing in with 2fa activated, while I think the controller tests for activation and deactivation are fine.

spec/controllers/two_factor_authentications_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+      it "shows the QR verification code" do
+        get :confirm_2fa
+        expect(response.body).to match I18n.t("two_factor_auth.confirm.title")
+        expect(response.body).to include("svg")
+        expect(response.body).to match(/#{@user.otp_secret.scan(/.{4}/).join(" ")}/)
+        expect(response.body).to match I18n.t("two_factor_auth.input_token.label")
+      end
+    end
+    context "2fa is already activated" do
+
```
diaspora-code-review: Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body beginning.

spec/integration/two_factor_authentication_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    context "when logging in with TOTP token" do
+      let!(:user) { FactoryGirl.create(:user) }
+
+      it "redirects to two-factor page" do
+        post "/users/sign_in",
+             params: {user: {username: user.username, password: user.password}}
+        expect(response).to render_template(:two_factor)
+        post "/users/sign_in",
+             params: {user: {username: user.username,
```
diaspora-code-review: Layout/AlignHash: Align the elements of a hash literal if they span more than one line.

spec/integration/two_factor_authentication_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+                             password: user.password,
```
diaspora-code-review: Layout/AlignHash: Align the elements of a hash literal if they span more than one line.

features/step_definitions/two_factor_steps.rb
```diff
@@ -0,0 +1,67 @@
...
+
+When /^I fill in a recovery code to deactivate from "([^"]*)"$/ do |username|
+  @me = find_user username
+  @codes = @me.generate_otp_backup_codes!
+  @me.save!
+  fill_in "two_factor_authentication_code", with: @codes.first
+end
+
+When /^I click on OK$/ do
+  find(".btn-primary", match: :first).click
```
Flaburgan: Steps are available everywhere, I feel like this is too generic and can cause problems. We already have generic steps like "I validate the modal". Can they be used there? If no, can we make that step more specific?

app/views/two_factor_authentications/_confirm.haml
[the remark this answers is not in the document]
Flaburgan: I successfully activated OTP with the code but first did the mistake to enter the space displayed inside the secret here. Can we indicate somewhere that those are only for readability and not actually part of the secret?

app/views/two_factor_authentications/_deactivate.haml
```diff
@@ -0,0 +1,28 @@
...
+      %h4= t("two_factor_auth.activated.change_button")
+      %p= t("two_factor_auth.activated.change_label")
+
+      = form_for "two_factor_authentication", url: two_factor_authentication_path,
+        html: {method: :delete, class: "form-horizontal"} do |f|
+        .form-group
+          = f.label :code, t("two_factor_auth.input_token.label"), class: "control-label col-sm-6"
+          .col-sm-6
+            = f.text_field :code, placeholder: t("two_factor_auth.input_token.placeholder"), class: "form-control"
+            = t("two_factor_auth.recovery.reminder")
```
Flaburgan: I tried to disable 2FA with a recovery code and it didn't work (the page reloaded but nothing changed). I know the recovery code is fine as I just used it to log in. Does that work for you?

Flaburgan (changes requested):
Two minor remarks, and maybe a bug with the disabling and recovery code? I don't know what's going on there as this is covered by the tests... Otherwise it works very well! This should be integrated soon ;)

app/views/two_factor_authentications/_deactivate.haml
jhass: It did work for me, did you maybe use the same recovery code to login as to deactivate? It looks like they're one time use only: https://github.com/tinfoil/devise-two-factor/blob/master/lib/devise_two_factor/models/two_factor_backupable.rb#L40-L41

jhass: I clarified they're single use

Flaburgan (approved):
Ship it!

jhass:
🎉 🍪 🎊

SuperTux88:
Great that this is finally finished :tada: :cookie: 

@jhass you didn't add a changelog entry when you merged this, but I'm sure we all agree that this deserves one ;) But this leads me to the question: do we want to uplift that to the `next-minor` branch? @Flaburgan marked this with the 0.8.0.0 milestone [a while ago](https://github.com/diaspora/diaspora/pull/7751#event-1965968374), but I don't know why? I know this contains a migration to add the required columns to the user table, but I think that is small enough to be run in a minor upgrade? Are there other reasons to not add this to the `next-minor`? Do we want to test that on `develop` pods first?

jhass:
🤦‍♂ I knew using Github's squash merge was going to be too easy...

I'm fine with uplifting this to `next-minor`, in fact I'll do it right away.

lislis:
Thank you for finishing this!
:cookie: for everyone!

PR #7618, opened 2017-09-16
Update gems

SuperTux88:

PR #7772, opened 2018-04-08
Update gems

SuperTux88:


Gemfile
```diff
@@ -138,35 +139,35 @@ gem "rails-i18n",           "5.1.1"
...
 
 # Services
 
-gem "omniauth",           "1.6.1"
+gem "omniauth",           "1.8.1"
 gem "omniauth-facebook",  "4.0.0"
 gem "omniauth-tumblr",    "1.2"
 gem "omniauth-twitter",   "1.4.0"
-gem "twitter",            "6.1.0"
+gem "twitter",            "6.2.0"
```
diaspora-code-review: Gems should be sorted in an alphabetical order within their section of the Gemfile. Gem `omniauth-wordpress` should appear before `twitter`.

SuperTux88:
Hmpf ... travis fails again because of https://github.com/diaspora/diaspora/pull/7618#issuecomment-329989144 (and again only on travis, not locally). The bug is already fixed now, but I can't update to the fixed version because rails-assets doesn't have the newest versions ... :rage:

SuperTux88:
And travis is green :tada:

skyschub (approved):

SuperTux88:
Merged as ... many commits ... :tada: Everything was backported except `Bump rails-assets-cropperjs` which was only cherry picked to `develop`.

Thanks @denschub for the review :cookie: and thanks to @svbergerem for the help with the perfect-scrollbar refactoring :cookie:

SuperTux88:
I reverted `Bump twitter-text` on `next-minor` again, because it added a new dependency (`libidn11-dev` on debian), so it's only in `develop` now.

Flaburgan:
@SuperTux88 would it be possible to add a note to the `develop` changelog about this new dependency?

SuperTux88:
I already added it to the wiki for debian, and before the release we should find which packages are needed for other distros (for ubuntu it's probably the same package, but didn't test that yet), so we can add it to the upgrade guide.

Flaburgan:
I can confirm it is the same package for Ubuntu ;)

Issue #7770, opened 2018-04-07
Facebook cross-posting should contain link to Diaspora Post


HankG:
I'm experimenting with the Facebook cross-posting capability and it is working pretty well.  When cross-posting a picture it posts a link to the picture in Diaspora.  I think it would be good for it to post the link to the original Diaspora post in general, or at least make that something a sure can toggle.  For posts with links I was thinking it should embed it near the end in a way that does not limit Facebook's rendering system from rendering the requested end link.  So it would be something like this:

`<message> (<diaspora link>) <original link URL>`

I've experimented with this manually and it renders well in Facebook.  

I have looked at the code from the service and think I see how I'd make the change. I'd be willing to pick this up as a first change if this behavior is something that is agreed upon.

HankG:
I have an initial implementation.  The behavior I am seeking is the following: 

- If the post is a photo or just text, have the D* link posted at the end like
- If the post ends in a validly formed URL then we want the URL to stay at the end so it is rendered by Facebook correctly but we want to append a (via D* link) so that there is a trace back to the original D* post

So a post which reads "Some text." would post to Facebook as: 
`"Some text. http://<diasporaserver>/p/<postnumber>"`

Whereas a post which reads ""Some text and a link http://someserver1/testlink1 http://someserver2/testlink2"" would post to Facebook as:
`"Some text and a link http://someserver1/testlink1 (via http://<diasporaserver>/p/<postnumber> ) http://someserver2/testlink2"`

I have amended the spec to exercise this new functionality so there are appropriate tests for happy path and some errant condition testing.  I don't have a means for Facebook integration testing.  If this concept sounds good I'd be up for working with a seasoned developer on next steps towards the pull request.

SuperTux88:
> Whereas a post which reads ""Some text and a link http://someserver1/testlink1 http://someserver2/testlink2"" would post to Facebook as:
> `"Some text and a link http://someserver1/testlink1 (via http://<diasporaserver>/p/<postnumber> ) http://someserver2/testlink2"`

That looks very weird, why should the "via" be inserted in the middle of the post? Why not a `(via http://<diasporaserver>/p/<postnumber>)` always at the end of the post?

HankG:
I believe if you put it at the end then Facebook will not render any of the URLs but I can experiment with it manually. If it works then always putting it at the end makes sense to me.

HankG:
[...]

PR #7774, opened 2018-04-08
7770 facebook crosspost with links

HankG:
A review of submission against https://github.com/diaspora/diaspora/issues/7770

Spec has been updated for new cases, style check has been run locally, but I don't know what rules for integration testing are before integration into main line since I don't have ability to test Facebook integration directly. If this looks good and the spec updates alone are sufficient then this may be ready for merge rather than just review.  On the style check some code I didn't write wasn't passing and I couldn't figure out how to consistently get around the "Align the elements of a hash literal if they span more than one line" warning no matter what permutations I tried so I made everything that flagged it single line.  I'll look for guidance on that on Discourse channels...

### DIFF
--- a/app/models/services/facebook.rb
+++ b/app/models/services/facebook.rb
@@ -24,14 +24,18 @@ class Services::Facebook < Service
 
   def create_post_params(post)
     message = post.message.plain_text_without_markdown
-    if post.photos.any?
-      message += " " + short_post_url(post, protocol: AppConfig.pod_uri.scheme,
-                                            host: AppConfig.pod_uri.authority)
-    end
+    post_url = short_post_url(post, protocol: AppConfig.pod_uri.scheme, host: AppConfig.pod_uri.authority)
 
-    {message: message,
-     access_token: access_token,
-     link:  post.message.urls.first
+    message = if message.empty?
+                post_url
+              else
+                "#{message} (via #{post_url})"
+              end
+
+    {
+      message:      message,
+      access_token: access_token,
+      link:         post.message.urls.first
     }
   end
 
--- a/spec/models/services/facebook_spec.rb
+++ b/spec/models/services/facebook_spec.rb
@@ -22,39 +22,49 @@ describe Services::Facebook, :type => :model do
       post_params = @service.create_post_params(post)
     end
 
-    it 'does not add post link when no photos' do
+    it "adds (via <post URL>) when message has no URLs " do
       message = "Some text."
       post = double(message: double(plain_text_without_markdown: message, urls: []), photos: [])
       post_params = @service.create_post_params(post)
-      expect(post_params[:message]).not_to include "http"
+      expect(post_params[:message]).to include "(via http:"
     end
 
-    it 'sets facebook id on post' do
-      stub_request(:post, "https://graph.facebook.com/me/feed").
-	to_return(:status => 200, :body => '{"id": "12345"}', :headers => {})
+    it "sets facebook id on post" do
+      stub_request(:post, "https://graph.facebook.com/me/feed")
+        .to_return(status: 200, body: '{"id": "12345"}', headers: {})
       @service.post(@post)
       expect(@post.facebook_id).to match "12345"
     end
-
   end
 
   describe "with photo" do
     before do
-      @photos = [alice.build_post(:photo, :pending => true, :user_file=> File.open(photo_fixture_name)),
-                 alice.build_post(:photo, :pending => true, :user_file=> File.open(photo_fixture_name))]
+      @photos = [alice.build_post(:photo, pending: true, user_file: File.open(photo_fixture_name)),
+                 alice.build_post(:photo, pending: true, user_file: File.open(photo_fixture_name))]
 
       @photos.each(&:save!)
+    end
 
-      @status_message = alice.build_post(:status_message, :text => "the best pebble.")
-        @status_message.photos << @photos
+    it "should include post url in message with photos as (via... " do
+      @status_message = alice.build_post(:status_message, text: "the best pebble.")
+      @status_message.photos << @photos
 
       @status_message.save!
       alice.add_to_streams(@status_message, alice.aspects)
+
+      post_params = @service.create_post_params(@status_message)
+      expect(post_params[:message]).to include "(via http:"
     end
 
-    it "should include post url in message with photos" do
+    it "should include post url in message with photos when no text message" do
+      @status_message = alice.build_post(:status_message, text: "")
+      @status_message.photos << @photos
+
+      @status_message.save!
+      alice.add_to_streams(@status_message, alice.aspects)
+
       post_params = @service.create_post_params(@status_message)
-      expect(post_params[:message]).to include 'http'
+      expect(post_params[:message]).to include "http:"
     end
 
   end


app/models/services/facebook.rb
```diff
@@ -23,16 +23,16 @@ def post_to_facebook(url, body)
...
+    else
+      post_url = short_post_url(post, protocol: AppConfig.pod_uri.scheme, host: AppConfig.pod_uri.authority)
+      message = "#{post_message} (via #{post_url} )"
     end
 
-    {message: message,
-     access_token: access_token,
-     link:  post.message.urls.first
-    }
+    {message: message, access_token: access_token, link: post.message.urls.first}
```
SuperTux88: > Align the elements of a hash literal if they span more than one line

It means to write it like this:

```ruby
    {
      message:      message,
      access_token: access_token,
      link:         post.message.urls.first
    }
```

But I think it's OK to have it in one line here.

app/models/services/facebook.rb
```diff
@@ -23,16 +23,16 @@ def post_to_facebook(url, body)
   end
 
   def create_post_params(post)
-    message = post.message.plain_text_without_markdown
-    if post.photos.any?
-      message += " " + short_post_url(post, protocol: AppConfig.pod_uri.scheme,
-                                            host: AppConfig.pod_uri.authority)
+    post_message = post.message.plain_text_without_markdown
+
+    if post_message.empty? && post.photos.none?
```
SuperTux88: A post should never be empty and have no photos, there is a validation for that:

https://github.com/diaspora/diaspora/blob/beeb073dbe0d95b0dd38c91f8b129fc0039bd3a0/app/models/status_message.rb#L20-L21

app/models/services/facebook.rb
```diff
@@ -23,16 +23,16 @@ def post_to_facebook(url, body)
...
+      message = post_message
+    else
+      post_url = short_post_url(post, protocol: AppConfig.pod_uri.scheme, host: AppConfig.pod_uri.authority)
+      message = "#{post_message} (via #{post_url} )"
```
SuperTux88: When there is no text but only photos, I think `post_message` is empty. In this case I think it looks weird to have the post be just a "(via https://pod.example.org/p/1234)", so I think it's better to just post the link in that case, as it was before. As soon as the message isn't empty, add it with (via ...).

Also, does facebook really need the space before the closing brace to detect the link? As you can see above, GitHub and also diaspora does detect URLs perfectly without this space.

app/models/services/facebook.rb
```diff
@@ -23,16 +23,16 @@ def post_to_facebook(url, body)
   end
 
   def create_post_params(post)
-    message = post.message.plain_text_without_markdown
-    if post.photos.any?
-      message += " " + short_post_url(post, protocol: AppConfig.pod_uri.scheme,
-                                            host: AppConfig.pod_uri.authority)
+    post_message = post.message.plain_text_without_markdown
```
SuperTux88: Why did you rename this to `post_message` but then use again `message` below? I think one variable should be enough.

spec/models/services/facebook_spec.rb
```diff
@@ -45,7 +77,7 @@
 
       @photos.each(&:save!)
 
-      @status_message = alice.build_post(:status_message, :text => "the best pebble.")
+      @status_message = alice.build_post(:status_message, text: "the best pebble.")
```
SuperTux88: You can move everything from this line and below to the `it "should include post url in message with photos"`, because it is only used once there. That way this code doesn't run a second time for `it "should include post url in message with photos when no text message"` where it isn't needed (because you create a message without text there).

spec/models/services/facebook_spec.rb
```diff
@@ -22,14 +22,46 @@
...
+    it "adds URL of post to end if no link at end" do
       message = "Some text."
       post = double(message: double(plain_text_without_markdown: message, urls: []), photos: [])
       post_params = @service.create_post_params(post)
-      expect(post_params[:message]).not_to include "http"
+      expect(post_params[:message]).to include "(via http:"
     end
 
-    it 'sets facebook id on post' do
+    it "adds URL at the end with ending with valid HTTP url" do
```
SuperTux88: I don't think we need that many tests for different urls in different places in the message, because it just doesn't make any difference.

spec/models/services/facebook_spec.rb
```diff
@@ -22,14 +22,46 @@
       post_params = @service.create_post_params(post)
     end
 
-    it 'does not add post link when no photos' do
+    it "adds URL of post to end if no link at end" do
       message = "Some text."
       post = double(message: double(plain_text_without_markdown: message, urls: []), photos: [])
       post_params = @service.create_post_params(post)
-      expect(post_params[:message]).not_to include "http"
+      expect(post_params[:message]).to include "(via http:"
```
SuperTux88: You should check if the via link is acutally at the end if you write in the test-text "to the end".

SuperTux88 (dismissed):
Thanks for your pull-request, here is some feedback :)

Also, can you please squash the commits when you do style fixes?

app/models/services/facebook.rb
HankG: Oh ok, I didn't know that so wanted to cover that case.  I can adjust that along with the below suggestion.

app/models/services/facebook.rb
HankG: I tried that and when I ran style check it still complained.  I'll try it again, I much prefer it over the one line...

spec/models/services/facebook_spec.rb
HankG: for each test? ok.

spec/models/services/facebook_spec.rb
HankG: It made a difference when I was putting the "via" link before the last URL, but you are right they are superfluous now.

spec/models/services/facebook_spec.rb
HankG: ok

app/models/services/facebook.rb
HankG: Code readability since that is explicitly the post message versus the message that is being embedded into the return object.  I can use one variable throughout if that is preferred.

spec/models/services/facebook_spec.rb
SuperTux88: No, as mentioned in https://github.com/diaspora/diaspora/pull/7774#discussion_r180253983 you can remove most of the tests, so you con't need to check it in each test.

app/models/services/facebook.rb
SuperTux88: Yes, please. It's already the message that is going to be posted to facebook, you only add some more things to it before you post it.

app/models/services/facebook.rb
```diff
@@ -24,14 +24,18 @@ def post_to_facebook(url, body)
...
+    post_url = short_post_url(post, protocol: AppConfig.pod_uri.scheme, host: AppConfig.pod_uri.authority)
+
+    message = if message.empty? && post.photos.any?
+                post_url
+              else
+                "#{message} (via #{post_url})"
+              end
+
+    {
+      message: message,
```
diaspora-code-review: Layout/AlignHash: Align the elements of a hash literal if they span more than one line.

app/models/services/facebook.rb
```diff
@@ -24,14 +24,18 @@ def post_to_facebook(url, body)
...
+      access_token: access_token,
+      link: post.message.urls.first
```
diaspora-code-review: Layout/AlignHash: Align the elements of a hash literal if they span more than one line.

HankG: I switched this to this format as requested but I can't get rid of this style error.  I'm at a loss on how to make this pass the style check except making it one line.  I'm following the example in the documentation for this Rocket/Colon style (table).  Can someone point out what I'm doing wrong here?  http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Layout/AlignHash

SuperTux88:
Please don't create commit messages with `Squashed commit of the following: ...` because this information isn't important afterwards. Nobody needs to know how many commits these were before and the main commit message (the first line) should always describe what was done in this commit. Also style/review fixes should always been squashed to the commit where they are introduced (if possible, for bigger PRs this sometimes would end in complicated rebases with many conflicts, so in that case a separate fix-commit is OK). For this PR there should be only one commit with `Facebook plugin always appends (via D* post URL) to cross-posts`.

By they way: You can reorder and squash commits interactively with `git rebase -i upstream/develop` (the `fixup` keyword is useful here). You can also directly commit style-fixes with `git commit --fixup=<some commit ID to which the fixup belongs>` and then you can squash them with `git rebase -i --autosquash upstream/develop`.

HankG:
Ugh...through the wringer again...

HankG:
So per your request not only do you want me to squash the style changes you want me to squash **all** the changes into one enormous commit, including stuff through code reviews to the very end?

SuperTux88:
> I think this is what you were looking for now (latest commit) but I really am no longer sure anymore...

You now also included the last commit from the `develop` branch, because you somehow rebased it, that should be removed again (probably with rebasing on the `develop` branch).

> So per your request not only do you want me to squash the style changes you want me to squash all the changes into one enormous commit, including stuff through code reviews to the very end?

Not exactly, multiple commits are OK and even required in some cases (one very big commit is hard to review), but fixups for your changes (because of reviews and or style-guide) should be done in the same commit as they are introduced (so the change you did in the beginning is already clean in this commit). So you don't need the original commit and then 5 commits with "fix style", "fix style again", "maybe this time", ... ;)

Multiple commits are good when you do multiple changes in one PR that can be cleanly splitted in two commits. Maybe you do a refactoring as driveby, then the refactoring should be done in it's own commit (and style-fixes that are needed because of the refactoring should go to this commit, and other style-fixes to the other commit). Or you add a bigger feature that can be splitted in multiple parts. But this PR is a really small fix and is fine with a single commit.

Short: the commits should be done in a way they make sense in the future, when somebody reads the git log to find where some changes came from. And they don't want to find the last commit for a line to be with the message "style fixes" were `'` in this line were changed to `"`. They want to find a commit with a meaningful message where you added the line directly with `"`.

I hope it's clearer now, sorry for the confusion :)

HankG:
> You now also included the last commit from the develop branch, because you somehow rebased it, that should be removed again (probably with rebasing on the develop branch).

I am a bit lost on how to address this, or how I would have done this in the first place.  I did the last step of the rebase -i one more time but something tells me it didn't address your comment. 

While I've used git for many years rebasing is not something we've done any of the projects I've worked on, in fact it was often verboten to do it under normal circumstances.  I logged into the dev IRC for a more interactive experience for a little while.

spec/models/services/facebook_spec.rb
```diff
@@ -22,39 +22,57 @@
...
     end
 
-    it 'sets facebook id on post' do
-      stub_request(:post, "https://graph.facebook.com/me/feed").
-	to_return(:status => 200, :body => '{"id": "12345"}', :headers => {})
+    it "adds (via <post URL>) when message has URLs" do
```
SuperTux88: This test is only a duplicate of the test above, because the fact that there is an URL doesn't make a difference here. So you can just remove this and remove the "when message has no URLs " from the test above.

app/models/services/facebook.rb
```diff
@@ -24,14 +24,18 @@ def post_to_facebook(url, body)
...
-      message += " " + short_post_url(post, protocol: AppConfig.pod_uri.scheme,
-                                            host: AppConfig.pod_uri.authority)
-    end
-
-    {message: message,
-     access_token: access_token,
-     link:  post.message.urls.first
+    post_url = short_post_url(post, protocol: AppConfig.pod_uri.scheme, host: AppConfig.pod_uri.authority)
+
+    message = if message.empty? && post.photos.any?
```
SuperTux88: When the message is empty, there always must be photos, so you can just remove the `&& post.photos.any?` here, because it's always true if the `message.empty?` is true.

SuperTux88 (changes requested):
Thanks for your corrections :+1: I was on holiday last week and didn't have the time to do a review, sorry for the delay. Your PR looks almost OK, this is mostly nitpicking.

spec/models/services/facebook_spec.rb
HankG: OK. I like having this to make sure that if someone changes this later it doesn't get hosed up with embedded URLs but I'm fine making these changes.

app/models/services/facebook.rb
HankG: Isn't that implied behavior versus explicit behavior?  We only want the post format to be solely the URL if the message is empty and there is a photo.  The way the code reads it states that directly, even if the second check is not necessary.  I'll implement it either way but it seems having more explicit code on intent here is better.

spec/models/services/facebook_spec.rb
SuperTux88: If somebody changes it later, they should add a test for the changed behavior. But if somebody changes it later so it doesn't contain the "via <post URL>" when it contains a mention (just a random example which makes no sense), then you don't have a test for it. And you shouldn't add test for every random case somebody maybe could do (which doesn't make a difference now). If it makes sense to change something in the future, then we can add a test for that later (and then this test maybe should fail and will be removed).

app/models/services/facebook.rb
SuperTux88: > We only want the post format to be solely the URL if the message is empty and there is a photo. 

Well, in fact, I only want to the post format to be solely the URL if the message is empty. Because as soon as there is no message, the "via" format doesn't make sense anymore. The fact that there are photos in the original don't make any difference. At the moment it's only valid with photos when the message is empty, but maybe that changes in the future. What if it's valid to have an empty message when you have a poll for example. Then nobody would change this line here to `message.empty? && post.photos.any? || post.poll.any?` (first because everybody forgot about it, and second because it would make any sense, because the important part for the decision is only `message.empty?`) and it would look very strange if you then have an "empty" post on facebook only containing "(via <post URL>)".

HankG: Ah true, I had been too laser focused on thinking about it with respect to the photo thing.

HankG:
All of these elements were passing before, and pass on my machine and 7 of the 8 CI build types (in an area that has nothing to do with my code). These automated checks all worked before my last tweak, which was to half a line in the facebook.rb code.  I'm not sure what the path forward is from here...looking for guidance...

SuperTux88:
Thanks for updating the PR, but this comment is still open: https://github.com/diaspora/diaspora/pull/7774#discussion_r185962601

> 7 of the 8 CI build types (failing in an area that has nothing to do with my code)

Some tests fail randomly from time to time, but it's not a problem when all other tests are green and the failing test has nothing to do with your change.

HankG:
@SuperTux88 correct. I just don't know what the desired course of action is.  Is it preferred I just re-initiate the CI process and hope another random error doesn't creep in?  What is the preferred mechanism for doing that?  

I also need to figure out why editing one line made the code coverage drop since all the testing harness is unchanged.

HankG:
I think all comments are addressed now...

spec/models/services/facebook_spec.rb
```diff
@@ -22,39 +22,49 @@
       post_params = @service.create_post_params(post)
     end
 
-    it 'does not add post link when no photos' do
+    it "adds (via <post URL>) when message has no URLs " do
```
SuperTux88: > I think all comments are addressed now...

The `when message has no URLs ` can be removed here, but otherwise it's OK :+1:

HankG: Is there a way to do that through here?  This death by a thousand cuts thing is killing me.

SuperTux88:
> Is there a way to do that through here?

I did it myself while merging it now, so I could include it to the next release. So merged as 5a8bdc1e71628238753d73e654c4eb22d3a9852f

Thank you :cookie:

PR #7776, opened 2018-04-08
Show more explicit error messages on failed login

svbergerem:
Related to #3540. We already have different error messages for unknown usernames and invalid passwords in some languages. As far as I can tell there is also no way to login using an email address, so I don't see any privacy or security issues related to this change.

### DIFF
--- a/config/locales/devise/devise.en.yml
+++ b/config/locales/devise/devise.en.yml
@@ -12,10 +12,10 @@ en:
     failure:
       already_authenticated: You are already signed in.
       inactive: Your account is not activated yet.
-      invalid: Invalid %{authentication_keys} or password.
+      invalid: Invalid password.
       locked: Your account is locked.
       last_attempt: You have one more attempt before your account is locked.
-      not_found_in_database: Invalid %{authentication_keys} or password.
+      not_found_in_database: Invalid %{authentication_keys}.
       timeout: Your session expired. Please sign in again to continue.
       unauthenticated: You need to sign in or sign up before continuing.
       unconfirmed: You have to confirm your email address before continuing.


SuperTux88 (dismissed):
Looks good to me :cookie:

jhass:
I'm actually not sure about this, devise intentionally sets these strings up that way so one doesn't leak whether an account exists or not.

There's two vectors here: First you can find out whether a particular person has an account if you know the email address they used with a high likelihood.

Second if you have a couple of potential mail addresses you can validate which one they used before starting to brute force their password.

Admittedly this is less critical for us since we also allow the username for sign in, which is quite public. But I'd like to here some more opinions with those concerns in mind.

SuperTux88:
I wasn't able to login with the email address (I checked that before I approved), so as @svbergerem said, it's only possible to login with the username which is "public" (you can check if a username exists via https://<pod>/u/<username>). So there shouldn't be a problem with this change?

jhass:
Oh, did we change that at some point? I was quite sure we allowed that.

No objections then.

svbergerem:
Yesterday I wasn't able to login in on my test pod using the email address. Today it worked on my production pod. This needs some more investigation.

Flaburgan:
I know users who use the email address to log in. It definitely works in production currently.

svbergerem:
Ok, I made a mistake in my test yesterday. Users are able to login using their email address because of [some override in user.rb](https://github.com/diaspora/diaspora/blob/ae17248b2ecb462d7b7471d3173b27c0cb6c64a8/app/models/user.rb#L215). As far as I know we don't tell anyone in the front-end that this is possible. I see two possible solutions:

1. We keep the login via email. Then I think we should make sure that the error messages are the same in all languages, because otherwise you can find out whether a user registered on a pod with a given email address. Then we might also want to enable the [paranoid mode](https://github.com/diaspora/diaspora/blob/ae17248b2ecb462d7b7471d3173b27c0cb6c64a8/config/initializers/devise.rb#L95) for devise, because otherwise we leak some information on the password reset page.

2. We remove the login via email. Then we can tell users whether the username exists that they were trying to use for login. We could also change the password reset page to use a username instead of the email address. This might even allow us to drop the uniqueness of email addresses in the database.

SuperTux88:
I think we should keep the password-reset with email, because otherwise everybody can spam me with password-reset mails with only knowing my username (which is public). But I agree, we should activate paranoid mode then.

I'm not sure if we should remove email logins, because as Fla said, some people are using it. So we should only remove it with a major release. If we don't want to remove it, I agree that we should make sure that the translations are all the same, so we don't leak the information in some languages. But I don't know how to ensure this, so the easiest is probably to just remove it.

denschub:
+1 to solution 1 and activating paranoid mode.

denschub:
Enabling Paranoid mode is tracked in #8001, and this PR is changing language files. Let's close.

PR #7778, opened 2018-04-09
Remove fileuploader.css

svbergerem:
I wasn't able to find any of the described classes.

cc @Fensterbank


Fensterbank:
We replaced the legacy fileuploader by fineuploader in https://github.com/diaspora/diaspora/pull/7083.
But I don't think these classes were used at this time and I don't know, when these class have been used.

It seems the CSS describes the default look of the file uploader, but we are having our own style and markup when uploading files. We use the library for performing the upload, but we don't use its default user interface.

I assume this css can safeley be removed.

SuperTux88 (approved):
I also can't find any usage of this, so lets remove it :+1:

SuperTux88:
Merged as 53d58a8bcb6da5120e8b95764dc45182dbb34ac4

Thanks for the cleanup @svbergerem :cookie: and thanks for the review @Fensterbank :cookie:

PR #7781, opened 2018-04-10
Enable Content-Security-Policy header by default

SuperTux88:
After 0.7.4.1 I think it's a good time to enable this by default ;) It's working without any problems in productions for a long time now, so I think it's save to enable it everywhere. Since this is a config-change and some podmins may have customization which could be blocked by this, I think we should still only change it in the next major release.

### DIFF
--- a/config/diaspora.yml.example
+++ b/config/diaspora.yml.example
@@ -571,10 +571,10 @@ configuration: ## Section
     ## is blocked by CSP.
     csp:
 
-      ## Report-Only header (default=true)
-      ## By default diaspora* adds only a "Content-Security-Policy-Report-Only" header. If you set
-      ## this to false, the "Content-Security-Policy" header is added instead.
-      #report_only: false
+      ## Report-Only header (default=false)
+      ## By default diaspora* adds a "Content-Security-Policy" header. If you set
+      ## this to true, the "Content-Security-Policy-Report-Only" header is added instead.
+      #report_only: true
 
       ## CSP report URI (default=)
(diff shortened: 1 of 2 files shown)


skyschub (approved):
I have no objections here, so I'm approving this PR. However, I suggest not merging it now, but waiting for a few days, so others have a chance to jump in as well.

config/diaspora.yml.example
```diff
@@ -571,7 +571,7 @@ configuration: ## Section
     ## is blocked by CSP.
     csp:
 
-      ## Report-Only header (default=true)
+      ## Report-Only header (default=false)
       ## By default diaspora* adds only a "Content-Security-Policy-Report-Only" header. If you set
       ## this to false, the "Content-Security-Policy" header is added instead.
```
Flaburgan: This comment also needs to be updated.

Flaburgan (changes requested):

config/diaspora.yml.example
SuperTux88: Of course, fixed.

pacohope:
I've been running with the CSP enabled and I found a bug. Fonts were not covered by the CSP if they were distributed through an asset host. I created a PR for a one-line change that fixes it. #7796 You might want to incorporate it and/or merge it with this.

SuperTux88:
Since #7796 is merged now, I think we can merge this too? Can you please approve your review @Flaburgan since your requested change is already fixed.

Flaburgan (approved):

jhass (approved):

PR #7783, opened 2018-04-11
Fix link to "Make yourself an admin" and simplify /podmin redirect

SuperTux88:
The link to "Make yourself an admin" is broken, because the title changed. I created now a redirect in the wiki which redirects to this section. That way we can simply change the target of the redirect once something changes again and don't need to change the link here again (which needs a release).

Also I changed how the redirect to /podmin works because some podmins were confuse how they can disable this redirect and I think the rule with two users can actually be a little confusing. I think the
main goal of this page to give the podmin a little start and I think after they configured everything, the pod works and they found the link to the wiki to make themself an admin, it is OK to remove the redirect.

Also it's bad for single-user pods where this page always stays active, even if they are an admin, but have only one user. It's more useful for single-user pods to have the login on the home page.

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -21,7 +21,7 @@ class HomeController < ApplicationController
           partial_dir.join("_show.html.erb").exist? ||
           partial_dir.join("_show.haml").exist?
       render :show
-    elsif User.count > 1 && Role.where(name: "admin").any?
+    elsif Role.admins.any?
       render :default
     else
       redirect_to podmin_path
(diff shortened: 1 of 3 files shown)


skyschub (approved):
Nice test optimization.

SuperTux88:
Merged as e31ca1fd3b96e1a84927fc1f27b18ef3acf44cfc and 686310fb8e3514e5e827b272a4e44b14f5711ab7

Thanks @denschub for the review :cookie:

PR #7787, opened 2018-04-15
Randomize start times of daily Sidekiq-Cron jobs

skyschub:
Some recurring background jobs can take a lot of resources, and others even include pinging other pods, like recurring_pod_check. Having all jobs run at 0 UTC causes a high local load, as well as a little bit of DDoSing through the network, as pods try to ping each other.

As Sidekiq-Cron does not support random offsets, we have to take care of that ourselves, so let's add jobs with random work times.


config/initializers/sidekiq_scheduled.rb
```diff
@@ -0,0 +1,80 @@
...
+      "class": "Workers::RecurringPodCheck"
+    }
+  }
+end
+# rubocop:enable Metrics/MethodLength
+
+def valid_config?(path)
+  return false unless File.exist?(path)
+
+  current_config = YAML::load_file(path)
```
diaspora-code-review: Style/ColonMethodCall: Do not use `::` for method calls.

config/initializers/sidekiq_scheduled.rb
```diff
@@ -0,0 +1,80 @@
...
+
+  # If they key don't match the current default config keys, a new job has
+  # been added, so we nede to regenerate the config to have the new job
+  # running
+  return false unless current_config.keys == default_job_config.keys
+
+  # If recurring_pod_check is still running at midnight UTC, the config file
+  # is probably from a previous version, and that's bad, so we need to
+  # regenerate
+  return current_config[:recurring_pod_check][:cron] != "0 0 * * *"
```
diaspora-code-review: Style/RedundantReturn: Redundant `return` detected.

config/initializers/sidekiq_scheduled.rb
```diff
@@ -0,0 +1,80 @@
...
+
+def regenerate_config(path)
+  job_config = default_job_config
+  File.open(path, "w") do |schedule_file|
+    schedule_file.write(job_config.to_yaml)
+  end
+end
+
+if Sidekiq.server?
+  schedule_file_path = File.join(Rails.root, "config", "schedule.yml")
```
diaspora-code-review: Rails/FilePath: Please use `Rails.root.join('path', 'to')` instead.

config/initializers/sidekiq_scheduled.rb
```diff
@@ -0,0 +1,80 @@
...
+  unless valid_config?(schedule_file_path)
```
diaspora-code-review: Style/IfUnlessModifier: Favor modifier `unless` usage when having a single-line body. Another good alternative is the usage of control flow `&&`/`||`.

SuperTux88 (approved):
Looks good to me :cookie:

SuperTux88:
Merged as 6918dbc7610577080d092237df528a5837a63a63

Thank you :cookie:

Flaburgan:
Hi, sorry I checked my emails yesterday but didn't see that one. The current implementation is not good in my opinion. There are times where Cron should not be set, because they could run twice, or not run at all. For example, when switch to daylight saving time. I don't know about other countries but in France, twice per year, time is shift of one hour. So in the night of the 21th of October, all times between 2am and 3am exist twice. At the opposite, at the 21th of March, clock goes straight from 01:59:59 to 03:00:00 so if there is a Cron set between those time, it will not be run. I'm sure other countries do that differently (at least not at the same dates).
There also is the leap second case. 

So in my opinion, we should think more about it and this commit should not have been merged like that.

Flaburgan:
Hm, jobs running in UTC means that a server hosted in Asia for example will not send birthday notifications at midnight then?

SuperTux88:
Yes, that was always the case. Also for you in France notifications aren't sent at midnight. These birthday notifications are suboptimal anyway, because the best would be to send it depending on the timezone of the person who has birthday, but that's not possible (we don't have this data and we don't federate this data). Sending the notification at midnight for you in France wouldn't help when the person who has birthday lives in America and not has birthday yet.

denschub:
> At the opposite, at the 21th of March, clock goes straight from 01:59:59 to 03:00:00 so if there is a Cron set between those time, it will not be run

Just for the record, and as an addition to what has been outlined before: that is simply not true. If you'd schedule a cron job at 2:30, it would simply run at 3:00. Literally all sensible implementations have a minute-trigger and check deltas between the last cron run and the current time, so it literally does not matter if you travel through time.

Issue #7640, opened 2017-10-13
mention breaks if display name contains markdown


DeusFigendi:
(I think this is a good title to search for)

I recognized Juni Walet juni_walet@pod.geraspora.de could not be mentioned here are two examples:

https://pod.geraspora.de/posts/6994798
https://pod.geraspora.de/posts/9d5492a08f3601351c0f101b0e8ace24

![bildschirmfoto vom 2017-10-13 21-17-11](https://user-images.githubusercontent.com/1711504/31562513-033f7e94-b05c-11e7-8e1a-b8e91f6faca6.png)
![bildschirmfoto vom 2017-10-13 21-16-36](https://user-images.githubusercontent.com/1711504/31562514-035ba02e-b05c-11e7-8364-d5413bed81b0.png)

I did not test it detailed but I'd guess the `[    ]` causes this. I wrote `@{juni_walet@pod.geraspora.de}` so that was converted correctly I think. Probably some escaping is needed?

I'll ask if Juni got notified, but I think so.

svbergerem:
This is the plain text: `@{Juni Walet   [^,..,^]; juni_walet@pod.geraspora.de}`. https://github.com/markdown-it/markdown-it-sup (an inline rule) recognizes `^,..,^` before the mention code is able to see the mention syntax.

SuperTux88:
The `^` is the problem (because it is parsed as markdown), and it will be solved automatically when we remove the server-side backporting for older versions and send the plain new syntax. You can fix this for now when you add a name yourself (so it is already the old syntax and the server doesn't touch it), so for example

DeusFigendi:
> it will be solved automatically when we remove the server-side backporting

I see… but it could also be escaped… `@{Juni Walet [\^,..,\^]; juni_walet@pod.geraspora.de}`

> You can fix this for now when you add a name yourself 

sure, that is a good and simple workaround until it's obsolete.

btw. Juni Walet confirmed she got notified.

SuperTux88:
Fixed by #7788

DeusFigendi:
This bug is back (at least on Geraspora). I saw it a short while ago but I thought it maybe wasn't fixed but now Ravenbird documented it again:

[...]

PR #7788, opened 2018-04-15
Remove backporting of mention syntax

### LINKED ISSUE #7276: Simplify mention syntax

SuperTux88:
I think enough pods support the new syntax now, and for those which don't, they should immediately upgrade anyway because of security hotfixes.

fixes #7276, #7392 and #7640

### DIFF
--- a/lib/diaspora/mentionable.rb
+++ b/lib/diaspora/mentionable.rb
@@ -71,20 +71,6 @@ module Diaspora::Mentionable
     }
   end
 
-  # Regex to find mentions with new syntax, only used for backporting to old syntax
-  NEW_SYNTAX_REGEX = /@\{[^\} ]+\}/
-
-  # replaces new syntax with old syntax, to be compatible with old pods
-  # @deprecated remove when most of the posts can handle the new syntax
-  def self.backport_mention_syntax(text)
-    text.to_s.gsub(NEW_SYNTAX_REGEX) do |match_str|
-      _, diaspora_id = mention_attrs(match_str)
-      person = find_or_fetch_person_by_identifier(diaspora_id)
-      old_syntax = "@{#{person.name.delete('{}')}; #{diaspora_id}}" if person
-      old_syntax || match_str
-    end
-  end
-
   private_class_method def self.find_or_fetch_person_by_identifier(identifier)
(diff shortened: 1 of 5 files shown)


skyschub (approved):
Let's 💀 it with 🔥!

PR #7793, opened 2018-04-23
Try to improve Polish translation

mkljczk:


SuperTux88:
Thanks for your contribution, but we use [WebTranslateIt](https://webtranslateit.com/en/projects/3020-Diaspora) for the translations, see: https://wiki.diasporafoundation.org/Contribute_translations

Please add your changes there or it would be overwritten again with the translations from WebTranslateIt.

PR #7794, opened 2018-04-24
Updated ja-KS translation

ciubotaru:
Enabled Kansai dialect of the Japanese language (ja-KS) and updated locale files from WebTranslateIt.


SuperTux88:
All translation files were updated before the release so this has conflicts now. But I also enabled the language via 6a3d9ede5261c837e83f0a4f0d6f0b796bbb70c7, so I close this.

PR #7798, opened 2018-04-25
Add unique index to poll participations on author_id and poll_id

cmrd-senya:
Previously we had only a Rails validation which ensured poll participation
uniqueness but this adds uniqueness control to the database level, so that
uniqueness is guaranteed even when changing data with avoiding Rails
validations.


SuperTux88 (changes requested):
When there wasn't a unique index before, you should remove potential duplicates (although there are probably none) before adding the index to make sure the index can be created.

CSammy:
I suggest that we do the index the other way around - `%i[poll_id author_id]` - and get rid of the index on `poll_id` in order to keep index overhead down on both storage and performance.

cmrd-senya:
Updated

SuperTux88 (approved):

SuperTux88:
Merged as e9f6dbdffd8b9c74238e1183cff3918a487e4a89

Thank you :cookie:

PR #7803, opened 2018-04-29
Account merging support

cmrd-senya:
Adds support for merging accounts with account migration by eliminating duplicate data in favor of the target user's data.


SuperTux88 (approved):
Looks good to me :cookie:

PR #7805, opened 2018-05-03
Add completed_at to account_migrations

cmrd-senya:
Use completed_at datetime field as an indication of a performed migration

### DIFF
--- a/db/migrate/20180430134444_add_completed_at_to_account_migration.rb
+++ b/db/migrate/20180430134444_add_completed_at_to_account_migration.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+class AddCompletedAtToAccountMigration < ActiveRecord::Migration[5.1]
+  def change
+    add_column :account_migrations, :completed_at, :datetime, default: nil
+
+    reversible do |change|
+      change.up do
+        set_completed_at_for_closed_accounts
+      end
+    end
+  end
+
+  def set_completed_at_for_closed_accounts
+    # rubocop:disable Rails/SkipsModelValidations
+    AccountMigration.joins(:old_person).where(people: {closed_account: true}).update_all(completed_at: Time.zone.now)
+    # rubocop:enable Rails/SkipsModelValidations
+  end
+end
--- a/app/models/account_migration.rb
+++ b/app/models/account_migration.rb
@@ -40,10 +40,11 @@ class AccountMigration < ApplicationRecord
 
     dispatch if locally_initiated?
     dispatch_contacts if remotely_initiated?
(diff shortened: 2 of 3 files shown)


db/migrate/20180430134444_add_completed_at_to_account_migration.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    reversible do |change|
+      change.up do
+        set_completed_at_for_closed_accounts
+      end
+    end
+  end
+
+  def set_completed_at_for_closed_accounts
+    AccountMigration.all.each do |account_migration|
```
SuperTux88: I think it's better to directly join with `old_person` and only select the migrations where the account is closed instead of select all and then select each `old_person` separately. Or maybe even use a raw `UPDATE` query, but I don't know if this works the same for mysql and postgresql, if it needs a different query for both I probably prefer the `update(...)` via rails.

SuperTux88 (changes requested):
Probably only nitpicking, because there are no account migrations yet and therefore it doesn't make a big performance difference, but maybe some people update later ... ;)

db/migrate/20180430134444_add_completed_at_to_account_migration.rb
cmrd-senya: We can also use `update_all`:
```ruby
AccountMigration.joins(:old_person).where(people: {closed_account: true}).update_all(completed_at: Time.zone.now)
```

It is performed in a single `UPDATE` query:
```sql
UPDATE "account_migrations" SET "completed_at" = '2018-05-03 18:10:14.918872' WHERE "account_migrations"."id" IN (SELECT "account_migrations"."id" FROM "account_migrations" INNER JOIN "people" ON "people"."id" = "account_migrations"."old_person_id" WHERE "people"."closed_account" = $1)  [["closed_account", "t"]]
```
(this is the query with postgresql)

SuperTux88: Oh, sure, I forgot about that, but that's even better :+1:

cmrd-senya: Pronto complains about it:
```
Rails/SkipsModelValidations: Avoid using `update_all` because it skips validations.
```.

How do you think, can we just disable the pronto validation here?

cmrd-senya:
Updated

SuperTux88 (approved):

SuperTux88:
Merged as cb294fd3f4a856b7ef2f503268061fd340265ef2

Thanks :cookie:

PR #7814, opened 2018-05-16
Link to the pod in the emails footer

Flaburgan:
So users can come back to the pod easily. This is mainly for users who could have forget where they registered.

### DIFF
--- a/app/views/layouts/notifier.text.haml
+++ b/app/views/layouts/notifier.text.haml
@@ -1,4 +1,4 @@
 != yield
-!= t('notifier.email_sent_by_diaspora', :pod_name => pod_name) 
-!= t('notifier.to_change_your_notification_settings')
+!= t("notifier.email_sent_by_diaspora", pod_name: pod_name)
+!= t("notifier.to_change_your_notification_settings")
 != edit_user_url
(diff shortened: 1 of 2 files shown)


Flaburgan:
Tests are green, please tell me if you need anything from me?

app/views/layouts/notifier.text.haml
```diff
@@ -1,4 +1,4 @@
 != yield
-!= t('notifier.email_sent_by_diaspora', :pod_name => pod_name) 
-!= t('notifier.to_change_your_notification_settings')
+!= t("notifier.email_sent_by_diaspora", pod_name: link_to(pod_name, AppConfig.pod_uri.site))
```
SuperTux88: `link_to` doesn't work in text emails. But most notifications contain a link to the post/comment including the domain, and the `edit_user_url` also includes the domain. So I think for users with text-mails this should be enough to find their pod. I think the `pod_name` like before is enough for text mails.

SuperTux88 (changes requested):

app/views/layouts/notifier.text.haml
Flaburgan: Updated.

SuperTux88 (approved):

SuperTux88:
Merged as 3d83fbdbe57daaaed6f869743a0d951d436bc609

Thank you :cookie:

PR #7815, opened 2018-05-18
Support fetching StatusMessage by Poll guid

cmrd-senya:
When public fetch is requested with Poll guid, return parent StatusMessage for it, which includes the Poll in its turn.

ref #7660


SuperTux88 (approved):

SuperTux88:
Merged as 87968284ba7ff0e2478162a5906024cb8911617c

Thank you :cookie:

PR #7821, opened 2018-05-27
Escape asterisks in README

radomirbosak:
Unescaped asterisks can cause markdown text to be misinterpreted. Also
syntax highlighters (e.g. in sublime text) freak out.

References: [markdown specification](https://daringfireball.net/projects/markdown/syntax#em)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# diaspora*
+# diaspora\*
 ### A privacy-aware, distributed, open source social network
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-# diaspora*
+# diaspora\*
```
radomirbosak: I'm not sure if escaping this one is needed, but it's better be consistent, right?

SuperTux88 (approved):

SuperTux88:
Merged as cfa7724d73db97dfcbc45214ba4d0cdb40dba037

Thank you :cookie:

PR #7823, opened 2018-06-05
Add a welcome email with address verification

Flaburgan:
So as discussed [on discourse](https://discourse.diasporafoundation.org/t/from-hearing-about-diaspora-to-becoming-a-recurrent-user/1356/17?u=flaburgan), this pull request adds a welcome email sent to the users once they filled the first form (with email and username). I only put the very basics in it, here is how it looks:

Subject is "Welcome aboard!"
Body:
```
You now are a member of the diaspora* community!

https://diaspora-fr.org is the diaspora* server (called a pod) where you registered. Log in there using your username, testwelcomeemail14. Once logged, you will be able to reach the whole diaspora* network, even users registered on other pods.

Your diaspora* id is testwelcomeemail14@diaspora-fr.org, share it with everyone so they can find you in the network.

If you need help, have a look at the tutorials or the help section. If you don't find anything, then post a public message with the #question tag so people can help you.

Welcome aboard, enjoy diaspora*!
```

@goobertron I'd love some feedback on the content, and what we could write.

I also still need to add links to the tutorials and help section text.

I guess I also need to write some tests...

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -1114,7 +1114,16 @@ en:
       find_pods: "There’s a list of pods you can sign up to at %{fediverse_observer}."
       other_questions: "If you have any other questions regarding choosing a pod, check out our %{wiki}."
     invalid_invite: "The invite link you provided is no longer valid!"
-
+    welcome_email:
+      subject: "Welcome to the diaspora* community!"
+      member: "Congratulations! You now are a member of the diaspora* community!"
+      part_of_network: "You have registered with %{pod_url}, one of the many nodes (called a “pod”) in the diaspora* network. To access the diaspora* network, you will need to sign in to that same pod using your username, %{username}, and password. Once signed in, you will be able to reach the whole diaspora* network, even people registered with other pods."
+      your_id: "Your diaspora* id is %{id}. This includes your home pod name and helps diaspora*'s software to locate you in the network. Share this with everyone you want to connect with in diaspora*."
+      verify: "Please confirm that this is really your email address. Until you do, we won’t send you any notification emails."
+      verify_link: "Verify my email address"
+      help: "If you need any help, have a look at our %{tutorials_link} or the in-app help section, which can be found via the user menu in the header bar. If you don’t find the information you need there, post a public message with the #question tag so that members of our amazing community can help you."
+      tutorials_link_text: "tutorials for new community members"
+      welcome: "Welcome aboard, and enjoy diaspora*!"
   reshares:
     create:
       error: "Failed to reshare."
@@ -1248,7 +1257,8 @@ en:
       your_email: "Your email"
       your_email_private: "Your email will never be seen by other users"
       change_email: "Change email"
-      email_awaiting_confirmation: "We have sent you an activation link to %{unconfirmed_email}. Until you follow this link and activate the new address, we will continue to use your original address %{email}."
+      email_awaiting_confirmation: "We have sent you an activation link to %{unconfirmed_email}. You will not receive any notifications until you follow this link and activate the new address."
+      email_not_verified: "Your email address is not verified. You will not receive notification emails from diaspora* until you verify it."
       change_password: "Change password"
       new_password: "New password"
       current_password: "Current password"
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -16,7 +16,7 @@ class RegistrationsController < Devise::RegistrationsController
       flash[:notice] = t("registrations.create.success")
       @user.process_invite_acceptence(invite) if invite.present?
       @user.seed_aspects
-      @user.send_welcome_message
+      @user.send_welcome_email
       sign_in_and_redirect(:user, @user)
       logger.info "event=registration status=successful user=#{@user.diaspora_handle}"
     else
--- a/app/mailers/welcome_mailer.rb
+++ b/app/mailers/welcome_mailer.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class WelcomeMailer < ApplicationMailer
+  def send_welcome_email(user)
+    @user = user
+    @podmin_message = podmin_message
+    mail(to: @user.email, reply_to: AppConfig.admins.podmin_email.presence,
+         subject: I18n.t("registrations.welcome_email.subject")) do |format|
+      format.text { render "registrations/welcome_email" }
+      format.html { render "registrations/welcome_email" }
+    end
+  end
+
+  private
+
+  def podmin_message
+    return unless AppConfig.settings.welcome_message.enabled?
+    AppConfig.settings.welcome_message.text.get
+  end
+end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -228,8 +228,17 @@ class User < ApplicationRecord
 
   def confirm_email(token)
     return false if token.blank? || token != confirm_email_token
-    self.email = unconfirmed_email
+
+    self.email = unconfirmed_email if unconfirmed_email.present?
+    self.confirm_email_token = nil
     save
+  end
+
+  # The address is considered verified while no confirmation is pending: both
+  # the welcome email (registration) and the change-email flow set this token
+  # and clear it once the link is followed
+  def email_verified?
+    confirm_email_token.blank?
   end
 
   ######## Posting ########
@@ -351,6 +360,7 @@ class User < ApplicationRecord
   ######### Mailer #######################
   def mail(job, *args)
     return unless job.present?
+
     pref = job.to_s.gsub("Mail::", "").underscore.sub(/_worker\z/, "")
     if(self.disable_mail == false && !self.user_preferences.exists?(:email_type => pref))
       job.perform_async(*args)
@@ -453,18 +463,9 @@ class User < ApplicationRecord
     aq
   end
 
-  def send_welcome_message
-    return unless AppConfig.settings.welcome_message.enabled? && AppConfig.admins.account?
-    sender_username = AppConfig.admins.account.get
-    sender = User.find_by(username: sender_username)
-    return if sender.nil?
-    conversation = sender.build_conversation(
-      participant_ids: [sender.person.id, person.id],
-      subject:         AppConfig.settings.welcome_message.subject.get,
-      message:         {text: AppConfig.settings.welcome_message.text.get % {username: username}}
-    )
-
-    Diaspora::Federation::Dispatcher.build(sender, conversation).dispatch if conversation.save
+  def send_welcome_email
+    update!(confirm_email_token: SecureRandom.hex(15))
+    WelcomeMailer.send_welcome_email(self).deliver_now
   end
 
   def encryption_key
--- a/app/views/registrations/welcome_email.html.erb
+++ b/app/views/registrations/welcome_email.html.erb
@@ -1,0 +1,70 @@
+<style>
+body {
+  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+  background-color: #eee;
+  padding: 1rem;
+}
+main {
+  margin: auto;
+  padding: 2rem;
+  background-color: #fff;
+  max-width: 800px;
+  border-radius: 10px;
+}
+h1 {
+  font-size: 1.4rem;
+  text-align: center;
+  margin-top: 0;
+  margin-bottom: 2rem;
+}
+strong {
+  white-space: nowrap;
+}
+.button-container {
+  text-align: center;
+  margin: 1.5rem;
+}
+.button-container a {
+  border-radius: 5px;
+  color: #fff;
+  background-color: #0088e6;
+  padding: 8px 16px;
+}
+p:last-child {
+  margin-bottom: 0;
+}
+</style>
+
+<main>
+  <h1>
+    <%= t("registrations.welcome_email.member") %>
+  </h1>
+  <p>
+    <% pod_url = AppConfig.pod_uri.site %>
+    <%= t("registrations.welcome_email.part_of_network",
+      pod_url: link_to(pod_url, pod_url), username: content_tag(:strong, @user.username)).html_safe %>
+  </p>
+  <p>
+    <%= t("registrations.welcome_email.your_id", id: content_tag(:strong, @user.person.diaspora_handle)).html_safe %>
+  </p>
+  <p>
+    <%= t("registrations.welcome_email.verify") %>
+  </p>
+  <p class="button-container">
+    <%= link_to t("registrations.welcome_email.verify_link"), confirm_email_url(token: @user.confirm_email_token) %>
+  </p>
+  <% if @podmin_message.present? %>
+  <p>
+    <%= @podmin_message %>
+  </p>
+  <% end %>
+  <p>
+    <%= t("registrations.welcome_email.help",
+      tutorials_link: link_to(
+        t("registrations.welcome_email.tutorials_link_text"), "https://diasporafoundation.org/tutorials")
+    ).html_safe %>
+  </p>
+  <p>
+    <em><%= t("registrations.welcome_email.welcome") %></em>
+  </p>
+</main>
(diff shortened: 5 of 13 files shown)


config/locales/diaspora/en.yml
```diff
@@ -1032,7 +1032,13 @@ en:
       success: "You’ve joined diaspora*!"
     closed: "Signups are closed on this diaspora* pod."
     invalid_invite: "The invite link you provided is no longer valid!"
-
+    welcome_email:
+      subject: "Welcome aboard!"
+      member: "You now are a member of the diaspora* community!"
+      part_of_network: "%{pod_url} is the diaspora* server (called a pod) where you registered. Log in there using your username, %{username}. Once logged, you will be able to reach the whole diaspora* network, even users registered on other pods."
```
jhass: Nitpick: We tend to prefer "sign in" for the terminology so far.

jhass:
I guess in light of the recent PGP bugs, more of our target audience switched to plain text only mails, so I think we should take care to provide a variant there :)

Flaburgan:
I thought about that, but this email is sent right when the user register so he didn't set any email preferences. How one does set that btw?

jhass:
The preference is within the mail client. We send a MIME multipart message with two parts, one being HTML one being plain text. As for the Rails side it should be just adding the plaintext template and adding it to the render call iirc.

Flaburgan:
Oh okay, I'll do it then. Should be done this afternoon.

goobertron:
Thanks for the ping. I'll look this over in the next day or two.

goobertron:
OK, I've had a look at the text. I think we should mention diaspora* in the email subject. I've also worked on the grammar and tried to make the message as clear as possible while remaining friendly and welcoming. I've also added a link to the tutorials and explained how to find the in-app help.

          subject: "Welcome to the diaspora* community!”
          member: “Congratulations! You now are a member of the diaspora* community!"
          part_of_network: “You have registered with %{pod_url}, one of the many nodes (called a “pod”) in the diaspora* network. To access the diaspora* network, you will need to sign in to that same pod using your username, %{username}, and password. Once signed in, you will be able to reach the whole diaspora* network, even people registered with other pods."
          your_id: "Your diaspora* id is %{id}. This includes your home pod name and helps diaspora*'s software to locate you in the network. Share this with everyone you want to connect with in diaspora*."
          help: "If you need any help, have a look at our %{tutorials_link} or the in-app help section, which can be found by via the user menu in the header bar. If you don’t find the information you need there, post a public message with the #question tag so that members of our amazing community can help you."
          tutorials_link_text: "tutorials for new community members"
          welcome: "Welcome aboard, and enjoy diaspora*!”

You'd then need to line 9 of app/views/registrations/welcome_email.haml to:

    != t("registrations.welcome_email.help", tutorials_link: link_to(t("registrations.welcome_email.tutorials_link_text"), "https://diasporafoundation.org/tutorials")).html_safe

(I think that's correct syntax, but do check it.)

I welcome comments on what I've done.

SuperTux88:
I set the milestone to the next major for this PR, because I think this new welcome mail isn't really compatible with the [already existing optional welcome message](https://github.com/diaspora/diaspora/blob/c0a4895854c9435ec17f88d06ca7124b7e5fecb6/config/diaspora.yml.example#L378-L396). When we now add a non-optional additional welcome mail, users on pods where the optional message is enabled would receive two mails (one notification mail for the PM, and the new welcome mail), and I think that's too much (at least I as a user would be already annoyed  when I would receive two different welcome messages).

There are multiple possibilities to solve this:
1. Only send the new welcome mail when the optional message is disabled (that would be compatible with a minor release, but wouldn't solve the problem of users forgetting their username on pods where the optional message is enabled).
2. Add a note to the changelog for next major, so podmins can disable or change their own welcome message.
3. Remove the optional welcome PM with the next major release.

Point 3 would allow some more options to replace the old welcome message better:
1. Make the new welcome mail customization (Include an optional "some words from your podmin" section?).
2. Send the new welcome mail as PM if the pod doesn't have mails enabled (that wouldn't solve the problem of users forgetting their usernames on pods without mails, but I think that can't be solved on pods without mails anyway ... but it would replace the old welcome-message on pods without mails ... maybe only do this if the welcome message contains "some words from your podmin" from the last point?)

What do others think?

goobertron:
Hmm, that's a good point.

Given that the current default of the optional welcome message is so bare, could we replace it with this new one? That would mean that new users would get the new, better welcome message as a PM, and get just one email notification if email has been enabled (and if they have registered a valid email address).

I don't know if a more complex message with links in it as I've suggested is compatible with the yml file.

My suggestions would be:

- Make this either non-optional or optional but with the default set to enabled.
- I'd like it to be configurable/editable by the podmin, but this means it wouldn't be localised.
- I like your idea to add an optional 'message from the podmin' at the end of the text. If the podmin adds a message to their `diaspora.yml` file, this would be added to the welcome message in the language they choose. That way, at least most of the message would be sent in the user's chosen language.
- The 'optional' setting in `diaspora.yml` would be just for the podmin's extra message, not for the new welcome message. (In which case, I would remove the default message text from `diaspora.yml.example`, as it is not needed.)

This would combine your options 2 and 3: we add a note to the changelog so that any podmins who have set an optional message have a chance to alter it to fit better with the new welcome message in this PR. However, no podmin would need to alter the `enabled` setting in yml, as that would be only for their own message. That should hopefully minimise the possibility of things going wrong when this is introduced.

SuperTux88:
> That would mean that new users would get the new, better welcome message as a PM, and get just one email notification if email has been enabled (and if they have registered a valid email address).

Fla wants to send this as Mail and not as PM, so users have a mail containing their username and pod URL for the case they forget this. A PM wouldn't work then, so I would only send the PM if email is disable on the pod.

> This would combine your options 2 and 3: we add a note to the changelog so that any podmins who have set an optional message have a chance to alter it to fit better with the new welcome message in this PR. 

Yes, that's more or less what I was thinking about. That way we can include the advantage of the old message (customizable by the podmin) with the new, translatable and better mail. And we can safely remove the old PM, so people don't receive two messages.

SuperTux88:
> we should not fill the user object's "email" property but the "unconfirmed_email" property when someone is registering.

No, you need to set an email, because it is `NOT NULL`. But you should set `disable_mail` to `true` until it is confirmed.

Flaburgan:
Oh, we have a `disable_mail` property? Didn't know that. Isn't that a bit overcomplicated?

I know users who put a wrong email on purpose when signing in because they want to remain anonymous, the email field could become optional imo. And that means `email` should not be `NOT NULL` any more. But we're getting out of scope for this PR.

goobertron:
Perhaps (getting very out of scope for this PR) if a user enters a non-existent email address, we could do the following:

The first time a mail notification is sent, and bounces back with an 'address not known' error, send the user a PM (but not an email confirmation, of course) to inform them that the email address was not recognised, and give them three options:

1. Correct the email address stored in the account (in case they had entered it incorrectly);
2. Retry an email to that address (possibly the receiving mail server was down when the notification was sent);
3. Continue to store this email address in the account but not send email notifications.

As I said, not for this PR but a possible idea for the future.

skyschub (changes requested):
* Let's make eMail address validation part of this, so that the welcome mail also includes the validation link (Devise supports that!)
* Needs a plain text version
* We should also remove the welcome private message feature then, there is no need to send two mails.
* Some language feedback was provided.

MasterOfTheTiger:
@denschub
> We should also remove the welcome private message feature then, there is no need to send two mails.

This should still be an option for those who want to use that.

SuperTux88:
No, as @denschub already said, there shouldn't be sent two mails after signup, and there is also no need to do so, the one should be good enough to cover everything.

MasterOfTheTiger:
> No, as @denschub already said, there shouldn't be sent two mails after signup, and there is also no need to do so, the one should be good enough to cover everything.

What about choosing one over another?

goobertron:
The welcome mail will be sent when someone signs up, before they've had chance to set preferences. The decision to send one email automatically is a good one, in my opinion.

I notice that I previously added mentions of the tutorials and in-app help to Fla's email text; as these features will be merged into the new guides, that text (and the link) will need to be updated.

MasterOfTheTiger:
@goobertron 
> The welcome mail will be sent when someone signs up, before they've had chance to set preferences. The decision to send one email automatically is a good one, in my opinion.

I know that. I misstated. What I was saying is that the podmin should be able to decide (in diaspora.yml) which version he or she wants to use. I personally like the direct message, that way they can easily respond to me with questions (which has happened on a few occasions).

SuperTux88:
> What I was saying is that the podmin should be able to decide (in diaspora.yml) which version he or she wants to use.

Since it's going to be used as email validation, private message won't work for that, so it needs to be email. And we still shouldn't spam the users with two mails.

> I personally like the direct message, that way they can easily respond to me with questions (which has happened on a few occasions).

As said, the new welcome email should cover all things needed to replace the old message, and for example to be able to answer we can just set the [`podmin_email`](https://github.com/diaspora/diaspora/blob/9bab794ea464ee602b2e315509adf17f0815b4f4/config/diaspora.yml.example#L670-L671) as `Reply-To`, so people can still reply to the podmin, which would even become handier later if there is a problem with the pod (down, expired SSL, or something else), so people can contact the podmin, even if they can't access their private messages anymore (and also can't get the podmin email from the sidebar, if the pod is down, and they didn't save it somewhere).

Flaburgan:
To simplify the changes on the codebase and for the podmin, I was wondering about another option, could we keep the welcome PM by the podmin, but simply disable the e-mail notification for that specific PM? That way, we don't have to change anything for the podmin, and the users don't receive two e-mails.

Flaburgan:
> Typo in the current text: "Once logged" should be "Once logged in"

Fixed, thanks. I also fixed @jhass remarks about Log in -> Sign in.

denschub:
> but simply disable the e-mail notification for that specific PM?

How would you implement that?

SuperTux88:
I don't like that there are two different messages, even if the podmin PM wouldn't also send an email notification. Can't we just combine both of them to just one single email/message, so the podmin message is also included there and we could set the podmin contact mail-address as `Reply-To` in the mail so people could still answer to this message to get in touch with the podmin?

Flaburgan:
So I would like to re-open registration on diaspora-fr, and to fight spam I want to have e-mail confirmation, so I need to finish my work on this PR. @skyschub you said above that Devise could be used for that, however reading the codebase I found that **we already have a verification feature** when the users are changing their e-mail in the settings. It is tested here: https://github.com/diaspora/diaspora/blob/develop/features/desktop/change_settings.feature#L11-L18
I just tried it in diaspora-fr and I never received any e-mail. I'm trying to debug it now. But anyway, what do we want? Re-use our custom validation code, or drop it and use Devise?

Flaburgan:
As I don't know Devise I went the easiest path for me, I reused the existing logic.

Regarding the "other notifications should not send e-mails if the address is not verified", I originally tried with @SuperTux88 suggestion of `disable_mail=true` but that does not really work: if a user is making a typo in his address and then fix it in the settings, the email to verify that second address would never be sent. ~~So I used the fact that for registration we directly fill the email in the db, instead of the unconfirmed email. That means if no unconfirmed email but a token, then we are at the registration step.~~ That did not work, someone registering with another person email address, and then changing to another unverified email in the settings, would be able to spam the first address. So I decided to change the logic: a user with a token is considered unverified. That means that users will stop receiving notifications on the old email while they are changing to a new one they did not verified yet, but that should last only a few minutes so it is an acceptable tradeoff imo.

So yeah hopefully that covers it. We now have verification, welcome message by podmin is incorporated in the e-mail instead of sending a PM, and an HTML and plain text version.

Flaburgan:
I just added a bit of design to the email. Here is how it looks now:

Screenshot_2026-07-27_12-34-00

Flaburgan:
While working on this I noticed that diaspora* is sending e-mails to unverified addresses still. The notifications are properly gated, but CSRF, account-removal, export and password reset are not. I just added the check for the CSRF one in a new commit as I think we want this, but I am not sure for the others, and not sure how to handle existing users in prod. Opinions?

skyschub:
> So I would like to re-open registration on diaspora-fr, and to fight spam I want to have e-mail confirmation, so I need to finish my work on this PR.

To combat spam signups (and make spam accounts harder in general by a bit), you'd need to gate the signups themselves behind a valid email, or at least signing in. Two common flows I see are:

- gating the registration entirely:
  1. You click a "sign up" link and see a form that only has a captcha and a field for an email address.
  2. Upon submitting it, you receive a link with a registration token.
  3. Clicking that link allows you to complete the registration flow.
- gating the first signin:
  1. You click a "sign up" link and see an entire "create account" flow like we have now.
  2. Upon submitting it, you cannot actually log in. Your account stays in an unusable locked state, but you receive an email with an "activate account" token link.
  3. Clicking that link automatically signs you in and allows future signins.

I don't think the former is possible with Devise's own methods, but in theory it's doable by hijacking the invite token mechanic. The latter, however, is what `Devise::Models::Confirmable` supports. As a bonus, its `reconfirmable` also handles email validations upon change, so it'd be a replacement for `validatable`.

What your PR is doing is only gating actually sending emails behind a validation. Spammers can still sign up for an account and use that account without having a valid email. Because of that, I'm honestly not sure how your PR helps with your stated goal of spam protection. What am I missing?

Flaburgan:
You are missing nothing. This PR is just the first step about ensuring that we verify the email when someone register. I have another branch based on it where I implement your option 2 above behind a setting for the podmin, but it is not ready yet so I did not push it.

skyschub:
Got it. Saw the other PR, thanks. In the interest of not dragging this PR along for another 8 years, let's try to get this in a mergable state asap. Merging this without the other PR done is risky, since it doesn't close an abuse vector but opens *another one* by immediately sending emails to signups which can be used to spam people. I do, however, trust that you're motivated enough to finish the other PR soon, too. 

Please remove all "has the user verified themselves" email gates from this PR. When the other PR is done, there will never be a state where a user could have an active account without a validated email address, so the entire gate is not needed. So I don't see a reason to introduce a few (half-deployed) checks, we can just ignore this. I'll give this a proper review ASAP, but from a quick glance, I don't expect big architectural feedback here and just a few nits about style.

Flaburgan:
>When the other PR is done, there will never be a state where a user could have an active account without a validated email address

Well, the gate in the other PR is behind an admin setting, false by default. So the default behaviour stays that people can register and use diaspora\* with an unverified email address. Do we want to change that?

SuperTux88:
> Well, the gate in the other PR is behind an admin setting, false by default. So the default behaviour stays that people can register and use diaspora* with an unverified email address. Do we want to change that?

I would vote to make that default true (at least as long as a mail-server is configured), because if it makes it harder for spammers to signup it creates a better experience on all pods. I'm not even sure if I would make it an option, but there might be people who think it's important to sign up without giving your email-address, but then we should maybe just make the email-field optional too (because why would we require to give an email when we don't care if it's valid?).

skyschub:
I wouldn't even vote tbh. "Only allow users with real email addresses to sign up" is so standard, we should have done that 10 years ago 🙃 but yes, turn it on by default everywhere (except in development, where it would be _really_ annoying).

Flaburgan:
OK so it means we keep the unverified checks here then? So this PR should be fine? Or you want the DB migration in it?

skyschub:
> OK so it means we keep the unverified checks here then?

No. There's no need to gate sending emails behind a "verified email check" if the entire account process is gated behind verified emails. If a podmin decides to turn off that check, that's on them, not on us. You already said that your gate only affects some emails and not all - and instead of adding checks to every email, we either could generalize that whole check away, or: just not check. There's no real need to.

Flaburgan:
Well a podmin may be fine with people not using a real email address but not fine with hardbounce. As you pointed above this PR originally deals with a different problem than spam. But you decide, on diaspora-fr email addresses will be verified at registration so I don't mind.

Flaburgan:
Updated. You did not answer me about [the db migration](https://github.com/diaspora/diaspora/pull/8487/changes#diff-c6642e3bf72caac2a4b033eb00ddd209abf3f6bb0de537d12ea3b2fd2c10fbc8), do you want it here or it's fine in the other PR?

skyschub:
I'm fine considering all existing users to be validated. It's not great, but it's a lot easier than having to deal with... nobody being able to sign in all of the sudden.

Flaburgan:
Sorry I misinterpreted your last message. I have now included the migration file in this PR. Please tell me if there is anything else I should do here before merging.

PR #7825, opened 2018-06-10
Allow fonts to be served from asset host in CSP.

### LINKED ISSUE #7796: font_src in Content Security Policy needs to include asset host
I have enabled CSP on my instance. Safari 11.1 reports the following in the develop tools:
```
Refused to load https://podname/assets/bootstrap/glyphicons-halflings-regular-...svg because it does not appear in the font-src directive of the Content Security Policy.
Refused to load https://podname/assets/entypo-...woff because it does not appear in the font-src directive of the Content Security Policy.
Refused to load https://podname/assets/entypo-...ttf because it does not appear in the font-src directive of the Content Security Policy.
Refused to load https://podname/assets/entypo-...svg because it does not appear in the font-src directive of the Content Security Policy.
Refused to load https://podname/assets/fonts/diaspora-custom-...ttf because it does not appear in the font-src directive of the Content Security Policy.
```

The fix was adding a single line to `secure_headers.rb`. I'm gonna create a PR and attach it to this issue. Might want to commit this at the same time if/when you accept #7781.

omnigrok:
Closes #7796.

I bumped into this setting up my pod (where I was brave / foolish and turned on CSP enforcement) and noticed that web fonts were being blocked by CSP. Seems like a simple fix.

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -23,6 +23,7 @@ SecureHeaders::Configuration.default do |config|
 
   if AppConfig.environment.assets.host.present?
     asset_host = Addressable::URI.parse(AppConfig.environment.assets.host.get).host
+    csp[:font_src] << asset_host
     csp[:script_src] << asset_host
     csp[:style_src] << asset_host
   end


omnigrok:
I'm new here, but I suspect that the CI build failure is unrelated to this change (it's related to the changing post defaults to public and appears to only happen on one of the build variants), and the test coverage decrease looks staggeringly large across a large number of files to have been caused by adding a single line. Let me know if I'm wrong though and I'll dig into it.

jhass (approved):

PR #7826, opened 2018-06-11
web+diaspora:// links handler

cmrd-senya:
Follow up for https://discourse.diasporafoundation.org/t/diaspora-links-and-webextensions/1026

In this PR I add a controller and a route `/links?q=%s` which handle diaspora links. Passing a `diaspora://` link in place of `%s` will redirect to an existing post which is referenced by the link. If the post is unknown it will attempt to fetch the post first. If it is not possible to fetch the post, `404` will be responded.

Links are accepted in 4 different forms:
`diaspora://:author/:type/:guid`
`web+diaspora://:author/:type/:guid`
`//:author/:type/:guid`
`:author/:type/:guid`

With this controller we can use `registerProtocolHandler` from HTML5 to register a diaspora pod as a handler for diaspora-specific links on external web sites. Due to limitations of `registerProtocolHandler` we can't use `diaspora://` protocol for links. It is only allowed to use protocol string prefixed with `web`, so it is `web+diaspora://`. In future when the limitation is resolved we can use `diaspora://` instead.

So as we run `registerProtocolHandler` at the pod load, your web browser will propose to set the pod (the new links controller) as a handler for `web+diaspora://` links. This is an example of a link which will be opened at your pod if you run this PR at your pod and agree to set the handler for the `web+diaspora://` links:
`web+diaspora://hq@pod.diaspora.software/post/77be9030367b0136f13b5254001bd39e`

GitHub doesn't want to show this link as clickable, so just copy-paste it to your address bar. When you put this into a `<a>` tag on an arbitrary website, it works totally fine.

TODO:
- [x] Cucumber test

### DIFF
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -48,3 +48,4 @@
 //= require jquery.are-you-sure
 //= require cropperjs/dist/cropper.js
 //= require pica
+//= require protocol-handler
--- a/app/assets/javascripts/protocol-handler.js
+++ b/app/assets/javascripts/protocol-handler.js
@@ -1,0 +1,17 @@
+// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-v3-or-Later
+
+function registerDiasporaLinksProtocol() {
+  var protocol = location.protocol;
+  var slashes = protocol.concat("//");
+  var host = slashes.concat(window.location.hostname);
+
+  if (location.port) {
+    host = host.concat(":" + location.port);
+  }
+
+  window.navigator.registerProtocolHandler("web+diaspora", host.concat("/link?q=%s"), document.title);
+}
+
+if (typeof (window.navigator.registerProtocolHandler) === "function") {
+  registerDiasporaLinksProtocol();
+}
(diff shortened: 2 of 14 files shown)


Flaburgan:
That's very nice, thank you! First thing that comes to my mind though is, will we be able to use it at some time? Because if the users don't set their browser as the application, then they will have an error returned by the browser saying it doesn't know how to handle those links. Is there a technical way to have a fallback of any kind to not display an error page to the user?

cmrd-senya:
TODO: a cucumber test

SuperTux88 (changes requested):
Great PR :cookie: :heart: 

(I would like to see also a special handling for just `web+diaspora://hq@pod.diaspora.software` where the user is fetched and then redirected to the profile. So you can link to your profile from your website/blog where everybody can open your profile on their pod :) But that can be done in a follow up :) )

SuperTux88 (approved):
LGTM :cookie:

SuperTux88:
Merged as a7cc0645e08b18aed2af9b091387794637046613 80dfdcd6afb600b400c7bccd427918658feebdf3 6f812a5b8f20c04ba1a7aa01dcec13d2a07ac1af and 2e0b3826997221289ce64d826c7a921541d7732d

Thank you :cookie:

PR #7830, opened 2018-06-28
Get processor count with ruby (on linux and mac)

SuperTux88:
`nproc` doesn't work on mac, but this ruby solution works since ruby 2.2 ([see documentation](http://ruby-doc.org/stdlib-2.2.0/libdoc/etc/rdoc/Etc.html#method-c-nprocessors)) and should work on linux and mac. Obviously I can only test this on linux, so maybe somebody with a mac can test if that works there too :)

### DIFF
--- a/script/configure_bundler
+++ b/script/configure_bundler
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require_relative "../config/bundler_helper"
+require "etc"
 
 rails_env = BundlerHelper.rails_env
 database = BundlerHelper.database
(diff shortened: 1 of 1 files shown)


SWW13:
```
$ ruby
require "etc"
print Etc.nprocessors
^D
8
```

skyschub (approved):
Yeah, this works fine.

SuperTux88:
Merged as ac868c9c0621296da046f5aca2cd8ce29ac2d985

Thanks for the test @SWW13 :cookie: and @denschub :cookie:

PR #7831, opened 2018-07-11
Install web+diaspora:// handler only when the user is logged in

SuperTux88:
I just realized that it even asks to install the handler when the user is not logged in, which I think is bad, because people should install the handler for their pod, not for every pod (when people view posts on other pods).


jhass (approved):

cmrd-senya (approved):

SuperTux88:
Superseded by #7833

Issue #7832, opened 2018-07-12
Improve the UX of adding the web-diaspora:// protocol handler


denschub:
PR #7826 implemented a call to `navigator.registerProtocolHandler` to register a handler for `web+diaspora://` URIs. Unfortunately, the implementation is not quite there yet, and probably needs some improvement:

* While we do not encourage it, we do allow non-https installations. In this case, the call to `registerProtocolHandler` would throw a `SecurityError`, possibly breaking the rest of the JS. Even if we ignore non-http pods, there are other cases where the call can throw - for example if the UA blocks the registration of any protocol handlers. TorBrowser, for example, is an example of a UA blocking all protocol handlers. We should at least catch the exception and throw it away.
* Unfortunately, browsers are weird and specs are even weirder. There are no UI implementation rules for custom protocol handlers, so stuff differs. In Chrome, we only get a rather unobtrusive icon in the address bar, while on Firefox, we receive a full-width bar below the address bar, asking the user to enable the handler.
  * In Chrome, I am concerned that people will never see the icon, and thus never enable the handler, which would limit our ability to use it, and that sounds sad.
  * In Firefox, everyone will notice it, for sure. Unfortunately, if the user decides not to install the handler and click the `x` button... They will get asked again on the very next request, which makes this really annoying.
* Although the representation differs across browsers, they have one thing in common: they are very indescriptive, and users have no idea what it means. In fact, some are confused by it and I already got several questions on both public and private channels.

As these issues are kinda large, and the PR got merged into next-minor, I'm marking this as blocking. Sorry for not looking at the PR at an earlier stage.

Based on the current situation, I have two possible solutions in my mind:

1. Do not attempt to register the handler on every page load. Instead, add a button to the user settings page calling the register function, and put some explanation next to it.
2. Do not directly try to register, but show a modal dialog explaining the protocol handler and asking the user if they want it. If they say yes, we can call out to `registerProtocolHandler` and be happy. Otherwise, we should store that decision somewhere and never ask the user again.

Given that idea 2 is a huge effort, I'd not suggest that as a solution for now. As we have no way of generating the links inside the UI anyway, and are not using them in the UI, it feels more like an experimental thing, so having it behind a button in the settings page may be fine for now. That would allow us to ship the handler, but without annoying our users by accident and without having to spend a lot of effort in a more elaborate UI solution.

/cc @SuperTux88 and @cmrd-senya

SuperTux88:
[...]

PR #7833, opened 2018-07-12
Make registering the Protocol Handler optional with an action in the user settings

skyschub:
This implements idea 1 as outlined in #7832 as a temporary solution until we have something better. It's mainly meant to stop annoying and confusing users whilst being able to keep the handler in place so we can do more with it.

I removed the global call to the register function and instead made it only pop up when the user clicked a button in the settings:

screen shot 2018-07-13 at 00 29 25

Not the prettiest thing in the world, but it get the job done.

To be quite honest, I am also somewhat unhappy with the controller name chosen, and super unhappy with the callback path. While we can happily rename `LinkController` to something more meaningful like `DiasporaProtocolLinkController`, the story is not so easy with the link segment. If a large number of users would click this button, we'd be commited to `/link`, which is way too generic. What does it do? Is it linking applications? Is it a special view I can use to share links with? Who knows. Something like `/diaspora_protocol_uris/inbound` would probably be a much, much better choice. But I'm leaving this one to @SuperTux88 and @cmrd-senya to decide.

---

Also, you have to excuse my first commit there, as it is kinda off-topic. But I had to hide the fact that I am a hipster now, and as we have excluded folders for RubyMine, it just seems fair...!


denschub:
This also completely obsoletes #7831. Sorry.

app/views/users/_edit.haml
```diff
@@ -219,3 +219,12 @@
           .btn.btn-danger{id: "close_account", data: {toggle: "modal", target: "#closeAccountModal"}}
             = t(".close_account_text")
           = render "close_account_modal"
+
+    .row
```
SuperTux88: I would put that above the export/close part, so that that's still the last thing. But at least split it with a `%hr`.

app/assets/javascripts/helpers/protocol_handler.es6
```diff
@@ -0,0 +1,21 @@
+// @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL-v3-or-Later
+
+Diaspora.ProtocolHandler = class {
```
SuperTux88: This file needs to be a normal `.js` file, because as far as I know the es6 stuff is only in `develop`.

SuperTux88 (changes requested):
> This implements idea 2 as outlined in #7832 

You mean idea 1, idea 2 would be a popup. But do you want to keep #7832 to track idea 2 or should we close it when this is merged?

> Not the prettiest thing in the world, but it get the job done.

I found two remarks, otherwise this looks good to me.

> To be quite honest, I am also somewhat unhappy with the controller name chosen, and super unhappy with the callback path.

Hmm, while we could rename the controller to describe better what it does, I think the path is OK (and with the routes file pointing to the renamed controller, it should be obvious what it does). I think a short path is OK.

app/assets/javascripts/helpers/protocol_handler.es6
denschub: ... tragic.

denschub:
Remarks addressed.

> I think the path is OK (and with the routes file pointing to the renamed controller, it should be obvious what it does)

I still don't see why we should assign such a short, memorable path to a feature that will never be accessed by anyone directly via the URL. :)

SuperTux88 (approved):

SuperTux88:
Merged as af500dabbccc0fca4344286901c9dc72ddf5114d f8a4a2c510a08c246ac957ac71d86ff5e86ffdb0 8709c45d30dc6ada3f065a8892f0aaf173cd997d and 4430f9e887fc50ca0722819f159111aa4c340786

Thanks for the fast fix :cookie:

milouse:
Even though this has been already merged, I would like to comment a bit. I think the protocol handler section should be in « Services » or « Applications » to be more discoverable. The « Account » section begins to be a bit crowded. Application has my preference as it is all about authorizing stuff. What do you think?

Flaburgan:
Well, it's not really related to Services nor Applications, but it should indeed not be in Account as it is specific to a browser, not an account.

PR #7834, opened 2018-07-15
Fix preview with uploaded photos

SuperTux88:
This fixes a regression I missed when reviewing #7655

### DIFF
--- a/app/assets/javascripts/app/views/publisher/uploader_view.js
+++ b/app/assets/javascripts/app/views/publisher/uploader_view.js
@@ -72,11 +72,19 @@ app.views.PublisherUploader = Backbone.View.extend({
     );
     // replace placeholder
     var placeholder = publisher.photozoneEl.find("li.loading#upload-" + id);
+
+    var imgElement = document.createElement("img");
+    imgElement.src = image.thumb_medium.url;
+    imgElement.classList.add("hidden");
+    imgElement.dataset.id = photoId;
+    imgElement.dataset.small = image.thumb_small.url;
+    imgElement.dataset.scaled = image.scaled_full.url;
+
     placeholder
       .prepend(
-        "<div class=\"x\"></div>" +
-        "<div class=\"circle\"></div>" +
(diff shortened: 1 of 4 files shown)


SuperTux88:
I also (hopefully) fixed the randomly failing test `back out of posting a photo-only post` which fails regularly since #7655 was merged.

app/assets/javascripts/app/views/publisher/uploader_view.js
```diff
@@ -76,7 +76,9 @@ app.views.PublisherUploader = Backbone.View.extend({
       .prepend(
         "<div class=\"x\"></div>" +
         "<div class=\"circle\"></div>" +
-        "<img src=\"" + image.url + "\" data-id=\"" + photoId + "\" alt=\"\" class=\"hidden\" />"
+        "<img src=\"" + image.thumb_medium.url + "\" data-id=\"" + photoId + "\"" +
+        " data-small=\"" + image.thumb_small.url + "\" data-scaled=\"" + image.scaled_full.url + "\"" +
+        " class=\"hidden\" />"
```
denschub: This... hurts my eyes. Can we replace this with something like...

```js
    var imgElement = document.createElement("img");
    imgElement.src = image.thumb_medium.url;
    imgElement.classList.add("hidden");
    imgElement.dataset.id = photoId;
    imgElement.dataset.small = image.thumb_small.url;
    imgElement.dataset.scaled = image.scaled_full.url;

    placeholder
      .prepend(
        '<div class="x"></div>' +
        '<div class="circle"></div>' +
        imgElement.outerHTML
      ).removeClass("loading");
```

It is more verbose, but also a lot more readable imho.

skyschub (changes requested):

app/assets/javascripts/app/views/publisher/uploader_view.js
SuperTux88: Fixed :+1:

skyschub (approved):

PR #7867, opened 2018-09-15
Remove reference to statistics.json

SuperTux88:
The `/statistics.json` is not a thing anymore (#7399), but it was still mentioned in the `diaspora.yml`.

### DIFF
--- a/config/diaspora.yml.example
+++ b/config/diaspora.yml.example
@@ -308,7 +308,7 @@ configuration: ## Section
 
     ## Statistics
     ## Your pod will report its name, software version and whether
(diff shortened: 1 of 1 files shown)


goobertron:
Good catch. Slightly off-topic, but statistics.json is also mentioned in the wiki for [your Munin plugin](https://wiki.diasporafoundation.org/Tools_to_use_with_Diaspora#Podmin_tools). Has the plugin been updated to use node-info? If so, I'll change the mention in the wiki.

jhass (approved):

PR #7870, opened 2018-09-29
Add a docker setup for development

CSammy:
# Docker Setup for Development

This is a first attempt at a Docker setup for development, in order to have a throw-away environment for easier testing, and (maybe in the future) to prevent installation of lots of stuff only needed for certain steps. I'm looking forward to your comments.

Related discussion: 
https://discourse.diasporafoundation.org/t/docker-image-to-set-up-a-development-environment-easily/1682

## Still To Do

- [x] Change "rights changed" message in entrypoint to output mismatching ownership IDs
- [x] Add --force to setup
    - [x] Add --force to build
    - [x] Add --force to config
- [x] Add --full to restart
- [x] Add --force to clean
- ~[ ] Make config checks better (not just "file exists?")~
- [x] Improve config creation and management
- [x] Add full support for pronto
- ~[ ] Improve detection of assets required for rspec~
- [x] Stop DB again if just started for 'exec' 
- [x] Make DB configurable
- [x] Wait for DB startup during migrate step (very slow during setup phase, first boot of DB)

### DIFF
--- a/docker/develop/Dockerfile
+++ b/docker/develop/Dockerfile
@@ -1,0 +1,62 @@
+FROM ruby:2.4.4-slim-stretch
+
+RUN DEBIAN_FRONTEND=noninteractive \
+    apt-get update && \
+    apt-get install -y -qq \
+        build-essential \
+        cmake \
+        curl \
+        ghostscript \
+        git \
+        imagemagick \
+        libcurl4-openssl-dev \
+        libidn11-dev \
+        libmagickwand-dev \
+        libmariadbclient-dev \
+        libpq-dev \
+        libssl-dev \
+        libxml2-dev \
+        libxslt-dev \
+        nodejs \
+        gosu \
+    && \
+    rm -rf /var/lib/apt/lists/*
+
+
+ARG DIA_UID
+ARG DIA_GID
+
+ENV HOME="/home/diaspora" \
+    GEM_HOME="/diaspora/vendor/bundle"
+
+RUN addgroup --gid $DIA_GID diaspora && \
+    adduser \
+        --no-create-home \
+        --disabled-password \
+        --gecos "" \
+        --uid $DIA_UID \
+        --gid $DIA_GID \
+        diaspora \
+    && \
+    mkdir $HOME /diaspora && \
+    chown -R diaspora:diaspora $HOME /diaspora
+
+
+RUN curl -L \
+        https://cifiles.diasporafoundation.org/phantomjs-2.1.1-linux-x86_64.tar.bz2 \
+        | tar -xj -C /usr/local/bin \
+            --transform='s#.*/##' \
+            phantomjs-2.1.1-linux-x86_64/bin/phantomjs
+
+
+ENV BUNDLE_PATH="$GEM_HOME" \
+    BUNDLE_BIN="$GEM_HOME/bin" \
+    BUNDLE_APP_CONFIG="/diaspora/.bundle"
+ENV PATH $BUNDLE_BIN:$PATH
+
+
+COPY docker-entrypoint.sh /entrypoint.sh
+COPY docker-exec-entrypoint.sh /exec-entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+
+CMD ["./script/server"]
--- a/script/diaspora-dev
+++ b/script/diaspora-dev
@@ -1,0 +1,554 @@
+#!/bin/bash
+
+
+# ----- Usage information -----
+
+print_usage() {
+  # Print help for the first argument
+  case "$1" in
+  # management
+    setup)
+      echo; echo "Set up the environment for diaspora*"
+      echo; echo "This command is an alias for the execution of the commands"
+      echo "build, config, bundle, migrate and setup-tests, in that order."
+      print_usage_header "setup [options]" \
+        "    --force    Rebuild image without using Docker's cache;" \
+        "               overwrite existing configuration" \
+        "    --mysql    Use MySQL as database (PostgreSQL is default)"
+      ;;
+    start)
+      echo; echo "Start diaspora* (includes database)"
+      print_usage_header "start [options]" \
+        "-d    Run in background"
+      ;;
+    stop)
+      echo; echo "Stop diaspora* (includes database)"
+      print_usage_header "stop"
+      ;;
+    restart)
+      echo; echo "Restart diaspora* using bin/eye (fast restart)"
+      print_usage_header "restart [options]" \
+        "    --full    Restart entire container"
+      ;;
+    logs)
+      echo; echo "Follow log output of the running diaspora* instance"
+      print_usage_header "logs [options]" \
+        "-a, --all     Follow all containers, including databases"
+      ;;
+    status)
+      echo; echo "Show currently running diaspora* Docker container(s) and related image(s)"
+      print_usage_header "status"
+      ;;
+    clean)
+      echo; echo "Delete diaspora* Docker containers and volumes (includes database)"
+      print_usage_header "clean [options]" \
+        "    --config    Delete configuration files as well"
+      ;;
+  # test & development
+    cucumber)
+      echo; echo "Run cucumber tests"
+      echo; echo "The specified cucumber tests will be executed. If none are given, all"
+      echo "tests are executed."
+      print_usage_header "cucumber [TEST...]"
+      ;;
+    jasmine)
+      echo; echo "Run all jasmine tests"
+      print_usage_header "jasmine"
+      ;;
+    rspec)
+      echo; echo "Run rspec tests"
+      echo; echo  "The specified rspec tests will be executed. If none are given, all"
+      echo "tests will be executed."
+      print_usage_header "rspec"
+      ;;
+    pronto)
+      echo; echo "Run pronto checks"
+      print_usage_header "pronto"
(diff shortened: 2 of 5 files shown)


docker/develop/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+        --gecos "" \
+        --uid $DIA_UID \
+        --gid $DIA_GID \
+        diaspora \
+    && \
+    mkdir /home/diaspora /diaspora
+
+
+RUN curl -L -O \
+        https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2 \
```
denschub: You may use `http://cifiles.diasporafoundation.org/phantomjs-2.1.1-linux-x86_64.tar.bz2`, as we've had issues with randomly disappearing downloads in the past...

SuperTux88: We had issues with randomly disappearing cifiles too :stuck_out_tongue: but at least we can fix that ourselves :+1:

script/diaspora-dev
```diff
@@ -0,0 +1,434 @@
...
+}
+
+dia_config() {
+  # Create rudimentary configuration files if they don't exist
+  [ ! -f "$DIASPORA_ROOT"/public/source.tar.gz ] && touch "$DIASPORA_ROOT"/public/source.tar.gz
+  if [ ! -f "$DIASPORA_CONFIG_DIA" ]; then
+    cp "$DIASPORA_CONFIG_DIA".example "$DIASPORA_CONFIG_DIA"
+  fi
+  if [ ! -f "$DIASPORA_CONFIG_DB" ]; then
+    sed -E '
```
denschub: From our off-the-record discussions, I know you don't really like this sed, neither do I. I just remembered that ERB will be evaluated within `database.yml`, so you could, in theory, provide a default `database.yml` containing something like

```yml
common: &common
  <<: *<%= ENV["DIASPORA_DOCKER_DB"] %>
```

to easily toggle between the chosen database engine. As the format of `database.yml` is not our invention, but in fact, Rails default behavior, it can be considered pretty stable, and thus I don't have much concerns with having a default `database.yml` shipped with the Docker image.

docker/develop/Dockerfile
denschub: ... hey! I'm doing my best!!1

script/diaspora-dev
CSammy: As far as I understand, this would break `script/configure_bundler` as it does not listen to environment variables.

Yes, I am not perfectly happy with this, but it is way better than the rudimentary YAML parser I wrote in awk ...

SuperTux88: I had the same idea and also suggested that, but if the sed works now, I think we can keep that, it should be as stable as the database.yml :)

Also the `script/configure_bundler` tries to parse the database.yml without ERB, so that wouldn't work at the moment.

I also suggested to create a default database.yml for docker that could be used with sed easier (so same as with ERB, but instead of `<%= ENV["DIASPORA_DOCKER_DB"] %>` use something like `%DIASPORA_DOCKER_DB%` which could be replaced with sed with a oneliner.

So as the sed works now, I think we can keep it, if it creates problems in the future, we can still replace it with an extra docker-specific file. :)

denschub: Hm, fair points. I am fine with keeping it, if it doesn't cause pain for anyone. :)

CSammy:
Except for the two TODO points left, I consider this ready-for-use. Please review.

It might be helpful to add links to the corresponding wiki pages (for example for explaining `rspec` or `pronto`), though they would need to be maintained manually. It would be nice to have another opinion on this.

denschub:
I played around with it, and it works nicely, so thanks again for your work.

> Make config checks better (not just "file exists?")

This is a development image, so I see no reason to do careful config checks here. If a file exists, it should be fine, and otherwise, something will probably throw an error (for example if there is a syntax error or important keys are missing), so I'm fine with you doing nothing. It's not like you made the current state any worse. :)

> Improve detection of assets required for rspec

I don't actually know what you mean here. I assume you are talking about detecting `404.html`  and calling `rake:generate_error_pages` based on the result? If so, I would consider this to be perfectly fine. If we ever run in issues with this, we can iterate further.

CSammy:
Using bindfs in Docker was surprisingly intricate, but it's working now, for me at least. @denschub I'd be happy if you could give it another go to improve my confidence on the portability of the solution.

CSammy:
Sadly, the performance impact of bindfs is too big to ignore. Right now, this works well for GNU/Linux, but OSX and Windows users will have a bad time. Apparently, just calling the list of tests with `bin/rake -T` takes over 3 minutes. I'm working on an alternative to directly bind mounting the directory to the container.

Context: Bind mounting on OSX makes calls to the filesystem apparently about 13 times slower. Rake especially is affected by this due to a strange startup behaviour.

CSammy:
We have a working solution now, speeds are okay, with a few seconds slowdown for OS X users.

To give a little bit of context: Docker for Mac uses their filesystem implementation osxfs to represent host data in a container. The implementation has, for various reasons, several drawbacks.
- Calls to the filesystem reportedly take about 13 times as much time as with native access, with varying effect in the field. To give you some perspective about the impact: An rsync dryrun of this diaspora project in such a container took about 2 seconds on a GNU/Linux OS, but about 44 seconds with Docker for Mac on OS X despite the latter running on better and more recent hardware. We have about 37k files in our repository, and most of them appear to be involved in a rake or rails start. In consequence, listing all tests using `bin/rake -T` took about 3:30 minutes in the optimal setup we could create, while it takes 11 seconds for the GNU/Linux machine.
- `inotify` messages are not passed through from the native filesystem of OS X to osxfs in a manner useful to us. While the important messages like `created` or `modified` are translated, some more intricate messages are not. For example, when saving a file in vim, no message is emitted even though the file is clearly modified. For a short period of time, I was hoping I could setup a second container that syncs the diaspora folder between a bind mount and a Docker volume. This is impossible to do reliably with the current state of `inotify` or the afore-mentioned performance of rsync.

[More context on `osxfs`.](https://docs.docker.com/docker-for-mac/osxfs/)

**The (Current) Solution**

The main problem is not size, but the amount of files called. The distribution of files in the diaspora folder shows about 27k files in `vendor/bundle` and about 7k files in `tmp` and the rest is <3k files. Both of these folders contain files either heavily specific to the distribution and/or not relevant for debugging in normal use cases. I opted to use a Docker bind mount like before. However, within the container, from an additional Docker volume, the folders `/diaspora/vendor/bundle` and `/diaspora/tmp` are bind-mounted to. As a result, the dominating chunk of files is kept in Docker volume infrastructure, and `bin/rake -T` showed to run in about 22 system compared to about 11 seconds on the GNU/Linux system. With further testing, it was determined that the slowdown for any rake or rails command was about the same amount of seconds, so it's a fixed slowdown and not relative to the execution time of any tested command.

Maybe we'll come up with an even better solution in the future, but this works for now.

CSammy:
It's now even simpler, I removed mount magic and related privileges.

@denschub Thank you very much for providing an OS X testing ground!

CSammy:
I was able to reproduce the issue with the `start` command using pure docker-compose. This appears to be a [known behaviour of docker-compose](https://github.com/docker/compose/issues/3347) and caused by an [old problem of pyinstaller](https://github.com/pyinstaller/pyinstaller/issues/208). pyinstaller should be fixed with their latest release v3.4, and as such docker-compose should benefit from it as soon as they bump their pyinstaller version.

Since this issue appears to get fixed in the near future, and is unrelated to the wrapper script, I will not build a fix around that but rather wait for the fix to arrive.

CSammy:
I believe this is merge-ready after squashing. Opinions?

denschub:
So, just to make sure everything works as expected, I pruned my Docker and retested everything.

- [x] `setup` working without issues (runs approx. 6 mins in total, including downloading images)
- [x] development server works after setup
- [x] detached server running works
- [x] live-code reloading seems to be working fine.
- [x] running all/a single spec works
- [x] running all/a single cuke works
- [x] running jasmine tests works
- [x] pronto is able to catch new style errors (thanks for those default parameters - I always fail to remember them)
- [x] db migrations get picked up and applied on demand

So, fat 👍 from me, :shipit: and 🎉 .

There is no command to attach logs if you run in detached mode. I don't consider that a blocker, since people who set `-d` may know enough Docker to attach logs themselves, but we probably should add a `log` command at some point. Your choice, either squash the current state and I'll merge, or implement `logs`, squash, and I'll merge after a quick test. Whatever your prefer, and whatever your time budget allows.

Thanks again for the work, this is huge.

skyschub (approved):

denschub:
Additional note for everyone reading this who was unable to hack into private verbal conversations: I am working on a bit of documentation cleanup/improvement, and these changes will also introduce accessible documentation on how to use this container for development purposes. These changes might be a few weeks out, but it's coming.

denschub:
Merged as dae1717da836b2606ffbeac14d5c7f11239e1e9d.

Thank you. Thank you. Will make sure to get you some cookies the next time we meet. :)

Flaburgan:
This is a massive step for one of the most important thing in free software projects: easing stepping of new contributors. Thank you very much to everyone who worked on it, and especially @CSammy and @denschub.

PR #7875, opened 2018-10-04
Added quotation marks for postgresql parameters.

lunagloaming:
As a new, novice user, I ran into an issue where my postgresql
password contained special characters.  After a brief conversation
with DenSchub on #diaspora, it was noted that the example
database configuration had quotes added for mysql but not for
postgresql, and it was suggested this be improved.

### DIFF
--- a/config/database.yml.example
+++ b/config/database.yml.example
@@ -1,9 +1,9 @@
 postgresql: &postgresql
   adapter: postgresql
-  host: localhost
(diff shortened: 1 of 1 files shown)


denschub:
Thank you so much for the fast PR! I've merged this as 45b7f0b11957a9e1f4608b7522d159920161ad05. Also, congrats on your first PR! 🎉

Issue #7812, opened 2018-05-15
Post does not open with guid


DeusFigendi:
A few days ago I had trouble opening a post with its GUID.

This is about this posting:
https://pod.geraspora.de/posts/8041990
https://node.hfrc.de/display/be47b074a97485fd3d5660da05407137d0414fbef8ce3300e9b8f4d187cccd8f

I can **NOT** open https://pod.geraspora.de/posts/be47b074a97485fd3d5660da05407137d0414fbef8ce3300e9b8f4d187cccd8f@node.hfrc.de it returns a 404, tested on 2 diaspora\*-pods. But I can open it for example here https://node.hfrc.de/display/be47b074a97485fd3d5660da05407137d0414fbef8ce3300e9b8f4d187cccd8f@node.hfrc.de which uses the exact same GUID.

It **seems** this is about the guid itself, but it is toatally valid

https://diaspora.github.io/diaspora_federation/federation/types.html#guid

> ## GUID
> 
> A network-wide, unique identifier. A random string of at least 16 and at most 255 chars. It contains only:
> *  Letters: a-z and A-Z
> *  Numbers: 0-9
> *  Special chars: `-`, `_`, `@`, `.` and :
> 
> Special chars aren’t allowed at the end.

The GUID be47b074a97485fd3d5660da05407137d0414fbef8ce3300e9b8f4d187cccd8f@node.hfrc.de is

* 77 chars long (>16; < 255)
* contains only letters, numbers @ and . which are all allowed
* ends with `e`

so I'd guess it is diasporas issue not hubzillas :)

I remembered deadsuperhero did something similar once, but I couldn't find a similar issue.

Oh

### How to reproduce:

* Search for holger@node.hfrc.de on your diaspora\*-pod
* Open a post in single-view
* Click the timestamp to get the post with the GUID in single view
* 404

SuperTux88:
GUIDs from hubzilla/redmatrix never worked for showing posts, I think it's at least the dots which are a problem for how rails handles URLs and the last dot is used to split the `format`, I don't know if there are more problems with it.

> It seems this is about the guid itself, but it is toatally valid

About valid and invalid:

[...]

PR #7876, opened 2018-10-04
Bump rubyzip, nokogiri and diaspora_federation

SuperTux88:
Fixes CVE-2018-1000544


SuperTux88:
I also just added another commit for nokogiri which was released two hours ago to fix:
* CVE-2016-9318
* CVE-2017-16932
* CVE-2017-18258
* CVE-2018-14404
* CVE-2018-14567

SuperTux88:
I also added a new version of `diaspora_federation` to this PR, because I don't want to open multiple PRs for the Gemfiles.

For diaspora relevant changes:
* Include `web+` prefix in `diaspora://` URL parsing (prevents `web+https://` links when replacing `diaspora://` URLs)
* Allow fetching of entities with dot in the GUID (fixes the problem of #7812 for fetching)
* Some fixes in the `federation_entities.json` (for the upcoming migration/import, but I think not relevant for 0.7.7.0 yet)

greysteil:
@SuperTux88 - would you be up for using [Dependabot](https://dependabot.com) to automatically generate dependency update PRs? I ran it on my fork and it generated [these PRs](https://github.com/greysteil/diaspora/pulls) (it caps itself at 5 PRs per run, and 10 open in total).

Hopefully a useful timesaver? It's used by a few thousand organisations, has been approved by GitHub's security team and the core source code is available [here](https://github.com/dependabot/dependabot-core). I built it and work on it full time.

SuperTux88:
No thanks, we usually do many gem updates at once (I just didn't have the time to do them all this time, that's why I only did it for the security relevant gems), and it's much easier to do them locally and review changelogs, instead of reviewing and testing a ton of PRs. What needs time is reading all the changelogs and testing if anything breaks, creating the PR isn't much work.

Automated PRs would only generate a lot of noise for the 500 watchers here and add unneeded load to travis for every new release of something.

greysteil:
Fair enough - thanks for considering, anyway. I always found finding changelogs and release notes a pain which is why I built them into Dependabot, but if you've got a good flow that works locally I guess that's not an issue.

SuperTux88:
Oh, you have changelogs in there :heart: I didn't see them at my quick look before (maybe too quick), I didn't notice that I need to expand them. That's actually a very useful feature, but I still don't want to create a PR for every single release of every single gem, because of the reasons mentioned above.

I used gemnasium before, which sadly doesn't exist anymore. It had a dashboard with all gems and I could see the changelogs between the current and the newest version, which was perfect. Last time I needed to search all changelogs myself, which works, but is a lot more work than having them all at one place.

So if you already have all the changelogs, providing them in a list/dashboard for all gems that aren't up-to-date would be a really cool thing and actually something I would happily use. I just don't like the PR thing, because I don't want to spam all the watchers with that, and I don't want 80% of the PRs be just gem updates.

PR #7881, opened 2018-10-08
Add to string when pass integer to person model function

VitorRibeiroCustodio:
Fixing function find_by_substring in Person model. When pass a integer to the function, that method doesnt, because, strip is a string method.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -104,7 +104,7 @@ class Person < ApplicationRecord
   # @param [String] search substring
   # @return [Person::ActiveRecord_Relation]
   scope :find_by_substring, ->(search_str) {
-    search_str = search_str.strip
+    search_str = search_str.to_s.strip
     if search_str.blank? || search_str.size < 2
       none
     else


VitorRibeiroCustodio:
Fixing function find_by_substring in Person model. When pass a integer to the function, that method doesnt, because, strip is a string method.

jhass:
Which issue is this fixing? 

Semantic wise I think it's fine for a method called `search_by_substring` to expect being passed a string or something that quacks like it.

CSammy:
Does it break if you search for a number in the frontend? If it doesn't, we don't need this.

HankG:
@CSammy  I have confirmed that the UI doesn't have an error.  I've looked at the logs and it is processed internally as a string from the REST handler so a path where the UI feeds it an integer doesn't exist.  The way I'd be coding up the search on API would also not have this artifact.  

@VitorRibeiroCustodio was there a code execution path you saw that this to error like this?

denschub:
This hasn't received more info within a month. @VitorRibeiroCustodio, please answer the questions asked in this PR if you ever come back. If this turns out to be a valid fix, we can reopen this PR.

Issue #7841, opened 2018-08-12
Unliking a post disables notifications for the post but the UI doesn't get fully updated


pskosinski:
1. Like a post → In the right upper corner you can see the bell icon getting black. Now you'll receive notifications for the post.
2. Unlike the post → The bell icon stays the same, even that you'll no longer receive notifications.
3. Refresh the page → The bell icon is grayed out, even that all you did was refreshing the page.

When one unlikes a post they will no longer be getting notifications but the UI doesn't get fully updated, at least not instantly.

HankG:
I think I see what you are stating but I'm not needing to refresh.  The problem when I try to reproduce it is just on the hover.  If I move the mouse off the bell icon, even if close enough to not force it to hide, then it properly updates.  However if the mouse is right on top of the icon it gets a bit muddled in the messaging.  Is this the behavior you are seeing or is it more dramatic than this, specifically are you sure you need to refresh the page? I tested this on macOS High Sierra running Firefox 62 with my pod running 0.7.6.0.

pskosinski:
Hovering the bell icon doesn't change anything in my case. Well, the icon toggles between black and greyed out on hover, but it's still wrong, it's black when it should be greyed out and vice versa. Yes, I'm sure I need to refresh the page. I have that issue on Firefox 62.0 and Google Chrome 69.0.3497.92 on Fedora 27, on pod with 0.7.6.0 software.

HankG:
Let me try it under Chrome to see what that behavior looks like for me on Chrome as well to try to reproduce it. Thanks!

HankG:
I've tested this on Chrome 69 under macOS 10 and Linux Mint MATE 19.  I'm not seeing the problem of the status not refreshing.  The behavior is consistent with my testing on FireFox 62 under macOS and Linux Mint MATE 19.  If I move the mouse slightly away from the icon the correct status shows when I try to toggle between the two.  If I move the mouse sufficiently away that the icons hide themselves again and come back they also retain a consistent status.  On your system you are seeing different behavior and requiring a refresh (I know you restated it in your previous response I just want to make sure we are doing the exact same thing in our browsers).  

I can see what you are saying about the iconography being a bit muddled though (because the hover color is hard to distinguish from the enabled/disabled color).  Without the text pop-up I'd probably be confused about what the status was until I moved away as well.  While I can't reproduce the status update part of this story I can understand discussing modifying the icons to be clearer.

pskosinski:
[...]

PR #7882, opened 2018-10-08
#7841 Post Unlike sets status icon correctly

HankG:


jeremyf:
👍

skyschub (changes requested):
Given that this was broken without us noticing, this probably needs a test.

HankG:
@denschub OK.

HankG:
I've updated the unit test to capture the errant behavior as well as to be as robust as the one for "like".

HankG:
Are there additional changes after my latest commit where I addressed the unit test or is something out of sync with the status?  I thought when I submitted new code it would reset the status with the new commit.  If I am looking at the main board view it looks like nothing has changed for two days and this needs changes, but in reality the PR changed a day ago and the new changes haven't been reviewed.  PS this is not meant as "why haven't the changes been reviewed?" this is a question about potentially stale status (or stale looking status) on the main PR view.

HankG:
This is ready for review.

skyschub (approved):
Thanks for adding the test! :)

denschub:
Merged!

Issue #7840, opened 2018-08-11
feature request: 'ignore' button in the mobile view


teslasmoustache:
I believe that the availability of an 'ignore' button in the mobile view is a necessary feature.

Thank you.

HankG:
When you say "ignore button" you are referring to on a profile not on an article, correct? I ask because none of the status action buttons exist on the mobile view (Enable/Disable alerts, report, ignore, hide).

teslasmoustache:
Either one, tbh. Both if possible. I suppose it would make the most sense to make the 'ignore' button available on profile pages, since what you would be trying to do is ignore a profile.

But, honestly, whatever's easiest. Some level of functionality is better than none.

HankG:
Starting work on implementing the ignore feature at the user level.  The notifications stuff on posts is a whole other topic and maybe should have a whole separate issue for tracking purposes.

I found a bug in the current mobile implementation of the profile viewer however.  The way the aspects drop down is coded in mobile it doesn't show a presently ignored user it just renders it as "Add Contact".  I know the Mobile code can do this because it does it correctly under privacy settings.  I'll work on addressing that at the same time as getting this working.

HankG:
I've got this working in a branch branch.  Posted some questions to Discourse before finishing it up and submitting a PR.

PR #7884, opened 2018-10-08
#7840 Ignore Users in Mobile UI

HankG:


HankG:
I believe that if this looks good as is, implementation wise, I need to edit the tests to reflect proper state of the new "block" field.

HankG:
I was able to do this without changing the controller so this I don't think this is a WIP any longer.  I've also attached screenshots of the UI in it's two states and the dialog box when selecting to block:
image

image

image

Flaburgan:
Thank you for your work! To make the UI consistent with the desktop version, maybe you could reuse the icon used there?

HankG:
@Flaburgan I played around with the icon but I was having problems getting it to render in the same way (sans label, same size).  If that is desired I can give that a whirl though.

Flaburgan:
Well it would be better I guess. If you have trouble with CSS please tell me, I should be able to help.

HankG:
Playing around a bit we have a version with a label and one without.  The one without is more consistent with the desktop version but there is no hover so therefore no  way for users to know what it means without clicking on it.  Thoughts?
image

image

Flaburgan:
My point here is "to ignore someone" is not an action you do everyday, and is not something we want to encourage, so the action icon should not be prominent. IMO, the icon alone is enough, but the alert text should be changed to "You are going to ignore that user. Are you sure?".

Also align it bottom right like on the desktop version please :)

HankG:
@Flaburgan It's also a bit of an edge case of someone who has never used the desktop version so they have no idea what that icon means. I'm having a problem getting it to go right without artifacts.  Using the same pull-right class (or variations of what I've seen in other controls doing that) is not producing what I want  or is doing it with artifacts.  If you have a suggestion for the way to do that that'd be great.  I'm just manipulating the class on the link button in app/views/people/show.mobile.haml and changing where I actually put that element.

app/views/people/show.mobile.haml
```diff
@@ -10,7 +10,20 @@
...
-      = render 'aspect_memberships/aspect_membership_dropdown'
+      - if @presenter.full_hash[:block]
+        .pull-right
+          = link_to t("users.privacy_settings.stop_ignoring"),
+            block_path(@presenter.full_hash[:block][:id]),
+            class: "btn btn-danger",
+            method: :delete
+      - else
+        = render "aspect_memberships/aspect_membership_dropdown"
+        = link_to t("users.privacy_settings.start_ignoring"),
```
Flaburgan: Try this before this line:
```
.pull-right.clear
   = link_to "",
     blocks_path(block: {person_id: @person.id}),
     data: {confirm: t("are_you_sure")},
     method: :post,
     confirm: true,
     class: "entypo-block"

Comment (Flaburgan) [reply app/views/people/show.mobile.haml]: And you'd need to change the translation and replace "are_you_sure" as I proposed in https://github.com/diaspora/diaspora/pull/7884#issuecomment-428485259

Comment (HankG) [reply app/views/people/show.mobile.haml]: Change the translation of what?

Comment (HankG) [reply app/views/people/show.mobile.haml]: The above didn't work but I was able to get what I think we are looking for.  I'm torn on whether the icon needs to be one size larger or not: 
image

Comment (HankG) [reply app/views/people/show.mobile.haml]: Changes to reflect the above have been pushed to the GitHub repo for new review...

**Comment by HankG** [comment]
This is ready for review now and captures the changes requested.

**Review by skyschub — CHANGES_REQUESTED** [review]
Looks good implementation-wise, thanks for your work. Could you please give a shot at vertically centering the icon? It's a bit suspended up in the air there.

screenshot 2018-11-08 at 04 45 51

**Comment by HankG** [comment]
I think I got it (preview below): 
image

**Review by Flaburgan — APPROVED** [review]
I just tested this locally and it works well, thank you!
The code also looks okay to me. I'm approving this and add the 0.7.9.0 milestone, this looks ready to be merged to me.

File: app/views/people/show.mobile.haml [hunk 2]
```diff
@@ -10,7 +10,23 @@
...
+        .pull-right
+          = link_to t("users.privacy_settings.stop_ignoring"),
+            block_path(@presenter.full_hash[:block][:id]),
+            class: "btn btn-danger",
+            method: :delete
+      - else
+        = render "aspect_memberships/aspect_membership_dropdown"
+    .clear
+  .mobile-icon-bar
+    - if user_signed_in? && @person != current_user.person && !@presenter.full_hash[:block]
```
SuperTux88: You should move this line one up, because it currently displays an empty icon-bar (with an double line) when you are already ignoring a user:

image

SuperTux88 (changes requested):
Only one last little thing :)

SuperTux88 (approved):

Issue #7664, opened 2017-11-14
Feature request: Larger images


SunyataZero:
It would be nice to have the option to upload and see larger images

Btw thank you all for what you are doing for the freedom of the internet

Kind Regards, Tord

DeadSuperHero:
Well, Diaspora's uploader already provides several different versions of the image after it has been uploaded and processed: https://github.com/diaspora/diaspora/blob/develop/app/uploaders/processed_image.rb

When you refer to "Having the option to upload and see larger images", what are you referring to here specifically? Would this be a user setting, or are you hoping to just have link access somewhere to the full-size image on statuses containing image attachments?

SunyataZero:
@DeadSuperHero Hi and thank you for your answer and interest in this. As it is right now when i click on an image it covers only a fairly small part of the screen for me:

![image](https://user-images.githubusercontent.com/10245688/32923898-4d0e95b6-cb3a-11e7-94d2-62454b393713.png)

(If i right click on the photo and choose "open in new tab" it has the same size)

It would be nice if there was an option to go to full screen, even if the file is smaller than full screen this may still look good

Another thing would be to increase the max pixel size of the allowed images, but i can understand that you have limited bandwidth and the hosting costs money

A third thing is to increase the size of the image in the view that you get by clicking on it (as in the screenshot above), this could be done either by "simply" increasing the max pixel size, or by scaling up the image

Again, thank you for what you're doing, i appreciate it and i'm trying to spread the word to my friends

Kind Regards,
Tord

Flaburgan:
As explained [here](https://github.com/diaspora/diaspora/issues/7888#issuecomment-429291214), there are two issues here: the size of the displayed images, in pixels, and the allowed size of uploaded images, in bytes.

Since https://github.com/diaspora/diaspora/pull/7734 if the uploaded image is bigger than 4MB then we try to bring it under that limit by resizing its biggest side to be 3072px max. IMO that's big enough for the raw file, we're not Flickr and if you want to share bigger images in diaspora*, you can still host them and embed them using markdown. (I guess the "needs discourse" label was there because of that problem? @denschub please confirm)

So this issue is now only about the size of the image on the screen. In my opinion, 700px (so our `scaled_full` files) is big enough for the stream (mainly to limit bandwidth consumption), but I agree that when the users click on the image (so they explicitly ask to display the image) we should show it bigger. We have the raw file, so we can do that.

I'm updating the issue title and will check the code, see if the change is hard to do.

[...]

PR #7890, opened 2018-10-12
Display raw images in gallery

Flaburgan:
Fixes https://github.com/diaspora/diaspora/issues/7664.

As said there, I feel like when the user explicitly open an image, we should give him the original one and not a resized.

### DIFF
--- a/app/assets/stylesheets/gallery.scss
+++ b/app/assets/stylesheets/gallery.scss
@@ -1,12 +1,24 @@
-$thumbnail-size: 12px;
-$thumbnail-margin: 2px;
-$thumbnail-active-size: $thumbnail-size + $thumbnail-margin;
+$thumbnail-size: 50px;
+$margin: 15px;
 
 #blueimp-gallery {
   .slides {
-    height: calc(100% - 40px);
-    padding: 20px 0 0 0;
     margin: 0;
+
+    .slide {
+      overflow-y: auto;
+    }
+
+    .slide-content {
+      bottom: $margin * 2 + $thumbnail-size;
+      top: $margin;
+    }
+
+    .too-tall {
+      margin-bottom: $margin * 2 + $thumbnail-size;
+      max-height: none;
+      position: static;
+    }
   }
 
   [class^="entypo-"], [class*="entypo-"] {
@@ -37,24 +49,28 @@ $thumbnail-active-size: $thumbnail-size + $thumbnail-margin;
   }
 
   .indicator {
-    margin: 8px 0;
-    position: unset;
-    height: $thumbnail-size + 5px;
+    bottom: 0;
+
     li {
-      border: none;
-      margin: $thumbnail-margin;
+      border: 0;
+      border-radius: $thumbnail-size / 2;
+      height: $thumbnail-size;
+      margin: 6px;
+      margin-bottom: $margin;
       vertical-align: middle;
       width: $thumbnail-size;
-      height: $thumbnail-size;
-      border-radius: $thumbnail-size / 2;
 
-      &.active, &:hover{
-        margin: $thumbnail-margin / 2;
-        width: $thumbnail-active-size;
-        height: $thumbnail-active-size;
-        border-radius: $thumbnail-active-size / 2;
-        transition: linear 0.2s;
-        transition-property: height, width, margin;
+      &.active,
+      &:hover {
+        opacity: 1;
+      }
+
+      &:hover::after {
+        opacity: 0; // We don't want another thumbnail
+      }
+
+      &:only-child {
+        display: none;
       }
     }
   }
--- a/app/assets/javascripts/app/views/gallery_view.js
+++ b/app/assets/javascripts/app/views/gallery_view.js
@@ -31,12 +31,20 @@ app.views.Gallery = app.views.Base.extend({
     return {
       index: link,
       event: event,
-      hidePageScrollbars: false,
-      disableScroll: true,
       continuous: true,
       toggleControlsOnReturn: false,
       onopened: this.preventHideControls,
-      slideshowInterval: 2000
+      slideshowInterval: 2000,
+      onslidecomplete: function(index, slide) {
+        // If the image is very tall (more than twice its width), then it is scrollable instead of resized
+        var image = slide.firstElementChild;
+        if (image.naturalHeight > window.innerHeight && image.naturalHeight > image.naturalWidth * 2) {
+          image.classList.add("too-tall");
+        } else {
+          var margins = 95; // Margins are 80px for thumbnails height and 15px for top image margin
+          image.style = "max-height: " + (window.innerHeight - margins) + "px";
+        }
+      }
     };
   }
 });
--- a/app/assets/templates/photo_tpl.jst.hbs
+++ b/app/assets/templates/photo_tpl.jst.hbs
@@ -19,7 +19,7 @@
     </div>
   {{/if}}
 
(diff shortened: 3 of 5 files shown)


denschub:
We've had too many discussion around image sizes, so I'm not even going to argue. If someone uploads an image that breaks UI/UX for some reason, we can fix that in the future. Showing the full-res image is probably fine, especially since it's clamped by the max. upload size.

However, I seriously dislike the larger thumbnails. Not only is it actually too large (the thumbnail shown is 50px wide, but the circle is 52px large), creating visible borders:

screenshot 2018-10-14 at 06 38 16

it also renderes the hover-on-thumbnail thingy kinda pointless:

screenshot 2018-10-14 at 06 40 11

The intention of those bobbles at the bottom is to give you an impression of where in the gallery one is at, and the hover is there to give you an idea on how the picture might look like. In addition, the larger thumbnails now overlap standard 3:2 photos on 16:10 and 16:9 screens, and I'd rather have small bubbles instead of obnoxious large bubbles that lay on top of photos.

Flaburgan:
To be honest, I only discovered the `:hover` feature while doing that PR. I was using the bubbles to navigate directly to a picture, and the current size definitely is too small to be able to see what's in. I never thought of "hovering" it. That's also something that doesn't work on mobile.

For those reasons, I suggest we drop the hover feature and increase the thumb size (maybe not as big as 50px, but if you think about it it's not that big). It also needs a `background-size: cover;` to adapt the image which isn't square most of the time. The thumbnails over the picture in some cases is not a problem IMO, if the picture is very tall, a scrollbar appears.

Do you think it is a good idea?

CSammy:
I'm always in favor of dropping features that are close to useless. It gives us more time to concentrate on important features.

denschub:
>  I suggest we drop the hover feature

Feel free to do so.

> It also needs a background-size: cover; to adapt the image which isn't square most of the time. 

Thumbnails [will be square every time all the time](https://github.com/diaspora/diaspora/blob/21f9a9d134a7869260cb8fa08f474cf1706c6037/app/uploaders/processed_image.rb#L22-L24).

> The thumbnails over the picture in some cases is not a problem IMO

Overlapping thumbnails in the case of the most common image aspect ratio (almost every digital camera shoots pictures in either 4:3 or 3:2, which both cause overlapping) in our most common screen aspect ratios (71% of Geraspora users were using diaspora\* maximized on 16:9 and 16:10 displays a year ago) is a problem IMO.

Flaburgan:
>Overlapping thumbnails in the case of the most common image aspect ratio (almost every digital camera shoots pictures in either 4:3 or 3:2, which both cause overlapping) in our most common screen aspect ratios (71% of Geraspora users were using diaspora* maximized on 16:9 and 16:10 displays a year ago) is a problem IMO.

Here are different mockups, to help us decide what to do.
**Current state:**
current

**With a max-height on the image:**
max-height

**Putting the thumbnails on the left:**
left
With this solution we need to be careful to not block the arrow left button when there are many thumbnails.

What do you think?

001101:
Would be nice if there is a option for the admin to select the quality and therefore determine if the pod has enough bandwidth for all the raw pictures to serve or if he prefers less good quality and more speed.

SuperTux88:
The images are always loaded from the pod of the author, not the pod of the viewer. And the podmin of the author can't control the pod of the viewer. So a setting wouldn't make much sense, if a pod does have less bandwidth, it should limit the number of users (the number of possible authors).

001101:
So the pod of the author does control the view on the viewers pod?

HankG:
I've played around with this versus the head of develop side by side and **really** like this.  The code changes look pretty straight forward too.  Because there is an upload size limit on images I think it looks good as is without worrying about someone having to pull down a 10MB image or something like that.

Flaburgan:
I continue to try things on this. Especially, I'd like to correctly deals with big height images like [this one](https://user-images.githubusercontent.com/930064/47871634-e2bfe200-de0c-11e8-9bd9-8f9200babb88.png) for example, which are currently very poorly displayed on diaspora:
big-height

Unfortunately, I can't find a nice solution to display them without the thumbnails to override them: 
long-bottom

I still think about displaying them on the left (and possibly detect rtl users and display it on the right then). However, I also had an idea: those kind of images are often shared alone. Maybe simply do not display the thumbnail when there is only one image would solve 90% of use cases? @denschub would that be fine for you?

SuperTux88:
> simply do not display the thumbnail when there is only one image

That sounds like a good idea to me anyway, because with only one image the thumbnail is pretty useless anyway.

Also I don't like the thumbnails at the side, that looks strange and probably also can collide with the arrow at the side.

Flaburgan:
Okay PR updated, I had to do code a bit more complicated it but the result is really cool.
So first of all, no more thumbnail if there is only one image. Other improvement: the image are now display with max-height set to 90%, bottom room is for the thumbnail. So you can see the whole image but it takes as much space as there is. However, if the image is very tall (read, height is more than twice width) then the image is displayed without being resized, with a scroll bar.

screenshot_2018-11-02 diaspora
screenshot_2018-11-02 diaspora 2
screenshot_2018-11-02 diaspora 1
screenshot_2018-11-02 diaspora

I'm going to put it on https://diaspora-fr.org to see how it works and ask feedback for users.

CSammy:
Is it possible to look at the very bottom of a very tall image without the thumbnails covering up parts of it?

app/assets/stylesheets/gallery.scss
```diff
@@ -37,24 +48,28 @@ $thumbnail-active-size: $thumbnail-size + $thumbnail-margin;
...
-      border-radius: $thumbnail-size / 2;
 
-      &.active, &:hover{
-        margin: $thumbnail-margin / 2;
-        width: $thumbnail-active-size;
-        height: $thumbnail-active-size;
-        border-radius: $thumbnail-active-size / 2;
-        transition: linear 0.2s;
-        transition-property: height, width, margin;
+      &.active,
```
diaspora-code-review: Nesting should be no greater than 3, but was 4

app/assets/stylesheets/gallery.scss
```diff
@@ -37,24 +48,28 @@ $thumbnail-active-size: $thumbnail-size + $thumbnail-margin;
...
+      &:hover {
+        opacity: 1;
+      }
+
+      &:hover::after {
```
diaspora-code-review: Nesting should be no greater than 3, but was 4

app/assets/stylesheets/gallery.scss
```diff
@@ -37,24 +48,28 @@ $thumbnail-active-size: $thumbnail-size + $thumbnail-margin;
...
+        opacity: 0; // We don't want another thumbnail
+      }
+
+      &:only-child {
```
diaspora-code-review: Nesting should be no greater than 3, but was 4

app/assets/stylesheets/gallery.scss
Flaburgan: I can't fix this one, when I put it outside of `.indicator` it asks me to merge it. IMO pronto should be set to 4, 3 is very low.

Flaburgan:
>Is it possible to look at the very bottom of a very tall image without the thumbnails covering up parts of it?

Hm, yeah, I should be able to do that.

Flaburgan:
@CSammy done

screenshot_2018-11-03 diaspora

skyschub (approved):
I really like this.

denschub:
Merged!

Issue #7838, opened 2018-07-19
Improve description of web+diaspora:// handler


Flaburgan:
When the user clicks on the Register button after he already registered, nothing happens. The button should not be displayed anymore. Instead, there should either be a button allowing to unregister, or a text explaining how to manage handlers inside the browser.

SuperTux88:
AFAIK there is neither a way to tell if a user installed the handler nor is there a way to unregister the handler, so I think we can't solve this?

But you can extend the current text if you like and write something about that the button only does something if the handler isn't installed already and that the users need to remove the handler from the browser settings.

Flaburgan:
Let's do that then.

denschub:
> AFAIK there is neither a way to tell if a user installed the handler nor is there a way to unregister the handler, so I think we can't solve this?

Just dropping by to say that this is true.

develaper:
Hi! I am planning to start to work on this issue but not very sure about how to reproduce it. Are we talking about the "register button" that takes the user to "/users/sign_up" ?

Flaburgan:
No, we're talking about the *web+diaspora:// protocol handler* section in the settings. You can read there what's this about.

cmrd-senya:
@Flaburgan does your comment missing a link at the _there_ word?

cmrd-senya:
> AFAIK there is neither a way to tell if a user installed the handler nor is there a way to unregister the handler, so I think we can't solve this?

If this is true then perhaps we should change the issue topic, shouldn't we?

Flaburgan:
> @Flaburgan does your comment missing a link at the _there_ word?

No, I was talking about the settings inside diaspora*. Sorry if that wasn't clear.

PR #7909, opened 2018-10-26
Reword the web handler setting

Flaburgan:
I tried to reword the web handler instruction a bit. It's hard to keep it short but clear at the same time. I was wondering if there should be an example, explaining the problem it solves: a link "joindiaspora.com/posts/xxx" would not allow users of other pods to be logged in and comment on it, the new syntax should be used instead. But it was too much. Maybe there should be a section in the help page, and which could be linked there?

Please check the english here, it can surely be improved.

Fixes https://github.com/diaspora/diaspora/issues/7838

### DIFF
--- a/app/views/users/_edit.haml
+++ b/app/views/users/_edit.haml
@@ -184,7 +184,8 @@
       .col-md-12
         %h3#protocol-handler
           = t(".protocol_handler.title")
-        %p= t(".protocol_handler.description")
+        %p= t(".protocol_handler.description", pod_url: AppConfig.pod_uri.site)
+        %p= t(".protocol_handler.browser")
         .form-group
           %button.btn.btn-default#register-protocol-handler
             = t(".protocol_handler.register")
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -1261,8 +1261,9 @@ en:
 
       protocol_handler:
         title: "web+diaspora:// protocol handler"
-        description: "Clicking this button will ask your browser to install a handler that allows us to open web+diaspora:// URLs on your home pod. This is currently experimental and interactions will depend on your browser."
-        register: "Register"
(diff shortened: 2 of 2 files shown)


goobertron:
Once the new message has been approved from a technical point of view, could someone ping me so I can polish the grammar etc?

HankG:
@goobertron @Flaburgan I've reviewed this and it is working.  I agree with the content from a technical POV and look forward to helping review the refined text.

HankG:
> I've done some rewording. I've made it a bit more prosaic so it explains a bit more clearly exactly what this handler does. If my understanding is incorrect, please let me know what needs to be changed!

I think the new text sounds better.  One question I have is if this text is supposed to be read by a complete newbie who has no idea why we would do this and has no concept of federation.  If the answer to that is "yes" then we may want to add a little more or reference a wiki offsite for more info on the protocol.  If "no" then I think it's fine.

> I also wonder about extracting the text 'web+diaspora://' into a separate string, webdiaspora, so that we can easily format it in an inline code block or in bold. I think it would look better that way.

From what I can tell this is not feasible in the I18n package without some hacking.  I'd advise against that.  If someone knows how to do it easily though I think it sounds like a fine idea...

Flaburgan:
I updated this PR to use your wording @goobertron, thank you!

goobertron:
@HankG yup, you can do something like:

in `config/locales/en.yml`:

>       webdiaspora: "web+diaspora://"
>       register: "Register the <code>%{webdiaspora}</code> handler on this browser"

and then in `app/views/users/_edit.haml`:

>        = t(".protocol_handler.register", webdiaspora).html_safe

(I think!)

goobertron:
You don't *need* to do it, but I think it could be useful for a repeated and somewhat complex and critical string such as this, as it would for one thing avoid the possibility of transcription errors by our translators.

SuperTux88:
I think it's simpler with just one plain string instead of multiple levels with placeholders. There also can be problems with placeholders (because of the higher complexity), like at the project site a few weeks ago. So I think placeholders should only be used when there is something dynamic that needs to be inserted.

SuperTux88 (approved):

SuperTux88:
Merged as 65e0f90b866fb44e321140ce0561fa13ba0dadac

Thanks to everybody who helped with this :cookie: :cookie: :cookie: :cookie:

PR #7912, opened 2018-10-30
Save a screenshot of the page in tmp/capybara when a cucumber test is failing

Flaburgan:
Not sure if I have to add something to avoid the CI doing it.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -77,3 +77,8 @@ Before do |scenario|
   # Reset overridden settings
   AppConfig.reset_dynamic!
 end
+
+After do |scenario|
+  Capybara.save_path = ENV["SCREENSHOT_PATH"]
+  page.save_screenshot("#{Time.now.utc} #{scenario.name}.png", full: true) if scenario.failed? && ENV["SCREENSHOT_PATH"]
+end


jhass:
Maybe we want to explicitly turn this on via an environment variable?

Flaburgan:
Up to you (and @SuperTux88 and @denschub). It can be switched on / off by anything. As long as the code doesn't have to be modified and the Testing Workflow page of the wiki clearly explain to new devs how to debug tests, I'm fine ;)

SuperTux88:
Since we need to explain where the screenshots are found in the wiki anyway, I also suggest to only enable that behavior only with an env var, which then can be documented too.

Also, the `tmp/`folder isn't accessible with the new dev-docker-container, is it possible to set another path?

Flaburgan:
Alright, I solved both problem in one time: now screenshots are only saved if `SCREENSHOT_PATH` env var is added, and they are saved at the path given that way. So to have a screenshot of a failing tests, you need to launch it that way: `SCREENSHOT_PATH="." bin/cucumber features/desktop/change_settings.feature`

SuperTux88 (approved):

SuperTux88:
Merged as b977e9c39af1dc65c6d84382dd5812a8f0649cc9

Thank you :cookie:

Flaburgan:
I added the instructions in [the wiki page](https://wiki.diasporafoundation.org/Testing_workflow#Debugging)

PR #7913, opened 2018-10-31
Do not autofollow back a user you are ignoring

Flaburgan:
Fixes #7808 

I also added a test for the auto follow back feature as it looks we didn't have one.

### DIFF
--- a/app/models/user/connecting.rb
+++ b/app/models/user/connecting.rb
@@ -11,6 +11,8 @@ class User
     # @param [Aspect] aspect The aspect to add them to.
     # @return [Contact] The newly made contact for the passed in person.
     def share_with(person, aspect)
+      return if blocks.where(person_id: person.id).exists?
+
       contact = contacts.find_or_initialize_by(person_id: person.id)
       return false unless contact.valid?
 
--- a/features/desktop/auto_follow_back.feature
+++ b/features/desktop/auto_follow_back.feature
@@ -1,0 +1,36 @@
+@javascript
+Feature: auto follow back a user
+
+  Background:
+    Given following users exist:
+      | username       | email             |
+      | Bob Jones      | bob@bob.bob       |
+      | Alice Smith    | alice@alice.alice |
+    And I sign in as "bob@bob.bob"
+    And I have an aspect called "My main aspect"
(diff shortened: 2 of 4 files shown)


jhass:
Makes me wonder, should we fail following a user at all if they're ignored, so an abstraction level below this? Would that break anything UI side?

Flaburgan:
Yeah I asked myself the same question. I think we should not block userA for following userB, even if userB is blocking userA. diaspora* relationship is asymmetrical, if userA wants to share with userB (so make is private profile available, even limited posts available) then that's fine. UserB blocking userA can still go to userA profile sometimes and see posts there.

jhass:
Yeah sure, it's not what I meant, rather doing the same check as this, just generally in `User#share_with`, so failing there if you try to follow a user you ignore.

Flaburgan:
>if you try to follow a user you ignore.

The UI doesn't allow you to do that, the aspect dropdown selector becomes "Stop ignoring" when you ignore someone. But I guess it should be done in the controller anyway, as someone can manually send such a request.

jhass:
And instead of patching all the holes  (controller, autofollowback, what else did we forget), I wonder if it makes sense to just put the filter into the one place that all these call.

Flaburgan:
Agree.

I don't see a method `sharing_with` in the `user.rb` file though, can you please point me to the place where you want me to move the check? Thanks!

jhass:
I was thinking of https://github.com/diaspora/diaspora/blob/develop/app/models/user/connecting.rb

app/models/contact.rb
```diff
@@ -63,7 +63,8 @@ def in_aspect?(aspect)
 
   # Follows back if user setting is set so
   def receive(_recipient_user_ids)
-    user.share_with(person, user.auto_follow_back_aspect) if user.auto_follow_back && !receiving
+    return if !user.auto_follow_back || receiving || user.blocks.where(person_id: person.id).exists?
```
diaspora-code-review: Layout/EmptyLineAfterGuardClause: Add empty line after guard clause.

Flaburgan:
@jhass PR updated to do the test in the `share_with` method.

jhass (approved):

PR #7919, opened 2018-11-01
Use libjemalloc if it's available for better memory usage

### LINKED ISSUE #6763: Sidekiq or sidekiq-cron is bloating memory
I open an issue to be able to set this as a blocker for 0.6. This is a regression since we replaced sidetiq by sidekiq-cron and bumped sidekiq from 3 to 4, not sure which action caused it.

See the original discussion on the [pull request](https://github.com/diaspora/diaspora/pull/6616#issuecomment-182231026)

<bountysource-plugin>

---
There is a **[$15 open bounty](https://www.bountysource.com/issues/32103278-sidekiq-or-sidekiq-cron-is-bloating-memory?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SuperTux88:
Fixes #6763

### DIFF
--- a/script/server
+++ b/script/server
@@ -202,6 +202,21 @@ diaspora.yml.example
 "
 fi
 
+# Use libjemalloc if it's available for better memory usage
+command -v ldconfig > /dev/null 2>&1
+if [ $? -eq 0 ]; then
+  ldconfig=ldconfig
+elif [ -x /sbin/ldconfig ]; then
+  ldconfig=/sbin/ldconfig
+fi
+if [ -n "${ldconfig}" ]; then
+  jemalloc_path=$(${ldconfig} -p | grep jemalloc | tr ' ' '\n' | grep '^/' | head -1)
+
+  if [ -n "${jemalloc_path}" ]; then
+    export LD_PRELOAD="${jemalloc_path}"
+  fi
+fi
+
 # Start Diaspora
 printf "Starting Diaspora in $RAILS_ENV mode "
 if [ -n "$PORT" ]


jhass:
I think I'd prefer checking `ldconfig -p` output first if available, then `whereis` if available and then do the debian fallback even if both a unavailable. If you'd want to get extra fancy you  could even throw `pkg-config` into the cascade.

SuperTux88:
`ldconfig -p` only works as root, while hopefully `script/server` is never executed as root. That's why I didn't use `ldconfig`.

How can I get the library path from pkg-config?

jhass:
Weird seems to work without root on the machines I have access to.

Ah nvm about  `pkgconfig` then, looks like it won't tell us the soname, just the libdir :(

SuperTux88:
> Weird seems to work without root on the machines I have access to.

Oh, I just tried that again and it works. I thought I already tried that, but maybe I had another error. It's not in the `$PATH` as user, because it's in `/sbin/`, but with the full path it works as user too. So I just changed it to use `ldconfig` and since that even works on debian, I removed the other checks, because I think we don't need them anymore.

jhass:
Oh we cannot assume it always is, rather we should check if it's in path (with `command -v`) and then fallback to sbin. Most distros merged all of these into /usr/bin by now or at least got rid of the bin/sbin distinction in the rootfs

jhass:
I doubt so. The "most" came from systemd pushing for this change :)

jhass (approved):

SuperTux88:
Merged as b2712ebf1d28df977d7a9b33cf2f802dd907535e

PR #7920, opened 2018-11-01
Fix script domain for instagram in CSP header

SuperTux88:
I just noticed that a post embedding an instagram post wasn't displayed correctly and it looks like instagram changed the domain where the script is loaded.


skyschub (approved):
lgtm.

denschub:
Aaaaand merged.

PR #7921, opened 2018-11-02
Use sudo for docker if socket is not writable

SuperTux88:
If you don't want to add your user to the `docker` group (because the `docker` group grants privileges equivalent to the `root` user, see [Docker daemon attack surface](https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface)) the script detects this now and automatically asks you for the `sudo` password.

### DIFF
--- a/script/diaspora-dev
+++ b/script/diaspora-dev
@@ -158,6 +158,16 @@ print_usage_full() {
 
 # ----- Helper functions -----
 
+dia_docker_compose() {
+  # Check permissions of docker socket and use sudo if needed
+  if [ -r "/var/run/docker.sock" ] && [ -w "/var/run/docker.sock" ]; then
+    docker-compose "$@"
+  else
+    echo "Attention: Docker socket not writable, using sudo for the docker command. You might be asked for your password now." >&2
+    sudo -E docker-compose "$@"
+  fi
+}
+
 dia_fetch_upstream() {
   # Add and fetch upstream develop branch
   if ! git remote show | grep -q '^upstream$'; then
@@ -181,12 +191,12 @@ exit_if_unconfigured() {
 
 dia_is_running() {
   # Check if diaspora container is running
-  docker-compose ps --services --filter status=running | grep -qx 'diaspora'
+  dia_docker_compose ps --services --filter status=running | grep -qx 'diaspora'
 }
 
 dia_is_db_running() {
   # Check if db container is running
-  docker-compose ps --services --filter status=running | grep -qx $DIASPORA_DOCKER_DB
+  dia_docker_compose ps --services --filter status=running | grep -qx $DIASPORA_DOCKER_DB
 }
 
 dia_get_db() {
@@ -199,14 +209,14 @@ dia_get_db() {
 dia_build() {
   if [ $# -gt 0 ] && [ "$1" == "--no-cache" ]; then nocache="--no-cache"; fi
   # Build the diaspora Docker container (diaspora:dev-latest)
-  docker-compose build $nocache diaspora
+  dia_docker_compose build $nocache diaspora
 }
 
 dia_bundle() {
   # Run bundle in order to install all gems into $DIASPORA_ROOT
   # Do not start database, not required and sometimes not yet configured
   echo "Installing gems via bundler ..."
-  docker-compose run \
+  dia_docker_compose run \
     --rm \
     --no-deps $1 \
     -e DIA_NODB=1 \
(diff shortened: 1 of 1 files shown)


script/diaspora-dev
```diff
@@ -158,6 +158,15 @@ print_usage_full() {
 
 # ----- Helper functions -----
 
+dia_docker_compose() {
```
CSammy: Would be nice to keep consistent with the rest of the functions and add a little descriptive comment to the function.

script/diaspora-dev
```diff
@@ -158,6 +158,15 @@ print_usage_full() {
...
+  if [ -r "/var/run/docker.sock" ] && [ -w "/var/run/docker.sock" ]; then
+    docker-compose "$@"
+  else
+    echo "Attention: docker socket not writable, using sudo for the docker command, you may be asked for your sudo password now!" >&2
```
CSammy: How about:

> Attention: Docker socket not writable, using sudo for the docker command. You may be asked for your sudo password now.

CSammy:
I'd love if you could fix https://github.com/diaspora/diaspora/pull/7921/files#diff-1f391931d71a69d6f451e8652b9dba97R420 by moving that line to the start of the `dia_setup` function. I only noticed after the PR was merged ...

script/diaspora-dev
CSammy: Also, I'm wondering how this works out on OS X. Is there a sudo command by default, and does it have the same semantics? I can't seem to find anything about an OS X version supporting `-E`.

goobertron: Maybe 'you may be asked for your admin password'? Some people might think they'd have a separate 'sudo password', not realising that this means the root access password. Just a thought.

SuperTux88: > Is there a sudo command by default, and does it have the same semantics?

I think there is, but no idea about the semantics. But @denschub said that on OS X docker runs as app as user and the socket has 777 permissions, so I think we don't need sudo for OS X.

> this means the root access password

Well it's not the root password, the password used with sudo is the users password. But I don't know how to write this. I think users know what they need to enter, I just didn't want to ask for a password without the user knowing why the password is asked.

> as it's a question of probability rather than permission – although, confusingly, the warning is all about 'permissions'!

It's about permissions, but sudo can be configured to work without password, but I can't know this. Also sudo caches the password for some time, so when you just entered the password two minutes ago, it doesn't ask again. That's why it's a probability if the password is asked or not.

SuperTux88: I added your changes and I just removed the "sudo", so it's only "your password" now, because that's basically what it is and I think it should be clear enough. Thanks for the feedback @CSammy and @goobertron :)

goobertron: Sorry, what I meant is that 'may be' implies someone granting you permission; 'might be' implies possibility without the influence of someone's will. It's a simple grammatical point: 'might be' works better in this instance than 'may be'; but it's a very minor point! Sorry for clouding it by making a feeble joke about permissions.

> Well it's not the root password, the password used with sudo is the users password

My understanding is that it is the password of the 'super-user' or admin of the machine (sorry, it's not actually 'root' which is a different thing), rather than an ordinary user. Of course, anyone who has installed a docker image would have to be that super-user, so it's all the same in this case. I won't try and argue technical things with you because my understanding is about 1% of yours!

> I'm wondering how this works out on OS X. Is there a sudo command by default, and does it have the same semantics? I can't seem to find anything about an OS X version supporting -E.

I can't give a rock-solid answer to this, but I've not found a problem running sudo commands in Mac OS that I've found on Stack Overflow and other help sites when I've needed to do something, and those commands are usually given for Linux users. Whether or not Mac OS supports -E I can't say; is there a way I can test this for you?

SuperTux88: > Sorry for clouding it by making a feeble joke about permissions.

Thanks for explaining, now I understand :D 

> Whether or not Mac OS supports -E I can't say; is there a way I can test this for you?

Does `sudo --help` work on mac and does it list `-E`? Or you can just try it, if `env` exists on mac, you can compare `sudo -E env` and `sudo env` and with `-E` it should be more output than without it.

But as said, I don't think that we need this sudo logic on mac.

goobertron: `sudo -help` output:

`-E, --preserve-env          preserve user environment when running command`

> compare sudo -E env and sudo env and with -E it should be more output than without it.

Yup, confirmed. Thanks for the guidance.

denschub: > Is there a sudo command by default, and does it have the same semantics

Yes, and yes. :)

CSammy: Thank you for verifying!

skyschub (approved):
:shipit:

denschub:
Merged.

PR #7922, opened 2018-11-03
Ignore invalid URLs for camo

SuperTux88:
When people only write `foobar` this would fail with `Addressable::URI::InvalidURIError: Absolute URI missing hierarchical segment: 'https://'`.


jhass (approved):

SuperTux88:
Merged as 2c56e447ed468437d30871685389971d485f951d

Issue #7080, opened 2016-09-09
mobile: allow sharing with more than one aspect/option


paskalito:
right now when try to adjust the aspect of a new share in mobile i get this:
![screenshot_2016-09-09-11-46-14](https://cloud.githubusercontent.com/assets/8012793/18383171/8c5083c6-7684-11e6-93db-e8f4c63dda87.png)

the problem is that i can eighter choose one aspect or make my post entirely public(eighter way just one optino), but if i don't want something public i usually want to share it with more than one aspect.

so its just the wrong type of selection list i get. i would need one that lets me select multiple options. - i could not find any example right now, but two other ressouces:

https://github.com/davidstutz/bootstrap-multiselect
http://www.aspdotnet-suresh.com/2014/03/jquery-multiselect-dropdown-list-with-checkboxes.html

altough it looks like they solve that inlay - i don't mind to use the OS dialog for that but like this it's just wrong. especially since d\* is all about having more control over your privacy.

habruening:
This is really only related to mobile mode. It is only a small ui issue.

sebastienadam:
I will try to solve this issue. Be indulgent to me, I am newbie to Ruby, RAILS and diaspora*

Flaburgan:
@sebastienadam if you have any questions and need help please come on #diaspora-dev on IRC freenode.

HankG:
I found a Bootstrap way of doing this but before I embarked on it I was looking at the work that Sebastien did already.  Should I be building off of what they did or is that up to me?

goobertron:
It's up to you; you can use the old code if it will save you work; but if it's easier to start from scratch, feel free.

HankG:
I have this mostly working at this point. I used the Bootstrap Mult-select control It behaves the same as the desktop site.  It supports multi-select. It properly disables options when "public" or "all aspects" is selected.  It is sending back the proper selected values or "public" or "all_aspects" as selected by the user.  Functionally this is done except for needing to run through style checker and that sort of thing...except...

The snafu is that if I keep it locked to the right then the window then the drop down menu renders off screen a bit since the left side of the menu aligns with the left side of the button versus the right side of the menu aligning with the right side.  I tried some CSS stuff but couldn't get it working.  Here is a screenshot of what it looks like on the left side now. It's not ready for getting pulled in yet because of that (unless there aren't strong opinions on the alignment moving).  I could use some help getting over that hurdle.  

[...]

PR #7602, opened 2017-08-30
Mobile: select more than one aspect for status message

sebastienadam:
Resolve issue #7080

PR #7923, opened 2018-11-03
Mobile Post Writing With Multi-select dropdown using Bootstrap

HankG:
related to diaspora/diaspora#7080

### DIFF
--- a/app/views/publisher/_publisher.mobile.haml
+++ b/app/views/publisher/_publisher.mobile.haml
@@ -15,17 +15,16 @@
           = image_tag "social-media-logos/#{service.provider}-32x32.png",
             title: service.provider.titleize, class: "service_icon dim",
             id: "#{service.provider}", maxchar: "#{service.class::MAX_CHARACTERS}"
-
-    %select{id: "aspect_ids_", class: "form-control", name: "aspect_ids[]"}
-      %option{value: 'public'}
-        = t('public')
-
-      %option{value: 'all_aspects', selected: true}
-        = t('all_aspects')
-
-      - current_user.aspects.each do |aspect|
-        %option{value: aspect.id}
-          = "· #{aspect.name}"
+      .pull-right
+        %select{id: "aspect_ids_", name: "aspect_ids[]", multiple: "multiple"}
+          %option{value: "public"}
+            = t("public")
+          %option{value: "all_aspects", selected: true}
+            = t("all_aspects")
+          %option(data-role="divider")
+          - current_user.aspects.each do |aspect|
+            %option{value: aspect.id}
+              = aspect.name.to_s
 
     .clear
     #publisher-textarea-wrapper
--- a/Gemfile
+++ b/Gemfile
@@ -111,6 +111,7 @@ source "https://rails-assets.org" do
 
   gem "rails-assets-backbone",                            "1.3.3"
   gem "rails-assets-bootstrap-markdown",                  "2.10.0"
+  gem "rails-assets-bootstrap-multiselect",               "0.9.15"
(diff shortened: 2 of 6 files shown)


app/views/publisher/_publisher.mobile.haml
```diff
@@ -15,17 +15,16 @@
...
-      - current_user.aspects.each do |aspect|
-        %option{value: aspect.id}
-          = "· #{aspect.name}"
+      .pull-right
+        %select{id: "aspect_ids_", name: "aspect_ids[]", multiple: "multiple"}
+          %option{value: "public"}
+            = t("public")
+          %option{value: "all_aspects", selected: true}
+            = t("all_aspects")
+          %option(data-role="divider")
```
diaspora-code-review: Prefer the hash attributes syntax (%tag{ lang: 'en' }) over HTML attributes syntax (%tag(lang=en))

HankG: I haven't figured out how to get this to work with the requested syntax.  It seems that in an option block HAML is only allowing very specific fields.  If someone knows how to fix this I'll implement that. I need this field to create the divider.  Alternatively we could take the divider out if we don't want to ignore the style problem and don't care about the divider.

HankG:
Here are the screenshots on iOS and Android.  Sorry for the Android being a screen-of-a-screen.  I have to get a data transfer process off the Android device I scrounged.  I figured something was better than nothing though.  I've confirmed that the scroll behavior is as expected with a touch screen device.  Unfortunately neither draw the scrollbars very prominently unless the user is scrolling.  I don't think there is anything I can do about that though :(
photo nov 03 14 55 29
photo nov 03 15 52 14

SuperTux88:
I wonder why you choose to use `bootstrap-multiselect` instead of the approach which was used in #7602 (which used the same dropdown was we have everywhere else). While it is up to you if you want to continue an existing PR or start from scratch (as goob commented [here](https://github.com/diaspora/diaspora/issues/7080#issuecomment-435159438)), I actually like the old PR better (at least from a users point of view). It has the same look and feel as everywhere else (including the world icon for public or the lock icon for limited, which is missing here).

While your solution here has some "problems":
  * The checkboxes are too low on both android and iOS
  * It uses system checkboxes which don't adapt to the diaspora theme (white checkboxes on the dark theme)
  * When de-selecting an item, it has a white background which is then difficult to read on the dark theme because light-gray text on white background (I only have an android device and only tested that there, I don't know about iOS)
  * Your dropdown is smaller, so more difficult to click on a touch device
  * I don't like the margin on the left of the checkboxes
  * If multiple aspects are selected, the text isn't translated

While some points are only nitpicking, I still think this is a bit of work. While the other PR basically was working and finished and the only complaint was, that [it was a lot of code duplication](https://github.com/diaspora/diaspora/pull/7602#pullrequestreview-102612377), so only about the code (which I didn't even had a look at in this PR), the functionality and look and feel was good.

So maybe it is still worth to have a look again at the other PR and check if you can reduce the code duplication instead of fixing all the other things in this PR?

(Also, sorry that I didn't comment earlier, I didn't have the time to checkout the state of the old PR earlier)

HankG:
@SuperTux88 I looked at the volume of code in the original PR and it seemed like a lot of code changes compared to the Bootstrap approach. The translation problem is because I used a direct string, which I need to fix.  I have some control over the formatting of the controls by using some styling but the styling differences didn't bother me at all.  I'm a daily user of both the desktop and the mobile sites so I'm not speaking of it as someone who is unaccustomed to both systems.

HankG:
Haven't heard a peep so I'll take a look at resurrecting the old PR rather than continue on with this one...

PR #7924, opened 2018-11-04
Sharpen small and medium thumbnails

bennettgoble:
I noticed profile icons appeared fuzzy on Diaspora. Resizing images [often results in apparent blurring](http://www.imagemagick.org/Usage/filter/#blurring). This commit adds sharpening to 50 and 100px thumbnail sizes in order to compensate for perceived loss of detail. It uses [unsharpening](https://en.wikipedia.org/wiki/Unsharp_masking) parameters of `1.5x1+0.7+0.02` referenced in a section dedicated to the subject on  [ImageMagick's resize page](http://www.imagemagick.org/Usage/resize/#resize_unsharp).

### Examples
|Before|After|
|-|-|
| hq_small_default | hq_small_unsharp |
| sistine_small_default | sistine_small_unsharp |
| author_small_default |  author_small_unsharp |

|Before|After|
|-|-|
| hq_medium_default |  hq_medium_unsharp |
| sistine_medium_default |  sistine_medium_unsharp |
| author_medium_default | author_medium_unsharp |

#### Original Images

[1](https://user-images.githubusercontent.com/151138/47958995-5e4c9b00-df95-11e8-8562-d97fbe749ed9.png), [2](https://user-images.githubusercontent.com/151138/47958997-699fc680-df95-11e8-948b-c91c0e2ded15.jpg), [3](https://user-images.githubusercontent.com/151138/47958998-6dcbe400-df95-11e8-929a-cc468038060d.png)

### DIFF
--- a/app/uploaders/processed_image.rb
+++ b/app/uploaders/processed_image.rb
@@ -20,10 +20,10 @@ class ProcessedImage < CarrierWave::Uploader::Base
   end
 
   version :thumb_small do
-    process resize_to_fill: [50, 50]
+    process resize_to_fill: [50, 50, combine_options: {unsharp: "1.5x1+0.7+0.02"}]
   end
   version :thumb_medium do
-    process resize_to_limit: [100, 100]
+    process resize_to_limit: [100, 100, combine_options: {unsharp: "1.5x1+0.7+0.02"}]
   end
   version :thumb_large do
     process resize_to_limit: [300, 1500]


SuperTux88 (approved):
Looks great, thanks :+1:

SuperTux88:
Merged as 935ff66c962166c7f24ae8cfc80a899a6ea2acb7

Thank you for your first pull request :cookie:

Flaburgan:
Thanks for this improvement! 
However that means we should regenerate all the old images, shouldn't we?

nivardus:
It would be a great if existing images were reprocessed, though not strictly necessary. What is the precedent?

1. Is there currently an easy way for pod admins to kick off bulk reprocessing?
2. Should this be done automatically as part of a migration to a new version or left up to instance owners due to its comparatively large cost?

Flaburgan:
1 -> no, not as far as I know
2 -> in a migration, delegated to sidekiq in a worker with a very low priority? We can write the command in the changelog, so the podmins who don't want to run it don't run it.

SuperTux88:
> However that means we should regenerate all the old images, shouldn't we?

I don't think we need to regenerate all the thumbnails, they are only thumbnails. Maybe it's useful for the profile pictures, but even then, there are many inactive profiles, so I'm not sure if we should do it for all profiles. And active users who aren't happy with their thumbnails can just reupload it. But you can create a rake-task if you like.

PR #7926, opened 2018-11-05
Rename "toggle mobile" to "switch to light version"  or "switch to full version"

Flaburgan:
From [this discourse discussion](https://discourse.diasporafoundation.org/t/impressions-and-ideas-from-a-new-user/2049?u=flaburgan), it's been pointed that 

>There are two versions of the website that can be accessed via mobile, “desktop” and “mobile”. Confusingly, the desktop version changes when viewed on a smaller screen, so it appears different than when viewed on a desktop. There is no clear indicator for a new user which is which, and switching requires pressing a button labeled “Toggle Mobile” but doesn’t tell you which mode you’re in. 

As I answered there, the distinction between "desktop" and "mobile" is gradually disappearing as we're switching to a unique, responsive version. But still, we're not there yet.

This PR changes the wording of the "toggle mobile" link which now explicitly tells to which version we're switching: "full" (formerly the desktop one) or "light" (formerly the mobile one). Those new names clearly indicates the left purposes of the remaining versions until there's only one. I was wondering if "version" is better than "mode" though, what do you think?


HankG:
Is that really going to solve the confusion mentioned in the thread though? I think It was the responsive design creating something that sort of looks like the mobile site but wasn't the mobile site that was the source of the confusion.  As someone who has done mobile development and uses both it hadn't even occurred to me that we have a mobil-ish view in the responsive design until it was pointed out on that thread.

denschub:
I am unsure. This feels like bikeshedding, but to me, "light version" kinda reminds me of dark modes in applications, and I'm not sure if I would expect a mobile optimized version behind this link. I'll leave this up for some other opinions... :)

Flaburgan:
Yes, this is a wording issue, maybe @goobertron can help here. Would "lightweight" or "slight" be better? Also, maybe we can use "mode" instead of "version" to avoid a possible confusion between the versions (0.7.8.0) of the software?

goobertron:
> Would "lightweight" or "slight" be better? Also, maybe we can use "mode" instead of "version"

I'll get my thesaurus out!

Flaburgan:
@goobertron I should have time to work on this tomorrow morning, what do you suggest?

goobertron:
Oh, I'm sorry; there's been so much going on that I've not been able to think of anything better. I'm seriously tempted to recommend that the links be changed to 'drink slimline tonic' and 'eat full-fat yoghurt', but I will try to think of a proper term that works.

I'm exhausted this evening after another busy day, but I'll try to put some energy into this tomorrow morning. Please ping me again then if you don't hear from me, as I'm liable to forget.

goobertron:
Is there a real reason not to use 'mobile' for that version. I know it's not only available on mobile devices, but it was designed for them. My current favourite is:

> Use mobile mode
> Use standard mode

You could use 'version' instead of 'mode' if you prefer, but 'mode' seems to suit it a little better.

You could use 'Switch to' instead of 'use', which is perhaps slightly clearly; I suggested 'use' only because it's much shorter. If 'switch to' fits nicely in the text area (and is not going ot be too long in other languages – French is usually the benchmark for long translations!) feel free to use it.

Hope that helps. I'm not entirely satisfied with my suggestions, but they do at least give the correct indication, and I hope they convey the sense you want.

Flaburgan:
>Is there a real reason not to use 'mobile' for that version?

Well, the discourse discussion is more clear about it, but in short, as the "desktop" version is usable with a mobile, using "light" and "full" is a way to explain the differences between the two versions.

goobertron:
I understand that, and I've followed that discussion. But does that really block using 'mobile' for a version designed specifically for mobile devices? I think it will be understandable to the most people, and the fact that they can use the full 'desktop' version on their mobile is clear from the fact that there is a link to switch to this version. And 'mobile' conveys that the 'light' version is optimised for mobiles. It's only one person who queried the use of 'mobile' for the 'light' version, and they seemed satisfied once you had explained it to them; and I'm not aware of a major problem in diaspora\*'s community about that term.

I honestly can't think of a satisfactory term for that version. 'Light' is too confusing (suggesting a light colour theme); any other terms I can think of (reduced, light-weight, and so on) have negative connotations.

Flaburgan:
You are right that we're trying to answer two different questions: which version am I using? And What's the difference between the versions? 
The first question is answered thanks to the text now changing to indicate to what version we're switching. I'm not sure the second one is answered with your suggestion but maybe I overthink it. I like Lightweight, it doesn't sound negative in my mind and clearly indicates that this is not about the theme imo.

Flaburgan:
So, are "Switch to lightweight mode" and "Switch to standard mode" okay?

goobertron:
I still think that 

>Use mobile mode
Use standard mode

is better, for the reasons I've given above.

- I don't really think that 'lightweight' clearly answers your second question: what is the difference between the versions? The 'lightweight' mode was created for mobile devices, and that is still its primary purpose while it remains in existence.
- Also, while 'switch to' is good, I worry that it (and particularly some translations of it) will take up too much space. But you can test that.

Your suggestions are at least grammatically correct; and, apart from the reservations I've mentioned, I see nothing wrong with them.

So far it's been just your opinion and mine; it would be good to hear from others. I'm not going to make any more objections to your suggestions; so, if no one else has any opinion, or cares, go ahead.

Flaburgan:
Alright so let's go with mobile and standard modes :)

About switch to, the current french translation of "Toggle mobile" is *"Activer/désactiver la version mobile"* and it doesn't break the UI so... we should be good ;)

Flaburgan:
(Doing it, I still think that a user in "standard mode" using a mobile phone so in the responsive view will find the "Switch to mobile mode" confusing as what he's looking at already looks like a mobile mode. But that's already in improvement so I'm going with that at the moment.)

goobertron:
> I still think that a user in "standard mode" using a mobile phone so in the responsive view will find the "Switch to mobile mode" confusing as what he's looking at already looks like a mobile mode.

I imagine that most people in that situation would realise that they're not currently using the mode designed for mobiles, and might benefit from trying it. :-)

Let's see; if we get complaints, we can revisit this.

Flaburgan:
So, ready for review, the current state is "Switch to standard mode" or "Switch to touch-optimized mode".
I also moved the link at the bottom of the menu in the standard mode as it is better placed there.

SuperTux88 (approved):
Well, I don't know anything better and I think the current state is the best so far, so let's try this. :)

PR #7928, opened 2018-11-08
Change the port exposed by Docker to 3000

### LINKED ISSUE #7927: Dev-Docker: Mismatch between the exposed port and `url` config
The dev docker container exposes diaspora to port `8080`, but the default `url` config is `http://localhost:3000/`. This causes issues, for example when uploading images, as the URL will be prefixed with `http://localhost:3000/`, which ain't workin.

We have two options here: One is to change the port we run to 3000, the second one is to modify `diaspora.yml`.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65487626-dev-docker-mismatch-between-the-exposed-port-and-url-config?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

skyschub:
I see no reason to use port `8080`, and since changing the `diaspora.yml` is more painful, let's just expose port `3000` to the world. :)

Fixes #7927.

### DIFF
--- a/docker/develop/docker-compose.yml
+++ b/docker/develop/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       - dia_data_tmp:/diaspora/tmp
       - dia_data_bundle:/diaspora/vendor/bundle
     ports:
-      - 8080:3000
+      - ${DIASPORA_DOCKER_PORT:-3000}:3000
     depends_on:
       - "${DIASPORA_DOCKER_DB}"
 


SuperTux88:
I added a env variable to override the default port, because usually I have my local setup running on port 3000 while I want to run tests in docker for example. And docker needs the port even when I don't want to access the webinterface, so I need to be ably to change it to another free port.

denschub:
I would approve this as this is basically @SuperTux88's PR now, as he changed the implementation, but...

screenshot 2018-11-18 at 12 54 15

SuperTux88:
Merged as a6655c187d33c9a902cfeaa0f8dec21ab768c4f7

PR #7929, opened 2018-11-08
Bump ruby version to 2.6, drop support of 2.3 and 2.4

Flaburgan:
I'm opening this to see if travis is green with ruby 2.5. If so, I'll switch ruby to 2.5 on diaspora-fr.org and use it for a week or two and report if any problem is found.


Flaburgan:
Travis is green with 2.5.3, I'm switching my pod to it. Here is [the changelog](https://www.ruby-lang.org/en/news/2017/12/25/ruby-2-5-0-released/) btw.

denschub:
I'm running 2.5 since December of last year, without any issues.

Flaburgan:
Nice!

HankG:
If we release 0.8.0.0 in August or something like that I'd say we should experiment with that now and see how Diaspora does with it.

HankG:
> 2.6 has been released since I open this PR. Do we want to skip 2.5 and support 2.4 and 2.6 directly?

If we release 0.8.0.0 in August or something like that I'd say we should experiment with that now and see how Diaspora does with it.

SuperTux88:
> I'd say we should experiment with that now

Usually we need to wait at least a few weeks until most gems have updates where the new version had some breaking changes, before we can start testing it. But since the next major is still a bit away, we can probably start testing 2.6 in some weeks (and skip 2.5 if everything works). :)

denschub:
Let's skip that and make the PR upgrading to 2.6. Also, let's drop 2.4 as well, because it's practically end-of-life.

skid9000:
Ruby 2.6.1 seems to work fine on our pod.

Flaburgan:
I changed to 2.6 and travis is green :)

jhass (approved):

SuperTux88:
There is even ruby 2.7 already, did anybody already do tests with it? Maybe we could upgrade directly to it and skip 2.5?

Flaburgan:
I saw it yeah but I didn't try it yet.

jhass:
I didn't try but I while I would expect it to work for there being not too many changes, it added a ton of deprecation warnings in preparation to ruby 3, and the ecosystem really didn't catch up to that yet. So it wouldn't look pretty for sure.

denschub:
2.7 is working fine on Geraspora since its release, so we *could* do it. That being said, as Jonne said, there are a ton of deprecation warnings, which might seriously confuse people. We could ask podmins to set `RUBYOPTS="-W0"`, but I'm not sure if this is a great idea.

denschub:
I might have to change my opinion. Production runs fine on 2.7, but development is not. There are a lot of interesting spec failures, mainly caused by frozen strings and our prehistoric `active_model_serializers` version. In theory, we need to update `active_model_serializers` to 0.10.x, or get rid of it entirely. Both of which are ... not fun.

denschub:
Yeah, that's a good point. Let's stick with 2.6 for now, as 2.7 causes significant logspam. @Flaburgan, do you mind rebasing this?

denschub:
As discussed with folks on IRC, let's merge this into `next-minor`. We can't quickly release 0.8.0.0 as the API and the account migration are not yet ready for prime time, but we'd like to get this change out to people soon, so this is fine.

Merged as 7193099902486429b4fcbbb567bbca2f3cbef6d4.

PR #7930, opened 2018-11-10
Update gems

SuperTux88:

PR #7931, opened 2018-11-17
Update gems

SuperTux88:
Second part of #7930 with only the gems for `develop` branch.


skyschub (approved):
Approved as soon as tests are green.

SuperTux88:
Tests are green, merged, thanks for the review @denschub :cookie:

PR #7934, opened 2018-11-22
Update sq translations

ujdhesa:
Please pull the lastest updates to sq translations.


HankG (approved):
I do not speak or read this language so I cannot comment on the new text.  I did confirm that the YAML structure of the keys is identical to the corresponding English in this commit version.  I'd say it therefore looks good.

HankG:
PS I made the above assessment that this language doesn't need to go through the standard WebTranselateIt process due to the fact that Blesnik is specifically called out as the maintainer of this translation.  Just in case someone looking to do translations in other languages finds this PR, the process for most languages is to not directly edit the YAML files.  See: https://wiki.diasporafoundation.org/Contribute_translations

ujdhesa:
Mislead from the head of those strings, which goes:

- zero:
- one:
- other:

I've translated those as dealing with the number of users doing something, rather than representing a list of them.

SuperTux88:
Merged in 0.7.8.0 as 5766a08b22fe8659d854acb8ea3da3dd266ce2d4, 440775f3b0cd87503ef5eceaac3c5c1df41ecfb2, 5ee85721f70abd4d32a6971905825b6617a77970 and 6c2b3a7b11464710c142c2f6bee263ceff747438.

Thanks @ujdhesa for the update :cookie: and thanks @HankG for the review :cookie:

PR #7943, opened 2018-12-04
Fix photos gallery when too many thumbnails by adding a scroll-x

### LINKED ISSUE #7942: New Photos View wrapping preview icons
I thought we tested this case but maybe we didn't.  User over on Diasp.org running 0.7.8.0 is having this artifact in their photo viewer.  https://diasp.org/posts/10957047
image

Flaburgan:
Fixes #7942


Flaburgan:
See https://github.com/diaspora/diaspora/issues/7942#issuecomment-444279781 for a screenshot of how it looks like.

Flaburgan:
I've got several users who complained about that issue, so a review of it would be cool :+1:

HankG (approved):

HankG:
> I've got several users who complained about that issue, so a review of it would be cool +1

I thought I already did that back when I experimented with it too...my bad...

skyschub (approved):

PR #7944, opened 2018-12-28
Speed up public stream and remove obsolete indexes

CSammy:
Speeds up the public stream by a factor of 1000 on a production PostgreSQL database, it's lightning fast now. MySQL shows improvement as well.

PR #7951, opened 2019-01-01
Speedy stream query

koehn:
I did [some analysis](https://explain.depesz.com/s/XFb9) of the `/stream` query against my Postgres 11 database with @CSammy's index on `posts.created_at`. The main problem with the query as written is that the Postgres (10/11) optimizer runs a full table scan on the `posts` table, not using the `created_at` index, because the index would be slower than just scanning the table. As you can see, on my relatively-capable machine (32GB RAM, 16GM free for caching, 8GB dedicated to postgres), it's taking [2800 milliseconds](https://explain.depesz.com/s/XFb9#l10) to do the scan. 

I recognized that there's no point in doing a full table scan on `/posts`, since I'm only returning the next fifteen posts behind whatever date is asked for, and with 3.8 million posts in my database, it's looking at a lot of additional posts for nothing. 

So I went to work modifying the query to look at only posts created in the past thirty days. Why thirty? It seemed like a reasonable tradeoff between minimizing the number of posts the query needs to consider, and assuring that the query would actually return the requested number of results with high probability. This involved changing the query to use a `between` expression instead of a `<` expression. 

The [results were impressive](https://explain.depesz.com/s/vBba). The new expression resulted in an ten-fold performance gain, taking only [250 milliseconds](https://explain.depesz.com/s/vBba#l9). It's faster because the DB knows based on its column statistics that using the index will be radically faster than a table scan, so it uses it. The query still returns the requisite 15 results. Pushing the "next" button slides the window back, just as before. Narrowing the query to 7 days reduced execution time to 70ms, again with no loss of posts. 

**Upside:** Radically faster stream performance. Query execution time is bounded by the density of posts over the window rather than number of posts in the pod's database. 

**Downside:** It's possible that looking at a thirty-day window of posts might be too small for some users to see all fifteen posts, and were there a thirty-day "gap" of posts between windows, the UI would be convinced there are no more posts to see for that stream. It's unlikely on the user's, but it's possible for brand-new users. The consequence is that they wouldn't see posts created before the "gap," but all posts after that would be displayed (and quickly!). 

The downside can be managed in a few ways:
1. **Acceptance.** It's a free social media platform, not a perfection-driven system. Users are unlikely to scroll very far back. Other social media platforms work this way. 
2. **Pod-level customization.** Have a property in `diaspora.yml` that allows the podmin to set the window size based on the performance they want. 
3. **User-level customization.** Have a user-settable property in the database, bounded by reasonable ranges, that allows each user to define the window size to make the speed/precision tradeoff that's best for them. 

I'm happy to amend the PR as needed.

While the PR as written only applies to the `/stream` API; for the others the window is effectively infinitely large, causing the DB to do the table scan to get its results. The same technique could be applied to the other APIs as well.


app/models/post.rb
```diff
@@ -110,7 +110,11 @@ def self.excluding_hidden_content(user)
   end
 
   def self.for_a_stream(max_time, order, user=nil, ignore_blocks=false)
-    scope = self.for_visible_shareable_sql(max_time, order).
+    self.for_a_ranged_stream(Date.parse('01-01-2008'), max_time, order, user, ignore_blocks)
```
diaspora-code-review: - Style/RedundantSelf: Redundant `self` detected.
- Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/post.rb
```diff
@@ -110,7 +110,11 @@ def self.excluding_hidden_content(user)
...
+  end
+
+  def self.for_a_ranged_stream(min_time, max_time, order, user=nil, ignore_blocks=false)
+    scope = self.for_visible_shareable_sql(min_time, max_time, order).
```
diaspora-code-review: - Style/RedundantSelf: Redundant `self` detected.
- Layout/DotPosition: Place the . on the next line, together with the method name.

lib/diaspora/shareable.rb
```diff
@@ -77,11 +77,15 @@ def from_person_visible_by_user(user, person)
       end
 
       def for_visible_shareable_sql(max_time, order, limit=15, types=Stream::Base::TYPES_OF_POST_IN_STREAM)
-        by_max_time(max_time, order).order(table_name + ".id DESC").where(type: types).limit(limit)
+        by_min_max_time(Date.parse('01-01-2008'), max_time, order).order(table_name + ".id DESC").where(type: types).limit(limit)
```
diaspora-code-review: - Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
- Metrics/LineLength: Line is too long. [129/120]

lib/diaspora/shareable.rb
```diff
@@ -77,11 +77,15 @@ def from_person_visible_by_user(user, person)
...
       end
 
-      def by_max_time(max_time, order="created_at")
-        where("#{table_name}.#{order} < ?", max_time).order("#{table_name}.#{order} DESC")
+      def for_ranged_visible_shareable_sql(min_time, max_time, order, limit=15, types=Stream::Base::TYPES_OF_POST_IN_STREAM)
```
diaspora-code-review: Metrics/LineLength: Line is too long. [124/120]

lib/evil_query.rb
```diff
@@ -86,7 +86,7 @@ def community_spotlight_posts!
     end
 
     def ids!(query)
-      fetch_ids!(query.for_a_stream(@max_time, @order), 'posts.id')
+      fetch_ids!(query.for_a_ranged_stream(@max_time - 30, @max_time, @order), 'posts.id')
```
diaspora-code-review: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

denschub:
As @CSammy mentioned in your very own post, there might be ways to make all stream queries faster without building something that breaks UX for people([1]). This solution looks like something tailored for one specific case with specific assumptions in mind, and it will not address any other issues we have (like, for example, tag streams if looking at very infrequent tags. or new users adding contacts only rarely sharing public posts without yet seeing any private posts).

If we have other options (and unless @CSammy tells me his ideas do not work out, or do not yield improvements, I assume that is the case), the potential issues do not seem worth it. And even if we decide that limiting the base data set is the only way to go, then we'd need actual discussion on that. Tagging as such.

---

[1]: Yes, you *might* be right and it will never cause any issues. Given our history, I doubt that, and I see no reason to say "yeah, let's ignore some amount of data, ignore the correctness, and ignore predictability because look how fast this is" if we don't need to.

denschub:
Also, [here](https://wiki.diasporafoundation.org/Styleguide#Automatic_local_review) is documentation on how to run style checks before submitting a PR.

HankG:
One idea I have off the cuff, may be bad but since we are discussing why not bring it up...

What if there was a bounded and unbounded option.  The bounded has the new lower bound window and the unbounded is the original.  The streams can query first for bounded and if it returns values it returns the result set.  However if it gets nothing it can then re-query unbounded and then return the results from that.  If 80-90% of the time these queries are returning something because there isn't a gap problem we'd get the 100x speed up and not have the gap problem. The only time it'd truly be duplicative is when you hit the actual end of the result set.  Since the first query takes milliseconds it's not the worst price to pay...

denschub:
#7944 improved the public stream by a factor of almost 1100 on my machine. We worked a bit on tag streams last week in person, and already achieved a 28000ms to 14ms reduction in some early tests for tag streams, but with some gotchas, so more work has to be done on that. That is without introducing any artificial limits, bounds, or anything else that makes the situation worse than it currently is.

> It's not changing the behavior of tag streams

Correct, which is why I was saying that adding a workaround that brings new implications is not a really good idea if it only tackles one very isolated case, with no re-usability in other cases. Things like indexes are fine, they don't make behavior different, but if the multi stream is suddenly time limited, while other streams are not, ... oh well. Better performance is nice, but better performance on the cost of consistency and predictability is not.

> I question the utility new users get out of seeing old posts

So, if someone adds a contact they know from other channels, and can clearly see some posts on their profile, albeit older posts (which is fine, since the contact told them they only post private most of the time, and thus there are no posts visible yet), and they still have zero posts in their stream after adding the contact, don't you think that's a horrible experience right there, and a source for many issue reports and confused users ahead? That's just one of the multiple cases I could think of where your workaround would fail horribly. Let's not do such things unless we absolutely, without alternative, have to. Which as of right now, we certainly don't.

The same goes for maintaining two sets of queries, one bounded and one unbounded "just in case". That's something that adds tons of complexity, and if we think about implementing something like that, we are, most likely, not working on the right problem. 

Anyway, a GitHub PR is not the right place for these discussions. The bigger picture "should we introduce potential fallacies for enhanced performance" is a topic for Discourse, not below a PR. But there certainly is no harm in having those arguments here, as well.)

> I'll update run them and make updates accordingly.

Also check the guides on running tests. Your PR is burning like a christmas tree on travis, and christmas-season is over. :)

goobertron:
If a solution based on this approach is used (and I see some debate about that already), would this act as a solution to the following:

> **Downside:** It's possible that looking at a thirty-day window of posts might be too small for some users to see all fifteen posts, and were there a thirty-day "gap" of posts between windows, the UI would be convinced there are no more posts to see for that stream. It's unlikely on the user's, but it's possible for brand-new users.

If a 30-day scan returns fewer than 15 posts, scan the previous 30 days. Repeat until there are 15 posts to display. As pointed out, the stream for the majority of users would be populated by posts more recent than 30 days old, but this gives scopes for populating the stream for other users, albeit with a slightly longer scanning time.

Flaburgan:
I agree with Dennis that changing the behaviour for the end user is not a good idea.

However, we do have a big performance problem, and I would almost call it a bug. So, about

>The same goes for maintaining two sets of queries, one bounded and one unbounded "just in case". That's something that adds tons of complexity, and if we think about implementing something like that, we are, most likely, not working on the right problem

sometimes, complex code needs to be written. Yes, it will probably be a bit harder to maintain, but if you're thinking about the hours lost everyday by everyone using diaspora* when you sum all the 10 seconds waiting at each stream loaded, the cost in $$ and the energy wasted. it is worth it in my opinion.

But if you have another idea, please share it :) (About that, this is clearly a technical discussion so I thought it would belong to github. But maybe not in a PR yes. Should we continue in a "Streams are too long to load" issue?)

HankG:
> sometimes, complex code needs to be written. Yes, it will probably be a bit harder to maintain, but if you're thinking about the hours lost everyday by everyone using diaspora* when you sum all the 10 seconds waiting at each stream loaded, the cost in $$ and the energy wasted. it is worth it in my opinion

I don't believe my suggestion creates substantial code complexity.  Maybe I'm wrong with how deep some of those call stacks get in the EvilQueries and such though. 

As was stated elsewhere maybe the ideal thing is to have this in a Discourse discussion.  That's especially true if @CSammy  and @denschub are actively working on a solution as well.  PR's in that case don't allow a side-by-side discussion which is going to be warranted.

HankG:
> > Also check the guides on running tests.
> 
> Will do, as we get closer to a solution. Not much point doing it now if you're not going to accept it anyway, is there?

The problem is every time you do the push it's throwing the same errors.  It may be best to close this PR, pick up the conversation in a discourse channel, and then after more discussion and work open up a new PR.

koehn:
Per @HankG , closing to pick this up in Discourse.

denschub:
> How is the work that you're going (with gotchas) any different than what I'm doing here?

The gotchas I was talking about were implementation-specific gotchas (i.e. more complex AR usage), not user-visible changes.

PR #7955, opened 2019-01-06
Downgrade bootstrap-switch

### LINKED ISSUE #7940: Visibility of extended profile always shows limited
When changing the visibility of your extended profile to public the switch will still show as limited though your extended profile will be visible to public.

SuperTux88:
It looks like 3.3.4 is broken :( See https://github.com/Bttstrp/bootstrap-switch/issues/691 for example and I also found other issues where people had problems with 3.3.4 and downgrading helped, so maybe there are more issues we didn't notice yet.

The [changelog](https://github.com/Bttstrp/bootstrap-switch/blob/master/CHANGELOG.md#334) is not really helpful at all, it only contains one thing, and I trusted it when I upgraded it and didn't expect something to break. However the changes between these minor versions seems to be [bigger](https://github.com/Bttstrp/bootstrap-switch/compare/v3.3.3...v3.3.4) and I don't understand what exactly broke it. Also 3.3.4 was released two years ago (and is still the newest version), but `bootstrap-switch-rails` was just released half a year ago, so it only hit us with the [last big gems update](https://github.com/diaspora/diaspora/pull/7930). So this doesn't look very active anymore, so in long term we should probably replace this.

Fixes #7940


HankG:
Wow!  I figured they'd have a unit test to pick up something like that.  That's a shame :(.  Good catch though!

skyschub (approved):

PR #7761, opened 2018-03-31
Added a scroll bar Aspects Dropdown

gayathriravipati:
#7696 

Added a scroll bar to the Dropdown

Issue #7696, opened 2018-01-13
UI: add to Aspects Dropdown needs autoscroll


paskalito:
or else at some point you cannot create a new aspect anmore for some new contact. ( and i guess also not put them into the ones you've already mafe ) 

![bildschirmfoto vom 2018-01-13 15-48-59](https://user-images.githubusercontent.com/8012793/34907289-7e69652c-f87c-11e7-8d72-f46c3f19d839.png)

i hope one understands despite the massive blurring..

Flaburgan:
Wow, I didn't imagine one could have that much aspects.

paskalito:
same thing is actually also the case when using the https://share.diasporafoundation.org/about.html thing just more drastic ( since the window  has just minimal height at opening)
![bildschirmfoto vom 2018-01-13 16-50-44](https://user-images.githubusercontent.com/8012793/34907667-5532f1f4-f882-11e7-9aab-988bec85249b.png)

PR #7958, opened 2019-01-10
Aspects Scrollbar 

HankG:
This relates to Issue #7696 continuing the work @Gayathrir1666 did in PR #7761 .  With this change we get scrollbars on Aspect buttons/menus on:

- User Search Page
- Contacts Page
- Post Publisher
- Hovercard


app/assets/stylesheets/aspects.scss
```diff
@@ -21,3 +21,7 @@
     }
   }
 }
+.aspect_membership {
```
diaspora-code-review: Selector `aspect_membership` should be written in lowercase with hyphens

HankG: When I change this to aspect-membership several cucumber and Jasmine tests fail.  One Jasmine test it was very straight forward since it was referencing it directly. The others I can't tell why the style change would create a change in behavior.  The errant behavior that is being flagged, such as not getting the notification when adding a person to the aspect or mis-count of aspects, doesn't seem to be happening in the User interface and again I'm not sure why changing the style name and its subsequent references throughout is creating the test failures.

SuperTux88: It's OK to ignore these, if it's not just a small change at a few places. Sometimes stuff like this needs to be changed in many places because it's used everywhere (in tests to select stuff, in js to select stuff, at other places in css, ...), and then it isn't worth to do it. If you still like to change this, please do it in a separate commit :)

HankG:
Assuming this gets approved I'll squash my commits into one but leave the original squashed commit from the original contributor so that there is attribution to their contribution in the commit log unless there is an objection.

Siedlerchr:
@HankG  You can use CoAuthored by https://help.github.com/articles/creating-a-commit-with-multiple-authors/

HankG:
> HankG You can use CoAuthored by https://help.github.com/articles/creating-a-commit-with-multiple-authors/

I never heard of that before.  Thanks!

denschub:
Superseded by #8213.

PR #7964, opened 2019-01-16
Update factory_girl_rails to 4.9.0

utkarsh2102:
This bumps the factory_girl_rails gem from 4.8.0 to 4.9.0.
#3855


HankG:
Factory Girl and Factory Girl Rails 4.9 is two years old so this change doesn't hurt my head necessarily.  However it did remind me that I wanted to look into switching to the renamed version of the project Factory Bot.  There is a transition guide which looks relatively painless to do.  Is that something we should explore here or on a separate commit/PR?

https://github.com/thoughtbot/factory_bot/blob/4-9-0-stable/UPGRADE_FROM_FACTORY_GIRL.md

SuperTux88:
It doesn't make much sense to upgrade to factory_girl 4.9.0, because the only difference in this version is a warning that is printed to upgrade to factory_bot. So if anything, it should be upgraded to factory_bot directly, which already was started in #7657, but there were still open points which then weren't continued.

utkarsh2102:
@HankG, @SuperTux88, would it make sense if I update it to `factory_bot`?
Or should I close this PR, altogether?

SuperTux88:
> would it make sense if I update it to `factory_bot`?

Yes, but please have a look at the discussion in the other PR I linked to see what that includes:

* try to not add new rubocop offenses, so not just do a simple search and replace (see for example [here](https://github.com/diaspora/diaspora/pull/7657/files#diff-b709bad60390eff76d8e893b0e31911dR76) for changed indentations). If you want to also fix existing offenses on the lines that are changed because of the replacement anyway, that would be also fine to me, but it's not needed.
* create shorthand methods like `create(:factory)` for `FactoryBot.create(:factory)`, which abstracts away the underlying framework. So we don't need to touch all these lines again if we want to replace `factory_bot` with something else in the future.

utkarsh2102:
>     * create shorthand methods like `create(:factory)` for `FactoryBot.create(:factory)`, which abstracts away the underlying framework. So we don't need to touch all these lines again if we want to replace `factory_bot` with something else in the future.

Could you point me to a particular example?
Where would that go?

SuperTux88:
> Could you point me to a particular example?

See the `factory_bot` documentation: https://www.rubydoc.info/gems/factory_bot/file/GETTING_STARTED.md#RSpec

You can just include `FactoryBot::Syntax::Methods`

denschub:
Let's close this PR. We absolutely need to do this, but this PR basically only updated the Gem quite a while ago, which is now causing conflicts in the Lockfile.

Issue #7910, opened 2018-10-29
Add line mentioning diaspora* on the splash page


goobertron:
It turns out that a lot of confusion that people coming over from G+ have had understanding that diaspora* is a federated network with lots of different nodes is because the line on the splash page that they reach says, in big letters:

> # Welcome to pluspora
> #### The online social world where you are in control

We show only the pod name there, which does not necessarily have any relation to the diaspora\* name.

To remove at least this point of confusion, I'd like to add the following line between them:

> #### Part of the diaspora\* federated network

Linked to https://diasporafoundation.org/

So that the page reads:

> # Welcome to %{pod_name}
> #### Part of the [diaspora\* federated network](https://diasporafoundation.org/)
> #### The online social world where you are in control

Everyone OK with this?

Flaburgan:
@gooberton do you want to jump in and do it? You now have a docker image to easily set up a dev environment ;)

goobertron:
Not for a couple of weeks, at least. There's too much else going on. Maybe after that, if no one else has. I guess it would be a good incentive for me to try that Docker image to finally get my local install working again!

PR #7966, opened 2019-01-21
#7910 Clarify pod is part of diaspora in splash screen

ghost:
This is my very first contribution, so I decided to start with a very simple change from issue #7910 .

Some points about the change:

-The guideline asks to include tests for every change. But it didn't seem right to include a test for a simple change in html. To be sure, I searched in the features folder and didn't find tests for this kind of addition, so it seemed right to leave tests aside for this one.

-I am not skilled at CSS, so probably someone will want to make changes in the style. Bellow follows a picture of the rendered html.

-It only makes sense to clarify the pod is part of diaspora* when the name of the pod is not diaspora* already, like it is in development. So I added a conditional.

-I don't know how the workflow for i18n translations goes. I added the proper english text, and I could add for pt-BR also, but I thought it would be better to wait for some feedback on this.

Example of the rendered html:

screenshot from 2019-01-20 22-58-22

### DIFF
--- a/app/views/home/default.haml
+++ b/app/views/home/default.haml
@@ -5,7 +5,9 @@
         .col-md-8.text-center
           %h1= t("home.default.headline", pod_name: pod_name)
           - if pod_name != "diaspora*"
-            %h2.clarification-header= t("home.default.clarify")
+            %h2.part-of-diaspora
+              != t("home.default.part_of_diaspora",
+                diaspora_site_link: link_to(t("home.default.diaspora_site_link"), "https://diasporafoundation.org/"))
           %h2= t("home.default.byline")
         .col-md-4.login-form
           = render partial: "sessions/form", locals: {mobile: false, resource: User.new, resource_name: :user}
--- a/app/views/home/default.haml
+++ b/app/views/home/default.haml
@@ -4,6 +4,8 @@
       .row
         .col-md-8.text-center
           %h1= t("home.default.headline", pod_name: pod_name)
+          - if pod_name != "diaspora*"
+            %h2.clarification-header= t("home.default.clarify")
           %h2= t("home.default.byline")
         .col-md-4.login-form
           = render partial: "sessions/form", locals: {mobile: false, resource: User.new, resource_name: :user}
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -42,6 +42,16 @@
     padding: 15px;
   }
 
+  .clarification-header {
+    background-color: $grey;
+    border: 1px solid $border-grey;
+    display: inline-block;
+    margin-bottom: 10px;
+    margin-top: 10px;
+    padding: 5px;
+    text-align: center;
+  }
+
   .login-form {
     fieldset { background: none; }
 
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -42,14 +42,12 @@
     padding: 15px;
   }
 
-  .clarification-header {
-    background-color: $grey;
-    border: 1px solid $border-grey;
-    display: inline-block;
-    margin-bottom: 10px;
-    margin-top: 10px;
-    padding: 5px;
-    text-align: center;
+  .part-of-diaspora {
+    font-style: italic;
+
+    a {
+      color: $white;
+    }
   }
 
   .login-form {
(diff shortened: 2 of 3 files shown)


HankG:
I honestly never thought about this before.  This could be an interesting tweak.  If we are making this change we may want to make part of that hyperlinked to the diaspora project page while we are at it.  

I agree that the styling needs some work but I see what you are trying to do in terms of making it a biline style thing.  It just doesn't feel right to me for some reason I can't put my finger on (I'm not good at graphic design).

goobertron:
Thank you for doing this! I think it will significantly help reduce the confusion for people who have just signed up.

A few points:

- I'd like there not to be a background to the new text. That looks a little clunky. I'd suggest no background. You could try putting it in italics, perhaps, to differentiate it from the line below.
- Thanks a lot for adding the conditional. Not knowing the syntax for that is the reason I was unable to fix this myself.
- It would be really good if 'diaspora\* federated network' was a link to the project site. I'll paste the code for this in line comments.
- It would be good to call the new string something like `part_of_diaspora` rather than `clarify`, to 'clarify' a bit more what the string is.
- Translations for any language other than English should be added via [WebTranslateIt](https://webtranslateit.com/en/projects/3020-diaspora/) rather than in PRs. (There are three exceptions to this, [listed here](https://wiki.diasporafoundation.org/Contribute_translations#Github), but all other translating is done via WTI.)
- I can't say with absolute certainty, but I'm pretty sure tests wouldn't be required for this PR, which is some HTML/ERB, CSS and a bit of i18n.

Thanks again.

app/views/home/default.haml
```diff
@@ -4,6 +4,8 @@
       .row
         .col-md-8.text-center
           %h1= t("home.default.headline", pod_name: pod_name)
+          - if pod_name != "diaspora*"
+            %h2.clarification-header= t("home.default.clarify")
```
goobertron: Try `t('home.default.part_of_diaspora', diaspora_site_link: link_to(t('diaspora_site_link_text'), 'https://diasporafoundation.org/')).html_safe`

(I think this is correct, but please test it!)

Flaburgan:
Hi @davidmoseler and thank you for that first contribution!

Do you have time to tackle the remarks made by @goobertron or should we handle them ourselves? (@denschub you assigned @SuperTux88 on this during the hackathon but I don't remember if it was to finish the PR. I can do it if you're fine with that).

app/views/home/default.haml
```diff
@@ -4,6 +4,9 @@
       .row
         .col-md-8.text-center
           %h1= t("home.default.headline", pod_name: pod_name)
+          - if pod_name != "diaspora*"
+            %h2.part-of-diaspora
+              != t("home.default.part_of_diaspora", diaspora_site_link: link_to(t('home.default.diaspora_site_link_text'), 'https://diasporafoundation.org/'))
```
diaspora-code-review: - Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
- Line is too long. [158/120]

Flaburgan:
No answer from @davidmoseler at the moment so I did the remarks made by goob. The only point which bothers me a bit is that there is visually no difference of the link at the moment:

Screenshot_2019-05-08 diasporaaa

That changes when it is hovered (it got a text underline) but is that enough?

goobertron:
@Flaburgan how about using the `$blue` colour used in-app for links? And disabling underlining on-hover.

SuperTux88:
I think it's OK as it is, adding color would make it too "noisy". The important information is already written there, the link is only bonus. If you want to know what diaspora is, you can click it, or click the link in the footer or just google it.

Flaburgan:
>@Flaburgan how about using the $blue colour used in-app for links? And disabling underlining on-hover.

That was the default and it was too much. 

>I think it's OK as it is, adding color would make it too "noisy". The important information is already written there, the link is only bonus. If you want to know what diaspora is, you can click it, or click the link in the footer or just google it.

I kinda agree, that's why I pushed it like that, but I'm still not 100% satisfied. If a better styling idea comes to my mind, I'll open a follow-up pull request.

goobertron:
> I think it's OK as it is, adding color would make it too "noisy". The important information is already written there, the link is only bonus. If you want to know what diaspora is, you can click it, or click the link in the footer or just google it.

Good point. The information in the text alone is what matters.

SuperTux88 (approved):

SuperTux88:
Merged as 6826e89a95f5ab6e3e6c0bffce0d68a1daf6594a and e04ddd0bccba2b54912af59ae27b88653f9d3ad0

Thanks to everybody who worked on this :cookie: :cookie: :cookie: :cookie: :cookie:

Issue #7967, opened 2019-01-21
clarify between multiple reshares


BEG65:
Currently, when I reshare a post I get notifications for it along the lines of "Reshare by (me)".

If I reshare more than one post, they are ALL called "Reshare by (me)".

This makes the notification list completely incomprehensible. I'd like to see the original title of the original post be retained in the "Reshare..." notification. So that I'd get something like

"(person) liked your reshare of (original title)" and "(person) commented on your reshare of (original title)"

As it is, I wind up posting links to items I would otherwise reshare in order to get some sanity in my notification list. 

I looked through the issues and googled around for any potential discussion of this, but if there's any, I missed it. Thanks!

Flaburgan:
Thank you for opening an issue @BEG65, that's indeed a very good suggestion!

goobertron:
This looks the same as the long-standing #5066 – feel free to add your suggestions there!

davidmoseler:
I am working on this one. The main feature is basically implemented. There is a problem though. What should happen if the original post is deleted?

There is a spec (which I broke) that reads "falls back to the title, if the root post was deleted". But the spec implementation doesn't make much sense, because, before, the text was only "Reshare by ${author}", thus not mentioning the deleted root post, and nothing changed before. Now that I made a change, I need to update de spec.

So, which behaviour would be the best one when the post root is deleted?

BEG65:
Obviously I can't speak for everyone, but myself I would be satisfied if original post title lost its link upon deletion. It could also change to say "reshare of (post deleted)" directly. I don't know that this happens terribly often, so I think its fine if the link to the original post goes away. 

What I'm not clear on is whether a reshared post is a copy of the original post or if it's actually the original post, but showing only the likes and comments generated by my reshare of it? If the original post is deleted then I think it's fine if the notification says something like "reshare of (deleted post)".

If I'm sharing multiple posts that all get deleted, t here's some other issue going on at that point ;-)

BEG65:
Goobertron, issue #5066 deals with reshares of a reshare... Seems to be slightly different (but yes related). Thank you!

MasterOfTheTiger:
@BEG65 
> Goobertron, issue #5066 deals with reshares of a reshare... Seems to be slightly different (but yes related). Thank you!

[...]

PR #7968, opened 2019-01-21
#7967 clarify multiple reshares [WIP]

ghost:
Adressing the issue #7967 .

I would say that there is still some work to do. But it would be nice to have revisions already.

Some points of attention:

-I changed the "reshare_by" logic to "reshare_of". This means that, instead of refering to the reshared post as a "Reshare by ${user}#, it now reads "reshare of **original post title**".

-The **original post title** is a link to the original post. If the original post is deleted, this link turns into plain text.

-Specs where updated.

-I need to make some experimentation on the UI to see how things behave. But I imagine that, when someone likes the reshared post, the notification reads something like "(person) liked your post reshare of (original title)". At first I would add logic to the translations so that "post" turned into "reshare", instead of "post reshare", but it turns out to work well with simpler code in the english language, as a "post reshare" means the same thing as "reshare", but less code is required. But this might be a problem in other languages.

-I should make manual tests and think about what happens when someone reshares a reshared post.

-Although all tests passed, I should pay more attention if code elsewhere could be broken by the changes. Maybe a test missed something important.

### DIFF
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -28,10 +28,7 @@ module NotificationsHelper
   end
 
   def opts_for_post(post)
-    {
-      post_author: html_escape(post.author_name),
-      post_link: post_link(post)
-    }
+    {post_author: html_escape(post.author_name), post_link: post_link(post)}
   end
 
   def opts_for_mentioned(mentioned)
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -30,10 +30,7 @@ module NotificationsHelper
   def opts_for_post(post)
     {
       post_author: html_escape(post.author_name),
-      post_link:   link_to(post_page_title(post),
-                           post_path(post),
-                           data:  {ref: post.id},
-                           class: "hard_object_link").html_safe
+      post_link: post_link(post)
     }
   end
 
--- a/app/helpers/posts_helper.rb
+++ b/app/helpers/posts_helper.rb
@@ -9,7 +9,12 @@ module PostsHelper
     if post.is_a?(Photo)
       I18n.t "posts.show.photos_by", :count => 1, :author => post.status_message_author_name
     elsif post.is_a?(Reshare)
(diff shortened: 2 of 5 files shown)


HankG:
This looks like a good start to what you are trying to achieve.  I'd like to play around with this to see what it looks like side by side but appreciate the way you are attempting to tackle this.  Thanks!

> -Although all tests passed, I should pay more attention if code elsewhere could be broken by the changes. Maybe a test missed something important.

Looking at the CI it seems there are Cucumber tests that need to be addressed.  It's bombing out on the behavior of a deleted post.

goobertron:
Thanks for working on this.

I envisaged the notification being

> X has interacted on your reshare Original post title

rather than

> X has interacted on your post Reshare by you

That means detecting whether the post concerned is a reshare; and, if so, replacing 'post' with 'reshare'. Otherwise the notification will read 

> X has interacted on your post reshare of Original post title

which looks somewhat clunky and is longer, so might not fit well in the notifications drop-down or page.

Is that something you could do? It looks as though you are already detecting whether or not the post is a reshare, so hopefully this would be feasible.

davidmoseler:
It is feasible. I just need to create a new key in I18n for the interactions on reshared posts. A new key is probably necessary because the wide range of grammars in the languages supported by diaspora.

davidmoseler:
I thought about using a variable like "post_type" to distinguish between "post" and "reshare". I even wrote the code, but then I erased it for some reason. Maybe because I will need to change all calls to all of these translations anyway. But maybe this is the best way to do it.

goobertron:
As there is already an `if post.is_a?(Reshare)` query, I think all you'd need to do is to tell it there to change the word 'post' to 'reshre' in the notifications strings.

That is, if that line 12 in `app/helpers/posts_helper.rb` isn't used in any other parts of the app. If it is, doing this might have undesired effects elsewhere. Hope you can find that out easily.

denschub:
We had some discussions about reshares and identified more need for further discussions. :) We will be having a long chat about how reshares work in the future, and we will make sure to include the points raised here. Thanks much! :)

PR #7978, opened 2019-02-10
Fixed typo and added missing word

ebner:


goobertron:
Hello! Thank you very much for your contribution.

Translations for diaspora\* are not handled through GitHub PRs. We use WebTranslateIt instead. Here is the [list of Swedish translations](https://webtranslateit.com/en/projects/3020-Diaspora/locales/en..sv/strings).

If you'd like to fix these for us, feel free to request an invite there (you should see a link on the WTI site) and you can join our Swedish translation team. You will see that there are 23 strings yet to be translated, and several not yet verified, so if you have some extra time, you could help our Swedish community by doing some of those as well.

ebner:
OK, I will request an invite. I'm closing my PR.

PR #7998, opened 2019-04-27
Add a manifest.json file as a first step to make diaspora* a PWA

Flaburgan:


Changelog.md
```diff
@@ -11,6 +11,7 @@
 ## Features
 * Add client-side cropping of profile image uploads [#7581](https://github.com/diaspora/diaspora/pull/7581)
 * Add client-site rescaling of post images if they exceed the maximum possible size [#7734](https://github.com/diaspora/diaspora/pull/7734)
+* Add a manifest.json file as a first step to make diaspora* a Progressive Web App [#7997](https://github.com/diaspora/diaspora/issues/7997)
```
SuperTux88: Please don't add changelogs entries (this can be merged to the `next-minor`, so it's even in the wrong section ;) ).

app/views/layouts/application.html.haml
```diff
@@ -15,6 +15,8 @@
     = render "head"
     = translation_missing_warnings
 
+    %link{rel: "manifest", href: "/manifest.json"}
```
SuperTux88: This can be added to [_head.haml](https://github.com/diaspora/diaspora/blob/develop/app/views/application/_head.haml) which is used for both, desktop and mobile.

Also, can/should we remove the old `mobile-web-app-capable` now?

https://github.com/diaspora/diaspora/blob/d6225daaf3503f5935133e0e9f109c7bee485e6d/app/views/application/_head.haml#L7

public/manifest.json
```diff
@@ -0,0 +1,16 @@
+{
+  "short_name": "diaspora*",
+  "name": "diaspora*",
```
SuperTux88: The old "Add to Home" used the page title, which used [`pod_title`](https://github.com/diaspora/diaspora/blob/9bab794/config/diaspora.yml.example#L356-L358) from the `diaspora.yml`. Should we use a controller to generate the json so we can also use the dynamic title?

SuperTux88 (changes requested):

SuperTux88 (approved):

SuperTux88:
Merged as 2073791ceeca5c63554ff58b1d17bc1767e53e79 and 5b6cef0679e7d69d6bbdcca6b1a808c92efdfb39

Thanks :cookie:

cmrd-senya:
So cool to have a way to create a diaspora shortcut on mobile now! Thank you!

app/controllers/manifest_controller.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class ManifestController < ApplicationController
+  def show # rubocop:disable Metrics/MethodLength
+    render json: {
+      short_name:       AppConfig.settings.pod_name,
+      name:             AppConfig.settings.pod_name,
+      description:      "diaspora* is a free, decentralized and privacy respectful social network",
```
goobertron: It's a bit late now, but I would suggest 'privacy-respecting' rather than 'privacy respectful'.

SuperTux88: It's never to late, I just changed that now in 035b6f39fcc91f5f18c1d2cc6357d1f9a37001fd :+1:

goobertron: Thanks (thumbs up).

PR #8003, opened 2019-04-28
Enable paranoid mode for devise

### LINKED ISSUE #8001: Enable Devise's paranoid mode
See https://github.com/diaspora/diaspora/pull/7776#issuecomment-379807297, and we really should do it.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/73444076-enable-devise-s-paranoid-mode?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SuperTux88:
fixes #8001

### DIFF
--- a/app/views/devise/passwords/new.haml
+++ b/app/views/devise/passwords/new.haml
@@ -9,10 +9,6 @@
         = AppConfig.settings.pod_name
 
     = form_for(resource, as: resource_name, url: password_path(resource_name), html: {class: "form-horizontal block-form"}, autocomplete: 'off') do |f|
-      - unless resource.errors.empty?
-        %legend
-          %i
-            = t('devise.passwords.new.no_account') # this is an error message and should not be displayed as a legend
       %fieldset
         %label#emailLabel.sr-only{for: "user_email"}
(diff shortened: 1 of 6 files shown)


skyschub (approved):

goobertron:
Apologies for a slightly off-topic comment, but with the change in text here, would it be worth (in a separate PR) adding a warning to people when they sign up that if they don't provide a genuine email address, there will be no way to send a password recovery email at a later date?

SuperTux88:
Merged as 607659939d1113c2840e7b36eced7d6f0c6c637f, thanks for the review @denschub :cookie:

SuperTux88:
@goobertron that doesn't change, that is already the case (without valid email it was never possible to recover your password), and since we want to validate email addresses in the future, it also wouldn't be a problem anymore.

PR #8004, opened 2019-04-29
Minor 2fa form style improvements

skyschub:
It bugs me more than I dare to admit. :) Here's a before/after:

Screenshot 2019-04-29 at 23 41 00

### DIFF
--- a/app/assets/stylesheets/forms.scss
+++ b/app/assets/stylesheets/forms.scss
@@ -93,5 +93,10 @@ textarea {
     }
 
     ::placeholder { text-transform: uppercase; }
+
+    p {
+      margin-top: .5rem;
+      text-align: center;
+    }
   }
 }
--- a/app/views/sessions/two_factor.html.haml
+++ b/app/views/sessions/two_factor.html.haml
@@ -15,7 +15,7 @@
     %fieldset
       %label.sr-only#otp-label{for: "otp_attempt"}
         = t("two_factor_auth.input_token.label")
-
+      %i.entypo-lock
       = f.text_field :otp_attempt,
       type: :text,
       placeholder: t("two_factor_auth.input_token.placeholder"),


SuperTux88 (approved):
looks sooooo much better now :D :cookie:

SuperTux88:
Merged as ef137f09f2717b61970555c87c414aa367d9a1c4 and 3c06bb2f4c1b2c8135a6f83359301ac0e117b93e, thank you :cookie:

goobertron:
Anal!

But it does look better.

Is the change from 'Forgot your password?' to 'Create an account' at the bottom intentional, or significant? There's also 'Create account' top right. I'm guessing that as 2fa is always about authenticating an existing account, these are not necessary.

@SuperTux88 was so fast off the mark that he approved and merged this while I was typing! I'm too slowwwwwww....

SuperTux88:
> Is the change from 'Forgot your password?' to 'Create an account' at the bottom intentional, or significant?

This didn't change in this PR, maybe he changed something in his local `diaspora.yml` between the screenshots, because they depend on settings.

> There's also 'Create account' top right. I'm guessing that as 2fa is always about authenticating an existing account, these are not necessary.

You're right, they don't make much sense on that page, since you already have an account, and you also already typed in your password (and you can't recover 2FA), so it can be removed.

denschub:
> This didn't change in this PR, maybe he changed something in his local diaspora.yml between the screenshots, because they depend on settings.

Ah, sorry. As Tuxxi rightfully suggested, the before screenshot was taken from Geraspora, and the right one was my local dev pod. Geraspora has registrations closed, so there is no "Create account" link, but my local dev pod has emails disabled, so the reset password feature is disabled. Should've cut that out of the screenshot. :)

Good catch about those links, though! Thanks. <3

PR #8005, opened 2019-04-29
Remove password reset and sign up link below two factor form

SuperTux88:
They don't make sense on that page, because at this stage, the user already has an account and also has already entered their password.

As pointed out by @goobertron in https://github.com/diaspora/diaspora/pull/8004#issuecomment-487766771


skyschub (approved):

SuperTux88:
Merged as ecda6eccf6b87a499960b8caf9d869b961132637, thanks for the review @denschub :cookie:

PR #8006, opened 2019-04-30
Use password to disable 2FA instead of a token

SuperTux88:
Using token doesn't make much sense when you can generate new tokens right below.


jhass (approved):
Good thinking :)

SuperTux88:
Merged as 54fd4846c0607ea1a3de07f65e9a39920cf4ba8a, thanks for the review @jhass :cookie:

PR #8010, opened 2019-05-06
Fix archive edge cases

cmrd-senya:
These are two fixes for the archive import process which fix the issues which I discovered while I was trying to import the archive from the old pod provided by @Flaburgan. With this two fixes the archive import doesn't fail.

### DIFF
--- a/lib/archive_importer/own_relayable_importer.rb
+++ b/lib/archive_importer/own_relayable_importer.rb
@@ -2,6 +2,8 @@
 
 class ArchiveImporter
   class OwnRelayableImporter < OwnEntityImporter
+    class NoParentError < RuntimeError; end
+
     def entity
       fetch_parent(symbolized_entity_data)
       entity_class.new(symbolized_entity_data)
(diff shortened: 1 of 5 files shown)


jhass (approved):

lib/archive_importer/own_relayable_importer.rb
```diff
@@ -19,6 +21,8 @@ def fetch_parent(data)
         break entity_class::PARENT_TYPE if entity_class.const_defined?(:PARENT_TYPE)
       }
       entity = Diaspora::Federation::Mappings.model_class_for(type).find_by(guid: data.fetch(:parent_guid))
+      raise NoParentError if entity.nil?
```
SuperTux88: Shouldn't relayables where the parent can't be fetched be invalid during validation? And didn't you remove invalid entities from the collection during validation, so they shouldn't be imported anymore? I think it would be interesting what has gone wrong in that case, does the remove not work at all? Or is it only in an edge case where it doesn't work?

SuperTux88: And to answer my own question here, I finally found what may can cause a problem during import even when it should already have been validated during the validation: The validation checks if a parent for relayables already exists in the database, or exists in the archive. But a parent can exist in the archive (so validation succeeds), but then when it fails to import (for example a reshare), then when the relayable tries to find it, it isn't there.

PR #8012, opened 2019-05-09
fix old photos without a remote_photo_path or remote_photo_name

jhass:


db/migrate/20190509125709_fix_missing_remote_photo_fields.rb
```diff
@@ -0,0 +1,9 @@
+class FixMissingRemotePhotoFields < ActiveRecord::Migration[5.1]
```
diaspora-code-review: Style/FrozenStringLiteralComment: Missing magic comment `# frozen_string_literal: true`.

SuperTux88 (approved):
:cookie:

db/migrate/20190509125709_fix_missing_remote_photo_fields.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class FixMissingRemotePhotoFields < ActiveRecord::Migration[5.1]
+  def up
+    Photo.where(remote_photo_path: nil).each do |photo|
+      photo.write_attribute(:unprocessed_image, photo.read_attribute(:processed_image))
+      photo.update_remote_path
```
SuperTux88: Oh, just something that came to my mind now, should we add a migration model here with that method, in case we remove it from the real model but the migration still exists?

jhass:
I thought about that but figured I skip it, given this stuff wasn't touched in years and the possibility that we forget about the migration when changing it as it is when dropping it :)

SuperTux88:
Merged and uplifted to 0.7.11.0 as 2d23a2601e744719f0696db209275cd85fdbb140, thank you :cookie:

PR #8014, opened 2019-05-11
Don't encrypt the OTP secret

SuperTux88:
It doesn't add any security to have this encrypted, but it adds complexity for podmins, because they need to backup the key.

Also, we had problems with suddenly getting `OpenSSL::Cipher::CipherError` during testing on local dev pods (not on production yet), were only manually changing the DB fixed it (and looking at the issues of [attr_encrypted](https://github.com/attr-encrypted/attr_encrypted), other people had problems with that too), so I think it's safer not to have the secrets encrypted.


skyschub (approved):
Thanks for addressing this! <3

SuperTux88:
Merged and uplifted to 0.7.11.0 as 165b8f4f6e4e0a0c7a4577c1f3a9ffcb3d8a7a09, thanks for the review @denschub :cookie:

db/migrate/20190511150503_decrypt_two_factor_secret.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def twofa_encryption_key
+    if AppConfig.heroku?
+      ENV["TWOFA_ENCRYPTION_KEY"]
+    else
+      key_file = File.expand_path("../../config/initializers/twofa_encryption_key.rb", File.dirname(__FILE__))
+
+      if File.exist? key_file
+        require key_file
+        File.delete(key_file)
```
jhass: Uh, kinda dangerous... Well I'm too late :D

SuperTux88: Well, I thought if I should make it safer (wait with deletion until after the successful migration, but couldn't find a nice solution without making everything a lot more complex). But since this only affects pods which were on `develop` (since this was never in a release), it shouldn't affect many pods (most pods are probably already upgraded), and if it really fails, their podmins should be able to fix it in the DB and tell the few people who already enabled 2fa to re-enable it (or maybe they even have a backup of the config folder since there are other important files in there too).

PR #8020, opened 2019-05-20
Don't relay others relayable on archive import

SuperTux88:
Make sure we send nothing during archive import.

### DIFF
--- a/lib/archive_importer/entity_importer.rb
+++ b/lib/archive_importer/entity_importer.rb
@@ -11,7 +11,7 @@ class ArchiveImporter
     end
 
     def import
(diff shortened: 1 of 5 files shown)


jhass (approved):
LGTM  in general.

I would consider swapping the name `import`here for something more specific like `skip_relaying` in order to not have the receiver part know about an import semantically and encourage  more purposee built  flags in the  future rather than one big one which you don't know what it all does anymore at some point.  But that's no  dealbreaker for me in case you disagree.

SuperTux88:
Yes, thought about that and wasn't sure if something more generic is better, so in case we need to do something else different for import, we could reuse the same option instead. But I think you're right, a more specific option is probably better, so I changed it. :+1:

PR #8029, opened 2019-06-11
Harmonize markdown titles sizes

### LINKED ISSUE #7757: Reduce font-size used to display titles in users posts
We only kept the `h1`, `h2` and `h3` titles in markdown posts a while ago and I think this is enough. However, the font-sizes used to display these titles are, in my opinion, too big. The h1 really looks like yelling out loud. So I only use the h3 title for my main title. But that means I can't have any subtitles!
There is plenty of room between the h3 size and the normal font size, I would suggest reduce the size used to display those titles so they all can be used.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56389500-reduce-font-size-used-to-display-titles-in-users-posts?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Flaburgan:
Fixes #7757

### DIFF
--- a/app/assets/stylesheets/markdown-content.scss
+++ b/app/assets/stylesheets/markdown-content.scss
@@ -16,6 +16,30 @@
     }
   }
 
+  h1 {
+    font-size: 2.7rem;
+  }
+
+  h2 {
+    font-size: 2.3rem;
+  }
+
+  h3 {
+    font-size: 2rem;
+  }
+
+  h4 {
+    font-size: 1.8rem;
+  }
+
+  h5 {
+    font-size: 1.6rem;
+  }
+
+  h6 {
+    font-size: 1.4rem;
+  }
+
   .img-responsive {
     display: inline;
   }


skyschub (changes requested):
Please use `rem`, not absolute font sizes. (Since our `html` font size is `10px` as defined by bootstrap, it's actually simple: Divide all numbers by 10.)

Flaburgan:
I thought about it but I was wondering if the user content size should be linked to the interface size, and I was not sure about it. But if you think it should then sure I'll do it.

Flaburgan:
Or maybe we want to use `em` not `rem` so we can separate the interface size from the user content size if needed?

denschub:
No, we don't want to use `em`, because then it suddenly matters where headlines are contained in, and it's possible that the private conversations for example have a different font-size because someone set a size in one of it's parents. There never should be a reason to make the user content larger while leaving the rest of the UI as small as it is, but having everything enlarged would make sense. In addition, all browser's "text-only" zoom features only enforce a new font-size on the root, and that's why rem are such a nice thing to use.

Flaburgan:
Thank you for that detailed explanation, I updated the PR.

skyschub (approved):
Thanks :)

SuperTux88:
Merged as 78538d034a3fe90295bf9c4514359b2a882d3d13

Thank you :cookie: and thanks @denschub for the review :cookie:

Issue #8018, opened 2019-05-20
Do not show getting started after account import


jhass:
Make sure to set the flag to false when creating a user from an account migration

Flaburgan:
Why do we want to set it to something else than what was in the DB?

SuperTux88:
There is no "what was in the DB" since it is a "new user", so it sets it to default true (as for new users), but the default doesn't know that this is an imported user, so in that case it should be set to false, which this issue is about.

Flaburgan:
I was talking about the DB of the "from" pod. I thought this info could be in the archive, so can be restored. But I agree in 99.99% it will be false.

SuperTux88:
I think settings aren't exported, and also can't be fully imported, because your target pod maybe doesn't have the same themes as your old pod.

Danielshow:
Hi, @Danielshow here, I am a newcomer and will like to contribute to this project by working on this.

jhass:
Cool, just ask if  you need any pointers :)

Danielshow:
Sure @jhass

theRealNG:
@jhass & @SuperTux88 Can one of you point me where the exact feature to import users from another pod is, i'm unable to find it.

SuperTux88:
The import is currently only a rake task [`accounts:migration`](https://github.com/diaspora/diaspora/blob/develop/lib/tasks/accounts.rake), and since it is neither completely finished nor released yet, it isn't well documented yet.

But @Danielshow also said he wants to work on this, so you should talk to each other before you both start working on this issue.

theRealNG:
@Danielshow are you working on this?

Danielshow:
@theRealNG You can pick it up, finding it hard to set up the projects. I will pick another once I am done

Flaburgan:
@Danielshow use `script/diaspora-dev` to set up a development env easily with docker.

Danielshow:
Thanks, @Flaburgan, I will try that now

PR #8036, opened 2019-06-26
8018 Do not show getting started after account import

theRealNG:
Copy 'getting_started' value for user when he is being migrated. 
https://github.com/diaspora/diaspora/issues/8018

### DIFF
--- a/lib/archive_importer.rb
+++ b/lib/archive_importer.rb
@@ -25,10 +25,12 @@ class ArchiveImporter
       email strip_exif show_community_spotlight_in_stream language disable_mail auto_follow_back
     ]
     data = convert_keys(archive_hash["user"], allowed_keys)
+    # setting getting_started to false as the user doesn't need to see the getting started wizard
     data.merge!(
       username:              attr[:username],
       password:              attr[:password],
       password_confirmation: attr[:password],
+      getting_started:       false,
       person:                {
         profile_attributes: profile_attributes
       }
--- a/spec/lib/archive_importer_spec.rb
+++ b/spec/lib/archive_importer_spec.rb
@@ -155,6 +155,7 @@ describe ArchiveImporter do
       expect(archive_importer.user.language).to eq("ru")
       expect(archive_importer.user.disable_mail).to eq(false)
       expect(archive_importer.user.auto_follow_back).to eq(true)
+      expect(archive_importer.user.getting_started).to be_falsey
 
       expect(archive_importer.user.profile.first_name).to eq("First")
       expect(archive_importer.user.profile.last_name).to eq("Last")


spec/lib/archive_importer_spec.rb
```diff
@@ -139,7 +139,8 @@
           "show_community_spotlight_in_stream" => false,
           "language"                           => "ru",
           "disable_mail"                       => false,
-          "auto_follow_back"                   => true
+          "auto_follow_back"                   => true,
+          "getting_started"                    => false,
```
diaspora-code-review: Style/TrailingCommaInHashLiteral: Avoid comma after the last item of a hash.

spec/lib/archive_importer_spec.rb
```diff
@@ -155,6 +156,7 @@
       expect(archive_importer.user.language).to eq("ru")
       expect(archive_importer.user.disable_mail).to eq(false)
       expect(archive_importer.user.auto_follow_back).to eq(true)
+      expect(archive_importer.user.getting_started).to eq(archive_hash['user']['getting_started'])
```
diaspora-code-review: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

Flaburgan:
Hi @theRealNG and thank you for your work! Did you check that `getting_started` was actually part of the exported archive? I'm not sure it is included.

jhass:
Looking at https://github.com/diaspora/diaspora/blob/develop/app/serializers/export/user_serializer.rb I don't think it's imported.

I think it's also useful to default to `false` for archive's  that don't contain it.

Flaburgan:
>I think it's also useful to default to false for archives that don't contain it.

Yep, we definitely don't want to break compatibility with archives generated until now, so either add it to the archive but don't fail if not present by setting it to false, or just always set it to false, that looks fine to me.

theRealNG:
Thinking of going with always setting it to false

theRealNG:
@Flaburgan & @jhass Pushed changes that always set the `getting_started` as false while importing a user profile.

jhass (approved):
This is fine to me, @SuperTux88 wished for transfering the status of the source account so he should say if this is a no-go for him for some reason :)

lib/archive_importer.rb
```diff
@@ -25,10 +25,13 @@ def create_user(attr)
       email strip_exif show_community_spotlight_in_stream language disable_mail auto_follow_back
     ]
     data = convert_keys(archive_hash["user"], allowed_keys)
+    # setting getting_started to false as we are importing an existing user profile
+    # so the user need not see the getting_started wizard
```
Flaburgan: I'm not a native speaker but I feel like "the user doesn't need to see the getting started wizard" sounds better.

Flaburgan (changes requested):

SuperTux88:
> @SuperTux88 wished for transfering the status of the source account so he should say if this is a no-go for him for some reason :)

Oh no, I didn't wish that, I even wrote that setting it to always false would be a solution, so that's perfectly fine for me. :+1: Fla wanted it copied from the old pod.

SuperTux88 (approved):
LGTM :cookie: 

(maybe squash your commits to a single one, ask if you don't know how to do that, but we can also do that when merging it)

Flaburgan (approved):

SuperTux88:
Merged, thanks for your contribution @theRealNG :cookie: 

(I also just saw that you developed that change on your `develop`-branch, you should do that on a separate branch next time, so `develop` can stay clean to branch of it)

PR #8037, opened 2019-06-29
Remind the username in the reset password e-mail

Flaburgan:


app/views/devise/mailer/confirmation_instructions.markerb
```diff
@@ -1,4 +1,4 @@
-<%= t('devise.mailer.welcome', :email => @resource.email) %>
+<%= t('devise.mailer.welcome', :email => @resource.username) %>
```
Flaburgan: I feel like username is warmer than email here. I didn't change the key to avoid all translation to change.

app/views/devise/mailer/reset_password_instructions.markerb
```diff
@@ -1,11 +1,13 @@
-<%= t('devise.mailer.hello', :email => @resource.email) %>
+<%= t('devise.mailer.hello', :email => @resource.username) %>
```
Flaburgan: Ditto.

app/views/devise/mailer/unlock_instructions.markerb
```diff
@@ -1,4 +1,4 @@
-<%= t('devise.mailer.hello', :email => @resource.email) %>
+<%= t('devise.mailer.hello', :email => @resource.username) %>
```
Flaburgan: Ditto.

app/views/devise/mailer/confirmation_instructions.markerb
SuperTux88: I think we can change the key and then change all translations via the API. I just need to find out how, but if I remember correctly, we have already done this before.

denschub: We might be fine [just renaming the key before pushing the new files inside the WTI interface](https://webtranslateit.com/en/docs/faq#rename-a-key), but if that does not happen to affect translations, doing the same via the API is something we can do with not too much work.

@Flaburgan might changing the key here, and changing the key in `en.yml` as well? :)

config/locales/devise/devise.en.yml
```diff
@@ -30,6 +30,7 @@ en:
         someone_requested: Someone has requested a link to change your password. If
           it was you, you can do this through the link below.
         change: Change my password
+        then_connect: Once changed, you will be able to connect using your username "%{username}" and your chosen password
```
denschub: I don't really like the word "connect" here, and would much rather have "sign in". Also, I'd probably expand this a little bit, so something along the lines of

> After setting a new password, you will be able to sign into diaspora* again using your username "%{username}" and your new password.

skyschub (changes requested):
I like the idea, especially since we sometimes receive questions by confused users who receive notification mails but can't remember their username. :)

Minor comments here.

Flaburgan:
I completely forgot about that pull request sorry. Remarks are fixed.

SuperTux88 (changes requested):
Besides this one thing, it looks good to me. Also please squash your commits when you're done.

Flaburgan:
Fixed and squashed ;)

SuperTux88 (approved):

PR #8039, opened 2019-07-02
Remove commentable functionality from photos

SuperTux88:
This is not (and as far as I know, was never) used. If we want to make standalone photos commentable, we can always add it back, but it would also need to change federation for it to work, because comments support only posts there. But for now it makes the code cleaner and easier to remove it.


skyschub (approved):
Clean up all the things!

PR #8041, opened 2019-07-02
Cleanup pending photos which were never posted with cronjob

SuperTux88:
Only delete photos older than a day, so we don't delete photos for posts which were uploaded 10 minutes ago and the author is still writing the post for it.

### DIFF
--- a/app/workers/cleanup_pending_photos.rb
+++ b/app/workers/cleanup_pending_photos.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module Workers
+  class CleanupPendingPhotos < Base
+    sidekiq_options queue: :low
+
+    def perform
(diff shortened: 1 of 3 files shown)


skyschub (approved):

denschub:
@SuperTux88 specs are failing for MySQL 😞

SuperTux88:
... and travis is finally green (after 3 hours of waiting for all builds to start and finish ...).

It wasn't MySQL, because on this PC I still have a MySQL dev env, but tests were green ... but I'm an idiot and I don't want to talk about it ;)

denschub:
Ohhh, that was a good one! My approval is still valid. :)

SuperTux88:
Merged as df4e79b842fa16b370c8c4b07b82bdd328f3ef7e

Thanks @denschub for the review :cookie:

PR #8056, opened 2019-09-08
Handle missing parent when deleting comment

jhass:
When a post is deleted, a comment retraction might race with a post retraction,
causing the comment deletion to raise because of a missing parent.

Fixes
```
NoMethodError: undefined method `update_comments_counter' for nil:NilClass
  from app/models/comment.rb:50:in `block in <class:Comment>'
  from lib/diaspora/federation/receive.rb:177:in `retraction'
  from config/initializers/diaspora_federation.rb:112:in `block (3 levels) in <top (required)>'
  from app/workers/receive_private.rb:13:in `block in perform'
  from app/workers/receive_base.rb:11:in `filter_errors_for_retry'
  from app/workers/receive_private.rb:10:in `perform'

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -47,8 +47,8 @@ class Comment < ApplicationRecord
   end
 
   after_destroy do
-    self.parent.update_comments_counter
-    participation = author.participations.find_by(target_id: post.id)
+    parent&.update_comments_counter
+    participation = author.participations.find_by(target_id: commentable_id)
     participation.unparticipate! if participation.present?
   end
 
--- a/spec/models/comment_spec.rb
+++ b/spec/models/comment_spec.rb
@@ -24,6 +24,14 @@ describe Comment, type: :model do
       participations = Participation.where(target_id: comment_alice.commentable_id, author_id: comment_alice.author_id)
       expect(participations.first.count).to eq(1)
     end
+
+    it "should work if parent is gone" do
+      comment_alice.parent.delete
+      comment_alice.reload
+      expect {
+        comment_alice.destroy
+      }.to_not raise_error
+    end
   end
 
   describe "#subscribers" do

### CODE REVIEW

File: app/models/comment.rb [hunk 1]
```diff
@@ -47,8 +47,8 @@ class Comment < ApplicationRecord
   end
 
   after_destroy do
-    self.parent.update_comments_counter
-    participation = author.participations.find_by(target_id: post.id)
+    parent.update_comments_counter if parent
```
diaspora-code-review: Style/SafeNavigation: Use safe navigation (`&.`) instead of checking if an object exists before calling the method.

SuperTux88: I actually agree with @diaspora-code-review here, this is supported since ruby 2.3 and we don't support older versions anymore.

SuperTux88 (commented):
I'm a bit confused, when does this happen? The stacktrace says it's failing here:

https://github.com/diaspora/diaspora/blob/d5b7c6d779774454e60e8a6ed7bd1e2c001a43da/lib/diaspora/federation/receive.rb#L177

But deleting a comment would go here:

https://github.com/diaspora/diaspora/blob/d5b7c6d779774454e60e8a6ed7bd1e2c001a43da/lib/diaspora/federation/receive.rb#L167-L175

So do we receive a retraction for the post which then destroys all comments which then fails? If this fails, I would expect a bigger problem, because then probably no post with comment can't be deleted?

Or do we receive a rectraction for both, a comment and it's parent at the same time? But then I'm not sure if this really fixes it, or if it again fails later because deleting the parent also deletes all it's comment, so a comment can't be deleted twice?

jhass:
Well, honestly I don't know.

> Or do we receive a rectraction for both, a comment and it's parent at the same time? But then I'm not sure if this really fixes it, or if it again fails later because deleting the parent also deletes all it's comment, so a comment can't be deleted twice?

Seems the more likely scenario to me and that the parent and comment retraction actually run at the same time, so this being a race condition, which is why earlier checks may pass.

Either way I could write a spec for this so this is a potential state the system can be in, right? We have no database constraints preventing this, so we should handle it IMO if for nothing but fault tolerance.

If it fails differently down the line again we'll get the next stacktrace and can handle it then :)

PR #8065, opened 2019-10-02
Refactoring code style.

feruzoripov:
Changing `unless` condition to `if`, because of `else`. It's not good practice to use `unless` with `else`.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -2,7 +2,7 @@
 
 module Admin
   class UsersController < AdminController
-    before_action :validate_user, only: %i(make_admin remove_admin make_moderator remove_moderator make_spotlight remove_spotlight)
+    before_action :validate_user, only: %i[make_admin remove_admin make_moderator remove_moderator make_spotlight remove_spotlight]
 
     def close_account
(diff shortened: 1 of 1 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 module Admin
   class UsersController < AdminController
-    before_action :validate_user, only: %i(make_admin remove_admin make_moderator remove_moderator make_spotlight remove_spotlight)
+    before_action :validate_user, only: %i[make_admin remove_admin make_moderator remove_moderator make_spotlight remove_spotlight]
```
diaspora-code-review: Metrics/LineLength: Line is too long. [131/120]

SuperTux88:
Thank you for your effort, but I'm afraid we won't accept this. Please don't create PRs just to fix the code style, only fix the code style if you need to change the line anyway. See #5561 and [this article in the Hound FAQ](http://help.houndci.com/en/articles/67602-can-i-have-hound-check-my-entire-repository-all-at-once) for the reasoning, it was also mentioned in [this post in our discourse](https://discourse.diasporafoundation.org/t/styleguides/467).

PR #8066, opened 2019-10-02
Fixing openid_connect files with rubocop

feruzoripov:


SuperTux88:
Thank you for your effort, but I'm afraid we won't accept this. Please don't create PRs just to fix the code style, only fix the code style if you need to change the line anyway. See #5561 and [this article in the Hound FAQ](http://help.houndci.com/en/articles/67602-can-i-have-hound-check-my-entire-repository-all-at-once) for the reasoning, it was also mentioned in [this post in our discourse](https://discourse.diasporafoundation.org/t/styleguides/467).

(also this one broke tests)

Issue #8062, opened 2019-09-28
Stop using rails-assets.org for CSS and JS assets


denschub:
`rails-assets.org` has, unfortunately, [a sad history](https://github.com/tenex/rails-assets/issues?utf8=%E2%9C%93&q=is%3Aissue+expired) of letting their SSL-certs and even the domain expire, causing installation and upgrade issues for our users. In May, I made a significant donation towards setting up reliable SSL certification renewals and a monitoring on such after a downtime in https://github.com/tenex/rails-assets/issues/445, but this was not very productive, as [they managed to have their domain's DNS responses be NXDOMAINs for a while in July](https://github.com/tenex/rails-assets/issues/450), and, just now, [the SSL certificate expired again](https://github.com/tenex/rails-assets/issues/457).

The fact that whoever runs `rails-assets.org` these days fails to both keep the service running and accept help if they are unable to do so, I feel like it is a fair conclusion to say we cannot and should not rely on this service. Downtimes in their service cause our CI, new installation, and existing pod's upgrades to fail, and I have no reason to accept that.

As much as I like'd the Rubygems-only approach of assets management, I think it's time for us to move away from that, and switch to the "more standard" approach of `rails/webpacker` in combination with `yarn`.

denschub:
I'll let this block the next minor release for now, and will work on removing it immediately. Although it will be a bit of work to backport, I feel like it's worth it. But I'm happy if someone disagrees.

PR #8071, opened 2019-10-31
Replaced FineUploader by Uppy library. Refactored things.

Fensterbank:
Fine uploader [is not maintained anymore](https://github.com/FineUploader/fine-uploader/issues/2073) and we are still stuck with the problem, that image resizing is not working on Firefox for Android since a [PR](https://github.com/FineUploader/fine-uploader/pull/2007) of mine never was brought into a released version.

So I decided to look around for alternatives and found [Uppy](https://uppy.io), a powerful library and successor of FineUploader. Since we are already using Pica for resizing, it's no problem to not have any resizing plugins in Uppy included.

I refactored major parts of the code to fully integrate uppy using a custom Resizer plugin with the following rules:

- if the file is smaller then the allowed max size, nothing is done and the file is uploaded as it is
- if the file is bigger then the allowed max size, but the dimensions of the longest edge is smaller than 3072px, we just recompress it as JPEG without resizing, it's propably a huge png file.
- if the file is bigger then the allowed max size and the dimentions of the longest edge is greater than 3072px, we resize it before uploading.

The whole logic stayed the same, so it's no difference for the user experience.
I also improved a few thing, e.g. the input file buttons are now created on server side via HAML insteald of adding them by JavaScript afterwards.

Unfortunately Uppy is not available in rails-assets, so I had to add the whole package to the codebase. As soon as #8062 is finished, we could change this to use uppy from npm.
But if you have better solutions for this, I'm happy to see them.


app/assets/javascripts/helpers/uppy.min.js
```diff
@@ -0,0 +1,2 @@
...}};var r={id:"uppy",autoProceed:!1,allowMultipleUploads:!0,debug:!1,restrictions:{maxFileSize:null,maxNumberOfFiles:null,minNumberOfFiles:null,allowedFileTypes:null},meta:{},onBeforeFileAdded:function(e,t){return e},onBeforeUpload:function(e){return e},store:_$lib_163(),logger:nullLogger};if(this.opts=___extends_93({},r,e),this.opts.restrictions=___extends_93({},r.restrictions,this.opts.restrictions),e&&e.logger&&e.debug?this.log("You are using a custom `logger`, but also set `debug: true`, which uses built-in logger to output logs to console. Ignoring `debug: true` and using your custom `logger`.","warning"):e&&e.debug&&(this.opts.logger=__debugLogger_93),this.log("Using Core v"+this.constructor.VERSION),this.opts.restrictions.allowedFileTypes&&null!==this.opts.restrictions.allowedFileTypes&&!Array.isArray(this.opts.restrictions.allowedFileTypes))throw new TypeError("`restrictions.allowedFileTypes` must be an array");this.translator=new _$Translator_199([this.defaultLocale,this.opts.locale]),this.locale=this.translator.locale,this.i18n=this.translator.translate.bind(this.translator),this.i18nArray=this.translator.translateArray.bind(this.translator),this.plugins={},this.getState=t
```
denschub: This file lacks a license header. Please use an unminified build of said library (our pipeline compresses it anyway), and leave the license header intact. Also, third-party dependencies belong into `vendor/assets/javascripts`.

denschub: If the unminified version (see above) still references a sourcemap, please make sure it's actually there.

skyschub (changes requested):

app/assets/javascripts/helpers/ResizerPlugin.es6
```diff
@@ -0,0 +1,100 @@
...
+      if (file.type.split('/')[0] !== 'image' || (file.size <= this.maxFileSize)) {
+        return;
+      }
+      this.uppy.log(`[Resize] ${file.name}: File size ${file.size} exceeds max file size of ${this.maxFileSize}.`);
+
+      return this.resize(file).then((blob) => {
+        // since size and eventually type are changing, we have to overwrite several redundant information in the file object
+        const filename = file.name.replace(/\.png$/i, '.jpg')
+        const processedFile = {
+          ...file,
```
denschub: [Yo, listen up here's a story, about a contributor who tried to use new features](https://www.youtube.com/watch?v=68ugkg9RePc)...

`pronto` will complain about this, and it will complain about this very hard. In fact, it will claim this is a syntax error, although we both know the object spread operator is production-ready and shipped in all major browsers except for Edge (which I, quite frankly, no longer care about, as it works in Edgeium).

The reason `pronto` will yell at you is because `pronto` is not using the `eslint` out of your path. Instead, it calls to `pronto-eslint`, which in turns calls to `eslintrb`, [which hasn't been updated since 2016](https://github.com/zendesk/eslintrb), which still is using`eslint` 2.7.0. Obviously, object spreads are quite foreign to a eslint parser coming straight out of the early Mesolithic Age.

There is nothing I can do to make `pronto` happy with this, besides forking `eslintrb` and forking `pronto-eslint`, which is not going to happen to preserve the minds of everyone working on this project.

I hate writing this suggestion as much as you will hate it when reading, but can you rewrite that using `Object.assign({}, file, { ... })`, please?

We'll be in a better place, one day. I promise.

app/assets/javascripts/helpers/ResizerPlugin.es6
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  install() {
+    this.uppy.addPreProcessor(this.prepareUpload);
+  }
+
+  uninstall() {
+    this.uppy.removePreProcessor(this.prepareUpload);
+  }
+}
```
diaspora-code-review: Missing semicolon.

PR #8077, opened 2019-12-03
Create diaspora.yml

Snowf2000:


SuperTux88:
Since you added a wordpress secret here and it's not secret anymore, make sure to revoke that and use a new secret for your pod.

Issue #8083, opened 2020-01-07
Lost tab


LarsWl:
There is no tab in Diaspora::Commentable::last_three_comments

PR #8084, opened 2020-01-07
fix lost tab

LarsWl:


lib/diaspora/commentable.rb
```diff
@@ -13,12 +13,12 @@ def self.included(model)
...
 
     # @return [Array<Comment>]
-  def last_three_comments
-    return [] if self.comments_count == 0
-    # DO NOT USE .last(3) HERE.  IT WILL FETCH ALL COMMENTS AND RETURN THE LAST THREE
-    # INSTEAD OF DOING THE FOLLOWING, AS EXPECTED (THX AR):
-    self.comments.order('created_at DESC').limit(3).includes(:author => :profile).reverse
-  end
+    def last_three_comments
+      return [] if self.comments_count == 0
```
diaspora-code-review: - Layout/EmptyLineAfterGuardClause: Add empty line after guard clause.
- Style/RedundantSelf: Redundant `self` detected.

lib/diaspora/commentable.rb
```diff
@@ -13,12 +13,12 @@ def self.included(model)
...
+      # DO NOT USE .last(3) HERE.  IT WILL FETCH ALL COMMENTS AND RETURN THE LAST THREE
+      # INSTEAD OF DOING THE FOLLOWING, AS EXPECTED (THX AR):
+      self.comments.order('created_at DESC').limit(3).includes(:author => :profile).reverse
```
diaspora-code-review: - Style/RedundantSelf: Redundant `self` detected.
- Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
- Style/HashSyntax: Use the new Ruby 1.9 hash syntax.

goobertron:
Linked to issue #8083.

denschub:
Hi there, thanks for your pull request! However, we generally don't accept pull request just fixing codestyle issues. We changed our rules when a lot of code was already written, so there is something to fix basically everywhere. While I agree that this case looks very weird as the indentation is wrong, that's still just a codestyle change.

The issue with codestyle-only changes is they kinda ruin the Git history of a file. Instead of being able to see the last logic-change in `git blame`, we'd only see a lot of style changes and would have to filter those out or skip over it. I'm sure you understand. :)

PR #8087, opened 2020-01-22
Drop rails-assets.org dependency.

### LINKED ISSUE #8062: Stop using rails-assets.org for CSS and JS assets

skyschub:
... and replace it with our own Gem cache. This isn't pretty, as we have to manage that one ourselves, and pushing updates requires a core-team member, but rails-assets.org has become way too unreliable for us to be a viable option. This closes #8062.

We're currently into a already 20-hour long downtime (see https://github.com/tenex/rails-assets/issues/462), with no solution on the horizon. This is not only risking lots of trouble for people setting up new pods or updating existing, it is also actively blocking us from merging a PR at the moment, because we have no CI coverage.

This move is... kinda sad, especially since I made a significant financial contribution to rails-assets.org last year in the hopes of helping stabilizing the service. It only got worse, so this has to be done.

This solution is not perfect, as any asset changes now require manual action from the core team. Ideally, when rails-assets.org is back up, this is as easy as downloading the gem from there and pushing it to our host. But since the server requires authentication for uploads, this can only be done by someone with access to our infrastructure. I think that's fair, though, as we're in the process of merging the API and then eventually rebuilding our frontend anyway.

For whoever merges this (r? @jhass and/or @SuperTux88):

* I populated the cache with gems currently in use in `develop` *and* `master`/`next-minor`, so this can be uplifted to all branches. But please verify that, as gems can be... weird.
* I used Geraspora's gem cache as a source for populating the cache. If you SSH into `starsupport`, you can find all source files in `/srv/geminabox/gemsrc`.
* The server is running `geminabox`, see [this](https://github.com/geminabox/geminabox#client-usage) for instructions on how to push gems. However, uploading/changing things requires HTTP auth. Check `/srv/geminabox/config.ru` for details, and ping me if you need help.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -99,7 +99,7 @@ gem "js-routes",         "1.4.7"
 gem "js_image_paths",    "0.1.1"
 gem "sprockets-es6",     "0.9.2"
 
-source "https://rails-assets.org" do
+source "https://gems.diasporafoundation.org" do
   gem "rails-assets-jquery",                              "3.4.1" # Should be kept in sync with jquery-rails
   gem "rails-assets-jquery.ui",                           "1.11.4"
 
@@ -312,7 +312,7 @@ group :development, :test do
   # Jasmine (client side application tests (JS))
   gem "jasmine",                   "3.4.0"
   gem "jasmine-jquery-rails",      "2.0.3"
-  gem "rails-assets-jasmine-ajax", "4.0.0", source: "https://rails-assets.org"
+  gem "rails-assets-jasmine-ajax", "4.0.0", source: "https://gems.diasporafoundation.org"
   gem "sinon-rails",               "1.15.0"
 
   # For `assigns` in controller specs
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GEM
   remote: https://rubygems.org/
-  remote: https://rails-assets.org/
+  remote: https://gems.diasporafoundation.org/
   specs:
     actioncable (5.1.7)
       actionpack (= 5.1.7)


denschub:
And just as I hit submit, rails-assets.org is up again. However, I'm not going to back down this time, I already decided to ignore the issue in September because it went back up. Given our record of running things, the risk of running it ourselves is lower than continuing shipping a rails-assets.org dependency to all pods.

denschub:
I'd also like to highlight [this comment](https://github.com/tenex/rails-assets/issues/462#issuecomment-577159095). The security implications of suggesting that a random stranger should set up a server and the official domain could be pointed to that server is beyond me. I'm honestly left speechless by the fact this was even considered.

jhass (approved):

joshjordan:
@denschub no need for dramatics. Certainly, there is much more to the process than *just* DNS changes. Remember, whenever you use a third party service to pull dependencies, you are *already* trusting your software to strangers. Whether its the Rails Assets maintainers or someone else, you need to verify the dependencies if you want to trust them.

I'm glad to see you removing this dependency. That's generally what we recommend as much better solutions have superceded. Rails Assets is kept alive by the folks who need it and can't/won't change, but where possible, you should be vendoring/hosting privately/moving to something like webpacker.

denschub:
> Remember, whenever you use a third party service to pull dependencies, you are already trusting your software to strangers

**Exactly**. I have spent many hours on manually checking the gems we pulled from rails-assets.org, and I even wrote a script that automatically checks a couple known gems occasionally, for precisely this reason, after I was able to observe many targeted attacks on npm and rubygems. I do the same for other services I use, like npm and cargo. Doing these manual and automatic checks results in me being able to trust a given service. Having these checks in place does not mean a project is protected, but it provides good indicators that a given service is *probably* doing fine.

The fact that you, as apparently one of the people in charge, decided it's fine for external people to host a service because your instance is down is a direct violation of the trust I had initially, because it shows me that this trust was probably misplaced in the first place. Even *if* you did some manual checking before switching over DNS records, I can think of a large number of very real attack scenarios where your manual tests would completely fail, because you'd actually see a valid gem, while a portion of users won't.

That's not dramatics, that's risk assessment.

joshjordan:
Just so other folks that find this thread (which is linked from Rails Assets' issue board) have context: that assessment doesn't jive with any potential failover plan we would've considered. However, I do agree with you @denschub that Rails Assets is not the right choice for him. Cheers!

SuperTux88 (approved):
:tada: :cookie:

jhass:
Merged as 1d72f957059e19d86b29f1111cacde3f283eb10b

PR #8088, opened 2020-01-24
API: Return a default token_endpoint_auth_method when the client gives none in its OpenID Connect registration request

jhass:
Since we announce it in the supported metadata, some clients expect to be told what to use and don't fallback to the spec standard of client_secret_basic on their own.


jhass:
Pushed two more small commits triggered by the randomish CI failure. If you prefer them separately just say so :)

skyschub (approved):
Good catch on those specs <3

PR #8089, opened 2020-01-24
Re-add sizes.raw to the photo JSON

skyschub:
... because we need it for showing the raw image in the lightbox. This got lost, as the photo extension was made after API development started.

There are no integration tests for the lightbox, and since I don't want to spend two days on hotfixing a regression, we won't have integration tests after this PR either. But at least we have spec coverage now, so.. yay?

### DIFF
--- a/app/presenters/photo_presenter.rb
+++ b/app/presenters/photo_presenter.rb
@@ -12,7 +12,8 @@ class PhotoPresenter < BasePresenter
       sizes:      {
         small:  url(:thumb_small),
         medium: url(:thumb_medium),
-        large:  url(:scaled_full)
+        large:  url(:scaled_full),
+        raw:    url
       }
     }
   end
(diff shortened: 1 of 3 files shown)


SuperTux88 (approved):
Nice catch :cookie:

app/presenters/photo_presenter.rb
```diff
@@ -12,7 +12,8 @@ def base_hash
       sizes:      {
         small:  url(:thumb_small),
         medium: url(:thumb_medium),
-        large:  url(:scaled_full)
+        large:  url(:scaled_full),
+        raw:    url,
```
diaspora-code-review: Style/TrailingCommaInHashLiteral: Avoid comma after the last item of a hash.

app/presenters/photo_presenter.rb
```diff
@@ -26,7 +27,8 @@ def as_api_json(full=false)
...
```
diaspora-code-review: Style/TrailingCommaInHashLiteral: Avoid comma after the last item of a hash.

app/presenters/photo_presenter.rb
denschub: That's what I get for fixing the stylebot, eh?

jhass (approved):

PR #8100, opened 2020-02-04
Re-add `raw` to a photo's json.

skyschub:
This fixes the current issue with lightboxes not working.

To make sure we don't acidentially drop it again, I added `raw` as a required field to the schema. Unfortunately, this broke CI, as we did not expose `raw` for profile pictures, which do reference the same API schema.

Initially, I wanted to expose the actual raw image on the API, and I still plan on doing this, but this change turns out to be a magical journey through all parts of diaspora, and will take me a bit of time. To get the lightboxes fixed in the meantime, this PR exposes the `scaled_full` version as the "raw" image.

### DIFF
--- a/app/presenters/avatar_presenter.rb
+++ b/app/presenters/avatar_presenter.rb
@@ -6,7 +6,8 @@ class AvatarPresenter < BasePresenter
     avatar = {
       small:  small(with_default),
       medium: medium(with_default),
-      large:  large(with_default)
+      large:  large(with_default),
+      raw:    raw(with_default) # rubocop:disable Rails/OutputSafety
     }.compact
 
     avatar unless avatar.empty?
@@ -21,6 +22,11 @@ class AvatarPresenter < BasePresenter
   end
 
   def large(with_default=false)
+    image_url(size: :scaled_full, fallback_to_default: with_default)
+  end
+
+  def raw(with_default=false)
+    # TODO: Replace me with the actual raw photo.
     image_url(fallback_to_default: with_default)
   end
 end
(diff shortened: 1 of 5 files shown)


app/presenters/avatar_presenter.rb
```diff
@@ -6,7 +6,8 @@ def base_hash(with_default=false)
     avatar = {
       small:  small(with_default),
       medium: medium(with_default),
-      large:  large(with_default)
+      large:  large(with_default),
+      raw:    raw(with_default)
```
diaspora-code-review: Rails/OutputSafety: Tagging a string as html safe may be a security risk.

app/presenters/avatar_presenter.rb
```diff
@@ -21,6 +22,11 @@ def medium(with_default=false)
   end
 
   def large(with_default=false)
+    image_url(size: :scaled_full, fallback_to_default: with_default)
+  end
+
+  def raw(with_default=false)
+    # ToDo: Replace me with the actual raw photo.
```
diaspora-code-review: Style/CommentAnnotation: Annotation keywords like `ToDo` should be all upper case, followed by a colon, and a space, then a note describing the problem.

app/presenters/avatar_presenter.rb
denschub: That's a different `raw`, you funny bot!

jhass (approved):

Flaburgan:
Should we open an issue as a follow-up to be sure to not forget about that then? @denschub maybe you want to do it as it looks like a good place to explain what you went through when beginning to do it?

denschub:
@Flaburgan it's high on my ToDo list - both filing the issue, and fixing it. :)

PR #8103, opened 2020-02-08
API: Don't return notifications target unless it's a post

jhass:
API docs imply that's how it should be since the "target" can be inferred from
the event creator in the non post notifications case.

Also fixed the `only_unread` parameter to only listen to `true`. (Can send that separately if you prefer.)

### DIFF
--- a/app/presenters/notification_presenter.rb
+++ b/app/presenters/notification_presenter.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class NotificationPresenter < BasePresenter
-  def as_api_json(include_target=true)
+  def as_api_json
     data = base_hash
-    data = data.merge(target: target_json) if include_target && linked_object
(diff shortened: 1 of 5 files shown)


app/presenters/notification_presenter.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class NotificationPresenter < BasePresenter
-  def as_api_json(include_target=true)
+  def as_api_json
     data = base_hash
-    data = data.merge(target: target_json) if include_target && linked_object
+    data = data.merge(target: target_json) if linked_object && linked_object.is_a?(Post)
```
diaspora-code-review: Style/SafeNavigation: Use safe navigation (`&.`) instead of checking if an object exists before calling the method.

skyschub (approved):

PR #8123, opened 2020-05-24
Update for sq translations

ujdhesa:


jhass:
Unfortunately this branch does not merge without conflicts. Could you resolve these? 

Also if you want to see updates in a release  any time soon, it probably would be better to base your work on the `next-minor` branch and send a pull request to there.  That's why we recommend using WebTranslateIt, because then translation contributors don't need to deal or worry with any of these details.

ujdhesa:
> Unfortunately this branch does not merge without conflicts. Could you resolve these?
I’ve imported the translations to the translation platform. This pull request is no longer necessary.

PR #8125, opened 2020-06-05
Add FreeBSD example for CA bundle location in configuration

deepy:


denschub:
As discussed on IRC, we have no plans of supporting FreeBSD, and if people want to install their pod using FreeBSD, they'll figure these things out themselves. I'll close this PR, but I'll encourage you to publish the steps you have to take to a blog post or whatever! If FreeBSD turns out to be a platform that sees high adoption, we'll probably be willing and able to make that an officially supported target. :)

PR #8130, opened 2020-06-13
Bump Rails

skyschub:
This technically is a security-release, but none of the issues affect us. Let's update anyway in the next minor, just to mute the alerts. :)


SuperTux88 (approved):
LGTM :cookie:

PR #8131, opened 2020-07-08
bundler: disable post install messages in production

jhass:


SuperTux88 (approved):
Cool, I didn't know this is possible :partying_face: 

We can then remove the "Ignore any migration notes this command gives you, they're already done for you or explicitly advised in our changelog!" from our update-guide, but we should add to rerun `script/configure_bundler` at every update, so this setting is added to the config.

jhass:
Yeah, I kinda was pointed to this by upstream :D https://github.com/rubygems/rubygems/issues/3801

denschub:
Merged as 9dff9642b6ec3e27e166983280fb341c8a6cc808.

denschub:
> but we should add to rerun `script/configure_bundler` at every update, so this setting is added to the config.

[Docs updated](https://wiki.diasporafoundation.org/wiki/index.php?title=Updating&type=revision&diff=10440&oldid=10437).

PR #8133, opened 2020-07-10
Drop spring

jhass:
it does more harm than good at this point


skyschub (approved):

SuperTux88 (changes requested):
There is also `config/spring.rb` which can be deleted now.

Otherwise looks good to me :cookie:

SuperTux88 (approved):

jhass:
Merged as ce82ba3dde7178b6e707d83bc6246fe033a328c9

PR #8145, opened 2020-09-16
Update rubyzip from 1.3.0 to 2.3.0

pravi:
Major change is dropping support for older ruby versions https://github.com/rubyzip/rubyzip/blob/master/Changelog.md#200-2019-09-25


jhass:
Any reason to stay behind and not update to 2.3.0 directly?

pravi:
Nothing specific, just wanted to be safe as it was a major bump. But I will update to 2.3.0.

pravi:
Done and green.

PR #8147, opened 2020-09-16
Use jquery-ui-rails instead of rails-assets-jquery.ui

### LINKED ISSUE #8144: Suggestion to switch from rails-assets-jquery.ui to jquery-ui-rails
Would it be a good idea to switch to jquery-ui-rails from  rails-assets-jquery.ui to reduce the number of gems we have to cache?

pravi:
Fixes #8144

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,13 +96,13 @@ gem "entypo-rails", "3.0.0"
 
 gem "handlebars_assets", "0.23.8"
 gem "jquery-rails",      "4.3.5"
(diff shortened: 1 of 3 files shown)


jhass:
That wasn't too hard to guess: https://travis-ci.org/github/diaspora/diaspora/jobs/727807701#L1357

pravi:
Fixed and ruby 2.6 tests already passed in ci.

PR #8149, opened 2020-09-16
Update bootstrap-switch-rails 3.3.3 -> 3.3.5

pravi:
The issue with 3.3.4 is now closed


SuperTux88 (changes requested):
The issue is closed, but the bug still exists ... either they didn't release it yet (3.3.5 was released before the bug was closed, there doesn't exist a new release yet) or the issue was closed without fixing the bug :shrug:

pravi:
Ouch! Making mistake once is okay, but repeating the same mistake...

PR #8160, opened 2020-10-04
Improve Service::Tumblr

AllanKlaus:
### Motivation
I saw we have a lack in the coverage of `Service::Tumblr` class. During the development I saw that we have issues on rubocop so I fixed it too.

### Changelog
[x] Fix rubocop styles
[x] Improve specs coverage
[x] Move some public methods that was public to private (They are only being used by the class)

### Obs.
I notice that on `coveralls` the coverage is different of the `simplecov` so I based during the task on `simplecov` report

#### before
Screenshot from 2020-10-04 10-01-20

#### after

Screenshot from 2020-10-04 09-56-46

### DIFF
--- a/app/models/services/tumblr.rb
+++ b/app/models/services/tumblr.rb
@@ -1,69 +1,71 @@
 # frozen_string_literal: true
 
-class Services::Tumblr < Service
-  MAX_CHARACTERS = 1000
+module Services
+  class Tumblr < Service
+    MAX_CHARACTERS = 1000
 
-  def provider
-    "tumblr"
-  end
-
-  def consumer_key
-    AppConfig.services.tumblr.key
-  end
-
-  def consumer_secret
-    AppConfig.services.tumblr.secret
-  end
-
-  def post(post, url='')
-    body = build_tumblr_post(post, url)
-    user_info = JSON.parse(client.get("/v2/user/info").body)
-    blogs = user_info["response"]["user"]["blogs"]
-    primaryblog = blogs.find {|blog| blog["primary"] } || blogs[0]
-    tumblr_ids = {}
-
-    blogurl = URI.parse(primaryblog["url"])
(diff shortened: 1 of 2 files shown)


denschub:
So, weird question, but: does that service even work? As in, can we still post things to Tumblr? There is noone actively maintaining it, and I'm not aware of any user, so I don't know if this even works (we know that other services did break over time, and so we removed them).

annando:
The Tumblr API itself does work now, we are using it on Friendica. So I guess it should work on Diaspora as well.

denschub:
>  So I guess it should work on Diaspora as well.

Yeah, sure, it *should*. The last significant change to that service was in 2015, so... it's save to say that's unmaintained. I also am not personally aware of anyone actually using this integration. We've had services being broken for years without anyone noticing, so... it'd be good to test, at least.

AllanKlaus:
Do we have any monitoring service like sentry to check errors on platform? If don't I can work to implement sentry.

I only don't if I can create a [full free license](https://sentry.io/for/open-source/) but i can work using a test account and after get the license someone can just change the keys

AllanKlaus:
@denschub @annando any idea how we can follow with this?

denschub:
diaspora\* is not a managed software, it's an open source project that everyone can install and run. There is no central instance, and so nobody can run a sentry on that single instance. Adding support is also not needed, as this is literally installing one Gem and adding a new file into the initializes directory - podmins who are keen on adding Sentry can surely figure that out, but there is no need to add that Gem for everyone.

The way forward here is to set up a test pod, activate the tumblr integration, and test if it actually still works. If it works, cool; if it doesn't work, we'll have to talk about what we do with that.

Flaburgan:
Superseded by https://github.com/diaspora/diaspora/pull/8244, thank you @AllanKlaus for your work here ;)

PR #8169, opened 2020-10-24
OpenID Connect: ensure consistent issuer URL

jhass:
root_url does not know the right protocol in all contexts,
some clients are strict when validating this.


tclaus:
This works in my environment.

skyschub (approved):
lgtm

PR #8172, opened 2020-10-31
Slightly polish the OAuth Scope descriptions.

skyschub:
This should be considered a draft to further improve upon, but this should already be much better than what's there right now.

@goobertron, do you have time to provide feedback here? :)

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -917,72 +917,72 @@ en:
           no_requirement: "%{name} requires no permissions"
         no_applications: "You have no authorized applications"
         revoke_autorization: "Revoke"
-        tos: "See the application's terms of service"
-        policy: "See the application's privacy policy"
+        tos: "See the application’s terms of service"
+        policy: "See the application’s privacy policy"
       scopes:
         openid:
-          name: "basic profile"
-          description: "This grants read access to basic profile information to the application."
+          name: "Basic profile information"
+          description: "This grants read-only access to your basic profile information."
         sub:
-          name: "sub"
-          description: "This grants sub permissions to the application."
+          name: "Unique identifier"
+          description: "This grants read-only access to your unique identifier."
         name:
-          name: "name"
-          description: "This grants read access to name data to the application."
+          name: "Full name"
+          description: "This grants read-only access to your full name."
         nickname:
-          name: "nickname"
-          description: "This grants read access to nickname data to the application."
+          name: "Username"
+          description: "This grants read-only access to your username."
         picture:
-          name: "picture"
-          description: "This grants read access to user's profile picture data to the application."
+          name: "Profile picture"
+          description: "This grants read-only access to your profile picture."
         'contacts:read':
-          name: "contacts:read"
-          description: "This grants read permissions to contacts and related data (like aspects) to the application."
+          name: "Contacts (Read-only)"
+          description: "This grants read-only access to contacts and related data (such as aspects)."
         'contacts:modify':
-          name: "contacts:modify"
-          description: "This grants write permissions to contacts and related data (like aspects) to the application."
+          name: "Contacts (Write)"
+          description: "This grants write access to contacts and related data (such as aspects)."
         conversations:
-          name: "conversations"
-          description: "This grants read/write permission to private messaging to the application."
+          name: "Conversations"
+          description: "This grants read and write permission to private messages."
         email:
-          name: "email"
-          description: "This grants read access to the user's email address to the application."
+          name: "E-Mail"
+          description: "This grants read-only access to your email address."
         interactions:
-          name: "interactions"
-          description: "This grants read and write permissions on interacting with posts to the application."
+          name: "Interactions"
+          description: "This grants access to interact with posts, for instance liking or submitting a comment."
         notifications:
-          name: "notifications"
-          description: "This grants read and write permissions on notifications for the user to the application."
+          name: "Notifications"
+          description: "This grants read and write access to your notifications."
         'private:read':
-          name: "private:read"
-          description: "This grants read privileges on private posts and streams to the application."
+          name: "Private Posts (Read-only)"
+          description: "This grants read-only access to your and your contacts’ private posts."
         'private:modify':
-          name: "private:modify"
-          description: "This grants modification privileges on private posts and streams, including allowing private sharing, to the application."
+          name: "Private Posts (Write)"
+          description: "This grants access to publish private posts."
         'public:read':
-          name: "public:read"
-          description: "This grants read permissions on public posts, interactions, and user data to the application."
+          name: "Public Posts (Read-only)"
+          description: "This grants access to your public posts, including interactions and related data."
         'public:modify':
-          name: "public:modify"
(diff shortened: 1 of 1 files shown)


denschub:
It looks like some scopes (like `email`) aren't even in use. Also, exposing the `sub` scope and other generic fields to the user probably makes no sense, as an app cannot not-request it. It probably makes more sense to collapse those under a "generic profile information", like other platforms are doing.

goobertron:
Well the English is basically perfect, as I'd expect. I'll make a few minor line comments. <3

config/locales/diaspora/en.yml
```diff
@@ -921,68 +921,68 @@ en:
...
         picture:
-          name: "picture"
-          description: "This grants read access to user's profile picture data to the application."
+          name: "Profile picture"
+          description: "This grants read-only access to your profile picture."
         'contacts:read':
-          name: "contacts:read"
-          description: "This grants read permissions to contacts and related data (like aspects) to the application."
+          name: "Contacts (Read-only)"
+          description: "This grants read-only access to contacts and related data (like aspects)."
```
goobertron, line 940: A minor style point: for clarity, I would use 'such as' rather than 'like' in 'related data (like aspects)'. But 'like' in this context is very common, and there's nothing wrong with it.

config/locales/diaspora/en.yml
```diff
@@ -921,68 +921,68 @@ en:
...
         'contacts:modify':
-          name: "contacts:modify"
-          description: "This grants write permissions to contacts and related data (like aspects) to the application."
+          name: "Contacts (Write)"
+          description: "This grants write access to contacts and related data (like aspects)."
```
goobertron, line 943: Ditto: 'like' –> 'such as'.

config/locales/diaspora/en.yml
```diff
@@ -921,68 +921,68 @@ en:
...
         email:
-          name: "email"
-          description: "This grants read access to the user's email address to the application."
+          name: "E-Mail"
+          description: "This grants read-only access to your email address."
         interactions:
-          name: "interactions"
-          description: "This grants read and write permissions on interacting with posts to the application."
+          name: "Interactions"
+          description: "This grants access to interact with posts, like liking or submitting a comment."
```
goobertron, line 952: Here, I'd replace 'like' with 'for instance'.

config/locales/diaspora/en.yml
```diff
@@ -921,68 +921,68 @@ en:
...
+          description: "This grants read-only access to your and your contacts' private posts."
         'private:modify':
-          name: "private:modify"
-          description: "This grants modification privileges on private posts and streams, including allowing private sharing, to the application."
+          name: "Private Posts (Write)"
+          description: "This grants access to publish private posts."
         'public:read':
-          name: "public:read"
-          description: "This grants read permissions on public posts, interactions, and user data to the application."
+          name: "Public Posts (Write)"
```
goobertron, line 963: Should this be 'Public Posts (Read-only)'?

config/locales/diaspora/en.yml
```diff
@@ -921,68 +921,68 @@ en:
...
+          description: "This grants access to your public posts, and their interactions and related data."
```
goobertron, line 964: How about either 'access to your public posts, their interactions and related data' or 'access to your public posts, including interactions and related data'?

config/locales/diaspora/en.yml
```diff
@@ -921,68 +921,68 @@ en:
...
         'public:modify':
-          name: "public:modify"
-          description: "This grants write permissions on public posts (including public post writing), interactions, and user data to the application."
+          name: "Public Posts (Write)"
+          description: "This grants write access to publish private posts and related data (like votes)."
```
goobertron, line 967: 'e.g. votes' rather than 'like votes'?

config/locales/diaspora/en.yml
```diff
@@ -921,68 +921,68 @@ en:
...
         profile:
-          name: "extended profile"
-          description: "This grants read access to extendend profile data to the application."
+          name: "Extended profile (Read-only)"
+          description: "This grants read-only access to your extendend profile data."
```
goobertron, line 970: Typo: 'extendend' –> 'extended'.

goobertron:
A couple of questions: I assume that in the interface in which these strings will be used, it is better to use a straight apostrophe ( ' ) rather than a curly one ( ’ ) as is used in the main diaspora UI. But I raise it just in case not.

https://github.com/diaspora/diaspora/pull/8172/files#diff-4697c6c6eafd9d8f97e9a2724105d333fc1a99370b1ce420887e683da7869f13R975 : what is the difference between 'Private profile data' and 'Extended profile'? This seems to duplicate line 969. Also, should there be a Basic profile (Read-only) and Basic profile (Write) lines?

config/locales/diaspora/en.yml
denschub: Fair suggestion, I'll change that.

config/locales/diaspora/en.yml
denschub: Oh yes, absolutely!

config/locales/diaspora/en.yml
denschub: That would be an incomplete list, but I changed that to "such as votes and attached media"

denschub:
I addressed your feedback, thank you so much. :)

> I assume that in the interface in which these strings will be used, it is better to use a straight apostrophe ( ' ) rather than a curly one ( ’ ) as is used in the main diaspora UI.

Actually no, there is no internal use for those strings, and they're just used to explain the permissions to the user during the app authorization flow. You're right in pointing this out, I changed the apostrophes.

> [Line 975](https://github.com/diaspora/diaspora/pull/8172/files#diff-4697c6c6eafd9d8f97e9a2724105d333fc1a99370b1ce420887e683da7869f13R975): what is the difference between 'Private profile data' and 'Extended profile'? This seems to duplicate line 969. Also, should there be a Basic profile (Read-only) and Basic profile (Write) lines?

I actually don't know.

There are some internal fields in a user's profile that can never be changed, like one's encryption key. Those things have to be read-only, but it appears like that scope isn't even implemented, so it's probably completely useless. I'll file a follow-up bug to investigate and clean up scopes...

jhass (approved):

goobertron:
> There are some internal fields in a user's profile that can never be changed, like one's encryption key.

If that field is to cover such elements, perhaps something like 'Private user data' or 'Private account data' would be better than 'Private profile data'. I associate the word 'profile' with information viewable by other users about that account ('write something to describe yourself') rather than data that enable the account to function in the network.

config/locales/diaspora/en.yml
goobertron: 'E.g.', like 'such as', implies an example or examples, not a complete list. (You might be thinking of 'i.e.', meaning 'that is'.) But 'such as' works just as well, and an extra example is, of course, lovely and wonderful.

denschub:
Good suggestion! I'll keep that in mind in case we keep that scope, but it looks like that's just a copy-paste error or an idea that was never implemented.

PR #8187, opened 2020-11-30
Enforce GUID being present for notifications

### LINKED ISSUE #8186: Notifications guid sometimes null in database
... which should never happen because there's an `after_initialize` always setting it. Which also makes this bug quite opaque, as a new random value is set each time one of those records are loaded.

However it breaks the API which is using the guid as identifier, so the backend would not find those records next time around the client wants to do something with them.

I've been testing to just enforce a `NOT NULL` constraint on the column on my pod for a few weeks, in hope of uncovering the origin of these. So far without any success. Put positively, adding the constraint seems to fix the error without any downsides.

jhass:
fixes #8186

### DIFF
--- a/db/migrate/20201130142541_change_notifications_guid_not_null.rb
+++ b/db/migrate/20201130142541_change_notifications_guid_not_null.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class ChangeNotificationsGuidNotNull < ActiveRecord::Migration[5.2]
+  def up
+    Notification.where(guid: nil).find_in_batches do |batch|
+      batch.each do |notification|
+        notification.save!(validate: false, touch: false)
+      end
+    end
+
+    change_column :notifications, :guid, :string, null: false
+  end
+
+  def down
+    change_column :notifications, :guid, :string, null: true
+  end
+end


db/migrate/20201130142541_change_notifications_guid_not_null.rb
```diff
@@ -0,0 +1,16 @@
+
```
diaspora-code-review, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

db/migrate/20201130142541_change_notifications_guid_not_null.rb
```diff
@@ -0,0 +1,16 @@
...
+class ChangeNotificationsGuidNotNull < ActiveRecord::Migration[5.2]
```
diaspora-code-review, line 2: Layout/LeadingEmptyLines: Unnecessary blank line at the beginning of the source.

db/migrate/20201130142541_change_notifications_guid_not_null.rb
jhass: Hush, there's not even a string in this file.

skyschub (approved):

PR #8188, opened 2020-11-30
API: Conversations read handling through explicit route

jhass:
ref https://github.com/diaspora/api-documentation/issues/52
ref https://github.com/diaspora/api-documentation/pull/60


skyschub (approved):
Sorry, somehow missed that one!

PR #8194, opened 2020-12-27
Fix syntax error for default_color_theme in diaspora.toml.example

SuperTux88:
I have no idea why, but this is the only one where the `:` wasn't replaced with a `=` in the example.


jhass:
Eh, just missed it :D

PR #8201, opened 2021-02-07
Bulk replace of insecure protocols: http->https

ford--prefect:
This is a bulk replace of all http://… to https://… with only leaving localhost to the unencrypted protocol.
This assumes, that all resources linked are indeed available over https, which has not been explicitly checked (yet).


ford--prefect:
The bulk replace seems to have overshot a bit, I'll improve and propose a better one.

tclaus:
That was also my first Idea - I've tried this some days ago, and started checking and replacing each URL individually. 
Be careful with URLs in *spec files, example.com samples, as well as "bad urls and good urls file".

ford--prefect:
Yep, I've just spent an hour to go through all of the affected files semi-manually and I'll try to feed it to the CI first before making a new PR.

tclaus:
I stuck with running tests locally on my attempt.. (macOS)
But running tests locally is a good idea.

ford--prefect:
I see. Seems I tackled this a bit overly oblivious. Sorry for that, I'm learning while doing it. Hope it does not bother you guys too much. I try to not break anything.

PR #8202, opened 2021-02-08
Delete user invitation code during account deletion

fantashley:
Disclaimer: I've never written Ruby, so please let me know if there's a better way of doing this.

If a Podmin closes an account on an invite-only server and that account holder still knows their invitation code, they can currently recreate an account using that invitation code. This PR deletes the original invitation code for that user during account deletion so they can't create another account against a Podmin's will. I realize the user can still recreate an account if they themselves were invited by another user (and still have that user's code), but this at least helps in situations where a server might have previously allowed registrations but no longer does.

### DIFF
--- a/lib/account_deleter.rb
+++ b/lib/account_deleter.rb
@@ -58,7 +58,7 @@ class AccountDeleter
   end
 
   def delete_user_invitation_code
-    InvitationCode.find_by(user_id: user.id).destroy
+    InvitationCode.find_by(user_id: user.id).try(:destroy)
   end
 
   def normal_ar_person_associates_to_delete
--- a/lib/account_deleter.rb
+++ b/lib/account_deleter.rb
@@ -39,6 +39,7 @@ class AccountDeleter
     remove_share_visibilities_on_contacts_posts
     disconnect_contacts
     delete_standard_user_associations
+    delete_user_invitation_code
     tombstone_user
   end
 
@@ -54,6 +55,10 @@ class AccountDeleter
         User.reflect_on_association(asso).class_name.constantize.where(id: ids).destroy_all
       end
     end
+  end
+
+  def delete_user_invitation_code
+    InvitationCode.find_by(user_id: user.id).destroy
   end
 
   def normal_ar_person_associates_to_delete
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -85,6 +85,7 @@ FactoryGirl.define do
     after(:create) do |u|
       u.person.save
       u.person.profile.save
+      FactoryGirl.create(:invitation_code, user: u)
     end
   end
 
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -85,7 +85,6 @@ FactoryGirl.define do
     after(:create) do |u|
       u.person.save
       u.person.profile.save
-      FactoryGirl.create(:invitation_code, user: u)
     end
   end
 
--- a/spec/lib/account_deleter_spec.rb
+++ b/spec/lib/account_deleter_spec.rb
@@ -76,6 +76,7 @@ describe AccountDeleter do
   describe "#close_user" do
     user_removal_methods = %i[
       delete_standard_user_associations
+      delete_user_invitation_code
       remove_share_visibilities_on_contacts_posts
       disconnect_contacts tombstone_user
     ]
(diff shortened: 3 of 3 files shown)


lib/account_deleter.rb
```diff
@@ -56,6 +57,10 @@ def delete_standard_user_associations
     end
   end
 
+  def delete_user_invitation_code
+    InvitationCode.find_by_user_id(user.id).destroy
```
diaspora-code-review, line 61: Rails/DynamicFindBy: Use `find_by` instead of dynamic `find_by_user_id`.

denschub:
Oh, that's a good catch, thanks for noticing and thanks for the PR! :)

Do you mind extending the AccountDeletions test (https://github.com/diaspora/diaspora/blob/develop/spec/lib/account_deleter_spec.rb) for that?

denschub:
We also discovered a CI failure, apparently caused by GitHub changing something to the Ubuntu GHA image. We fixed that in https://github.com/diaspora/diaspora/pull/8204, the tests in this PR should all be happy again if you do a rebase. Sorry for that! :)

fantashley:
No worries! And sure, I'll try to get the tests figured out this week!

spec/lib/account_deleter_spec.rb
```diff
@@ -92,7 +93,7 @@
   end
 
   describe "#delete_standard_user_associations" do
-    it 'removes all standard user associaltions' do
+    it 'removes all standard user associations' do
```
diaspora-code-review, line 96: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/lib/account_deleter_spec.rb
```diff
@@ -107,11 +108,18 @@
     end
   end
 
+  describe "#delete_user_invitation_code" do
+    it 'deletes user invitation code' do
```
diaspora-code-review, line 112: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

spec/lib/account_deleter_spec.rb
```diff
@@ -107,11 +108,18 @@
...
+      @account_deletion.delete_user_invitation_code
+    end
+  end
+
   describe "#delete_standard_person_associations" do
     before do
       @account_deletion.person = bob.person
     end
-    it 'removes all standard person associaltions' do
+    it 'removes all standard person associations' do
```
diaspora-code-review, line 122: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

fantashley:
Okay @denschub I think I figured it out!

lib/account_deleter.rb
```diff
@@ -56,6 +57,10 @@ def delete_standard_user_associations
     end
   end
 
+  def delete_user_invitation_code
+    InvitationCode.find_by(user_id: user.id).destroy
```
denschub, line 61: With the comment below (at the test) taken into consideration, 

```suggestion
    InvitationCode.find_by(user_id: user.id).try(:destroy)
```

should make it unnecessary to add an invitation code to testing users, and it should make sure this always works for "real" users.

spec/lib/account_deleter_spec.rb
```diff
@@ -107,11 +108,23 @@
     end
   end
 
+  describe "#delete_user_invitation_code" do
+    it "deletes user invitation code" do
+      expect(bob.invitation_code).not_to be_nil
```
denschub, line 113: Note: if you apply my suggestion about not adding invitation codes to all test users, `bob` doesn't get one. However, I don't think you have to change anything here, because this like requests the invitation code, and [accessing a user's `.invitation_code` creates one if it doesn't have one yet](https://github.com/diaspora/diaspora/blob/ecd2b87475d96672c4c01d7735b2a0fd4fc68766/app/models/user.rb#L123), so I think this should pass. If it doesn't pass, let me know and I'll take a closer look!

skyschub (commented):
I have one comment about the logic here; I think we can make this a bit safer, but this is looking great already!

spec/factories.rb
```diff
@@ -85,6 +85,7 @@ def r_str
     after(:create) do |u|
       u.person.save
       u.person.profile.save
+      FactoryGirl.create(:invitation_code, user: u)
```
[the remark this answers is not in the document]
fantashley: Ah, good point, thank you!

fantashley:
Awesome, that makes sense, thanks so much for the help!

skyschub (approved):
Looks good, will merge once CI is green!

denschub:
Squashed and merged into the next minor release as d5d53baa1cc2bbcf1beace83885f5406462e9237.

Thanks for your Pull Request! 🍪

PR #8204, opened 2021-02-08
Install ghostscript into the CI environment.

skyschub:
Let's see if this fixes the [current cucumber brokenness](https://github.com/diaspora/diaspora/runs/1851101000?check_suite_focus=true#step:10:1421).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
           - 5432:5432
     steps:
       - name: Install system dependencies
-        run: sudo apt-get install -y libidn11-dev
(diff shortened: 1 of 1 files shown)


denschub:
Cukes passed with `ghostscript`, but jasmine timed out. Let's try one more time, this time using just all deps from the wiki (so that we have a documented environment), and let's also disable `fail-fast`. It's not worth the hassle to re-trigger things if one component fails, and we might get useful insights from not cancelling.

denschub:
Build is green, let's :shipit:.

SuperTux88 (approved):
LGTM :cookie:

denschub:
Merged as 7a7c48a470012ddfc426066da3549aa717da6b7c and 875f54b8460b660ec719aae5d4982cf5e83f7293.

PR #8206, opened 2021-02-13
Enforce lowercase tags and migrate existing uppercase tags for Postgres.

skyschub:
... and also create a lower-case tag name index to possibly speed up ActsAsTaggableOn...

As discussed with @SuperTux88 on IRC. Hopefully the inline-documentation is enough, but please ping me if not. :)

### DIFF
--- a/db/migrate/20210213000000_lowercaseify_tags.rb
+++ b/db/migrate/20210213000000_lowercaseify_tags.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+class LowercaseifyTags < ActiveRecord::Migration[5.2]
+  # We have to disable the global transaction for migrations, otherwise we
+  # can't rescue from duplicates below. Checking if there are duplicates before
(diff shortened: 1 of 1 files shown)


SuperTux88 (approved):
LGTM :cookie:

db/migrate/20210213000000_lowercaseify_tags.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+class LowercaseifyTags < ActiveRecord::Migration[5.2]
+  # We have to disable the global transaction for migrations, otherwise we
+  # can't rescue from duplicates below. Checking if there are duplicates before
+  # running the update would make this migration even more complicated, so...
+  disable_ddl_transaction!
+
+  # rubocop:disable Metrics/MethodLength
+  def up
```
diaspora-code-review, line 10: Metrics/AbcSize: Assignment Branch Condition size for up is too high. [<3, 21, 3> 21.42/20]

Issue #8198, opened 2021-01-21
non-secure http links in diaspora main page


ford--prefect:
I noticed two items in the diaspora main page in version 0.7.14.0.

The first one appears in the head section of the html:
> `<head` prefix='og: http://ogp.me/ns# article: http://ogp.me/ns/article# profile: http://ogp.me/ns/profile#'>

And the second one are the links to the diaspora foundation in the lower left column:
> `<a href="http://wiki.diasporafoundation.org/Welcoming_Committee">`
> `<a target="_blank" href="http://wiki.diasporafoundation.org">Wiki</a>`

While I don't understand the purpose of the first one, the second appearance is an easy fix.

tclaus:
Good finding - There are lots of other places with legacy (http) protocols. most of them now have the secure variant. 
Some others are not available any more.

ford--prefect:
Sounds like you would appreciate some help finding all the other insecure protocol uses? I feel I can do that. Merge request coming.

ford--prefect:
Created a pull request (#8201) for this.

denschub:
Please only change links you have a really good reason to change something (i.e. a link that is actually affecting users), because your mass-change would just completely ruin the git logs with no reason whatsoever.

denschub:
And just to clarify what I mean:

* Links like the sidebar links to the wiki: Great! Those get exposed to the user, so changing those is great.
* Links in templates, like the default front page: Also great, they can be clicked by users.
* HTML namespaces like the opengraph example: No need - those URLs are for spec-purposes only, they will not get queried by a browser.
* URLs in tests: No reason to change those, they have no user impact.

Please also **do not** touch translations. It's fine to edit the `en.yml`s, but everything else is handled outside of GitHub at WebTranslateIt - if you change the English main version, the languages will eventually sync.

PR #8207, opened 2021-02-20
#8198  Changed user facing unsecure protocol to secure

tclaus:
Solves #8198 - some Protocols are still http instead of https 
This PR changes the ones a user has direct contract with or are published in the mail start page.

### DIFF
--- a/app/assets/javascripts/app/views/help_view.js
+++ b/app/assets/javascripts/app/views/help_view.js
@@ -14,7 +14,7 @@ app.views.Help = app.views.StaticContentView.extend({
 
   initialize : function() {
     this.GETTING_HELP_SUBS = {
-      getting_started_a: {tutorial_series: this.linkHtml("http://diasporafoundation.org/getting_started/sign_up", Diaspora.I18n.t("getting_started_tutorial"))},
+      getting_started_a: {tutorial_series: this.linkHtml("https://diasporafoundation.org/getting_started/sign_up", Diaspora.I18n.t("getting_started_tutorial"))},
       get_support_a_website: {link: this.linkHtml("https://diasporafoundation.org/", Diaspora.I18n.t("foundation_website"))},
       get_support_a_tutorials: {tutorials: this.linkHtml("https://diasporafoundation.org/tutorials", Diaspora.I18n.t("tutorials"))},
       get_support_a_wiki: {link: this.linkHtml("https://wiki.diasporafoundation.org/Special:Search", Diaspora.I18n.t("wiki"))},
@@ -27,8 +27,8 @@ app.views.Help = app.views.StaticContentView.extend({
     this.POSTS_AND_POSTING_SUBS = {
       post_report_a: {community_guidelines: this.linkHtml("https://diasporafoundation.org/community_guidelines", Diaspora.I18n.t("community_guidelines"))},
       format_text_a: {
-        markdown: this.linkHtml("http://diasporafoundation.org/formatting", Diaspora.I18n.t( 'markdown' )),
-        here: this.linkHtml("http://daringfireball.net/projects/markdown/syntax", Diaspora.I18n.t( 'here' ))
+        markdown: this.linkHtml("https://diasporafoundation.org/formatting", Diaspora.I18n.t( 'markdown' )),
+        here: this.linkHtml("https://daringfireball.net/projects/markdown/syntax", Diaspora.I18n.t( 'here' ))
       }
     };
 
--- a/app/assets/javascripts/app/views/help_view.js
+++ b/app/assets/javascripts/app/views/help_view.js
@@ -12,6 +12,7 @@ app.views.Help = app.views.StaticContentView.extend({
     "click .faq-link-keyboard-shortcuts": "keyboardShortcuts"
   },
 
+  /* eslint-disable camelcase */
   initialize : function() {
     this.GETTING_HELP_SUBS = {
       getting_started_a: {tutorial_series: this.linkHtml("https://diasporafoundation.org/getting_started/sign_up", Diaspora.I18n.t("getting_started_tutorial"))},
@@ -27,10 +28,11 @@ app.views.Help = app.views.StaticContentView.extend({
     this.POSTS_AND_POSTING_SUBS = {
       post_report_a: {community_guidelines: this.linkHtml("https://diasporafoundation.org/community_guidelines", Diaspora.I18n.t("community_guidelines"))},
       format_text_a: {
-        markdown: this.linkHtml("https://diasporafoundation.org/formatting", Diaspora.I18n.t( 'markdown' )),
-        here: this.linkHtml("https://daringfireball.net/projects/markdown/syntax", Diaspora.I18n.t( 'here' ))
+        markdown: this.linkHtml("https://diasporafoundation.org/formatting", Diaspora.I18n.t("markdown")),
+        here: this.linkHtml("https://daringfireball.net/projects/markdown/syntax", Diaspora.I18n.t("here"))
       }
     };
+  /* eslint-enable camelcase */
 
     this.TAGS_SUBS = {
       filter_tags_a: {
--- a/app/helpers/meta_data_helper.rb
+++ b/app/helpers/meta_data_helper.rb
@@ -5,7 +5,7 @@ module MetaDataHelper
   include ActionView::Helpers::TagHelper
 
   def og_prefix
-    'og: http://ogp.me/ns# article: http://ogp.me/ns/article# profile: http://ogp.me/ns/profile#'
+    'og: https://ogp.me/ns# article: https://ogp.me/ns/article# profile: https://ogp.me/ns/profile#'
   end
 
   def site_url
--- a/app/helpers/meta_data_helper.rb
+++ b/app/helpers/meta_data_helper.rb
@@ -5,7 +5,7 @@ module MetaDataHelper
   include ActionView::Helpers::TagHelper
 
   def og_prefix
-    'og: https://ogp.me/ns# article: https://ogp.me/ns/article# profile: https://ogp.me/ns/profile#'
+    "og: https://ogp.me/ns# article: https://ogp.me/ns/article# profile: https://ogp.me/ns/profile#"
   end
 
   def site_url
--- a/app/models/open_graph_cache.rb
+++ b/app/models/open_graph_cache.rb
@@ -33,7 +33,7 @@ class OpenGraphCache < ApplicationRecord
   end
 
   def fetch_and_save_opengraph_data!
-    uri = URI.parse(url.start_with?("http") ? url : "http://#{url}")
+    uri = URI.parse(url.start_with?("http") ? url : "https://#{url}")
     uri.normalize!
     object = OpenGraphReader.fetch!(uri)
 
--- a/app/models/open_graph_cache.rb
+++ b/app/models/open_graph_cache.rb
@@ -33,7 +33,7 @@ class OpenGraphCache < ApplicationRecord
   end
 
   def fetch_and_save_opengraph_data!
-    uri = URI.parse(url.start_with?("http") ? url : "https://#{url}")
+    uri = URI.parse(url.start_with?("http") ? url : "http://#{url}")
     uri.normalize!
     object = OpenGraphReader.fetch!(uri)
 
--- a/app/views/terms/default.haml
+++ b/app/views/terms/default.haml
@@ -142,7 +142,7 @@
             Source code and materials
 
           %p
-            This Service runs on a diaspora* social network server. This source code is licensed under an <a href="http://www.gnu.org/licenses/agpl-3.0.html">AGPLv3</a> license which means you are allowed to and even encouraged to take the source code, modify it and use it.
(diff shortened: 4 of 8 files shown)


app/assets/javascripts/app/views/help_view.js
```diff
@@ -14,7 +14,7 @@ app.views.Help = app.views.StaticContentView.extend({
 
   initialize : function() {
     this.GETTING_HELP_SUBS = {
-      getting_started_a: {tutorial_series: this.linkHtml("http://diasporafoundation.org/getting_started/sign_up", Diaspora.I18n.t("getting_started_tutorial"))},
+      getting_started_a: {tutorial_series: this.linkHtml("https://diasporafoundation.org/getting_started/sign_up", Diaspora.I18n.t("getting_started_tutorial"))},
```
diaspora-code-review, line 17: - Identifier 'getting_started_a' is not in camel case.
- Identifier 'tutorial_series' is not in camel case.

app/assets/javascripts/app/views/help_view.js
```diff
@@ -27,8 +27,8 @@ app.views.Help = app.views.StaticContentView.extend({
     this.POSTS_AND_POSTING_SUBS = {
       post_report_a: {community_guidelines: this.linkHtml("https://diasporafoundation.org/community_guidelines", Diaspora.I18n.t("community_guidelines"))},
       format_text_a: {
-        markdown: this.linkHtml("http://diasporafoundation.org/formatting", Diaspora.I18n.t( 'markdown' )),
-        here: this.linkHtml("http://daringfireball.net/projects/markdown/syntax", Diaspora.I18n.t( 'here' ))
+        markdown: this.linkHtml("https://diasporafoundation.org/formatting", Diaspora.I18n.t( 'markdown' )),
```
diaspora-code-review, line 30: - There should be no spaces inside this paren.
- Strings must use doublequote.

app/assets/javascripts/app/views/help_view.js
```diff
@@ -27,8 +27,8 @@ app.views.Help = app.views.StaticContentView.extend({
...
+        here: this.linkHtml("https://daringfireball.net/projects/markdown/syntax", Diaspora.I18n.t( 'here' ))
```
diaspora-code-review, line 31: - There should be no spaces inside this paren.
- Strings must use doublequote.

app/helpers/meta_data_helper.rb
```diff
@@ -5,7 +5,7 @@ module MetaDataHelper
   include ActionView::Helpers::TagHelper
 
   def og_prefix
-    'og: http://ogp.me/ns# article: http://ogp.me/ns/article# profile: http://ogp.me/ns/profile#'
+    'og: https://ogp.me/ns# article: https://ogp.me/ns/article# profile: https://ogp.me/ns/profile#'
```
diaspora-code-review, line 8: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

config/routes.rb
```diff
@@ -208,7 +208,7 @@
   get 'help/:topic' => 'help#faq'
 
   #Protocol Url
-  get 'protocol' => redirect("http://wiki.diasporafoundation.org/Federation_Protocol_Overview")
+  get 'protocol' => redirect("https://wiki.diasporafoundation.org/Federation_Protocol_Overview")
```
diaspora-code-review, line 211: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/helpers/meta_data_helper.rb
denschub: https://github.com/diaspora/diaspora/issues/8198#issuecomment-775280343

> * HTML namespaces like the opengraph example: No need - those URLs are for spec-purposes only, they will not get queried by a browser.

app/models/open_graph_cache.rb
```diff
@@ -33,7 +33,7 @@ def self.find_or_create_by(opts)
   end
 
   def fetch_and_save_opengraph_data!
-    uri = URI.parse(url.start_with?("http") ? url : "http://#{url}")
+    uri = URI.parse(url.start_with?("http") ? url : "https://#{url}")
```
denschub, line 36: This assumes that every URL ever posted supports TLS - which is not a good assumption to make.

config/locales/diaspora/en.yml
```diff
@@ -3,7 +3,7 @@
 #   the COPYRIGHT file.
 
 # Sample localization file for English. Add more files in this directory for other locales.
-# See http://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting points.
+# See https://github.com/svenfuchs/rails-i18n/tree/master/rails%2Flocale for starting points.
```
denschub, line 6: No need to change this, there is no user-impact here whatsoever.

config/locales/diaspora/en.yml
```diff
@@ -456,7 +456,7 @@ en:
...
+      embed_multimedia_a: "You can usually just paste the URL (e.g. https://www.youtube.com/watch?v=nnnnnnnnnnn ) into your post and the video or audio will be embedded automatically. The sites supported include: YouTube, Vimeo, SoundCloud, Flickr and a few more. diaspora* uses oEmbed for this feature. If you post a direct link to an audio or video file, diaspora* will embed it using standard HTML5 player. We’re supporting more media sources all the time. Remember to always post simple, full links – no shortened links; no operators after the base URL – and give it a little time before you refresh the page after posting for seeing the preview."
```
denschub, line 459: This is a piece out of the documentation. That URL is never loaded, and it's not even clickable. It's also save to assume that no user would ever copy-paste that URL, because it's not a valid video URL. Changing this invalidates all translations of that string, and there is no reason to do that.

skyschub (changes requested):
This is missing a couple of things.

- https://github.com/diaspora/diaspora/blob/1f329991259dcaf201d3b0ca1e5ca3be306ce10b/app/views/streams/main_stream.html.haml#L66
- https://github.com/diaspora/diaspora/blob/1f329991259dcaf201d3b0ca1e5ca3be306ce10b/app/views/streams/main_stream.html.haml#L97
- https://github.com/diaspora/diaspora/blob/1f329991259dcaf201d3b0ca1e5ca3be306ce10b/app/assets/javascripts/app/helpers/locations.js#L9-L10
- https://github.com/diaspora/diaspora/blob/1f329991259dcaf201d3b0ca1e5ca3be306ce10b/app/assets/javascripts/app/helpers/locations.js#L17-L19
- https://github.com/diaspora/diaspora/blob/1f329991259dcaf201d3b0ca1e5ca3be306ce10b/app/views/terms/default.haml#L145

and potentially a few others (have not looked too closely). In addition, it changes some things that have no real user impact, or no impact at all, and for the sake of not touching things just because(tm), let's avoid that.

app/assets/javascripts/app/views/help_view.js
tclaus: I think, this can be ignored, since the identifier was not changed

app/helpers/meta_data_helper.rb
tclaus: These URLs provided the reason for this PR and Issue.
BTW, the ogp URLs are prefixes for the og: Protocols. In their documentation, every URL is a https - URL, so even this is not a technical need, changing (as many as possible) these URLs modernize the Diaspora output. Don't show a skeptical user, that the diaspora devs don't care about secure protocols. 

I herby declare a request to change this.

tclaus:
> This is missing a couple of things.
> * https://github.com/diaspora/diaspora/blob/1f329991259dcaf201d3b0ca1e5ca3be306ce10b/app/assets/javascripts/app/helpers/locations.js#L17-L19

"korona.geog.uni-heidelberg.de" - This is a dead link. (because of Korona?) 
What to do with this? 

> 
> and potentially a few others (have not looked too closely). In addition, it changes some things that have no real user impact, or no impact at all, and for the sake of not touching things just because(tm), let's avoid that.

Yes, a few others.. but I don't really now what's really needed. For me diaspora should change as much of these as possible. 

One of my favorites is this "create link" - prompt. But I was unable to find the generation if this. 

Bildschirmfoto 2021-02-21 um 09 34 18

config/locales/diaspora/en.yml
tclaus: I'll change this back. On the other Hand lets consider not to show the user old-fashioned url protocols as examples, even there is a bit effort in changing translations

app/views/terms/default.haml
```diff
@@ -142,7 +142,7 @@
             Source code and materials
 
           %p
-            This Service runs on a diaspora* social network server. This source code is licensed under an <a href="http://www.gnu.org/licenses/agpl-3.0.html">AGPLv3</a> license which means you are allowed to and even encouraged to take the source code, modify it and use it.
+            This Service runs on a diaspora* social network server. This source code is licensed under an <a href="https://www.gnu.org/licenses/agpl-3.0.html">AGPLv3</a> license which means you are allowed to and even encouraged to take the source code, modify it and use it.
```
diaspora-code-review, line 145: Line is too long. [275/120]

app/assets/javascripts/app/views/help_view.js
denschub: If you touch it, you own it. Which, by the way, is exactly the reason why I'm insisting so hard to only change things where it's needed. As annoying as it is, if we fix the existing codestyle issues step by step, we'll eventually end up with a consistent codestyle everywhere. Here, the scope is minimal - renaming those variable names and renaming their usages in one single template file is enough.

app/helpers/meta_data_helper.rb
denschub: > These URLs provided the reason for this PR and Issue.

That still doesn't mean it makes any sense. ;)

They changed their docs [some time in July last year](http://web.archive.org/web/20200702012303/https://ogp.me/) and didn't even note it somewhere because it has absolutely no relevance whatsoever.

Nobody cares about namespace URLs. They will not show up in your browser, they will not show up as a mixed content warning, they won't even show up in the devtools. The only reasonable response to someone looking at the HTML source and asking why there are http links is: "That absolutely has no relevance whatsoever.". I mean, what's next? People complaining that all our svg's have `xmlns="http://www.w3.org/2000/svg"` and `http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd` as their dtd?! Like, c'mon.

But fine, I don't care enough to spend any more time on discussing this. If you want to change it, sure.

denschub:
> "korona.geog.uni-heidelberg.de" - This is a dead link. (because of Korona?)
> What to do with this?

True, [they deprecated that tileserver in 2019](http://k1z.blog.uni-heidelberg.de/2019/02/08/the-future-of-openmapsurfers-tiles/). Ignore this for now, but I'd appreciate if you could file a bug to remove those tiles, they're no longer working.



> Yes, a few others.. but I don't really now what's really needed. For me diaspora should change as much of these as possible.

As I already commented in #8198: Change anything that is affecting users. All links than can be clicked or somehow followed by users. Changing anything else creates nothing but noise with no improvement whatsoever.



> One of my favorites is this "create link" - prompt. But I was unable to find the generation if this.

That's coming from an upstream dependency: https://github.com/refactory-id/bootstrap-markdown/blob/a496d34b9bd34451c8315a850472f794c8df7d53/js/bootstrap-markdown.js#L1182 which appears to be dead.

app/assets/javascripts/app/views/help_view.js
tclaus: Whats about 47 language files these identifiers are also referenced? 
Is it a good idea just to change this identifier and let the direct 'neighbors' untouched? 

Bildschirmfoto 2021-02-21 um 18 05 24

denschub: These keys have no reference to any language file. They're merely assignments to variables, they're passed as the variable object [here](https://github.com/diaspora/diaspora/blob/1f329991259dcaf201d3b0ca1e5ca3be306ce10b/app/assets/javascripts/app/views/help_view.js#L67), and as long as you change them both here and [in the template they're used](https://github.com/diaspora/diaspora/blob/1f329991259dcaf201d3b0ca1e5ca3be306ce10b/app/assets/templates/faq_getting_help_tpl.jst.hbs), you're fine. That's just an assignment for later use in the same unit, not relevant to what's stored in the language file.

`Diaspora.I18n.t()` is the function actually loading things from the language file. Changing those would be very bad indeed, but the bot is not complaining about that. :) 

> Is it a good idea just to change this identifier and let the direct 'neighbors' untouched?

It's not a good idea, no! Most likely the bot would yell at you anyway, because you're changing context lines.

denschub: Ah wait, you're talking about `tutorial_series`, not `getting_started_a`. I actually completely missed that.

Yeah, you're right, that's bad. Okay, fine, let's **not** change the keys then, but please enclose the entire block with a 

```js
/* eslint-disable camelcase */
...
/* eslint-enable camelcase */
```

To declare that we're violating camelcase here on purpose, and to avoid further but complaints for changes in the future! :)

denschub:
Squashed and merged as d7941230a4a420f35bb11214f8bb4ec9314e68ab.

Issue #7333, opened 2017-02-17
Use the default aspect in mobile version


ldidry:
I choose the "public" aspect by default in my settings, but the mobile version uses the old behavior, sharing with all aspect by default. It should respect the settings.

Diaspora* version: 0.6.3.0
Diaspora* pod: framasphere.org
Phone: Nexus 6P, Android 7.1.1
Browser: Firefox Nightly 54.0a1

Flaburgan:
As the default aspect can be multiple aspects, this is blocked by https://github.com/diaspora/diaspora/issues/7080 we need to be able to select multiple aspects in the mobile publisher first.

Flaburgan:
@pskosinski if you want to work on this please open a PR that fixes #7080 first, and then a new one for this issue. Smaller PRs are easier to review.

SuperTux88:
I also noticed that the section in the settings to set the default aspects is hidden on mobile (makes sense as it isn't working on mobile at the moment anyway), so this should also be changed when this gets fixed.

PR #8217, opened 2021-03-06
7080 multi select on aspects on mobile

tclaus:
Fixes Issue #7080 and updating Pr #7602 from sebastienadam .

Multiple aspects in mobile views. 

Moved Aspects in mobile view in one line with share and upload photos button, as in desktop view. 
This is what aspects selector look on mobile now: 
Bildschirmfoto 2021-03-06 um 20 27 39

**Notes on failing Styling Tests**: The one styling was about renaming a -style issue- to change _ to - and a follow-up not to use IDs at general, so this styling issue was there forever. 

**Notes on Tests**: There are timeouts, but the cases run green. No Idea what to do about this.

### DIFF
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -658,24 +658,18 @@ form#update_profile_form {
   .submit_block { margin-bottom: 20px; }
 }
 
-select#user_language,
-select#user_color_theme,
-#user_auto_follow_back_aspect_id {
-  padding: 3px;
-}
-
 .hero-unit-mobile {
-  padding: 10px;
-  font-size: 14px;
-  line-height: 18px;
-  color: inherit;
   background-color: $background-grey;
   border-radius: 10px;
+  color: inherit;
+  font-size: 14px;
+  line-height: 18px;
+  padding: 10px;
 }
 
 .search-mobile {
-  text-align: center;
   padding-top: 30px;
+  text-align: center;
 }
 
 input#q.search {
@@ -717,8 +711,9 @@ input#q.search {
   top: 0;
 }
 
-#publisher_mobile {
+#publisher-mobile {
   float: right;
+  margin-left: 5px;
 }
 
 #file-upload-publisher {
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -676,28 +676,10 @@ input#q.search {
   margin-bottom: 20px;
 }
 
-@mixin aspect-dropdown-link() {
-  $link-text-color: #333;
-
-  a {
-    cursor: pointer;
-    padding-left: 10px;
-
-    .text {
-      color: $link-text-color;
-      font-size: 128%;
-    }
-
-    &:hover {
-      background: $background-grey;
-    }
-  }
-}
-
 .aspect-dropdown {
   li {
     @include selectable-list;
-    @include aspect-dropdown-link;
+    @include aspect-dropdown-link(128%);
 
     .status-indicator {
       display: inline-block;
--- a/app/assets/stylesheets/mobile/mobile.scss
+++ b/app/assets/stylesheets/mobile/mobile.scss
@@ -658,7 +658,9 @@ form#update_profile_form {
   .submit_block { margin-bottom: 20px; }
 }
 
-select#user_language, select#user_color_theme, #user_auto_follow_back_aspect_id, #aspect_ids_ {
+select#user_language,
+select#user_color_theme,
+#user_auto_follow_back_aspect_id {
   padding: 3px;
 }
 
@@ -680,14 +682,39 @@ input#q.search {
   margin-bottom: 20px;
 }
 
-select#aspect_ids_ {
-  width: auto !important;
-  float: right;
-  margin: 0px;
+@mixin aspect-dropdown-link() {
+  $link-text-color: #333;
+
+  a {
+    cursor: pointer;
+    padding-left: 10px;
+
+    .text {
+      color: $link-text-color;
+      font-size: 128%;
+    }
+
+    &:hover {
+      background: $background-grey;
+    }
+  }
+}
+
+.aspect-dropdown {
+  li {
+    @include selectable-list;
+    @include aspect-dropdown-link;
+
+    .status-indicator {
+      display: inline-block;
+      height: 14px;
+      width: 19px;
+    }
+  }
 }
 
 #file-upload-spinner {
-  top: 0px;
+  top: 0;
 }
 
 #publisher_mobile {
--- a/app/assets/stylesheets/mobile/settings.scss
+++ b/app/assets/stylesheets/mobile/settings.scss
@@ -57,3 +57,9 @@
     }
   }
 }
+
+.user-language,
+.user-color-theme,
+.aspect-id {
+  padding: 3px;
+}
--- a/app/assets/stylesheets/mobile/settings.scss
+++ b/app/assets/stylesheets/mobile/settings.scss
@@ -56,10 +56,8 @@
       width: 50px;
     }
   }
-}
 
-.user-language,
-.user-color-theme,
-.aspect-id {
-  padding: 3px;
+  select {
+    padding: 3px;
+  }
 }
--- a/app/views/aspects/_aspect_dropdown.mobile.haml
+++ b/app/views/aspects/_aspect_dropdown.mobile.haml
@@ -1,0 +1,36 @@
+-# Note: all_aspects is a global in the ApplicationController
+:ruby
+  dropdown_css = {"data-toggle" => "dropdown"}
+  if current_user.getting_started?
+    dropdown_css[:title] = popover_with_close_html("2. #{t('shared.public_explain.control_your_audience')}")
+    dropdown_css["data-content"] = t("shared.public_explain.visibility_dropdown")
+  end
+
+.btn-group.aspect_dropdown
+  %button.btn.btn-default.dropdown-toggle{dropdown_css}
+    %i.entypo-lock.small#visibility-icon
+    %span.text
+      = t("all_aspects")
+    %span.caret
+  %ul.dropdown-menu.pull-right{unSelectable: "on"}
+
+    %li.public.radio{"data-aspect_id" => "public"}
+      %a
+        %span.status_indicator
+          %i.glyphicon.glyphicon-ok
+        %span.text
+          = t("public")
+    %li.all_aspects.radio{"data-aspect_id" => "all_aspects"}
+      %a
+        %span.status_indicator
+          %i.glyphicon.glyphicon-ok
+        %span.text
+          = t("all_aspects")
+    %li.divider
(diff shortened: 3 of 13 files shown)


app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -658,39 +658,60 @@ form#update_profile_form {
...
+  }
 }
 
 #file-upload-spinner {
-  top: 0px;
+  top: 0;
 }
 
-#publisher_mobile {
+#publisher-mobile {
```
diaspora-code-review, line 714: Avoid using id selectors

tclaus: Old existing Selector - changed '_ 'to '-'

Flaburgan: Yes you can ignore this one.

tclaus:
@Flaburgan Whats about the time-out tests? Can these be triggered again? Is it something what has its cause in the code?

Flaburgan:
> @Flaburgan Whats about the time-out tests? Can these be triggered again? Is it something what has its cause in the code?

I was checking that yesterday, when we were using Travis I was able to re-run builds, but looks like I can't do that with Github Actions, I don't know if that's a permission problem or a feature not available, can you tell @SuperTux88 ?

SuperTux88:
I can re-trigger it, but if all 4 cucumber tests failed with the same test, it's probably a real error that doesn't go away with re-running it, so there is no point of triggering the tests again with the current state (and the timeout problem with jasmine can just be ignored for this run, it will re-run again on next commit anyway).

tclaus:
> , it will re-run again on next commit anyway).

OK, I will check this again. Thanks for the analysis.

tclaus:
All tests runs well locally - so if any of these still fail, I request some help.

features/step_definitions/web_steps.rb
```diff
@@ -31,7 +31,7 @@ def with_scope(locator)
 
 When /^(?:|I )press "([^"]*)"(?: within "([^"]*)")?$/ do |button, selector|
   with_scope(selector) do
-    click_button(button)
+    find_button(button).trigger('click')
```
diaspora-code-review, line 34: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

features/support/env.rb
```diff
@@ -23,8 +23,13 @@
 
 Capybara.server = :webrick
 
+options = {
+  window_size: [1920, 1080],
+  timeout: 30
```
diaspora-code-review, line 28: Layout/HashAlignment: Align the keys and values of a hash literal if they span more than one line.

features/support/env.rb
```diff
@@ -23,6 +23,11 @@
 
 Capybara.server = :webrick
 
+options = {
```
diaspora-code-review, line 26: Lint/UselessAssignment: Useless assignment to variable - `options`.

features/mobile/posts_from_main_page.feature
```diff
@@ -21,8 +21,10 @@ Feature: posting from the mobile main page
   Scenario: post and delete some text
     Given I visit the mobile publisher page
     And I append "I am eating yogurt" to the publisher
-    And I select "Unicorns" from "aspect_ids_"
-    And I press "Share"
+    And I press the aspect dropdown
+    And I toggle the aspect "Unicorns"
```
Flaburgan, line 25: Could you please add a test where you post to two different aspects and check that the post is visible in these two and only these two?

Flaburgan (changes requested):
Great job, it is working well and the design is nice. Only a comment about the test coverage and then we should be good to go.

features/mobile/posts_from_main_page.feature
tclaus: Oui.

tclaus: @Flaburgan Can it be that there is no test for setting more than one aspect at same time? Even in Desktop-Mode?

Flaburgan: They are: https://github.com/diaspora/diaspora/blob/develop/features/desktop/posts_from_main_page.feature#L58

tclaus: Help requested - cucumber tests still won't work. I don't get test to work - the Aspects button will (in test) not expand, or aspects are not 'visible' However - on real browser everything seems to work.

features/step_definitions/aspects_steps.rb
```diff
@@ -112,6 +116,12 @@ def aspect_dropdown_visible?
   click_aspect_dropdown
 end
 
+When /^I press the share button$/ do
+  # There were issues 'clicking' the share button on mobile
+  click_share_button
+  sleep 1 # Wait a little bit - especially on mobile views shaing dont works every time
```
jhass, line 122: Ideally there's some UI change we can observe and wait for, something like `expect(page).to have_selector("#dropdown.visible")`, did you have any luck with something like this? :)

app/views/aspects/_aspect_dropdown.mobile.haml
```diff
@@ -0,0 +1,36 @@
+-# Note: all_aspects is a global in the ApplicationController
+:ruby
+  dropdown_css = {"data-toggle" => "dropdown"}
+  if current_user.getting_started?
+    dropdown_css[:title] = popover_with_close_html("2. #{t('shared.public_explain.control_your_audience')}")
+    dropdown_css["data-content"] = t("shared.public_explain.visibility_dropdown")
+  end
```
jhass, line 7: I went through the sign up flow on mobile and even reaching the publisher page already had `getting_started` be `false` on my new user. Manually setting it back to `true` showed for this to have no effect. Let's drop this.

tclaus:
@Flaburgan @jhass  I added cucumber tests. 
Note: I went into serous problems creating and running these tests because I had to learn to run `rake assets:precompile ` whenever I change templates before running cucumber tests! Cucumber then runs on something 'old'.

tclaus:
Ping @jhass / @Flaburgan 
I added tests and fixed code.  Can you approve it now?
(The whining pronto/scss is a "Avoid using id selectors" which is from the legacy code and built more deep into architecture to change to in passing by)

tclaus:
Issues are solved, tests added, everything is green. Is there anything more to do for an approval? 
@Flaburgan , @jhass

SuperTux88 (changes requested):
There are some mostly minor things I found, but I think it should be quick to fix them. Can you please also squash some of the commits together (style fixes, cleanup and reverts should just be squashed with the commit where it was initially changed, so we only end up with commits with actual changes and meaningful commit messages).

Sidenote: this doesn't fix #7333 yet, and that is fine and can be done in a separate PR. But I think it should be fixed in the same release as this one gets merged, otherwise it will be even more confusing for users as it already is.

app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -658,40 +658,62 @@ form#update_profile_form {
   .submit_block { margin-bottom: 20px; }
 }
 
-select#user_language, select#user_color_theme, #user_auto_follow_back_aspect_id, #aspect_ids_ {
-  padding: 3px;
-}
-
```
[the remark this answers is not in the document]
tclaus: I have added this again, except the aspect_ids_. A bad way to define styles for programmatic built Ids.. I found the f.select .. part which builds this Ids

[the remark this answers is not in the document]
SuperTux88: Not really the scope of this PR, but maybe it can be replaced by just a class then instead, which then doesn't need to be programmatically built and can be easier found? Or just select all `select` elements on the settings page (no idea why it's only these specific elements)? :thinking:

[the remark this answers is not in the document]
tclaus: I changed it to class settings. But I do not "see" any change when a select element has a padding.

app/assets/stylesheets/mobile/mobile.scss
```diff
@@ -658,40 +658,62 @@ form#update_profile_form {
...
+    .text {
+      color: $link-text-color;
+      font-size: 128%;
+    }
+
+    &:hover {
+      background: $background-grey;
+    }
+  }
+}
```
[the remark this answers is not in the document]
tclaus: This is always inside a mixin. I don't change it (for know)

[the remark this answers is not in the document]
SuperTux88: Yes, it's a mixin, but it's almost the same mixin defined twice, and both time it's only used once. Defining it once and using it twice would be better (and the one little difference can then just be added as an override where it needs to be different).

app/assets/stylesheets/mobile/mobile.scss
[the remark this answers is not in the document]
SuperTux88: With the `padding: 3px`:
image
Without the `padding: 3px`:
image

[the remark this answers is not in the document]
tclaus: Ah, now I see it. Thanks.

app/assets/stylesheets/mobile/mobile.scss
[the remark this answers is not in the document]
tclaus: @SuperTux88 I moved this into a nice mixin.

app/assets/stylesheets/mobile/settings.scss
```diff
@@ -57,3 +57,9 @@
     }
   }
 }
+
+.user-language,
+.user-color-theme,
+.aspect-id {
```
SuperTux88, line 63: With "replaced by just a class" I meant using one class for all of them together, so you don't need a separate class for each element. Like for example `.settings-dropdown` or something like that. Or as said, it probably also works if you just select all the `select` elements on the settings page, so no extra class is needed at all? That way it would also automatically work if for some reason we would add another dropdown which should also be styled that way on mobile.

```suggestion
.settings-container select {
```

(or better move it with only `select` inside the `.settings-container` block above :) )

SuperTux88 (changes requested):
Looks good :+1: only one last little thing.

app/assets/stylesheets/mobile/settings.scss
tclaus: I now have set the property to all selects inside the settings-container

app/views/users/_edit.haml
```diff
@@ -71,7 +71,7 @@
         %h3= t(".change_color_theme")
         = form_for "user", url: edit_user_path, html: {method: :put} do |f|
           .form-inline.clearfix
-            = f.select :color_theme, available_color_themes, {}, class: "form-control form-group"
+            = f.select :color_theme, available_color_themes, {}, class: "form-control form-group color-theme"
```
SuperTux88, line 74: There is still a leftover class here.

tclaus: Removed, thanks for being such an eagle-eye.

SuperTux88 (approved):
Looks good now :cookie:

PR #8224, opened 2021-03-21
Installing missing libffi6 in Git CI Action

tclaus:
On Latest OS Version v6 of libffi is missing. This add it back to the CI


tclaus:
Back to the workshop

Issue #6640, opened 2016-01-15
Add possibility to blacklist pods from the admin


jaywink:
At some point the first rogue pod will come up and spam 10000 spam messages a day to all the pods it can find, with a gazillion of hashtags. Podmins can currently block things with firewalls of course, but a simple list in the admin would be better for podmins to handle.

SansPseudoFix:
Pods or users from remote pods?

jaywink:
Pods. I guess users would be a valid use case too, but a different feature from this.

piraz:
Maybe users could report a pod like a post or comment.

PR #8228, opened 2021-04-04
Adds ability to block a pod in Admin-View

tclaus:
It uses the already existing 'blocked' information of Pods. 

On Adminview a 'blocked/unblock' UI element is added. 
A blocked pod is then filtered out in any stream views. No Database element is deleted or changed. 

This Feature was requested in Issue #6640 and also discussed/requested in https://discourse.diasporafoundation.org/t/add-better-spam-controls/296

Is this a viable approach? 
On https://societas.online this is already rolled out and ready to test. 

Block / Unblock
Bildschirmfoto 2021-04-04 um 13 32 20

Added blocked pod count in pod list - header
Bildschirmfoto 2021-04-04 um 13 31 56

Any remarks / ideas or help is appreciated.

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -336,6 +336,13 @@ class Person < ApplicationRecord
 
   # discovery (webfinger)
   def self.find_or_fetch_by_identifier(diaspora_id)
+    # pod blocked?
+    if diaspora_handle_from_blocked_pod?(diaspora_id)
+      logger.info "Rejecting #{diaspora_id}, from blocked pod"
+      raise DiasporaFederation::Discovery::DiscoveryError,
+            "Failed discovery for #{diaspora_id}: blocked pod"
+    end
+
     # exiting person?
     person = by_account_identifier(diaspora_id)
     return person if person.present? && person.profile.present?
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -336,7 +336,6 @@ class Person < ApplicationRecord
 
   # discovery (webfinger)
   def self.find_or_fetch_by_identifier(diaspora_id)
-
     return nil if diaspora_handle_from_blocked_pod?(diaspora_id)
 
     # exiting person?
@@ -348,7 +347,6 @@ class Person < ApplicationRecord
     DiasporaFederation::Discovery::Discovery.new(diaspora_id).fetch_and_save
 
     by_account_identifier(diaspora_id)
-
   rescue DiasporaFederation::Discovery::InvalidDocument
     logger.info "#{diaspora_id} returns not as a valid document"
     nil
@@ -393,6 +391,12 @@ class Person < ApplicationRecord
     self
   end
 
+  def self.diaspora_handle_from_blocked_pod?(diaspora_handle)
+    host = diaspora_handle.split("@").last
+    pod = Pod.find_by(host: host)
+    !pod.nil? && pod.blocked
+  end
+
   private
 
   def fix_profile
@@ -409,11 +413,4 @@ class Person < ApplicationRecord
     diaspora_id = Person.where(guid: guid).where.not(diaspora_handle: diaspora_handle).pluck(:diaspora_handle).first
     errors.add(:base, "Person with same GUID already exists: #{diaspora_id}") if diaspora_id
   end
-
-  def self.diaspora_handle_from_blocked_pod?(diaspora_handle)
-    host = diaspora_handle.split('@').last
-    pod = Pod.find_by_host(host)
-    return !pod.nil? && pod.blocked
-  end
-
 end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -36,7 +36,7 @@ class Person < ApplicationRecord
   end
 
   has_many :contacts, :dependent => :destroy # Other people's contacts for this person
-  has_many :posts, :foreign_key => :author_id, :dependent => :destroy # This person's own posts
+  has_many :posts, foreign_key: :author_id, inverse_of: :author, dependent: :destroy # This person's own posts
   has_many :photos, :foreign_key => :author_id, :dependent => :destroy # This person's own photos
   has_many :comments, :foreign_key => :author_id, :dependent => :destroy # This person's own comments
   has_many :likes, foreign_key: :author_id, dependent: :destroy # This person's own likes
@@ -257,6 +257,9 @@ class Person < ApplicationRecord
 
     query = query.where(contacts: {sharing: true, receiving: true}) if mutual
 
+    # return only unblocked or local persons
+    query = query.includes(:pod).where("pods.blocked = false or pods.blocked is null").references(:pod)
+
     query.where(closed_account: false)
          .order([Arel.sql("contacts.user_id IS NULL"), "profiles.last_name ASC", "profiles.first_name ASC"])
   end
@@ -335,13 +338,24 @@ class Person < ApplicationRecord
   def self.find_or_fetch_by_identifier(diaspora_id)
     # exiting person?
     person = by_account_identifier(diaspora_id)
-    return person if person.present? && person.profile.present?
+    if person.present? && person.profile.present?
+      # Return if pod is not blocked
+      return person if person.pod.nil?
+      return person unless person.pod.blocked
 
+      nil
+    end
     # create or update person from webfinger
     logger.info "webfingering #{diaspora_id}, it is not known or needs updating"
     DiasporaFederation::Discovery::Discovery.new(diaspora_id).fetch_and_save
+    person = by_account_identifier(diaspora_id)
+    return person if person.pod.nil?
+    return person unless person.pod.blocked
 
-    by_account_identifier(diaspora_id)
+    nil
+  rescue DiasporaFederation::Discovery::InvalidDocument
+    logger.info "#{diaspora_id} returns not as a valid document"
+    nil
   end
 
   def self.by_account_identifier(diaspora_id)
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -336,23 +336,19 @@ class Person < ApplicationRecord
 
   # discovery (webfinger)
   def self.find_or_fetch_by_identifier(diaspora_id)
+
+    return nil if diaspora_handle_from_blocked_pod?(diaspora_id)
+
     # exiting person?
     person = by_account_identifier(diaspora_id)
-    if person.present? && person.profile.present?
-      # Return if pod is not blocked
-      return person if person.pod.nil?
-      return person unless person.pod.blocked
+    return person if person.present? && person.profile.present?
 
-      nil
-    end
     # create or update person from webfinger
     logger.info "webfingering #{diaspora_id}, it is not known or needs updating"
     DiasporaFederation::Discovery::Discovery.new(diaspora_id).fetch_and_save
-    person = by_account_identifier(diaspora_id)
-    return person if person.pod.nil?
-    return person unless person.pod.blocked
 
-    nil
+    by_account_identifier(diaspora_id)
+
   rescue DiasporaFederation::Discovery::InvalidDocument
     logger.info "#{diaspora_id} returns not as a valid document"
     nil
@@ -413,4 +409,11 @@ class Person < ApplicationRecord
     diaspora_id = Person.where(guid: guid).where.not(diaspora_handle: diaspora_handle).pluck(:diaspora_handle).first
     errors.add(:base, "Person with same GUID already exists: #{diaspora_id}") if diaspora_id
   end
+
+  def self.diaspora_handle_from_blocked_pod?(diaspora_handle)
+    host = diaspora_handle.split('@').last
+    pod = Pod.find_by_host(host)
+    return !pod.nil? && pod.blocked
+  end
+
 end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -336,8 +336,6 @@ class Person < ApplicationRecord
 
   # discovery (webfinger)
   def self.find_or_fetch_by_identifier(diaspora_id)
-    return nil if diaspora_handle_from_blocked_pod?(diaspora_id)
-
     # exiting person?
     person = by_account_identifier(diaspora_id)
     return person if person.present? && person.profile.present?
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -6,6 +6,7 @@
 
 class Person < ApplicationRecord
   include Diaspora::Fields::Guid
+  include Diaspora::Federation
 
   # NOTE API V1 to be extracted
   acts_as_api
--- a/app/workers/fetch_webfinger.rb
+++ b/app/workers/fetch_webfinger.rb
@@ -10,6 +10,8 @@ module Workers
 
     def perform(account)
       person = Person.find_or_fetch_by_identifier(account)
+      return if person.nil?
+      return if person.pod.blocked
 
       # also, schedule to fetch a few public posts from that person
       Diaspora::Fetcher::Public.queue_for(person)
--- a/app/workers/receive_base.rb
+++ b/app/workers/receive_base.rb
@@ -26,6 +26,7 @@ module Workers
            DiasporaFederation::Salmon::InvalidEncoding,
            Diaspora::Federation::AuthorIgnored,
            Diaspora::Federation::InvalidAuthor,
+            Diaspora::Federation::PodBlocked,
            Diaspora::Federation::RecipientClosed => e
       logger.warn "don't retry for error: #{e.class}"
     end
--- a/app/workers/receive_base.rb
+++ b/app/workers/receive_base.rb
@@ -26,7 +26,7 @@ module Workers
            DiasporaFederation::Salmon::InvalidEncoding,
            Diaspora::Federation::AuthorIgnored,
            Diaspora::Federation::InvalidAuthor,
-            Diaspora::Federation::PodBlocked,
+           Diaspora::Federation::PodBlocked,
            Diaspora::Federation::RecipientClosed => e
       logger.warn "don't retry for error: #{e.class}"
     end
--- a/config/initializers/diaspora_federation.rb
+++ b/config/initializers/diaspora_federation.rb
@@ -15,48 +15,44 @@ DiasporaFederation.configure do |config|
   config.define_callbacks do
     on :fetch_person_for_webfinger do |diaspora_id|
       person = Person.where(diaspora_handle: diaspora_id, closed_account: false).where.not(owner: nil).first
-      if person
-        unless person.pod&.blocked
-          DiasporaFederation::Discovery::WebFinger.new(
+      unless person.nil? || person.pod&.blocked
+        DiasporaFederation::Discovery::WebFinger.new(
+          {
+            acct_uri:      "acct:#{person.diaspora_handle}",
+            hcard_url:     AppConfig.url_to(DiasporaFederation::Engine.routes.url_helpers.hcard_path(person.guid)),
+            seed_url:      AppConfig.pod_uri,
+            profile_url:   person.profile_url,
+            atom_url:      person.atom_url,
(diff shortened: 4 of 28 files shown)


lib/evil_query.rb
```diff
@@ -63,7 +65,10 @@ def make_relation!
       logger.debug("[EVIL-QUERY] make_relation!")
       post_ids = aspects_post_ids! + ids!(followed_tags_posts!) + ids!(mentioned_posts)
       post_ids += ids!(community_spotlight_posts!) if @include_spotlight
-      Post.where(:id => post_ids)
+      logger.debug("Make blocked query")
+      Post.left_outer_joins(author: [:pod])
+          .where(id: post_ids)
+          .where("(pods.blocked = false or pods.blocked is null)")
```
jhass, line 71: So here's the hard part about this :/ Our main stream query already is notoriously complex and thus slow. A lot of optimization went into it already in order to make it not too terribly slow. Extending it needs to be done very considerate and carefully, ideally doing some performance benchmarking on it, but at least comparing query plans before and after.

Any of that happened here? :)

Flaburgan: Actually, why do we register posts from blocked pod at all? We could just drop them. That would mean no way back of course but it would be much simpler.

tclaus: @jhass I checked the queries: 
Normal public posts query looks like this: 
```
SELECT  "posts".* FROM "posts" 
    WHERE "posts"."public" = true 
        AND (posts.created_at < '2021-04-11 17:49:52') 
        AND "posts"."type" IN ('StatusMessage', 'Reshare') AND (posts.author_id NOT IN (6,2)) 
    ORDER BY posts.created_at DESC, posts.id DESC LIMIT 15  
```

This results in this explanation (unsuprisingly): 
Bildschirmfoto 2021-04-11 um 20 13 45

Query for Blocked posts: 
```
SELECT  "posts".* FROM "posts" 
    LEFT OUTER JOIN "people" ON "people"."id" = "posts"."author_id" 
    LEFT OUTER JOIN "pods" ON "pods"."id" = "people"."pod_id" 
    WHERE (pods.blocked is null or pods.blocked = false) AND "posts"."public" = true 
        AND (posts.created_at < '2021-04-11 17:51:47') 
        AND "posts"."type" IN ('StatusMessage', 'Reshare') AND (posts.author_id NOT IN (6,2)) 
    ORDER BY posts.created_at DESC, posts.id DESC LIMIT 15 
```
Explain: 
Bildschirmfoto 2021-04-11 um 20 20 22


Its, well-  more, but indexed. Is this a _real_ problem? 
The real bottleneck and time-consuming is IMHO the tons of extra queries done after this, beginning with 'like_posts_for_stream', Photos, Mentions, polls and so on. (Likes could also be joined to the posts query)

tclaus: @Flaburgan 
> Actually, why do we register posts from blocked pod at all? We could just drop them. That would mean no way back of course but it would be much simpler.

That was also my first Idea, but consider: 
A Podmin might be retain to require the data for legal reasons. (might be an extra function)
A Podmin might only temporary block - and unblock a pod after more investigation.
A Podmins can track what is currently blocked. 

My idea was not to delete any data.

SuperTux88: If a pod gets blocked the messages should not only be dropped on federation level, it should also be rejected/retracted for relayables like when the user is blocked. Otherwise users would federate content on their posts which they might want to delete.

Also as @jhass said, the stream query is already complex enough and was a reason for slow streams multiple times, so lets not make that more complex than it needs to be. And filtering/rejecting at federation level would remove the need of adding more stuff to the stream query.

config/locales/javascript/javascript.de.yml
```diff
@@ -23,6 +23,7 @@ de:
         no_info: "Zurzeit keine weiteren Informationen verfügbar"
         not_available: "nicht verfügbar"
         offline_since: "offline seit:"
+        blocked: "Blockiert"
```
SuperTux88, line 26: Please don't edit locale files other than for English, all other files are managed by webtranslateit and will be overwritten.

(also, sidenote, it's probably better if you post screenshots here with English, because not everybody understand German, so you don't need to change anything but English anyway.)

lib/evil_query.rb
tclaus: I got stuck stopping relay messages complete, here my knowledge about diaspora internal comes to an end. 

@SuperTux88 : Do you mean, to delete all posts( and user and comments)  from foreign Pods if this gets blocked? Thats quite a hard step. If I get you right, should we consider to 'archive' these deleted posts somewhere? No to recover this, but to have a copy for later. - maybe for legal reasons? 

What do you think about the Query explanation?

SuperTux88: Comments/likes/poll are relayed through the authors pod, if you hide them from the author, the blocked pod could still spam on their posts, but the author can do nothing against it which would be really bad. So if a pod is blocked, these messages need to be properly rejected (and not just dropped) and especially not just silently being forwarded to other pods and not telling the post author about it.

And you don't need to delete existing entities when blocking a pod (can be an optional step, if a podmin only wants to block future messages or also delete past stuff), but if it's in the database, it should also be shown to the users to give them the options to moderate their posts. So only filtering on incoming message, **no filtering at displaying** (and therefore also no need to change that query).

If you want to still store stuff for legal reasons take a backup or a screenshot or whatever you need.

tclaus: @SuperTux88 
In the meantime I have blown up my developer Database to 4Million posts and I see the problem.  I agree to reject any (future) requests to this pods and not changing the posts query. (So old entries are still visible).
Has this to be done in the federation-code or still on the diaspora site? Starting points for me? 

Another function (or job) may be manual triggered or automatic scheduled to clean up posts, comments etc from such blocked pods.

app/models/person.rb
```diff
@@ -332,6 +335,9 @@ def exported_key
 
   # discovery (webfinger)
   def self.find_or_fetch_by_identifier(diaspora_id)
+
```
diaspora-code-review, line 338: Layout/EmptyLinesAroundMethodBody: Extra empty line detected at method body beginning.

app/models/person.rb
```diff
@@ -341,6 +347,10 @@ def self.find_or_fetch_by_identifier(diaspora_id)
     DiasporaFederation::Discovery::Discovery.new(diaspora_id).fetch_and_save
 
     by_account_identifier(diaspora_id)
+
```
diaspora-code-review, line 350: Layout/EmptyLinesAroundExceptionHandlingKeywords: Extra empty line detected before the `rescue`.

app/models/person.rb
```diff
@@ -398,4 +408,11 @@ def other_person_with_same_guid
     diaspora_id = Person.where(guid: guid).where.not(diaspora_handle: diaspora_handle).pluck(:diaspora_handle).first
     errors.add(:base, "Person with same GUID already exists: #{diaspora_id}") if diaspora_id
   end
+
+  def self.diaspora_handle_from_blocked_pod?(diaspora_handle)
```
diaspora-code-review, line 412: Lint/IneffectiveAccessModifier: `private` (on line 395) does not make singleton methods private. Use `private_class_method` or `private` inside a `class << self` block instead.

app/models/person.rb
```diff
@@ -398,4 +408,11 @@ def other_person_with_same_guid
...
+    host = diaspora_handle.split('@').last
```
diaspora-code-review, line 413: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/person.rb
```diff
@@ -398,4 +408,11 @@ def other_person_with_same_guid
...
+    pod = Pod.find_by_host(host)
```
diaspora-code-review, line 414: Rails/DynamicFindBy: Use `find_by` instead of dynamic `find_by_host`.

app/models/person.rb
```diff
@@ -398,4 +408,11 @@ def other_person_with_same_guid
...
+    return !pod.nil? && pod.blocked
```
diaspora-code-review, line 415: Style/RedundantReturn: Redundant `return` detected.

app/models/person.rb
```diff
@@ -398,4 +408,11 @@ def other_person_with_same_guid
...
+  end
+
```
diaspora-code-review, line 417: Layout/EmptyLinesAroundClassBody: Extra empty line detected at class body end.

app/workers/fetch_webfinger.rb
```diff
@@ -10,9 +10,14 @@ class FetchWebfinger < Base
 
     def perform(account)
       person = Person.find_or_fetch_by_identifier(account)
+      return if person.nil?
 
-      # also, schedule to fetch a few public posts from that person
-      Diaspora::Fetcher::Public.queue_for(person) unless person.nil?
+      unless person.pod.nil?
+        return if person.pod.blocked
```
diaspora-code-review, line 16: Style/SoleNestedConditional: Consider merging nested conditions into outer `unless` conditions.

config/initializers/diaspora_federation.rb
```diff
@@ -16,43 +16,47 @@
...
-            hcard_url:     AppConfig.url_to(DiasporaFederation::Engine.routes.url_helpers.hcard_path(person.guid)),
-            seed_url:      AppConfig.pod_uri,
-            profile_url:   person.profile_url,
-            atom_url:      person.atom_url,
-            salmon_url:    person.receive_url,
-            subscribe_url: AppConfig.url_to("/people?q={uri}")
-          },
-          aliases: [AppConfig.url_to("/people/#{person.guid}")],
-          links:   [
+        unless person.pod&.blocked
```
diaspora-code-review, line 19: Style/SoleNestedConditional: Consider merging nested conditions into outer `if` conditions.

config/initializers/diaspora_federation.rb
```diff
@@ -16,43 +16,47 @@
...
-          url:              AppConfig.pod_uri,
-          photo_large_url:  person.image_url,
-          photo_medium_url: person.image_url(size: :thumb_medium),
-          photo_small_url:  person.image_url(size: :thumb_small),
-          public_key:       person.serialized_public_key,
-          searchable:       person.searchable,
-          first_name:       person.profile.first_name,
-          last_name:        person.profile.last_name
-        )
+        unless person.pod&.blocked
```
diaspora-code-review, line 45: Style/SoleNestedConditional: Consider merging nested conditions into outer `if` conditions.

config/initializers/diaspora_federation.rb
```diff
@@ -82,7 +86,10 @@
     end
 
     on :fetch_public_key do |diaspora_id|
-      Person.find_or_fetch_by_identifier(diaspora_id).public_key
+      person = Person.find_or_fetch_by_identifier(diaspora_id)
+      if person
+        person.public_key unless person.pod&.blocked
```
diaspora-code-review, line 91: Style/SoleNestedConditional: Consider merging nested conditions into outer `if` conditions.

lib/diaspora/fetcher/public.rb
```diff
@@ -17,6 +17,10 @@ module Diaspora; module Fetcher; class Public
   Status_Unfetchable = 6
 
   def self.queue_for(person)
+    unless person.pod.nil?
+      return if person.pod.blocked
```
diaspora-code-review, line 21: Style/SoleNestedConditional: Consider merging nested conditions into outer `unless` conditions.

lib/diaspora/fetcher/public.rb
```diff
@@ -49,6 +53,10 @@ def qualifies_for_fetching?
         return false
       end
 
+      unless @person.pod.nil?
+        return false if @person.pod.blocked
```
diaspora-code-review, line 57: Style/SoleNestedConditional: Consider merging nested conditions into outer `unless` conditions.

tclaus:
@SuperTux88 with ef7d20e is this what you mean "block on federation level"? 
Is there another source-line to block "relayables"? 

I have rolled out this code since first commit and it works fast (in stream) and reliable. There are no (old and new) posts from blocked pods. 

> So if a pod is blocked, these messages need to be properly rejected (and not just dropped) and especially not just silently being forwarded to other pods and not telling the post author about it

Do you have a source line that supports me? 

Maybe we can think this PR as a tandem to #8249 (Block/Close user) ?

SuperTux88:
> @SuperTux88 with ef7d20e is this what you mean "block on federation level"?

Well, yes and no ... it kinda works ... but as far as I know returning `nil` then fails in the federation-code, which then means it retries 10 times to receive the message. Also, this blocks all messages from a user/pod if you just block there.

A better place for blocking is probably here:

lib/diaspora/federation/entities.rb
```diff
@@ -223,6 +223,28 @@ def self.related_entity(entity)
           parent: entity.respond_to?(:parent) ? related_entity(entity.parent) : nil
         )
       end
+
+      def self.should_perform(entity)
+        author = entity.try(:diaspora_handle)
+        if author.present?
+          return false if Person.diaspora_handle_from_blocked_pod(author.diaspora_id)
```
diaspora-code-review, line 230: Style/SoleNestedConditional: Consider merging nested conditions into outer `if` conditions.

lib/diaspora/federation/entities.rb
```diff
@@ -223,6 +223,28 @@ def self.related_entity(entity)
...
+        end
+
+        root_author = entity.try(:root_diaspora_id)
+        if root_author.present?
+          return false if Person.diaspora_handle_from_blocked_pod(root_author)
```
diaspora-code-review, line 235: Style/SoleNestedConditional: Consider merging nested conditions into outer `if` conditions.

lib/diaspora/federation/entities.rb
```diff
@@ -223,6 +223,28 @@ def self.related_entity(entity)
...
+        end
+
+        return false if closed_account(author)
+        return false if closed_account(root_author)
```
diaspora-code-review, line 239: Layout/EmptyLineAfterGuardClause: Add empty line after guard clause.

lib/diaspora/federation/entities.rb
```diff
@@ -223,6 +223,28 @@ def self.related_entity(entity)
...
+        true
+      end
+
+      def self.closed_account(diaspora_handle)
+        return false if diaspora_handle.nil?
```
diaspora-code-review, line 244: Layout/EmptyLineAfterGuardClause: Add empty line after guard clause.

app/models/person.rb
```diff
@@ -332,6 +335,13 @@ def exported_key
 
   # discovery (webfinger)
   def self.find_or_fetch_by_identifier(diaspora_id)
+     # pod blocked?
+     if diaspora_handle_from_blocked_pod?(diaspora_id)
```
diaspora-code-review, line 339: Layout/IndentationWidth: Use 2 (not 3) spaces for indentation.

app/models/person.rb
```diff
@@ -332,6 +335,13 @@ def exported_key
...
+      logger.info "Rejecting #{diaspora_id}, from blocked pod"
```
diaspora-code-review, line 340: Layout/IndentationWidth: Use 2 (not 1) spaces for indentation.

app/models/person.rb
```diff
@@ -332,6 +335,13 @@ def exported_key
...
+      raise DiasporaFederation::Discovery::DiscoveryError,
+            "Failed discovery for #{diaspora_id}: blocked pod"
+    end
```
diaspora-code-review, line 343: Layout/EndAlignment: `end` at 343, 4 is not aligned with `if` at 339, 5.

app/workers/receive_base.rb
```diff
@@ -26,6 +26,7 @@ def filter_errors_for_retry
            DiasporaFederation::Salmon::InvalidEncoding,
            Diaspora::Federation::AuthorIgnored,
            Diaspora::Federation::InvalidAuthor,
+            Diaspora::Federation::PodBlocked,
```
diaspora-code-review, line 29: Layout/ArrayAlignment: Align the elements of an array literal if they span more than one line.

SuperTux88 (changes requested):
I have to give some feedback here, as I don't think this is going in the right direction, as it looks like you are just adding blocks everywhere you find something could be blocked, but that's not how this is working. Some places don't make sense, as they never can happen (as for example you can't block your own pod), and some places just shouldn't be blocked, as they would create a mess in the federation and inconsistent states and take away control from the users.

Blocking pods might sound easy at the beginning, but it really isn't, there is a lot going into it. 

There are workarounds out there to break federation with some users from other pods ("block" them), but these are workarounds and they do something similar as you did so far, just hard break federation. But that has side-effects, isn't transparent to users and lead to further problems, so they should only be used at own risk if you understand what's happening. But as this PR is something that should be merged in the main code and be used by all podmins, it needs to work robust and also handle all the cases properly, so it can't be done with just "break federation" and ignore all the problems that generates.

So blocking pods needs to work in a as clean as possible way, so it doesn't just create a huge mess. And as you can unblock a pod again, it also needs to continue working when a pod gets unblocked again, so some stuff still needs to continue working, even if it's only in the background.

I think you can probably undo most of the changes (at least those affecting streams and federation), and then add again:

* If it's a normal entity (like a status message, or a reshare), ignore it in [receive.rb](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/federation/receive.rb) (not processing them is fine)
* If it's a relayable (comment, like and poll participation), and the pod is blocked and your pod is the owner of the root entity, then reject it sending a retraction back (like when the author [if a comment is ignored](https://github.com/diaspora/diaspora/blob/develop/lib/diaspora/federation/receive.rb#L286))
* There are still a lot of messages you don't want to block, as they aren't normal content of a user, for example retractions, and account deletions, so only block it, if it's actual content. For example contact changes (adding/removing to aspects) also still needs to work to not create inconsistent state if the pod gets unblocked again, but you wouldn't want to user to receive notifications about it, as they can't interact with the user when the pod is blocked, so process the entity as normal, but then return `nil` [here](https://github.com/diaspora/diaspora/blob/d4079070edc8b750257fffe5da9b0494504d5576/lib/diaspora/federation/receive.rb#L61-L69) to not trigger notifications.
* You probably also don't want to send your content to those pods anymore (as people from those pods couldn't reply anyway), this can be done by filtering the people [here](https://github.com/diaspora/diaspora/blob/739fc780c3b4cff075311e200b3639ad586ddf30/lib/diaspora/federation/dispatcher.rb#L54), but again, you can only do that for actual content (posts, comments, ...), other messages still need to be sent to those pods (retractions, account deletions, unshare, ...), otherwise you people from your pod who sent posts/comments to the now blocked pods can't delete them anymore from there, which shouldn't happen.
* For the frontend you maybe want to block showing existing content, so for example block the profile of a person from a blocked pod (or probably show a message for users from your pod, that that pod is blocked), and block the single post view of a post from person from a blocked pod (show 404), but those are just simple checks when viewing these pages, no filtering of complicated stream queries required.
* If you really want old posts not to be included in streams anymore, build a way how a podmin can optionally delete all posts from a pod (but that can't be undone when you want to unblock a pod again, so it should be optional if a podmin wants to do that)
* For better UX, it should be shown when somebody search for a diaspora ID from a blocked pod, that this pod was blocked, instead of just not working, otherwise users are confused and then create bugreports or support requests.
* Also for better UX, if users already have contacts with people from blocked pods, it should mark those contacts as blocked in the UI and also not let them select those contacts anymore when write private messages. So your users understand what's happening instead of things they expect to work just don't work.

That's basically all that's needed, and all that should be done and not more. There are things that need to work, otherwise you create a huge mess in a federated system like diaspora.

If you have any more questions (or if you think I missed something in the list above, which is totally possible) please discuss them first, as again, this is a complicated topic and not as easy as it might initially look.

Issue #8220, opened 2021-03-11
[Feature request] See local posts only


stemy2:
Hello,

It seems not possible to see only the post locally published on my pod (i mean, by users with "@mypod.tld" in their usernames), it make the moderation really difficult. Is it possible to implement a such functionality ?

tclaus:
As a Podmins or moderator, auch a function would be nice and useful.
Could be realized by an Admin-function or by a search term.

tclaus:
After thinking about it - 
Only local (public) posts on a pod would mean on some pods: Just my own posts, if the pod has closed registrations and the podmin (and invites) are the only guests. 

The other idea is, that local posts mean to view a more streamlined list of posts, more like the 'people in my favorite pup', then the 'public posts list'. The later means ( in my view) a more or less random list of posts linked in any case with the pod. 

So should this be better an option a podmin might switch on and off? Should booth options be visible or just one of them? 
Any ideas about this?

jhass:
I'm pretty sure this idea came up previously but I'm unable to find any prior discussions currently. If anyone stumbles upon them, please cross-link.

tclaus:
Eric Wright requests here: https://discourse.diasporafoundation.org/t/newhere-posts-or-local-posts-highlighted/1979 to 'see' local #newhere - posts to send a message . (2018)

PR #8232, opened 2021-04-10
8820 Show Local public posts

tclaus:
This addresses #8220 

* It adds a link to the sidebar above 'Public activity' that filtered only local public posts in the stream 
* It is configurable in diaspora.toml file, defaults to : not visible
* It can be enabled for admins, moderators or everybody

### DIFF
--- a/app/controllers/streams_controller.rb
+++ b/app/controllers/streams_controller.rb
@@ -25,6 +25,10 @@ class StreamsController < ApplicationController
     stream_responder(Stream::Public)
   end
 
+  def local_public
+    stream_responder(Stream::LocalPublic)
+  end
+
   def activity
     stream_responder(Stream::Activity)
   end
--- a/app/controllers/streams_controller.rb
+++ b/app/controllers/streams_controller.rb
@@ -26,7 +26,18 @@ class StreamsController < ApplicationController
   end
 
   def local_public
-    stream_responder(Stream::LocalPublic)
+    return stream_responder(Stream::LocalPublic) if AppConfig.settings.enable_show_local_post_link == "always"
+
+    if AppConfig.settings.enable_show_local_post_link == "admin" && Role.is_admin?(current_user)
+      return stream_responder(Stream::LocalPublic)
+    end
+
+    if AppConfig.settings.enable_show_local_post_link == "moderator" &&
+       (Role.moderator?(current_user) || Role.is_admin?(current_user))
+      return stream_responder(Stream::LocalPublic)
+    end
+
+    stream_responder(Stream::Public)
   end
 
   def activity
--- a/app/controllers/streams_controller.rb
+++ b/app/controllers/streams_controller.rb
@@ -26,18 +26,11 @@ class StreamsController < ApplicationController
   end
 
   def local_public
-    return stream_responder(Stream::LocalPublic) if AppConfig.settings.enable_show_local_post_link == "always"
-
-    if AppConfig.settings.enable_show_local_post_link == "admin" && Role.is_admin?(current_user)
-      return stream_responder(Stream::LocalPublic)
+    if AppConfig.local_posts_stream?(current_user)
+      stream_responder(Stream::LocalPublic)
+    else
+      head :not_found
     end
-
-    if AppConfig.settings.enable_show_local_post_link == "moderator" &&
-       (Role.moderator?(current_user) || Role.is_admin?(current_user))
-      return stream_responder(Stream::LocalPublic)
-    end
-
-    stream_responder(Stream::Public)
   end
 
   def activity
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -42,6 +42,17 @@ module ApplicationHelper
       "bookmarklet('#{bookmarklet_url}', #{width}, #{height});"
   end
 
+  def show_local_posts_link?(current_user)
+    return true if AppConfig.settings.enable_show_local_post_link == "always"
+    return true if AppConfig.settings.enable_show_local_post_link == "admin" &&
+                   Role.is_admin?(current_user)
+
+    return true if AppConfig.settings.enable_show_local_post_link == "moderator" &&
+                   (Role.moderator?(current_user) || Role.is_admin?(current_user))
+
+    false
+  end
+
   def all_services_connected?
     current_user.services.size == AppConfig.configured_services.size
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -42,17 +42,6 @@ module ApplicationHelper
       "bookmarklet('#{bookmarklet_url}', #{width}, #{height});"
   end
 
-  def show_local_posts_link?(current_user)
-    return true if AppConfig.settings.enable_show_local_post_link == "always"
-    return true if AppConfig.settings.enable_show_local_post_link == "admin" &&
-                   Role.is_admin?(current_user)
-
-    return true if AppConfig.settings.enable_show_local_post_link == "moderator" &&
-                   (Role.moderator?(current_user) || Role.is_admin?(current_user))
-
-    false
-  end
-
   def all_services_connected?
     current_user.services.size == AppConfig.configured_services.size
   end
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -51,6 +51,12 @@ class Post < ApplicationRecord
 
   scope :all_public, -> { where(public: true) }
 
+  scope :all_local_public, -> {
+    left_outer_joins(author: [:pod])
+      .where("pods.host is null") # local posts have no host in pods
+      .where(public: true)
+  }
+
   scope :commented_by, ->(person)  {
     select('DISTINCT posts.*')
       .joins(:comments)
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -52,9 +52,10 @@ class Post < ApplicationRecord
   scope :all_public, -> { where(public: true) }
 
   scope :all_local_public, -> {
-    left_outer_joins(author: [:pod])
-      .where("pods.host is null") # local posts have no host in pods
-      .where(public: true)
+    where(" exists ( 
+      select 1 from people where posts.author_id = people.id
+      and people.pod_id is null)
+      and posts.public = true")
   }
 
   scope :commented_by, ->(person)  {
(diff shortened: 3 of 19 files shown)


lib/stream/local_public.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+#   Copyright (c) 2010-2011, Diaspora Inc.  This file is
+#   licensed under the Affero General Public License version 3 or later.  See
+#   the COPYRIGHT file.
```
jhass, line 5: We skip including this for new files >.>

config/diaspora.toml.example
```diff
@@ -288,6 +288,12 @@
 ## (or commented out) to enable the first registration (you).
 #enable_registrations = true
 
+## Show local posts (default="disabled")
```
jhass, line 291:
```suggestion
## Show local posts stream (default="disabled")
```

config/diaspora.toml.example
```diff
@@ -288,6 +288,12 @@
...
+## If any other setting than disabled local public posts
+## created on this pod can be showed.
```
jhass, line 293:
```suggestion
## created on this pod can be shown.
```

config/diaspora.toml.example
```diff
@@ -288,6 +288,12 @@
...
+## Choose one setting
+# enable_show_local_post_link= "disabled"|"admin"|"moderator"|"always"
```
jhass, line 295:
```suggestion
# enable_local_posts_stream= "disabled"|"admins"|"moderators"|"everyone"
```

config/diaspora.toml.example
```diff
@@ -288,6 +288,12 @@
 ## (or commented out) to enable the first registration (you).
 #enable_registrations = true
 
+## Show local posts (default="disabled")
+## If any other setting than disabled local public posts
+## created on this pod can be showed.
+## Choose one setting
```
jhass, line 294:
```suggestion
## Setting this admins shows the local posts stream only to users with the admin role.
## Setting this to moderators shows the local posts stream only to users with the moderator or admin role.
## Setting this to everyone shows the local posts stream to all users.
```

config/locales/diaspora/de.yml
```diff
@@ -1245,6 +1245,8 @@ de:
       title: "Stream"
     public:
       title: "Öffentliche Aktivität"
+    local_public:
+      title: "Lokale Posts"
```
jhass, line 1249:
```suggestion
```

Please don't provide anything but the English translation. Everything else is handled through WebTranslateIt.

spec/models/post_spec.rb
```diff
@@ -58,6 +58,23 @@
...
+        post1 = FactoryGirl.create(:status_message, author: alice.person, public: true)
+        post2 = FactoryGirl.create(:status_message, author: bob.person, public: true)
+        expect(Post.all_local_public.ids).to match_array([post1.id, post2.id])
+      end
+
+      it "doesn't include any posts from other pods" do
+        pod = FactoryGirl.create(:pod)
+        external_person = FactoryGirl.create(:person, pod: pod)
+        FactoryGirl.create(:status_message, author: alice.person, public: true)
+        FactoryGirl.create(:status_message, author: bob.person, public: true)
```
jhass, line 72: Please rebase against the factory_bot update :)

app/helpers/application_helper.rb
```diff
@@ -42,6 +42,17 @@ def bookmarklet_code(height=400, width=620)
...
+  def show_local_posts_link?(current_user)
+    return true if AppConfig.settings.enable_show_local_post_link == "always"
+    return true if AppConfig.settings.enable_show_local_post_link == "admin" &&
+                   Role.is_admin?(current_user)
+
+    return true if AppConfig.settings.enable_show_local_post_link == "moderator" &&
+                   (Role.moderator?(current_user) || Role.is_admin?(current_user))
+
+    false
+  end
```
jhass, line 54: How about moving this to `lib/configuration_methods.rb` like so

```rb
def has_local_posts_stream?(user)
    return true if settings.enable_local_posts_stream == "admins" &&
                   Role.is_admin?(user)
    return true if settings.enable_local_posts_stream == "moderators" &&
                   (Role.moderator?(user) || Role.is_admin?(user))

    settings.enable_local_posts_stream == "everyone"
end
```

Then we can reuse the logic in the controller.

app/controllers/streams_controller.rb
```diff
@@ -25,6 +25,21 @@ def public
...
+      return stream_responder(Stream::LocalPublic)
+    end
+
+    if AppConfig.settings.enable_show_local_post_link == "moderator" &&
+       (Role.moderator?(current_user) || Role.is_admin?(current_user))
+      return stream_responder(Stream::LocalPublic)
+    end
+
+    stream_responder(Stream::Public)
+  end
```
jhass, line 41: I think this should 404 instead of returning other content than promised.

With above change we could then write this as (untested):

```rb
if AppConfig.has_local_posts_stream?(current_user)
  stream_responder(Stream::LocalPublic)
else 
  render status: :404
end
```

jhass (changes requested):

jhass:
Should this be added as an API route too? If yes, how does a client discover whether it's available or not?

tclaus:
> Should this be added as an API route too? If yes, how does a client discover whether it's available or not?

Hm - it was not directly intended to be available as an API function. I let the core Team decide if its a need.  - I think this feature has its benefits on a desktop system for filtering a smaller part of the public post. 

If this should be an API function, a system to get enabled features should be established.

jhass:
If it's intended for user consumption I feel like it should be in the API, though admittedly that could be a second step.

If it's not intended for user consumption then enabling it for everyone shouldn't be an option maybe.

app/helpers/application_helper.rb
tclaus: Great Idea - I did not know this class until now.

tclaus:
> If it's intended for user consumption I feel like it should be in the API, though admittedly that could be a second step.
> 
> If it's not intended for user consumption then enabling it for everyone shouldn't be an option maybe.

I'll take answer a) For User AND Admins (by configuration)

tclaus:
@jhass @SuperTux88 
Same problem here (one join less in query) then in #8228  - to get just local Posts a join must be made over Posts -> People to see if the "People .pod_id" is null. 

This takes on huge databases a bit of time. 
Two Ideas for this: 
a) Get all public posts as usual ( in 15 pieces batches), test every single post if its local or not and fill up the batch if some posts are filtered out to keep the size. '(query, filter, loop if posts are filtered out until batch is full)

b) Introduce a 'isLocal' attribute to the posts table

What do you think?

tclaus:
In the meantime.. 
With this scope in post.rb
```
  scope :all_local_public, -> {
    where(" exists ( 
      select 1 from people where posts.author_id = people.id
      and people.pod_id is null)
      and posts.public = true")
  }
```

and this index: 
```
CREATE INDEX local_people
    ON public.people USING btree
    (id ASC NULLS LAST)
    TABLESPACE pg_default
    WHERE pod_id IS NULL;

COMMENT ON INDEX public.local_people
    IS 'Index for people without a host';
```

a Query over 4Million Posts takes: 56ms
(without this index ~200ms)

Explantation: First (naive) idea was a left join, but this means, all Rows of Posts has to be joined with all rows of people, then the query checks which one is null.
This solution (exists)  loops (inside the database) as long enough rows exists which has a pod_id = null aka "local posts". 

With Index the Query plan look like this: 

Bildschirmfoto 2021-04-14 um 18 44 11

This is the execution Plan without the new index (Fast as lightning, but seq scans are usually not the best idea) 
Bildschirmfoto 2021-04-14 um 18 49 32


Is it allowed to migrate a new index into the database?

jhass:
Generally adding new indexes is ok, yes. However we should look at the trade-offs, how much faster does it make some vs. how much disk space does it consume? Then we have the soft rule of avoiding long running migrations in minor-releases unless they solve a serious bug. So it's also interesting how long generating the index might take.

Keep in mind all of this has to work on MariaDB/MySQL too :)

tclaus:
On my Test database (>100K people) the index creation took only some 100ms.
In MySQL the Index should be a regular index (MySQL doesn't know partial indexes (Index for only a part of the data)). 

The query outcome is 200ms / 50ms - at least the Peoples Table must be scanned sequentially (Full Table Scan in MySQL Terms) - So using an index is the better Idea than not using it. 

Is it a need to test index on a MySQL environment? ( I don't have such at the moment)

By The way: It seems easy to start without a general extra index, but monitor it in real live or recommend the Podmins to use such?

jhass:
Of course, adding an index later is easy. Maybe 200ms is not too bad, I doubt this functionality will actually be used a lot. By the way we also need to be careful how we communicate about this feature. Diaspora is a bad tool for creating local communities and we should take care to not take the impression that this would change this.

I would strongly vote against recommending podmins to manually add indexes or mess with the database in any other way.

tclaus:
> I would strongly vote against recommending podmins to manually add indexes or mess with the database in any other way.
Strongly agree with this (and rest of text) 

Lets communicate this feature as a feature for Podmins with an option to open this for everyone.

tclaus:
@jhass Do you have a clue what's going on with the CI? On local machine there are no issues, I've rebased and pushed the rebased version. (I think so..)

jhass:
Maybe your local `diaspora.toml` still defines `enable_local_posts_stream`

app/models/post.rb
```diff
@@ -51,6 +51,13 @@ class Post < ApplicationRecord
 
   scope :all_public, -> { where(public: true) }
 
+  scope :all_local_public, -> {
+    where(" exists (
```
diaspora-code-review, line 55: Layout/TrailingWhitespace: Trailing whitespace detected.

lib/stream/local_public.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# rubocop:disable Style/ClassAndModuleChildren
+class Stream::LocalPublic < Stream::Base
+  def link(opts={})
+    Rails.application.routes.url_helpers.local_public_stream_path(opts)
+  end
+
+  def title
+    I18n.translate("streams.local_public.title")
```
diaspora-code-review, line 10: Rails/ShortI18n: Use `t` instead of `translate`.

tclaus:
The defaults.yml has had a false key. On tests the local configuration is read, so it didn't get noticed.

jhass (approved):

SuperTux88:
Merged to `develop` (I tried `next-minor`, but had too many conflicts)

Thanks @tclaus for the PR :cookie: and @jhass for the review :cookie:

Issue #7878, opened 2018-10-07
Directly pasting image into a post when composing


HankG:
A feature I use on a lot of places which I can't do in Diaspora right now is copy/pasting an image from the clipboard into the composition area and have it become part of the post automatically.  In the current Diaspora a user needs to save it to the disk then add upload it to attach it to the post...

HankG:
It turns out that the FineUploader we are already using supports this option.  You just have to set up the paste options and it works.  The big question is if we want the paste-focus to be on the text area or the entire page.  I coded it up against the text area first.  Thoughts?

Flaburgan:
It has to be the textarea as you can have the publisher open several times in the page to comment.

Flaburgan:
We also have the drag and drop feature to upload an image. It currently only works on the camera icon, maybe you can extend it to works on the whole textarea.

HankG:
I'll look into that.  Right now drag and drop into the text area drops the file path to text box.

HankG:
@Flaburgan yes I got your suggestion working but had to bring in a bit more of the FineUploader library (we only brought in the core before). It's only a net couple hundred K if I'm reading it correctly.  I've tested this manually and am looking at if I need to make an update to the Jasmine tests.

HankG:
I'm seeing an inconsistency between the mobile and desktop calls into the photos service.  This goes back to the earliest versions of both.  The mobile one adds the aspects fields to the POST arguments while the desktop one does not when calling the photos upload service.  I'm going to do some experimenting but does this discrepancy make any sense to anybody?

HankG:
I have a PR that captures the original request, plus the enhanced drag and drop requested by @Flaburgan , and with the change to more consistently call into the file uploaded between the two. https://github.com/diaspora/diaspora/pull/7883

Siedlerchr:
Drag and Drop  Support as an intermediate step would be also nice.

HankG:
@Siedlerchr the PR for this does drag and drop as well. Drag and drop already worked if you put it on top of the camera icon.  In the current version it works if you drop it into the text field or the thumbnail bar too.

PR #8237, opened 2021-04-15
7878 direct image pasting

tclaus:
Picking up  Issue #7878 
and adding this to PR  https://github.com/diaspora/diaspora/pull/7883
 this auto-expands the publisher when dropping a file to it. 

I also checked the extra log entry "Grabbed 1 dropped files". This line is inside FineUploader: 
`options.callbacks.dropLog("Grabbed " + files.length + " dropped files.");`
The second "level" parameter is missing, so this can not be filtered out.

### DIFF
--- a/app/assets/javascripts/app/views/publisher/uploader_view.js
+++ b/app/assets/javascripts/app/views/publisher/uploader_view.js
@@ -10,7 +10,7 @@ app.views.PublisherUploader = Backbone.View.extend({
     this.publisher.photozoneEl.on("click", ".x", _.bind(this._removePhoto, this));
 
     // Initialize the PostPhotoUploader and subscribe its events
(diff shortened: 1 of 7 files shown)


jhass (approved):
Did you have any chance to look into the issue of the filename being pasted when you paste an image? I'm still seeing that behavior.

Otherwise this looks good enough to me to give it some time in the wild. Nothing needs to land 100% perfect, we can always improve later :)

tclaus:
As usual: already deployed on societas.online. 

I'll check.

tclaus:
> As usual: already deployed on societas.online.
> 
> I'll check.

I am sorry - did not found the source of this issue (Filename pasted on pasting image) in a reasonable amount of time.

Issue #8139, opened 2020-08-25
Add a "contact the podmin" method


denschub:
Currently, there is no way for non-logged in people to contact a podmin on a unified channel. Some pods expose an email address in their Terms of Service, but not all do. NodeInfo exposes the username of the admin account, but that's not great, as you need to have a diaspora\* account to send messages to that.

We could build a simple "contact form" on a well-known URI that allows people to send an email to `podmin_email` if it's set. This would resolve some issues we have when people try to get in touch with their podmins for account-related questions, and it would also make reporting contents easier between podmins.

Spam protection might be a hard issue, though, as we can't rely on "more robust" solutions like ReCaptcha, so this will likely need some careful consideration.

SuperTux88:
> Spam protection might be a hard issue, though, as we can't rely on "more robust" solutions like ReCaptcha, so this will likely need some careful consideration.

Personally I'd rather have the mail-address listed public somewhere instead of a contact form, because I think it's easier to spam-filter plain mails instead of contact-form-mails sent by diaspora containing spam ... and I think we can all agree that a captcha wouldn't help much for spam. I have also all well-known common addresses (abuse@, webmaster@, postmaster@ and @hostmaster@) configured on my domains, and I don't receive much spam on them, but if we want to we could also hide the mail-address behind a captcha?

denschub:
Just listing the email address would also be significantly easier. Probably works as well, as long as it is placed on a known URL, maybe linked in the footer, and can't be accidentally removed. :)

PR #8242, opened 2021-05-09
8139 add a podmin mail to footer

tclaus:
As requested in #8139 it adds a podmin mail address to footer if set in configuration. 
Footer screenshot

### DIFF
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -32,6 +32,7 @@ en:
   no_results: "No results found"
   _contacts: "Contacts"
   _terms: "Terms"
+  _podmin_mail: "Podmin Mail"
   _statistics: "Statistics"
 
   # For reference translation; the real ActiveRecord English transations are actually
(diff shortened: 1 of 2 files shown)


tclaus:
A quick-win?

config/locales/diaspora/en.yml
```diff
@@ -32,6 +32,7 @@ en:
   no_results: "No results found"
   _contacts: "Contacts"
   _terms: "Terms"
+  _podmin_mail: "Podmin Mail"
```
SuperTux88, line 35: I think "Contact Podmin" would be a better text for most users, because "Contact" is what most people understand (also by people not very familiar with the project) and that's also what I'm looking for when I search how to contact somebody.

SuperTux88 (changes requested):
This will lead to more spam (everything that is shown in the internet that looks like a mail-address will be used to send spam eventually), but I also hope people have a spamfilter on their mails, so it shouldn't be a big problem. And it's just the easiest and best solution for this problem, since more and more people contact the project-mail-address because they have a problem with a specific pod.

Other than a little nitpick with the translation I think this is fine.

SuperTux88 (approved):

SuperTux88:
Thanks :cookie: Merged as 4c337c6952bcd6421f2e75bfabe4adb7be2c16d0

PR #8239, opened 2021-04-28
Enhance report overview

tclaus:
This PR enhances the admin/moderators report overview by 
 - adding a view that shows recent reports and the action done (Item Deleted or not) 
 - adding a statistics view (Who reports and who is the originator of a reported item) 
 - a once 'reviewed report can be deleted retrospectively
 - let podsmins show the originator even /after/ a reported item was deleted
 - links to the originator profile directly (and not to the Search view) 
 
**Heads up!**
It needs a migration! 
Two fields are added: a 'action' textfield that is used to describe what happened to the report. Its free text, but filled based on the action. 
The second fields holds the diaspora_handel for the originator. This is needed because after deleting a report this information is gone. 

**What is the outcome?**
A Podmin now has more insights in old reports and in most reporter and most originators. A valuable base for more decisions what to do. 

**Future**
List of reports grows without limits - a paginated view would be nice, but that was currently above my skill level.

Screenshots

 (Added date time)
Bildschirmfoto 2021-04-28 um 08 11 07

( deleted or reviewed reports) 
Bildschirmfoto 2021-04-28 um 08 10 56

(statistics with links to the authors / originators) 
Bildschirmfoto 2021-04-28 um 08 11 17

Comments, Ideas, critics... welcome!

PR #8246, opened 2021-05-22
Let podmin send private messages to local users

tclaus:
Hidden under this headline https://github.com/diaspora/diaspora/issues/7297 a podmin should be allowed to send a message to it's local users even when they don't share each others. 

Together with improved report system (see PR #8239 ), there are some cases when a podmin needs to get in contact for a specific topic to its users. 
There are also many other situations where a private message is very helpful between podmin and users.

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -20,7 +20,11 @@ class ContactsController < ApplicationController
       format.json {
         @people = if params[:q].present?
                     mutual = params[:mutual].present? && params[:mutual]
-                    Person.search(params[:q], current_user, only_contacts: true, mutual: mutual).limit(15)
+                    if current_user.admin?
+                      Person.search_for_admin(params[:q], current_user, only_contacts: true, mutual: mutual).limit(15)
+                    else
+                      Person.search(params[:q], current_user, only_contacts: true, mutual: mutual).limit(15)
+                    end
                   else
                     set_up_contacts_json
                   end
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -21,7 +21,7 @@ class ContactsController < ApplicationController
         @people = if params[:q].present?
                     mutual = params[:mutual].present? && params[:mutual]
                     if current_user.admin?
-                      Person.search_for_admin(params[:q], current_user, only_contacts: true, mutual: mutual).limit(15)
+                      Person.search_as_admin(params[:q], current_user, only_contacts: true, mutual: mutual).limit(15)
                     else
                       Person.search(params[:q], current_user, only_contacts: true, mutual: mutual).limit(15)
                     end
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -123,7 +123,7 @@ class ConversationsController < ApplicationController
   end
 
   def no_contacts?
-    # an admin always has at least all local contcts
+    # an admin always has at least all local contacts
     return false if current_user.admin?
 
     current_user.contacts.mutual.empty?
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -24,9 +24,8 @@ class ConversationsController < ApplicationController
     end
 
     gon.contacts = contacts_data
-
     respond_with do |format|
-      format.html { render "index", locals: {no_contacts: current_user.contacts.mutual.empty?} }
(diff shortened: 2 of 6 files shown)


app/models/person.rb
```diff
@@ -243,23 +248,43 @@ def self.search_query_string(query)
 
     [where_clause, q_tokens]
   end
-
+
```
diaspora-code-review, line 251: Layout/TrailingWhitespace: Trailing whitespace detected.

Flaburgan:
This PR should be renamed to "private message" instead of "mails", it confused me :stuck_out_tongue: 
Apart from that I'm all for it!

app/controllers/contacts_controller.rb
```diff
@@ -20,7 +20,11 @@ def index
       format.json {
         @people = if params[:q].present?
                     mutual = params[:mutual].present? && params[:mutual]
-                    Person.search(params[:q], current_user, only_contacts: true, mutual: mutual).limit(15)
```
Flaburgan, line 23: I'm not sure that's the good place to put that changes. If the comment above is right, "Used for mentions" means an admin will be able to @username the user when writing a post. I am not sure that gives access to the post to the user, and it will have to be put in an aspect etc, that's not practical. I think private messages are what we want to achieve here.

app/controllers/contacts_controller.rb
```diff
@@ -20,7 +20,11 @@ def index
...
+                    if current_user.admin?
+                      Person.search_for_admin(params[:q], current_user, only_contacts: true, mutual: mutual).limit(15)
```
Flaburgan, line 24: I would have said "search_as_admin"

app/controllers/conversations_controller.rb
```diff
@@ -37,7 +36,12 @@ def create
     # This will have to be removed when mobile autocomplete is ported to Typeahead
     recipients_param, column = [%i(contact_ids id), %i(person_ids person_id)].find {|param, _| params[param].present? }
     if recipients_param
-      person_ids = current_user.contacts.mutual.where(column => params[recipients_param].split(",")).pluck(:person_id)
+      # As an admin, I want to send a message to all local user
```
Flaburgan, line 39: Okay now I get why you modified the contacts controller, it was to get the autocomplete here to work. But as an admin is able to contact anyone, I guess to add a new route which returns every local user is fine, and this shouldn't be in the contacts controller I guess. Maybe it even already exists. Can you tell us @jhass, or where to put it if it doesn't exist?

app/controllers/conversations_controller.rb
```diff
@@ -118,4 +121,11 @@ def contacts_data
         {value: contact_id, name: ERB::Util.h(Person.name_from_attrs(*name_attrs)) }
       }
   end
+
+  def no_contacts?
```
Flaburgan, line 125: Can you make that a bit more explicit? `not_admin_and_no_contacts?` for example?

app/controllers/conversations_controller.rb
```diff
@@ -118,4 +121,11 @@ def contacts_data
...
+    # an admin always has at least all local contcts
```
Flaburgan, line 126: contActs

app/models/person.rb
```diff
@@ -244,22 +249,42 @@ def self.search_query_string(query)
...
+            else
+              query.searchable(user)
+            end
+    query.where(closed_account: false)
+         .order(["profiles.last_name ASC", "profiles.first_name ASC"])
+  end
+  # rubocop:enable Rails/DynamicFindBy
+
+  def self.exists_in_contacts(user, mutual)
+    return "select 1 from contacts where contacts.person_id = people.id and contacts.user_id = #{user.id}" unless mutual
```
Flaburgan, line 282: Nitpick, but SQL is usually written in uppercase.

Flaburgan (changes requested):

app/controllers/conversations_controller.rb
tclaus: Function differs if admin parses person_ids or a non-admin user parses person_ids. Different endpoints seems for me a bit complicated. (Check in frontend which role a user has, then change endpoints for same data with different filter) What do others think?

app/controllers/conversations_controller.rb
tclaus: "no_contacts_for_user" / "no_contacts_for_users_role"? .. For admin seems to specialized fo me for a methods' name?

Issue #7464, opened 2017-05-24
Allow to close remote accounts on own pod


SuperTux88:
In addition to #7463, a podmin should be able to also close remote spam-accounts on their own pod (also including all comments of that person and also sending retractions for comments on local posts, where the pod is allowed to send retractions).

Additionally a pod should block/reject new posts/comments from a locally closed remote person. So optimally they would still be processed and send back retractions for comments on local posts (like ignore/block does), but nothing is persisted.

Tectract:
We should not "ban remote accounts", as obviously that is not possible nor desireable.

Here's my suggested solution, it consists of:

1. The ability to block users by username and IP

2. The ability to share blocklists, via a voluntary subscription system between pods, when you can add a pod's blocklist to your own, via admin tools.

3. Blocklists should include the reason for the ban, and there should be some tool for reviewing these, or possible a STANDARDIZED TOS that many pods subscribe to, that could be referred to here. Maybe there is already such a standardized / proposed TOS?

4. Users on a pod's blocklist should not be allowed to post anything to pod from which they are "blocked", and their posts also should not be allowed to be RESHARED to the pod from which they are blocked

5. Consider a consensus or reputation system among pods. Pods with many many users on the blocklists of many many pods should have a lower reputation in this system

6. Right now it seems possible to de-federate from a single pod, simply be removing the DB entries, or even through the admin tools, currently. It may be desireable to share not only blocklists of users/IP addresses, but also PODS that have super-low "reputation" score or simply by a podmin's custom list. 


I want to point out in this thread, that, FAIR MODERATION is not censorship. Every pod should have a reasonable and FAIR terms of service and good moderation is all about enforcing those TOS fairly and equally to all users. Things like "insults" should not be banned, this thread is mainly in reference to keeping things like Islamic terror and radical hategroups off of D*, according to voluntary terms of service and fair moderation rules.

Tectract:
I think for most users the ability to ignore is plenty, however ignores should include reshares from the ignored user. I think that's a separate ticket, maybe it already exists?

SuperTux88:
No, that's not the topic here. Every pod is different and independent. You won't see a central censorship system on diaspora, and that was already discussed on discourse and I won't discuss this here again (and I need to ban you here too if you try to discuss it again).

This issue is about YOU as a podmin can do on YOUR pod what YOU want. It's to make moderation features better for podmins so they can block remote users easier to keep their pods clean.

Tectract:
[...]

Issue #7463, opened 2017-05-23
A podmin should be able to close an account including all comments from the admin panel


SuperTux88:
Often spammers use comments to spam, so a podmin should be able to retract all comments of a spammer before closing the account. That could be done with a "delete comments"-checkbox when closing an account from the admin panel.

For better backward compatibility, the comments should be retracted before closing the account, so other pods actually receive a `retraction` for the comments. We can add a `includeComments` boolean flag to the `account_deletion` message, so receiving pods know they should delete all comments from that person, but that would only work with pods knowing that flag.

Flaburgan:
@SuperTux88 do you plan to work on that? Do you have a script I can use at the moment? We just add another case with framasphere.

SuperTux88:
There is still [this](https://github.com/SuperTux88/diaspora/blob/nerdpol_ch/script/nerdpol.ch/diaspora_spam.rb) script which is a modified version of a script @jhass wrote once (and I don't know the link of the original anymore).

PR #8249, opened 2021-06-07
Block/Close and wipe data for all accounts on local pod in admin UI

tclaus:
It solves  #7464 and #7463 

- It adds the SuperTux88 / JHass' script to close and wipe an account - even remote ones. 
   For this function you now can search through all persons on your pod, not only the local user.
- No external ruby code to remove spammers!
- It changes the order of the lock/close buttons in Admin view by its severity. 
- It has some code refinements (Fixed Close / Lock/Unlock on Person and User classes) 
- It prevents an admin to remove its own admin status accidently. 

Remote / Local user in users' search view
125693555-80872d90-6811-424b-9fbe-ed96cbb640d2


tclaus:
Uh.. "Green" with first commit!

Flaburgan:
Awesome work, thank you! One suggestion: use "Block" instead of "Close" when you're talking about remote account. Also, is the pub key of the user stored somewhere or definitely replaced? If the account is blocked indefinitely (can't be unblocked), I would warn about it in the confirm popup (you kept it for remote account, did you?).

tclaus:
> "Block" instead of "Close" when you're talking about remote account.

It was intended to 'block' external accounts and 'close' local ones. Do you think to change button names for local and remote accounts?

tclaus:
> Awesome work, thank you! One suggestion: use "Block" instead of "Close" when you're talking about remote account. Also, is the pub key of the user stored somewhere or definitely replaced? If the account is blocked indefinitely (can't be unblocked), I would warn about it in the confirm popup (you kept it for remote account, did you?).

Yes, I kept the confirmation.
About to change the buttons caption.

tclaus:
It is intended that the serialized_public_key is set to "BLOCKED" ( but it don't work now)

Flaburgan:
Yeah I would love confirmation on this by others @jhass @SuperTux88 what do you think, should we store the pub key somewhere in case the podmin wants to restore it, or can we go like this with irreversible block?

app/views/admins/_user_entry.haml
```diff
@@ -1,131 +1,89 @@
...
-          = user.try(:person).try(:name)
+          = person.diaspora_handle
+        = t(".remote") if person.remote?
+        = t(".local") if person.local?
+        %br
+        - if person.closed_account?
+          .label.label-danger
+            Closed
+        - else 
+          -if person.locked_access?
```
diaspora-code-review, line 17: - Style/IfInsideElse: Convert `if` nested inside `else` to `elsif`.
- The - symbol should have one space separating it from code

app/views/admins/_user_entry.haml
```diff
@@ -1,131 +1,89 @@
...
         %h4.media-heading
-          = user.try(:person).try(:name)
+          = person.diaspora_handle
+        = t(".remote") if person.remote?
+        = t(".local") if person.local?
+        %br
+        - if person.closed_account?
+          .label.label-danger
+            Closed
+        - else
```
diaspora-code-review, line 16: Line contains trailing whitespace

app/views/admins/user_entry/_close_action.haml
```diff
@@ -0,0 +1,9 @@
+- if person != @current_user.person && !person.closed_account?
+  = link_to person.owner.present? ? t("admins.user_search.close_account") : t("admins.user_search.block_account"),
+    admin_close_account_path(person.id),
+    method: :post, data: {confirm: t("admins.user_search.are_you_sure")},
+    class: "btn btn-danger btn-block"
+  = link_to person.owner.present? ? t("admins.user_search.wipe_and_close_account") : t("admins.user_search.wipe_and_block_account"),
```
diaspora-code-review, line 6: Line is too long. [132/120]

app/views/admins/user_entry/_view_profile.haml
```diff
@@ -0,0 +1,4 @@
+- if !person.closed_account?
```
diaspora-code-review, line 1: Style/NegatedIf: Favor `unless` over `if` for negative conditions.

tclaus:
Changed the buttons label to "block" on external accounts and "close" for locals as it was before. 
Moved the "Account Active" - label to the header. It now shows the states "Active", "Locked" and "Closed" in green, orange and red colors. 


Bildschirmfoto 2021-06-10 um 08 22 42
Bildschirmfoto 2021-06-10 um 08 23 14

app/views/admins/user_entry/_close_action.haml
tclaus: Should be ignored, I don't like to split the ternary into multiple lines. Other Ideas?

SuperTux88:
> It is intended that the serialized_public_key is set to "BLOCKED" ( but it don't work now)

Don't do that. That's a hacky workaround from a manual script for people who know what they are doing and know how to fix it again when things breaks. It isn't guaranteed that manually changing keys won't cause problem in the future, so this is not something that should be done here (a similar hacky workaround with messing with the key in the past did break things eventually).

The clean solution would be just deny federation messages from closed accounts (maybe except `AccountMigration` messages, so you could still mark an account as merged even after it was closed?). And that way it then also could cleanly reject relayables (if a closed account tries to comment on a local post, like if the person is blocked).

And if the button closes an account it should say "close", not "block".

The federation blocking from closed accounts doesn't necessarily need to be done in the scope of this PR, but just don't modify the key here.

Flaburgan:
>And if the button closes an account it should say "close", not "block".

But what "Close" does mean for a remote account?

SuperTux88:
Maybe "Close locally" or something?

The alternative would be to introduce a "blocked" state of a person and use that instead. The advantage of that would be, that it can be undone, and can be differentiated from the closed state (so people trying to visit a blocked account could get a "this account was blocked" message instead of the message that the account was closed), and we could still see when the account was closed for real (by either the user itself or the remote podmin) ... and we could even keep the "blocked" state when somebody migrates a previously blocked account to a new pod, which we probably want to block again.

tclaus:
@SuperTux88 
Can you please give some more detailed answers on this: ? 

- Don't delete public Key,right? Remove: app/workers/wipe_account.rb:46 ?
- The Text on button: "Close" (as before), but "Close locally" on remotes. Is is OK to name "Wipe and close". Look into the UI: it's clear which user is local or remote, so it might be redundant to name the button "close locally". See first screenshots here.

- A "Block" state may come with a model change. Shouldn't be done in this PR.

- Can you please give a line to

> deny federation messages from closed accounts

"Closed" accounts still spill new posts or comments into the pod. But I am still not so deep into architecture to find the 'right' point in source.

SuperTux88:
> Don't delete public Key,right? Remove: app/workers/wipe_account.rb:46 ?

Yes

> The Text on button: "Close" (as before), but "Close locally" on remotes. Is is OK to name "Wipe and close". Look into the UI: it's clear which user is local or remote, so it might be redundant to name the button "close locally". See first screenshots here.

To people who know how diaspora works, it should be clear that "close" of a remote account doesn't allow to close the account on other pods. But people who are new to diaspora "close locally" maybe is cleaner.

> A "Block" state may come with a model change. Shouldn't be done in this PR.

If a blocked state is added later, the button could be changed to use that then (and not close the account anymore) ... the button then can be renamed to "block" and blocking can also be undone then. But I think it's fine to first do it with closing in this PR and do the separate block state later (also I think this can then be merged to 0.7.x while the blocking then can go to 0.8.x because of model change ... but if it's separated the 0.7.x people can also already profit from it)

> Can you please give a line to

See my comment here: https://github.com/diaspora/diaspora/pull/8228#issuecomment-860084940 (sorry, I accidentally submitted that comment too early, but it's now complete)

app/views/admins/user_entry/_close_action.haml
```diff
@@ -0,0 +1,9 @@
+- if person != @current_user.person && !person.closed_account?
+  = link_to person.owner.present? ? t("admins.user_search.close_account") : t("admins.user_search.close_local_account"),
+    admin_close_account_path(person.id),
+    method: :post, data: {confirm: t("admins.user_search.are_you_sure")},
+    class: "btn btn-danger btn-block"
+  = link_to person.owner.present? ? t("admins.user_search.wipe_and_close_account") : t("admins.user_search.wipe_and_close_local_account"),
```
diaspora-code-review, line 6: Line is too long. [138/120]

tclaus:
@SuperTux88 
Changed UI facing texts. 
Added person.closed_account? in Diaspora_federation. 

To block receiving messages is mainly implemented in #8228 and will come into this code after merging. (Added a TODO on the lines) Will clean this lines while expecting merge conflicts then. 
Bildschirmfoto 2021-06-15 um 07 58 54

goobertron:
This looks very useful, and thanks for working on it, @tclaus. Just one question: are the user accounts in your screen-shots real accounts or did you invent them? If they're real accounts, would you mind blurring the personal details in your screenshots, or replacing them with invented ones? Just for personal privacy.

Issue #8106, opened 2020-02-10
Contact status not properly preserved after account migration


jhass:
Or so is my guess.

I got the following state between two now local users:

```
# SELECT users.id, people.id FROM users JOIN people ON people.owner_id = users.id WHERE users.id IN (2, 3);
 id |  id  
----+------
  2 | 1988
  3 | 2558

# SELECT user_id, person_id, sharing, receiving FROM contacts WHERE (user_id = 2 AND person_id = 2558) OR (user_id = 3 AND person_id = 1988);
 user_id | person_id | sharing | receiving 
---------+-----------+---------+-----------
       3 |      1988 | f       | t
       2 |      2558 | t       | t
```

Where user ID 3 is a migrated account and user ID 2 is a locally created account. Note how the migrated account thinks the other user is not sharing with them but the other one does think so (and is right).

Might be worthwhile to try to recreate this in a test setup (I didn't yet).

Given two pods
Given a user on each pod
Ensure both are fully sharing with each other
Migrate one user to the other users pod
Check whether full sharing status is preserved  after migration.

PR #8254, opened 2021-06-17
8106 Migration: set share status in contacts

tclaus:
It sets the sharing attribute in contacts  from import file in migration

### DIFF
--- a/spec/integration/migration_service_spec.rb
+++ b/spec/integration/migration_service_spec.rb
@@ -287,17 +287,17 @@ describe MigrationService do
 
       contact = user.contacts.find_by(person: Person.by_account_identifier(contact1_diaspora_id))
       expect(contact).not_to be_nil
-      expect(contact.sharing).to be_falsey
+      expect(contact.sharing).to be_truthy
       expect(contact.receiving).to be_falsey
 
       contact = user.contacts.find_by(person: Person.by_account_identifier(contact2_diaspora_id))
       expect(contact).not_to be_nil
-      expect(contact.sharing).to be_falsey
+      expect(contact.sharing).to be_truthy
       expect(contact.receiving).to be_truthy
 
       contact = user.contacts.find_by(person: Person.by_account_identifier(migrated_contact_new_diaspora_id))
       expect(contact).not_to be_nil
-      expect(contact.sharing).to be_falsey
+      expect(contact.sharing).to be_truthy
       expect(contact.receiving).to be_truthy
 
       aspect = user.aspects.find_by(name: "Friends")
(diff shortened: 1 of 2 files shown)


tclaus:
Fixes #8106 Contact status not properly preserved after account migration
Share attribute was ignored  in import

spec/integration/migration_service_spec.rb
```diff
@@ -287,17 +287,17 @@ def setup_validation_time_expectations
...
       contact = user.contacts.find_by(person: Person.by_account_identifier(contact2_diaspora_id))
       expect(contact).not_to be_nil
-      expect(contact.sharing).to be_falsey
+      expect(contact.sharing).to be_truthy
       expect(contact.receiving).to be_truthy
 
       contact = user.contacts.find_by(person: Person.by_account_identifier(migrated_contact_new_diaspora_id))
       expect(contact).not_to be_nil
-      expect(contact.sharing).to be_falsey
+      expect(contact.sharing).to be_truthy
```
Flaburgan, line 300: Is there a case where this can actually be false? How does the import deals with ignored users for example?

tclaus: I'll check.

tclaus: The attribute should always be true (and always be set in migration.json), a user I don't share with, has no existence in the contacts table (as far as I analyzed this). 
A blocked user, don't alter the contacts entry - he/she can be blocked or unblocked at any time without loosing information. 

More interestingly.. I can't find any place where a blocked entry is added to the export file. @cmrd-senya Can you confirm this? In the  https://github.com/diaspora/diaspora/blob/1dd2382d0336c3514ec21e2c41e8c99f541dbbb5/app/serializers/export/user_serializer.rb file there is no "blocked".. or anything else? 

So if all agree, this PR might be OK and ready to approve, but the blocked mechanics needs a review.

SuperTux88: > The attribute should always be true (and always be set in migration.json), a user I don't share with, has no existence in the contacts table (as far as I analyzed this).

No, it's not always true, and a user you don't share with has an entry (`sharing: true, recieving: false`) ... I know this is really confusing, and I just noticed that the `ContactSerializer` also messed this up and `following` and `followed` are the wrong way around (but they aren't used for import yet, so it's "fine"?) ... I need to think about how to handle/fix this the best way, and we probably should un-confuse our database schema with these `sharing`/`receiving` stuff which don't work the way one would expect them to work ... In the federation layer this already got cleaned up and works in a non-confusing way, and the `ContactSerializer` tried to copy the federation model but got it wrong because in the database it's just confusing ...

Anyway ... this PR fixes the problem ... for now at least, so it's fine to merge it ... but we need to revisit this topic because the current state is just a big mess.

Issue #8256, opened 2021-06-18
Migration import stops on different issues


tclaus:
Some Issues while importing a random migration file

rails aborted!
DiasporaFederation::Federation::Fetcher::NotFetchable: Failed to fetch Post:6036f0c5f9f5f09761ab3ed9b3e1f03b8695d1b8 from neo@friendica.produnis.de: RuntimeError: Failed to fetch https://friendica.produnis.de/fetch/post/6036f0c5f9f5f09761ab3ed9b3e1f03b8695d1b8: 404
/Users/thorstenclaus/.rvm/gems/ruby-2.6.6@diaspora/gems/diaspora_federation-0.2.6/lib/diaspora_federation/federation/fetcher.rb:15:in `rescue in fetch_public'
/Users/thorstenclaus/.rvm/gems/ruby-2.6.6@diaspora/gems/diaspora_federation-0.2.6/lib/diaspora_federation/federation/fetcher.rb:10:in `fetch_public'
/Users/thorstenclaus/.rvm/gems/ruby-2.6.6@diaspora/gems/diaspora_federation-0.2.6/lib/diaspora_federation/entities/related_entity.rb:43:in `fetch'
/Users/thorstenclaus/.rvm/gems/ruby-2.6.6@diaspora/gems/diaspora_federation-0.2.6/lib/diaspora_federation/entities/reshare.rb:54:in `validate_root'
/Users/thorstenclaus/.rvm/gems/ruby-2.6.6@diaspora/gems/diaspora_federation-0.2.6/lib/diaspora_federation/entities/reshare.rb:66:in `tap'
/Users/thorstenclaus/.rvm/gems/ruby-2.6.6@diaspora/gems/diaspora_federation-0.2.6/lib/diaspora_federation/entities/reshare.rb:66:in `from_hash'
/Users/thorstenclaus/.rvm/gems/ruby-2.6.6@diaspora/gems/diaspora_federation-0.2.6/lib/diaspora_federation/entity.rb:124:in `from_json'
/Users/thorstenclaus/Documents/develop/web-projects/diaspora/lib/archive_importer/entity_importer.rb:27:in `entity'
/Users/thorstenclaus/Documents/develop/web-projects/diaspora/lib/archive_importer/entity_importer.rb:14:in `import'
/Users/thorstenclaus/Documents/develop/web-projects/diaspora/lib/archive_importer/own_entity_importer.rb:7:in `import'
[...]

Issue #8019, opened 2019-05-20
Optimize archive import with lots of dead (offline pods) contacts


jhass:
So my account migration took about 36 hours or so, that is because it tries to fetch all dead people in my contacts and reshares. If we can somehow  cache that a fetch failed from the validator pass, or just disable the fetch in the import pass, that should provide a tremendous speedup for the case of many dead contacts.

goobertron:
Would it make sense to prune these from the account data before it is packaged for export?

SuperTux88:
No, this needs to be handled during the import. Pods are maybe still up when you export your profile, but are down a week later, or if your export is really old, maybe even years later (and the import works also with an old archive, every archive created with 0.7.0.0 or newer should work for import).

goobertron:
I understand that in some cases an archive will be imported long after export, but I don't think that was the case in Jonne's test. I imagined that in this case it was a lot of long-closed accounts that had been sharing with his account and had never been pruned before.

If that's the case, or if that's a possible case, it might be worth either periodically pruning accounts from dead pods from an account's active sharing relationships (perhaps putting them in an 'inactive' status so they can be re-activated as contacts if the account becomes live again) or doing this pruning at the time of export as well as at import. There then might be a cleaner exported data package, and neither step would take a huge amount of time.

If I'm misunderstanding some technical detail, don't feel you need to spend time trying to explain it to me; just ignore me. I thought I'd make the suggestion in case I had thought of a useful method to think about. If not, that's fine.

SuperTux88:
You need to do it during import anyway, so there is no need to do it twice, and the export is also already a long enough process anyway. And there is the possibility that your new pod already knows the (then not missing) resource, even if the pod is down at the moment, so it can import your reshare, but couldn't if you would have filtered it during export.

PR #8257, opened 2021-06-23
Migration: Fixing import  issues / Speed up repeated unknown pod detection

tclaus:
It handles some migration issues listed in #8256, please check @cmrd-senya, @jhass, @Flaburgan .
(Import stops when posts comments or likes don't have a parent/author),
It handles #8019 (caching person fetches from remotes)

For this reason I changed the Person.find_or_fetch_by_identifier function. It now won't raise a NotFetchable exception, it just returns nil or the person object. Some callers rely on nil (which was never returned), some other rely on exception. IMHO this function should return an existing person from local store or by fetching, but never raise an exception. 
Now,  Person.find_or_fetch_by_identifier  returns a person, or nil. Callers are adapted to this behaviour.

@cmrd-senya : I left over a failing test, which I can't fix, maybe you have on eye on this? 
@jhass : can you check import - times with your archive file?

### DIFF
--- a/lib/archive_importer/own_relayable_importer.rb
+++ b/lib/archive_importer/own_relayable_importer.rb
@@ -2,6 +2,8 @@
 
 class ArchiveImporter
   class OwnRelayableImporter < OwnEntityImporter
+    class NoParentError < RuntimeError; end
+
     def entity
       fetch_parent(symbolized_entity_data)
       entity_class.new(symbolized_entity_data)
@@ -19,6 +21,8 @@ class ArchiveImporter
         break entity_class::PARENT_TYPE if entity_class.const_defined?(:PARENT_TYPE)
       }
       entity = Diaspora::Federation::Mappings.model_class_for(type).find_by(guid: data.fetch(:parent_guid))
+      raise NoParentError if entity.nil?
+
       data[:parent] = Diaspora::Federation::Entities.related_entity(entity)
     end
   end
--- a/spec/lib/archive_importer/post_importer_spec.rb
+++ b/spec/lib/archive_importer/post_importer_spec.rb
@@ -113,5 +113,34 @@ describe ArchiveImporter::PostImporter do
         end
       end
     end
+
+    context "with reshare" do
+      let(:guid) { UUID.generate(:compact) }
+      let(:entity_json) { JSON.parse(<<~JSON) }
+        {
+          "entity_data" : {
+             "created_at" : "2015-10-19T13:58:16Z",
+             "guid" : "#{guid}",
+             "author" : "#{new_user.diaspora_handle}",
+             "root_author" : "author@example.com",
+             "root_guid" : "#{UUID.generate(:compact)}"
+          },
+          "entity_type" : "reshare"
+        }
+      JSON
+
+      context "with unknown root author" do
+        it "handles missing person" do
+          allow(DiasporaFederation::Federation::Fetcher).to receive(:fetch_public)
+            .and_raise(DiasporaFederation::Federation::Fetcher::NotFetchable)
+
+          expect {
+            instance.import
+          }.not_to raise_error
+
+          expect(Reshare.find_by(guid: guid)).to be_nil
+        end
+      end
+    end
   end
 end
--- a/lib/archive_importer/entity_importer.rb
+++ b/lib/archive_importer/entity_importer.rb
@@ -15,12 +15,14 @@ class ArchiveImporter
     rescue DiasporaFederation::Entities::Signable::SignatureVerificationFailed,
            DiasporaFederation::Discovery::InvalidDocument,
            DiasporaFederation::Discovery::DiscoveryError,
+           DiasporaFederation::Federation::Fetcher::NotFetchable,
+           OwnRelayableImporter::NoParentError,
            ActiveRecord::RecordInvalid => e
       logger.warn "#{self}: #{e}"
+      self.persisted_object = nil
     end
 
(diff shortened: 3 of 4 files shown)


spec/federation_callbacks_spec.rb
```diff
@@ -231,14 +231,12 @@
       expect(key.to_s).to eq(person.serialized_public_key)
     end
 
-    it "forwards the DiscoveryError when the person can't be fetched" do
+    it "return nil if the person can't be fetched" do
       diaspora_id = Fabricate.sequence(:diaspora_id)
       expect(Person).to receive(:find_or_fetch_by_identifier).with(diaspora_id)
-        .and_raise(DiasporaFederation::Discovery::DiscoveryError)
+        .and_return(nil)
```
diaspora-code-review, line 237: Layout/MultilineMethodCallIndentation: Align `.and_return` with `.with` on line 236.

tclaus:
Reminder: To enable/disable local caching on your dev machine you can toggle it (without changing configuration files) with: 
$: rails dev:cache

I also added a statement to the spec helpers to clear any caches between each test.

spec/lib/archive_importer/post_importer_spec.rb
```diff
@@ -113,5 +113,39 @@
...
+        it "doesn't raise error" do
+          expect {
+            instance.import
+          }.not_to raise_error
+        end
+      end
+    end
   end
-end
+end
```
diaspora-code-review, line 151: Layout/TrailingEmptyLines: Final newline missing.

lib/archive_importer/own_relayable_importer.rb
```diff
@@ -19,7 +21,9 @@ def fetch_parent(data)
         break entity_class::PARENT_TYPE if entity_class.const_defined?(:PARENT_TYPE)
       }
       entity = Diaspora::Federation::Mappings.model_class_for(type).find_by(guid: data.fetch(:parent_guid))
-      data[:parent] = Diaspora::Federation::Entities.related_entity(entity)
+      raise NoParentError if entity.nil?
```
SuperTux88, line 24: The same question as in https://github.com/diaspora/diaspora/pull/8010#discussion_r281341950 still exists. In theory this shouldn't happen, because in the validate step it already tries to fetch all the entities here

https://github.com/diaspora/diaspora/blob/f85f167f505985c6c34dc77f5c29493e28d7aa20/lib/archive_validator/own_relayable_validator.rb#L15-L17

using

https://github.com/diaspora/diaspora/blob/f85f167f505985c6c34dc77f5c29493e28d7aa20/lib/diaspora/federated/fetchable.rb#L9-L17

which already returns nil when not being able to be fetched, and if the parent is nil, it's not valid:

https://github.com/diaspora/diaspora/blob/a3196a108946eb331d007add064182e558e448aa/lib/archive_validator/relayable_validator.rb#L16-L18

which then should remove all invalid entities during the validation phase, so they aren't tried to be actually imported:

https://github.com/diaspora/diaspora/blob/f85f167f505985c6c34dc77f5c29493e28d7aa20/lib/archive_validator/collection_validator.rb#L9-L13

If this doesn't work as expected, it also means a lot of the things are done twice, which isn't intended (and could also lead to longer runtimes). If you have steps to reproduce this and can debug why entities that don't exist still end up at this point and aren't filtered out earlier already, it would be good to know and should be fixed there instead.

SuperTux88 (changes requested):
This isn't a full review yet, just a few comments about some basic stuff which probably changes a lot of the rest of the PR anyway.

tclaus:
@SuperTux88 I give up. Sorry - I didn't understand the validation process in depth so that I can check all needed authors, parents and entries and fix tests and implement all as a validation.

lib/archive_importer/own_relayable_importer.rb
tclaus: The part: find_by(guid: data.fetch(:**parent_guid**))  makes the problem here, not detected by the validator. 
But we already talked that it seems to be inconvenient to fetch everything in the validation part.

Flaburgan:
> @SuperTux88 I give up. Sorry - I didn't understand the validation process in depth so that I can check all needed authors, parents and entries and fix tests and implement all as a validation.

If I understand correctly, you since had a call with @SuperTux88 to solve this point. So what is the status of that pull request now?

tclaus:
My status is, that first idea was to push checks into the validation part, which did not succeed because of needed code changes. Another (more inportant) point was that fetching everything during validation is not what a validation process should do. 

After some code attempts, I implemented the smallest possible solution in this PR. 
To get forward and have one working branch I also rebased the upload-migration file branch  to this work. 

Status is: it just works. (Until someone else Codes a „better“ solution).

tclaus:
@SuperTux88 As discussed, I have tried to make the smallest possible working solution without rebuilding the validation part. This solution gives what it should give.
With this in mind - could you check this again and possible approve (and merge) this? 
It will be let the #8274 working.

SuperTux88:
I ended up merging https://github.com/diaspora/diaspora/pull/8010 because it had separate commits for the two different problems and also had a test for both problems. But I merged in the test from here since this was more generic (tests generic fetch problem and not only discovery errors).

Thanks for the work on this anyway since we finally found an answer for the original question in that PR :cookie:

Flaburgan:
Awesome to finally see this merged!

tclaus:
Thanks for your analysis - to finally found a solution.

PR #8258, opened 2021-06-24
Migration: Adds user blocks to export/import profile

tclaus:
This PR adds archiving and importing the blocks model.
@Flaburgan @cmrd-senya

### DIFF
--- a/app/serializers/export/user_serializer.rb
+++ b/app/serializers/export/user_serializer.rb
@@ -10,7 +10,8 @@ module Export
                :show_community_spotlight_in_stream,
                :auto_follow_back,
                :auto_follow_back_aspect,
-               :strip_exif
+               :strip_exif,
+               :blocks
     has_one    :profile, serializer: FederationEntitySerializer
     has_many   :contact_groups, each_serializer: Export::AspectSerializer
     has_many   :contacts, each_serializer: Export::ContactSerializer
@@ -43,6 +44,10 @@ module Export
       [comments, likes, poll_participations].map {|relayable|
         relayable.find_each(batch_size: 20)
       }
+    end
+
+    def blocks
+      object.blocks.map(&:person_diaspora_handle)
     end
 
     %i[comments likes poll_participations].each {|collection|
--- a/lib/archive_importer.rb
+++ b/lib/archive_importer.rb
@@ -47,7 +47,7 @@ class ArchiveImporter
 
   def import_blocks
     blocks = archive_hash["user"]["blocks"]
-    return if blocks.present?
+    return if blocks.nil?
 
     blocks.each do |blocked_id|
       p = Person.find_or_fetch_by_identifier(blocked_id)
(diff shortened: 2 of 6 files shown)


lib/archive_importer.rb
```diff
@@ -37,12 +37,26 @@ def create_user(attr)
...
 
   private
 
   attr_reader :archive_hash
 
+  def import_blocks(attr, user)
+    return unless attr[:blocks]
+
+    attr[:blocks].each do |blocked_id|
+      p = Person.find_or_fetch_by_identifier(blocked_id)
```
SuperTux88, line 51: This raises an error when the person isn't known and can't be fetched, which would abort the whole import. Also, if the blocked account was migrated between the export-creation and the import, you want to block the new user, not the old one. So in the end you probably want a similar logic for blocks as it already exists for contacts here: https://github.com/diaspora/diaspora/blob/develop/lib/archive_validator/contact_validator.rb

lib/archive_importer.rb
```diff
@@ -37,12 +37,26 @@ def create_user(attr)
     )
     self.user = User.build(data)
     user.save!
+    import_blocks(attr, user)
```
SuperTux88, line 40: This should be moved to the `import` method above (line 13) where all the other stuff is imported to. If the user already exists `create_user` isn't called, so blocks wouldn't be imported.

app/serializers/export/user_serializer.rb
```diff
@@ -45,6 +46,10 @@ def relayables
       }
     end
 
+    def blocks
+      object.blocks.map(&:person).map(&:diaspora_handle)
```
SuperTux88, line 50: I think `.map(&:person).map(&:diaspora_handle)` should be done in one `map` step, to not need to first create a full list of persons. But you can add `diaspora_handle` to the delegated methods in the block class, so you can just do `.map(&:person_diaspora_handle)`.

SuperTux88 (changes requested):
Can you please split this up in two separate pull requests, one for the archive creation additions (for 0.7.x) and one for the import part (for 0.8.x).

Also I think the import part needs more work, to make it more robust, so it works similar to the contacts import (maybe can even be done to share code to have less duplication?).

app/serializers/export/user_serializer.rb
tclaus: Today I learned about delegate methods. Thanks.

lib/archive_importer.rb
tclaus: Do you mean the line: ` person.account_migration.newest_person` to fetch migrated user? 

The 'find_or_fetch_by_identifier' method will not throw a exception any more if person can not be fetched, (PR #8257 will introduce this). Many methods rely on returning nil if person is not found, so I globally made this even for callers. 
See:  https://github.com/diaspora/diaspora/blob/d27eefeb3471844179c86c8dc7c28ad99801f40f/lib/archive_validator/contact_validator.rb#L34

Until 8257 a Person-fetch might never return nil if a person is not found.

lib/archive_importer.rb
tclaus: I had to use both - a user could set this on create or on update an existing user

tclaus:
Closed, because of splitting import/export in two different branches. As requested.

PR #8260, opened 2021-06-27
Migration: Import compressed archive file handling

tclaus:
It fixes the '# TODO: archive is likely to be a .json.gz file' in migration_service.rb . 

A profile export is downloaded as a gz file, which can not be imported directly before this PR without unzipping.
By uncompressing a gz file, it also handles a zip file (maxOS and Windows create Zip on compressing files by default)
It won't overwrite any existing files. 
It removes the uncompressed file after handling it (without removing plain json file, if uncompressed file was given) 
@Flaburgan @cmrd-senya  What do you think?

### DIFF
--- a/spec/integration/migration_service_spec.rb
+++ b/spec/integration/migration_service_spec.rb
@@ -188,7 +188,7 @@ describe MigrationService do
   let(:new_username) { "newuser" }
   let(:new_user_handle) { "#{new_username}@#{AppConfig.bare_pod_uri}" }
 
-  let(:archive_file) { Tempfile.new("archive") }
+  let(:archive_file) { Tempfile.new(%w[archive .json]) }
 
   def setup_validation_time_expectations
     expect_person_fetch(contact2_diaspora_id, nil)
@@ -311,6 +311,49 @@ describe MigrationService do
       comment = Comment.find_by(guid: others_comment_entity.guid)
       expect(comment.author.diaspora_handle).to eq(others_comment_entity.author)
       expect(comment.parent.author.diaspora_handle).to eq(user.diaspora_handle)
+    end
+  end
+
+  context "compressed archives" do
+    it "uncompresses gz archive" do
+      gz_compressed_file = create_gz_archive
+      service = MigrationService.new(gz_compressed_file, new_username)
+      uncompressed_file = service.send(:archive_file)
+      json = uncompressed_file.read
+      expect {
+        JSON.parse(json)
+      }.not_to raise_error
+    end
+
+    it "uncompresses zip archive" do
+      zip_compressed_file = create_zip_archive
+      service = MigrationService.new(zip_compressed_file, new_username)
+      uncompressed_file = service.send(:archive_file)
+      json = uncompressed_file.read
+      expect {
+        JSON.parse(json)
+      }.not_to raise_error
+    end
+
+    def create_gz_archive
+      target_file = Tempfile.new(%w[archive .json.gz]).path
+      Zlib::GzipWriter.open(target_file) do |gz|
+        File.open(archive_file.path).each do |line|
+          gz.write line
+        end
+      end
+      target_file
+    end
+
+    def create_zip_archive
+      target_file = Tempfile.new(%w[archive .zip]).path
+      Zip::OutputStream.open(target_file) do |zip|
+        zip.put_next_entry("archive.json")
+        File.open(archive_file.path).each do |line|
+          zip.write line
+        end
+      end
+      target_file
     end
   end
 
--- a/app/services/migration_service.rb
+++ b/app/services/migration_service.rb
@@ -19,6 +19,8 @@ class MigrationService
     find_or_create_user
     import_archive
     run_migration
+  ensure
+    remove_intermediate_file
   end
 
   # when old person can't be resolved we still import data but we don't create&perform AccountMigration instance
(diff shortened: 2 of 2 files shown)


tclaus:
Will be integrated into bigger PR with UI

Flaburgan:
The smaller the PRs are, the easier it is to review them. If that works independently of the UI, I'm in favor a separate PR (I can re-open this one). You probably want to add tests to see it merged though ;)

tclaus:
Thats an argument. 
The mentioned "bigger" Branch currently evolves here: https://github.com/tclaus/diaspora/tree/upload_migration_files

weex:
>(But it will sit on top of this, I need the zip file handling)

I read this to mean this is WIP or draft.

tclaus:
> I read this to mean this is WIP or draft.

No, it's not really a WIP or Draft.

This PR is part of the "Migration" Feature. The zipped file support can be used / merged without the other parts.

Flaburgan:
So, I'm trying this locally with the rake task and it fails:
```
bin/rake accounts:migration 
WARNING: diaspora.yml is deprecated and will no longer be read in diaspora 0.9.
         Please copy over diaspora.toml.example to diaspora.toml and migrate your settings from diaspora.yml.
/var/lib/gems/2.7.0/gems/faraday-0.15.4/lib/faraday/options.rb:166: warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
/var/lib/gems/2.7.0/gems/faraday-0.15.4/lib/faraday/options.rb:166: warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
/var/lib/gems/2.7.0/gems/faraday-0.15.4/lib/faraday/options.rb:166: warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
/var/lib/gems/2.7.0/gems/faraday-0.15.4/lib/faraday/options.rb:166: warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
/var/lib/gems/2.7.0/gems/faraday-0.15.4/lib/faraday/options.rb:166: warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
/var/lib/gems/2.7.0/gems/activerecord-5.2.6/lib/active_record/type.rb:27: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/var/lib/gems/2.7.0/gems/activerecord-5.2.6/lib/active_record/type/adapter_specific_registry.rb:9: warning: The called method `add_modifier' is defined here
/var/lib/gems/2.7.0/gems/faraday-0.15.4/lib/faraday/rack_builder.rb:55: warning: Capturing the given block using Proc.new is deprecated; use `&block` instead
/var/lib/gems/2.7.0/gems/actionpack-5.2.6/lib/action_dispatch/middleware/stack.rb:37: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/var/lib/gems/2.7.0/gems/actionpack-5.2.6/lib/action_dispatch/middleware/static.rb:111: warning: The called method `initialize' is defined here
WARNING: Sidekiq::Web.sessions= is no longer relevant and will be removed in Sidekiq 7.0. /var/lib/gems/2.7.0/gems/sidekiq-6.2.1/lib/sidekiq/web.rb:75:in `set'
Account migration is requested
Enter the archive path: ../path_diaspora_data_l1hBiuYRhMBsbU0TgzZPwQ.json.gzip
Enter the new user name: test
Archive path: ../path_diaspora_data_l1hBiuYRhMBsbU0TgzZPwQ.json.gzip
New username: test
rake aborted!
NoMethodError: undefined method `content_type' for nil:NilClass
/home/fla/diaspora/diaspora/app/services/migration_service.rb:83:in `zip_file?'
/home/fla/diaspora/diaspora/app/services/migration_service.rb:76:in `archive_file'
/home/fla/diaspora/diaspora/app/services/migration_service.rb:72:in `archive_validator'
/home/fla/diaspora/diaspora/app/services/migration_service.rb:13:in `validate'
/home/fla/diaspora/diaspora/lib/tasks/accounts.rake:12:in `block (2 levels) in <top (required)>'
Tasks: TOP => accounts:migration
(See full trace by running task with --trace)
```

tclaus:
I see the problem. 
But in the meantime, test it with gz or zip file suffix.

Flaburgan:
Wow you're actually right, it's because the path is wrong, the file is named `.gz` but I typed `.gzip`. Well I guess we should have an error message or something saying the file can't be found instead of a stacktrace.

tclaus:
My workflow was to drag and drop the file from finder (Mac) directly into the terminal. The Filepath is then copied. 
But it's an issue that needs to get fixed.

lib/tasks/accounts.rake
```diff
@@ -26,6 +26,8 @@ namespace :accounts do
       puts "Errors in the archive found:\n#{exception.message}\n-----"
     rescue MigrationService::MigrationAlreadyExists
       puts "Migration record already exists for the user, can't continue"
+    ensure
+      service.remove_intermediate_file
```
SuperTux88, line 30: You can just write `ensure` in the `MigrationService` in `perform!` (above the line of calling `remove_intermediate_file` there), that way `remove_intermediate_file` doesn't need to be called twice (and it also ensures that the temp files get cleaned up when `perform!` is called from somewhere else). It also means `remove_intermediate_file` can be private since it doesn't need to be called from external.

spec/integration/migration_service_spec.rb
```diff
@@ -314,6 +314,50 @@ def setup_validation_time_expectations
     end
   end
 
+  context "compressed archives" do
+    let(:old_person) {
```
SuperTux88, line 318: This isn't used anywhere in this context.

spec/integration/migration_service_spec.rb
```diff
@@ -314,6 +314,50 @@ def setup_validation_time_expectations
...
+      FactoryBot.create(:person,
+                        profile:               FactoryBot.build(:profile),
+                        serialized_public_key: archive_private_key.public_key.export,
+                        diaspora_handle:       archive_author)
+    }
+
+    it "uncompresses gz archive" do
+      gz_compressed_file = create_gz_archive
+      service = MigrationService.new(gz_compressed_file, new_username)
+      expect(service.only_import?).to be_truthy
```
SuperTux88, line 328: Instead of testing `only_import?` (which has nothing to do with uncompressing) it would be better to test if `service.archive_file` returns the uncompressed file (differerent path than `gz_compressed_file` and has actual json content and isn't compressed anymore). You can use `service.send(:archive_file)` to access the private `archive_file` method from the test.

spec/integration/migration_service_spec.rb
```diff
@@ -314,6 +314,50 @@ def setup_validation_time_expectations
...
+
+    it "uncompresses zip archive" do
+      zip_ccompressed_file = create_zip_archive
+      service = MigrationService.new(zip_ccompressed_file, new_username)
+      expect(service.only_import?).to be_truthy
+    end
+
+    def create_gz_archive
+      target_path = File.dirname(archive_file.path)
+      target_file = File.join(target_path, "archive.json.gz")
```
SuperTux88, line 339: You should ensure that compressed files also are removed again after the test, so best is probably to also use a `Tempfile` instead.

SuperTux88 (changes requested):

spec/integration/migration_service_spec.rb
tclaus: Thanks for giving code snippets

lib/tasks/accounts.rake
tclaus: @SuperTux88 Thanks for reviewing this. In context of #8274 this rake file was more refactored to eliminate duplicate code. So I think for now this file can be as it is.

SuperTux88: This was more a request to also change `MigrationService.perform!` and add the `ensure` directly there (and when `remove_intermediate_file` doesn't need to be called here anymore it can also be made private).
```ruby
  def perform!
    find_or_create_user
    import_archive
    run_migration
  ensure
    remove_intermediate_file
  end
```

SuperTux88 (approved):

SuperTux88:
Squashed and merged, thanks :cookie:

PR #8262, opened 2021-06-28
Cleanup unused commenting_disabled? and can_comment?

SuperTux88:


jhass (approved):

Flaburgan:
Wow, I had no idea we had that

SuperTux88:
Merged as c203c1eb940779e439ed5b9591fb38991ab5497f

@Flaburgan probably a leftover from a looooooong time ago back from the beginning when comments weren't relayed yet, so there where more restrictions who can send comments to whom (because federation reasons). But I didn't really go back to check when this was in use the last time, so I could also be wrong with my guess.

PR #8263, opened 2021-07-02
Migration: Exporting user blocks

tclaus:

PR #8264, opened 2021-07-02
Migration: Importing user blocks

tclaus:


lib/schemas/archive-format.json
```diff
@@ -20,7 +20,12 @@
           ]
         },
         "strip_exif": { "type": "boolean" },
-
+        "blocks": {
```
SuperTux88, line 23: I added this change to `next-minor` when merging #8263 as 342d4f7fde6f12cbfc9cc2f438936d099548ee3e (so you can rebase and remove this here)

SuperTux88 (changes requested):

lib/archive_importer.rb
```diff
@@ -37,12 +38,35 @@ def create_user(attr)
...
 
   attr_reader :archive_hash
 
+  def import_blocks
+    blocks = archive_hash["user"]["blocks"]
+    return if blocks.nil?
+
+    blocks.each do |blocked_id|
+      p = Person.find_or_fetch_by_identifier(blocked_id)
+      if p.present?
```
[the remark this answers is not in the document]
tclaus: > `find_or_fetch_by_identifier` should never return nil

I've already read the longer explanation in the other PR. I'll take this as a rule of thump. 
(Remark: there were some callers lurking around which didn't respect this and assumed 'nil' when calling this. Thats why I was confused what is really expected)

lib/archive_importer/blocks_importer.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class ArchiveImporter
+  class BlocksImporter
+    include Diaspora::Logging
+
```
diaspora-code-review, line 6: Layout/TrailingWhitespace: Trailing whitespace detected.

lib/archive_importer/blocks_importer.rb
```diff
@@ -0,0 +1,32 @@
...
+class ArchiveImporter
+  class BlocksImporter
+    include Diaspora::Logging
+  
+    def initialize(json, user)
+      @json = json
+      @user = user
+    end
+
+    attr_reader :json
```
diaspora-code-review, line 12: Style/AccessorGrouping: Group together all `attr_reader` attributes.

lib/archive_importer/blocks_importer.rb
```diff
@@ -0,0 +1,32 @@
...
+    attr_reader :user
```
diaspora-code-review, line 13: Style/AccessorGrouping: Group together all `attr_reader` attributes.

lib/archive_importer/blocks_importer.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def import
+      p = Person.find_or_fetch_by_identifier(json)
+      migrant_person = handle_migrant_person(p)
+      user.blocks.create(person_id: migrant_person.id) if migrant_person.present?
+    rescue ActiveRecord::RecordInvalid,
+            DiasporaFederation::Discovery::DiscoveryError => e
```
diaspora-code-review, line 20: Layout/ArrayAlignment: Align the elements of an array literal if they span more than one line.

lib/archive_importer/blocks_importer.rb
```diff
@@ -0,0 +1,32 @@
...
+      logger.warn "#{self}: #{e}"
+    end
+
+    private
+
+    def handle_migrant_person(person)
+      return person if person.account_migration.nil?
+
```
diaspora-code-review, line 28: Layout/TrailingWhitespace: Trailing whitespace detected.

lib/archive_importer/blocks_importer.rb
```diff
@@ -0,0 +1,32 @@
...
+    private
+
+    def handle_migrant_person(person)
+      return person if person.account_migration.nil?
+  
+      person.account_migration.newest_person
+    end
+  end
+end
```
diaspora-code-review, line 32: Layout/TrailingEmptyLines: Final newline missing.

spec/lib/archive_importer_spec.rb
```diff
@@ -138,14 +139,24 @@
...
     }
 
     it "creates user" do
+      # an unknown blocked person
+      stub_request(:get, "https://unknown.pod/.well-known/webfinger?resource=acct:bad_guy@unknown.pod")
+        .to_return(status: 404, body: "", headers: {})
+      stub_request(:get, "https://unknown.pod/.well-known/host-meta")
+        .to_return(status: 404, body: "", headers: {})
+      stub_request(:get, "http://unknown.pod/.well-known/host-meta")
+        .to_return(status: 404)
```
SuperTux88, line 155: Instead of mocking the separate requests (which URLs are called in which order is part of the internals of the federation gem, and the old `host-meta` fallback will eventually be removed ... I know that other places like this already exist in other tests, which is probably where you copied that from, but still less to cleanup later), you maybe want to just make the discovery to directly raise a `DiscoveryError` (which is part of the public API of the federation gem). This is pretty easy if you don't care about which person is tried to be discovered and just want to make everything fail (which works when you move this to a `block_importer_spec` and only care about a single block per test, see my other comment about that). You can look about how this is already done here:

https://github.com/diaspora/diaspora/blob/671e8476ba30b78fa0478955bf2b41a03c26de9d/spec/lib/archive_validator/contact_validator_spec.rb#L59-L62

SuperTux88 (changes requested):
Only a few small things left, but otherwise looks good.

lib/archive_importer.rb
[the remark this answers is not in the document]
SuperTux88: If you find other places which expect `nil`, they probably need to be fixed. But as said, I think most places just can't do what they want to do anyway if the person can't be fetched, so they just let the `DiscoveryError` raise. Or are the other places which you are talking about also part of the importer which then break the import?

spec/lib/archive_importer_spec.rb
```diff
@@ -138,14 +139,24 @@
...
+
       expect {
         archive_importer.create_user(username: "new_name", password: "123456")
+        archive_importer.import
```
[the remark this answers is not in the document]
tclaus: I am not sure.. shouldn't a 'create_user' create everything what is directly attached to a user, so the blocks should also be imported here? Thats why I added import_blocks call in the create method first. The user import is correctly tested if also blocks are imported/created, or am I wrong?
If I create a new blocks_importer_spec, I have the feeling that there I would also test it during the create_user call?

spec/lib/archive_importer_spec.rb
tclaus: Thats cool - I saw it already somewhere.. but thanks for this snippet!

spec/lib/archive_importer_spec.rb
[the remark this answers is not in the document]
SuperTux88: No, otherwise you would need to test the whole import here. Contacts also aren't tested here and blocks are very similar to contacts (anti contacts?). The profile is a bit of a special case, since it's very closely attached to the user and is automatically created when you create the user anyway, so you as well can directly provide it with the correct data instead of first creating an empty profile and then insert the profile data later.

spec/lib/archive_importer_spec.rb
SuperTux88: This only works that way, when you only want to fetch and fail a single user, because it just makes all user fetches fail. That's also why I mentioned this should be moved to a `BlockImporter` spec, where each case is tested separately. If you would want to also test a successful fetch, this would fail here.

tclaus: (Hm.. a successful fetch means a 'real' http call.. Is this allowed during tests, did you mean this? ) But I'll see the need to move it to a separate class.

SuperTux88: You can mock a successful fetch without a http call in a similar way:

https://github.com/diaspora/diaspora/blob/671e8476ba30b78fa0478955bf2b41a03c26de9d/spec/lib/archive_validator/contact_validator_spec.rb#L49-L51

(but this conflicts with the first mocking, so it also only works with a single fetch, which isn't a problem when you test each case separately)

spec/lib/block_importer_spec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+describe "#import blocks" do
```
SuperTux88, line 3: The `block_importer_spec.rb` should also be in the `spec/lib/archive_importer` directory (as is the `block_importer.rb` is also in a `archive_importer` sub-directory), and then it should be wrapped in `describe ArchiveImporter::BlockImporter do`, and then only test the stuff from the `BlockImporter`.

So it doesn't create an instance of `ArchiveImporter`, it only creates instances of `BlockImporter` (and run `.import` on it) and then test the different scenarios you want to test for it.

You can have a look at the other tests in `spec/lib/archive_importer`, but since the `BlockImporter` is a simpler than other importers (it's basically just a single diaspora ID as string, and not a complex json structure as parameter), you tests probably also going to look simpler.

(also you still have to remove the modifications from the `archive_importer_spec.rb`)

spec/lib/archive_importer/block_importer_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+describe ArchiveImporter::BlockImporter do
+
```
diaspora-code-review, line 4: Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body beginning.

spec/lib/archive_importer/block_importer_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:target) { FactoryBot.create(:user) }
+
+  describe "#import" do
+
```
diaspora-code-review, line 8: Layout/TrailingWhitespace: Trailing whitespace detected.

spec/lib/archive_importer/block_importer_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+describe ArchiveImporter::BlockImporter do
+
+  let(:target) { FactoryBot.create(:user) }
+
+  describe "#import" do
+    
+    it "adds a block entry to the user" do
+      person_to_block = FactoryBot.create(:person)
+
```
diaspora-code-review, line 11: Layout/TrailingWhitespace: Trailing whitespace detected.

spec/lib/archive_importer/block_importer_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+        DiasporaFederation::Discovery::DiscoveryError, "discovery error reasons"
+      )
+
+      block_importer = ArchiveImporter::BlockImporter.new("unknown_person@bad_pod.tld", target)
+      block_importer.import
+
+      expect(target.blocks).to be_empty
+    end
+  end
+end
```
diaspora-code-review, line 30: Layout/TrailingEmptyLines: Final newline missing.

spec/lib/block_importer_spec.rb
SuperTux88: > (also you still have to remove the modifications from the `archive_importer_spec.rb`)

Removed that myself.

PR #8069, opened 2019-10-18
Remove the JSXC/Prosody integration.

skyschub:
As per discussion on https://discourse.diasporafoundation.org/t/removing-diaspora-s-current-chat-integration/2718, nobody raised serious concerns or objections. Given future plans, we do not think having an unfinished implementation of something that likely will not get finished in the current form is worth it. So let's get rid of it.

I deliberately ignored all codestyle warnings. As this PR is exclusively removing things, I felt better not touching anything. Happy to change things if someone feels like it.

When merged, this PR needs documentation in the Changelog that explains the implications.

PR #8265, opened 2021-07-03
Don't export chat_enabled flag anymore

SuperTux88:
The chat is already removed for 0.8, so there is no need to still export this data since it can't be imported anyway.

Related to #8069


jhass (approved):

SuperTux88:
Merged as 8f804e376afdadcdc3303bdbc2ae40c92fac8a3f

PR #8267, opened 2021-07-04
Downgrade ethon because of https://github.com/typhoeus/ethon/issues/194

SuperTux88:
There is no release with the "fix" (revert) yet, so just downgrading to "fix" the issue.


skyschub (approved):

SuperTux88:
Merged as 7e9987b7f568c3b6a44a398008a967515a84ff73

Thanks for the review and the help to find the cause of the problems @denschub :cookie:

PR #8268, opened 2021-07-07
Switch to terser gem from uglifier for JS compression

pravi:
Terser is better maintained and many projects like webpack switched to it.


config/initializers/sprockets.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+Sprockets.register_compressor 'application/javascript', :terser, Terser::Compressor
```
diaspora-code-review, line 3: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

pravi: Fixed

pravi:
Many tests fail with similar error (Uglifier seems hard coded here).

```
  1) StreamsController#aspects jasmine fixtures generates a jasmine fixture with a prefill
     Failure/Error: @body = Uglifier.compile(src)

     ActionView::Template::Error:
       uninitialized constant Uglifier
     # ./lib/bookmarklet_renderer.rb:33:in `compile'
     # ./lib/bookmarklet_renderer.rb:27:in `body'
     # ./app/helpers/application_helper.rb:41:in `bookmarklet_code'
     # ./app/views/streams/main_stream.html.haml:134:in `_app_views_streams_main_stream_html_haml__1243015712165079052_47019520168920'
     # ./app/controllers/streams_controller.rb:76:in `block (2 levels) in stream_responder'
     # ./app/controllers/streams_controller.rb:75:in `stream_responder'
     # ./app/controllers/streams_controller.rb:21:in `aspects'
     # ./spec/controllers/jasmine_fixtures/aspects_spec.rb:27:in `block (4 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # NameError:
     #   uninitialized constant Uglifier
     #   ./lib/bookmarklet_renderer.rb:33:in `compile'
```

config/initializers/sprockets.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+Sprockets.register_compressor "application/javascript", :terser, Terser::Compressor
```
SuperTux88, line 3: It looks like this isn't needed to register it manually.

SuperTux88 (changes requested):

Gemfile
```diff
@@ -47,7 +47,7 @@ gem "sidekiq-cron", "1.2.0"
 
 # Compression
 
-gem "uglifier", "4.2.0"
+gem "terser", "~> 1.0"
```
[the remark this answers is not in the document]
pravi: Updated, but isn't exact versions stored in Gemfile.lock already? `bundle update --conservative` will update only required gems even if the dependency is open.

config/initializers/sprockets.rb
pravi: Removed

SuperTux88:
Merged as 20a3abd86443365b2b4f8f63b1a1ab32389a29b2

Thanks :cookie:

PR #8273, opened 2021-07-20
Person.find_or_fetch_by_identifier never return nil

tclaus:
Person.find_or_fetch_by_identifier raises an exception if person is not found localy and not fetchable. It never returns nil. These code changes take care about this behaviour and changes specs and code to behave equally.

Its a very useful to keep in mind this for user-imports.


SuperTux88:
Merged as 8691e650dc726fb929df33409475594111b231da and cdb50edabceb43570b9845b224d82d26ec5b154d

Thanks for the cleanup :cookie:

PR #8275, opened 2021-07-24
Using galleries default border radius for indicator images

tclaus:
Small indicator images in image gallery are round ones, this doesn't correspond with preview shown in stream.
It also doesn't look good on real-world images. (may look OK for User profile images).
This PR restores the default image galleries' border radius for indicator images.

Before: 
Bildschirmfoto 2021-07-24 um 10 31 39

After: 
Bildschirmfoto 2021-07-24 um 11 08 49

### DIFF
--- a/app/assets/stylesheets/gallery.scss
+++ b/app/assets/stylesheets/gallery.scss
@@ -55,7 +55,6 @@ $margin: 15px;
 
     li {
       border: 0;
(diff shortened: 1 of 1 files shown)


SuperTux88:
Merged as 07a49de7e1aa3c37f17128bc2cb4aa180a3708ca

Thanks for the fix @tclaus :cookie:

PR #8276, opened 2021-07-26
Update repo data before trying to install packages

SuperTux88:


SuperTux88:
Self-merged as 7d0c29e99aab176f07a2006e6d7f44b3641dcfbe because this only affects CI which is green again now.

Issue #7960, opened 2019-01-15
Add support for Markdown footnote syntax


diewald:
I added Markdown footnotes to my initial post - and they did not work. That is, I wrote "blah blah[^1]" followed at the bottom of the post with "[^1]: more blahs." This should have created a superscript link "1" after blah, linking to the [^1] at the bottom of the post. The bottom should have looked like "1: more blahs" 

Instead, the original "[^1]" text strings are at both location, looking odd.

Reference used: https://en.support.wordpress.com/markdown-quick-reference/

Flaburgan:
Hi @diewald and thank you for taking time to open an issue!

If someone wants to work on that: diaspora* uses an external library to parse markdown: https://github.com/markdown-it/markdown-it. Some specific syntaxes like the foot notes are dealt with by external plugins. To support it, we would need to add https://github.com/markdown-it/markdown-it-footnote to diaspora*. See the plugins already added in [the Gemfile](https://github.com/diaspora/diaspora/blob/develop/Gemfile#L110)

tclaus:
@denschub 
To fix this 'https://github.com/markdown-it/markdown-it-footnote' should be added to https://gems.diasporafoundation.org before it can be added to gem file and 
Maybe fussilicode fails on this?

Buy the way: markdown-it is currently at version 12.x and some plugins also updated.

denschub:
> To fix this 'https://github.com/markdown-it/markdown-it-footnote' should be added to https://gems.diasporafoundation.org before it can be added to gemfile.

Done, `rails-assets-markdown-it-footnote (3.0.3)` now exists. If you need a different version, do let me know.

> Buy the way: markdown-it is currently at version 12.x and some plugins also updated.

If you're serious about updating other asset gems - and I shall warn you: that's going to be an annoying and painful process - then please reach out to me in private (check [here](https://overengineer.dev/me/card.html) for all available channels) and I can walk you through building asset gems locally to test things.

PR #8277, opened 2021-07-31
#7960 Add support for Markdown footnote syntax

tclaus:
It solves #7960 Add support for Markdown footnote syntax

A quick fix, but surely rarely used. 

It adds the possibility to add footnotes to posts. Syntax is defined here: https://pandoc.org/MANUAL.html#footnotes

Source example: 
Bildschirmfoto 2021-07-31 um 08 01 17

The outcome: 
Bildschirmfoto 2021-07-31 um 08 24 48

It would be nice if this could then be added to the diaspora markdown - doc: 
https://diasporafoundation.org/formatting

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -105,10 +105,11 @@ source "https://gems.diasporafoundation.org" do
 
   gem "rails-assets-highlightjs",                         "9.12.0"
   gem "rails-assets-markdown-it",                         "8.4.2"
-  gem "rails-assets-markdown-it-hashtag",                 "0.4.0"
   gem "rails-assets-markdown-it-diaspora-mention",        "1.2.0"
-  gem "rails-assets-markdown-it-sanitizer",               "0.4.3"
+  gem "rails-assets-markdown-it-footnote",                "3.0.3"
+  gem "rails-assets-markdown-it-hashtag",                 "0.4.0"
   gem "rails-assets-markdown-it--markdown-it-for-inline", "0.1.1"
+  gem "rails-assets-markdown-it-sanitizer",               "0.4.3"
   gem "rails-assets-markdown-it-sub",                     "1.0.0"
   gem "rails-assets-markdown-it-sup",                     "1.0.0"
 
--- a/lib/diaspora/message_renderer.rb
+++ b/lib/diaspora/message_renderer.rb
@@ -50,7 +50,10 @@ module Diaspora
       end
 
       def strip_markdown
-        renderer = Redcarpet::Markdown.new Redcarpet::Render::StripDown, options[:stripdown_options]
+        stripdown_options = options[:markdown_options]
+        # Footnotes are not supported in text-only outputs (mail, crossposts etc)
+        stripdown_options[:footnotes] = false
+        renderer = Redcarpet::Markdown.new Redcarpet::Render::StripDown, stripdown_options
         @message = renderer.render(message).strip
       end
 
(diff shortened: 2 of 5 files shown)


Gemfile
```diff
@@ -105,10 +105,11 @@ source "https://gems.diasporafoundation.org" do
 
   gem "rails-assets-highlightjs",                         "9.12.0"
   gem "rails-assets-markdown-it",                         "8.4.2"
-  gem "rails-assets-markdown-it-hashtag",                 "0.4.0"
   gem "rails-assets-markdown-it-diaspora-mention",        "1.2.0"
-  gem "rails-assets-markdown-it-sanitizer",               "0.4.3"
+  gem "rails-assets-markdown-it-footnote",                "3.0.3"
```
diaspora-code-review, line 109: Bundler/OrderedGems: Gems should be sorted in an alphabetical order within their section of the Gemfile. Gem `rails-assets-markdown-it-diaspora-mention` should appear before `rails-assets-markdown-it-hashtag`.

tclaus: It appears before ..-hashtag! Trouble with Gemfiles?

tclaus:
Nice.. https://societas.online/posts/81f78e60d40301392d8042010a9c0002

lib/diaspora/message_renderer.rb
```diff
@@ -50,7 +50,7 @@ def escape
       end
 
       def strip_markdown
-        renderer = Redcarpet::Markdown.new Redcarpet::Render::StripDown, options[:markdown_options]
+        renderer = Redcarpet::Markdown.new Redcarpet::Render::StripDown, options[:stripdown_options]
```
SuperTux88, line 53: Why does this need special `stripdown_options` without `footnotes`? So text-only doesn't support footnotes?

I think the reason for this should be commented here. And I'd also prefer still using `options[:markdown_options]` but just explicitly dropping `footnotes` from it (or set it to `false`?), so there is still only one `markdown_options` in `DEFAULTS` and if `markdown_options` needs to be changed it only needs to be changed once. Otherwise maybe in the future only `markdown_options` gets updated and `stripdown_options` is forgotten.

Gemfile
SuperTux88: The review-bot uses it's own Gemfile, so this can be ignored and probably also fixes that problem for the future

lib/diaspora/message_renderer.rb
```diff
@@ -50,7 +50,10 @@ def escape
       end
 
       def strip_markdown
-        renderer = Redcarpet::Markdown.new Redcarpet::Render::StripDown, options[:markdown_options]
+        stripdown_options = options[:markdown_options]
+        # Footnotes are not supported in text-only outputs (mail, crossposts etc)
+        stripdown_options[:footnotes] = false
```
SuperTux88, line 55: This actually changes the value in `options` since it's only a reference to the original options. I don't know if it's a problem right now (if the same message renderer is uses twice), but I think it's at least an unexpected side-effect. I'd just do `stripdown_options = options[:markdown_options].except(:footnotes)` which gives you a new hash without that key, but without changing the original.

SuperTux88 (changes requested):

lib/diaspora/message_renderer.rb
tclaus: Thanks for the code-suggestion and explanation ;-)

SuperTux88 (approved):

SuperTux88:
Merged as 122e12a0689856f0da227603d0bfe04bb5a814a3 c71caa5cb8479686a1ca3cac7521f2047de3cb0b 8f9ac33649e4d711ad2d39cd8994fe9c824a9902

Thanks for the PR :cookie:

PR #8280, opened 2021-08-15
Cleanup person_by_handle route

SuperTux88:
I think this is a leftover from a long time ago and I can't find anything that still uses that


tclaus (approved):
I can not see any call to this. So it looks good to me.

Flaburgan (approved):
Confirmed

SuperTux88:
Merged as 28f03f0a1156fe534cd77a4e99eb2c30fb147b31

PR #8284, opened 2021-08-28
Support to search for multiple tags

tclaus:
It allows to search for more than one tag in the search-field.

No frills, just search for "#cat #photos" or "#cat #painting" to see the feature. 

Can be tested on https://societas.online

### DIFF
--- a/app/helpers/stream_helper.rb
+++ b/app/helpers/stream_helper.rb
@@ -5,7 +5,7 @@
 #   the COPYRIGHT file.
 
 module StreamHelper
-  def next_page_path(opts ={})
+  def next_page_path(_opts={})
     if controller.instance_of?(TagsController)
       tag_path(name: @stream.tag_names, max_time: time_for_scroll(@stream))
     elsif controller.instance_of?(PeopleController)
@@ -22,6 +22,7 @@ module StreamHelper
   end
 
   private
+
   # rubocop:disable Rails/HelperInstanceVariable
   def next_stream_path
     if current_page?(:stream)
@@ -50,7 +51,7 @@ module StreamHelper
 
   def time_for_scroll(stream)
     if stream.stream_posts.empty?
-      (Time.now + 1).to_i
+      (Time.zone.now + 1).to_i
     else
       stream.stream_posts.last.send(stream.order.to_sym).to_i
     end
--- a/app/helpers/stream_helper.rb
+++ b/app/helpers/stream_helper.rb
@@ -7,13 +7,13 @@
 module StreamHelper
   def next_page_path(opts ={})
     if controller.instance_of?(TagsController)
-      tag_path(:name => @stream.tag_name, :max_time => time_for_scroll(@stream))
+      tag_path(name: @stream.tag_names, max_time: time_for_scroll(@stream))
     elsif controller.instance_of?(PeopleController)
-      local_or_remote_person_path(@person, :max_time => time_for_scroll(@stream))
+      local_or_remote_person_path(@person, max_time: time_for_scroll(@stream))
     elsif controller.instance_of?(StreamsController)
       next_stream_path
     else
-      raise 'in order to use pagination for this new controller, update next_page_path in stream helper'
+      raise "in order to use pagination for this new controller, update next_page_path in stream helper"
     end
   end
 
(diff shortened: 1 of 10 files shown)


app/helpers/stream_helper.rb
```diff
@@ -5,15 +5,15 @@
 #   the COPYRIGHT file.
 
 module StreamHelper
-  def next_page_path(opts ={})
+  def next_page_path(_opts={})
     if controller.instance_of?(TagsController)
-      tag_path(:name => @stream.tag_name, :max_time => time_for_scroll(@stream))
+      tag_path(name: @stream.tag_names, max_time: time_for_scroll(@stream))
```
diaspora-code-review, line 10: Rails/HelperInstanceVariable: Do not use instance variables in helpers.

app/helpers/stream_helper.rb
```diff
@@ -5,15 +5,15 @@
...
     elsif controller.instance_of?(PeopleController)
-      local_or_remote_person_path(@person, :max_time => time_for_scroll(@stream))
+      local_or_remote_person_path(@person, max_time: time_for_scroll(@stream))
```
diaspora-code-review, line 12: Rails/HelperInstanceVariable: Do not use instance variables in helpers.

tclaus: Its an old module - should this be changed?

app/views/tags/show.mobile.haml
```diff
@@ -5,13 +5,14 @@
...
+  -# rubocop:disable Style/IdenticalConditionalBranches
+  - if tag_followed?
+    .btn.btn-danger.tag_following_action{data: {name: @stream.tag_names}}
+      = t(".stop_following", tag: @stream.tag_names)
   - else
-    .btn.btn-danger.tag_following_action{data: {name: @stream.tag_name}}
-      = t(".stop_following", tag: @stream.tag_name)
-
+    .btn.btn-success.tag_following_action{data: {name: @stream.tag_names}}
+      = t(".follow", tag: @stream.tag_names)
```
cmrd-senya, line 14: This code is for mobile, but the problem I describe exist for both mobile and desktop implementations.

So when I search for "#photo #cats", what I see there on the button to follow and stop following the tags. If I interact with the button it subscribes me to the combined tag of `#photocats` which is probably not what the user would expect to happen

image
image

lib/stream/followed_tag.rb
```diff
@@ -16,7 +16,12 @@ def title
 
   # @return [ActiveRecord::Association<Post>] AR association of posts
   def posts
-    @posts ||= StatusMessage.user_tag_stream(user, tag_ids)
+    if tag_ids.empty?
+      @posts = StatusMessage.none
+    else
+      @posts ||= StatusMessage.any_user_tag_stream(user, tag_ids)
+    end
```
cmrd-senya, line 23: I'd rather move this condition under `any_user_tag_stream`. We probably want such behavior in any usage of `any_user_tag_stream`

lib/stream/tag.rb
```diff
@@ -5,46 +5,43 @@
...
   end
 
   def display_tag_name
-    @display_tag_name ||= "##{tag_name}"
+    @display_tag_name ||= tag_names.split(" ").map {|t| "##{t}" }.join(" ")
   end
 
   def tagged_people
-    @people ||= ::Person.profile_tagged_with(tag_name).paginate(:page => people_page, :per_page => people_per_page)
+    @tagged_people ||= ::Person.profile_tagged_with(tag_names).paginate(page: people_page, per_page: people_per_page)
```
cmrd-senya, line 26: This doesn't really work with multiple tag names, is it expected?

The search results shows nothing, even if there is some profiles with multiple tags.

image

spec/models/status_message_spec.rb
```diff
@@ -41,22 +41,22 @@
 
       describe ".tag_steam" do
```
cmrd-senya, line 42: The tests changes you're making here are :ok:

But I don't see any tests for the new functionality you're adding here.

I see your other pull requests add tests so I assume you just didn't want to invest time in the tests because you weren't sure it will get merged soon?

cmrd-senya (commented):
The code looks good to me and the search for multiple tags works fine.

Some parts of the overall feature look a bit incomplete - profile search and follow tags button in the multi tag case.

I don't know what's our policy for the case like that - are we okay shipping it like that and then improving in some follow up PRs?

PR #8286, opened 2021-09-01
Remove deprecated gem devise_lastseenable & Add devise_last_seen

brunoocasali:
Closes #4807

https://github.com/diaspora/diaspora/issues/4807

As @tclaus mentioned, the default interval time is still 5 minutes as the old implementation, but now is configurable by devise's setup initializer.

@jhass feel to free to checkout the https://github.com/brunoocasali/devise_last_seen gem ;)


tclaus:
I just pulled this PR and it works. 
You should have mention, that the default update rate is set to 5min, which is great.
So - it just works for me.  The last_seen field is updated every 5 minutes on user actions.

tclaus (approved):

brunoocasali:
Hey guys, should I review the conflicts or there is no intention to merge this at all? 😬

tclaus:
Gem deps are merged usually later and then in batches, so don't hold your breath.

PR #8287, opened 2021-09-04
ensure ./log folder exists

tclaus:
The log folder is not kept in this repo. A new user must manual create this for the first successful run. 
Lets help newcomers.


Flaburgan:
I never had to create that folder, why did you need to?

tclaus:
As a new developer, diaspora crashes right after the very first start, when fresh cloned from repo.  First, eye daemon wants to log into ./log but fails.
Sure, this might be configured to any other location, but out of the box - diaspora crashes.

See log here: https://gist.github.com/tclaus/71ab59899c360b433d141c2c21453537

Flaburgan:
That is surprising, which OS are you using ?

tclaus:
> That is surprising, which OS are you using ?

MacOSX (Intel)

BTW: in ./confiy/eye.rb is the reference: 

```
Eye.application("diaspora") do
  working_dir Rails.root.to_s
  env "RAILS_ENV" => rails_env
  stdout "log/eye_processes_stdout.log" unless rails_env == "development"
  stderr "log/eye_processes_stderr.log"
```


If you are using RubyMine or Docker, this might not apply, but think to the newcomers.

Flaburgan:
>If you are using RubyMine or Docker, this might not apply, but think to the newcomers.

I'm not using those tools, and in both dev and prod installations I did I never had to do that. But I was always running GNU/Linux.

SuperTux88 (approved):
The interesting thing is, I can reproduce this, but I never had that problem in the past (so same as Fla, I never had to create anything manually), but now it's there. Something created that in the past but I don't know what (and I don't know why it doesn't work anymore). So it's still kinda strange, but I guess having the directory in the repo doesn't hurt anyway.

SuperTux88:
Merged as 8392c894d2f714377214731673190958d4c09860

Thanks for the fix :cookie:

Issue #7902, opened 2018-10-20
diaspora should send AccountMigration message back when a federation message received for a moved account


cmrd-senya:
We discussed this and supported on the protocol level. It's time to implement it in diaspora.

The goal is to improve the spread of account migration info. 

Consider the following flow:
1. User A exports their archive
2. User B from another pod starts sharing with User A
3. User A import their archive to a new pod

If User B is not on the pod where some of the export-time contacts are, User B will not be notified about account migration event.

To resolve this issue I propose to resend `AccountMigration` message from the old home pod upon a receive of a message addressed to the old User A account.

This must be simple to implement.

Something like this should work:
```ruby
diff --git a/config/initializers/diaspora_federation.rb b/config/initializers/diaspora_federation.rb
index 1ed4e9c98..4aeb57a3e 100644
--- a/config/initializers/diaspora_federation.rb
+++ b/config/initializers/diaspora_federation.rb
@@ -104,6 +104,14 @@ DiasporaFederation.configure do |config|
 
     on :receive_entity do |entity, sender, recipient_id|
       Person.by_account_identifier(sender).pod.try(:schedule_check_if_needed)
+      User.find(recipient_id).tap do |user|
+        next unless user&.person&.account_migration
+        Diaspora::Federation::Dispatcher.build(
+          user,
+          user.person&.account_migration,
+          subscribers: [Person.by_account_identifier(sender)]
+        ).dispatch
+      end
 
       case entity
       when DiasporaFederation::Entities::AccountDeletion

```

I'll create a PR later if there's no objections.

SuperTux88:
The same should happen if a message for a closed account is received, by sending an `AccountDeletion` back. But no objections. :+1:

SuperTux88:
I'm reopening this because I just realized the current solution for that doesn't work, because the old pod would need to sign the message with both keys (old and new user) but can't do that at the moment (unless the user migrated from the same pod to itself with a different username and the pod literally has both users). I'll think about a solution for that, I'll keep it in the 0.7.16 milestone, so we either solve it before a release or revert the commit again.

PR #8288, opened 2021-09-04
diaspora should send AccountMigration message back when a federation …

tclaus:
…message received for a moved account 
From Issue #7902

I use Senya's Patch for this

### DIFF
--- a/config/initializers/diaspora_federation.rb
+++ b/config/initializers/diaspora_federation.rb
@@ -104,14 +104,16 @@ DiasporaFederation.configure do |config|
 
     on :receive_entity do |entity, sender, recipient_id|
       Person.by_account_identifier(sender).pod.try(:schedule_check_if_needed)
-      User.find_by(id: recipient_id).tap do |user|
-        next unless user.person.account_migration
+      unless recipient_id.nil?
+        User.find_by(id: recipient_id).tap do |user|
+          next unless user.person.account_migration
 
-        Diaspora::Federation::Dispatcher.build(
-          user,
-          user.person.account_migration,
-          subscribers: [Person.by_account_identifier(sender)]
-        ).dispatch
+          Diaspora::Federation::Dispatcher.build(
+            user,
+            user.person.account_migration,
+            subscribers: [Person.by_account_identifier(sender)]
+          ).dispatch
+        end
       end
 
       case entity
--- a/config/initializers/diaspora_federation.rb
+++ b/config/initializers/diaspora_federation.rb
@@ -104,16 +104,14 @@ DiasporaFederation.configure do |config|
 
     on :receive_entity do |entity, sender, recipient_id|
       Person.by_account_identifier(sender).pod.try(:schedule_check_if_needed)
-      if User.exists?(recipient_id)
-        User.find(recipient_id).tap do |user|
-          next unless user&.person&.account_migration
+      User.find_by(id: recipient_id).tap do |user|
+        next unless user.person.account_migration
 
-          Diaspora::Federation::Dispatcher.build(
-            user,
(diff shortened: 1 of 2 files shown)


config/initializers/diaspora_federation.rb
```diff
@@ -104,6 +104,17 @@
 
     on :receive_entity do |entity, sender, recipient_id|
       Person.by_account_identifier(sender).pod.try(:schedule_check_if_needed)
+      if User.exists?(recipient_id)
+        User.find(recipient_id).tap do |user|
```
SuperTux88, line 108: If you use `User.find_by(id: recipient_id)` it returns `nil` if the user doesn't exist. That way you can skip the extra exists query, since all users should exist at this point the exists is kinda redundant anyway.

config/initializers/diaspora_federation.rb
```diff
@@ -104,6 +104,17 @@
...
+          next unless user&.person&.account_migration
+
+          Diaspora::Federation::Dispatcher.build(
+            user,
+            user.person&.account_migration,
```
SuperTux88, line 113: The `user.person` should also exist at this point, so the `&` is also redundant.

SuperTux88 (changes requested):

config/initializers/diaspora_federation.rb
tclaus: @SuperTux88 This runs into many test-errors because user.person line 108 seems to be nil.
Is this because of the tests (the build invalid user entities) or is a person not present in any case?

SuperTux88: You also removed the `&` on the `next unless` line, they are needed there because they may be nil there. But if they are nil, it already exits there, so the additional `&` at the line below is never needed because at that point it's already ensured that the user and the person exists.

But now thinking about that, it might also only be an issue with the test and incomplete test data, and in the real world both the user and the person should (in theory) always be existing ... but it's a bit strange that so many tests fail. But it probably doesn't hurt to keep them in that line, but it isn't needed a few lines below when it was already handled here.

SuperTux88: I think I know what the problem is. `recipient_id` is sometimes `nil` (well is always `nil` for public messages). So checking if that's nil, and only do the whole thing when a message is sent to a user directly (and this user was migrated). But for public federation where `recipient_id` is `nil` we don't even need to do the `User.find_by(id: recipient_id)`.

SuperTux88: So `recipient_id` comes from here (if it isn't `nil`):
https://github.com/diaspora/diaspora/blob/16e754f4c71f1389a4fd6c5ef5e9a778d8285aa7/config/initializers/diaspora_federation.rb#L101

And if `recipient_id` isn't `nil`, then neither `user` nor `user.person` should ever be nil. So the only check that is needed is if `recipient_id` is `nil`, if it isn't, no `&` should be needed anymore.

tclaus:
rspec ./spec/federation_callbacks_spec.rb:386 # diaspora federation callbacks :receive_entity receives a entity for a recipient
rspec ./spec/federation_callbacks_spec.rb:355 # diaspora federation callbacks :receive_entity receives a Retraction

Both test expect a call of receive_entity but send a recipient_id of an unknown user - shouldn't the test also create a valid user then ? ("Id: 42") ?

SuperTux88:
> shouldn't the test also create a valid user then

Yes, just use `recipient_id = FactoryBot.create(:user).id`

SuperTux88 (approved):

SuperTux88:
Merged as 2a99cc93ba04d0676ae667e33ee98fbc13c37345

Thanks :cookie:

PR #8299, opened 2021-09-19
Fix running jasmine tests in docker

SuperTux88:
Google Chrome is now required to be installed to run jasmine tests

### DIFF
--- a/docker/develop/Dockerfile
+++ b/docker/develop/Dockerfile
@@ -20,6 +20,9 @@ RUN DEBIAN_FRONTEND=noninteractive \
         nodejs \
         gosu \
     && \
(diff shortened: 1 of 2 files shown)


Flaburgan:
Hm, wouldn't it work with chromium ?

SuperTux88:
Jasmine searches for the `google-chrome` binary, it doesn't even work on my local machine where I have chromium installed because it says it can't find google chrome. Maybe it works when you do a symlink, didn't test that yet, but I thought just installing the thing it wants is the better solution.

jhass (approved):

SuperTux88:
Merged as b2a56376cd26366004d9c673117e45531167af79

PR #8302, opened 2021-09-29
Fix sharing status of local contacts after an import

SuperTux88:
While checking if there still is a problem, with contacts sharing state after the import, I noticed that #8254 was actually the wrong fix for #8106, so I reverted that again. The problem in #8106 is only with contacts on the same pod as the user was imported to, all other contacts from other pods where working. Fixing that was actually pretty easy, so now also local contacts update the sharing status after the import.

#8254 trusted the content of the archive where it shouldn't be trusted, because the user can modify the archive, so it was intentionally ignored if somebody else was sharing with the imported user during the import, Contacts refresh their sharing status if they really share with the imported user after the import (and now also local contacts do this).

But unrelated to #8106, there were also "empty" contacts created, where both directions didn't share with each others. This created contacts in the database which neither were listed in an aspect, but also weren't listed under "only sharing with me", so the total number of contacts was wrong. I also fixed that here.

And since contacts only sharing with the imported user weren't created in the database during the import anymore, I took the contacts from the archive to also send the migration message to them.

### DIFF
--- a/app/models/account_migration.rb
+++ b/app/models/account_migration.rb
@@ -14,6 +14,8 @@ class AccountMigration < ApplicationRecord
   attr_accessor :old_private_key
   attr_writer :old_person_diaspora_id
 
+  attr_accessor :archive_contacts
+
   def receive(*)
     perform!
   end
@@ -29,10 +31,11 @@ class AccountMigration < ApplicationRecord
   # executes a migration plan according to this AccountMigration object
   def perform!
     raise "already performed" if performed?
+
     validate_sender if locally_initiated?
     tombstone_old_user_and_update_all_references if old_person
     dispatch if locally_initiated?
-    dispatch_contacts if remotely_initiated?
+    dispatch_contacts
     update(completed_at: Time.zone.now)
   end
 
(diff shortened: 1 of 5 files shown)


app/models/account_migration.rb
```diff
@@ -14,6 +14,8 @@ class AccountMigration < ApplicationRecord
   attr_accessor :old_private_key
   attr_writer :old_person_diaspora_id
 
+  attr_accessor :archive_contacts
```
diaspora-code-review, line 17: Style/AccessorGrouping: Group together all `attr_accessor` attributes.

SuperTux88: This is a new check, which wasn't there in the past (at most places in the code they aren't grouped), and I suggest disabling it, because I don't find it useful. Hacing the attributes separated and being able to order/group them freely makes the code better structured in my opinion.

tclaus: I agree with it.

tclaus:
@SuperTux88 Would it help If I merge this into societas pod to test it?

SuperTux88:
> Would it help If I merge this into societas pod to enable someone to test it?

Sure, why not :+1:

tclaus:
Looks good to me - but I can not create real world example with test data.

SuperTux88:
> but I can not create real world example with test data.

I tested the case from #8106 manually (before and after the fix, and it was working after the fix), and since it was already working for contacts not on the same pod, this should continue working.

SuperTux88:
Merged to `develop`, thanks for the review @tclaus :cookie:

PR #8304, opened 2021-10-02
domain change

dmorley:
poduptime to fediverse observer update


app/views/registrations/_registrations_closed.haml
```diff
@@ -5,7 +5,7 @@
   wiki: link_to(t("registrations.closed.another_pod"), "https://diasporafoundation.org/getting_started/sign_up"))
 
 != t("registrations.closed.find_pods",
-  poduptime: link_to("Poduptime", "https://diaspora.podupti.me/"))
+  poduptime: link_to("Fediverse Observer", "https://diaspora.fediverse.observer/"))
```
goobertron, line 8: Thanks for this update. For the same of future-proofing and to avoid confusion, it might be worth changing the name of this string from poduptime to fediverseobserver here and in https://github.com/diaspora/diaspora/blob/develop/config/locales/diaspora/en.yml#L1111 – what do you think?

dmorley: Yea, I was just not sure how all that translation stuff works, will the other languages just get the key change etc?

goobertron: That's a very good question. I can't remember whether the string name should be changed here or in WebTranslateIt. Apologies if I've suggested an incorrect action. Pinging @jhass for confirmation.

jhass: Let's keep the key name for now, changing it indeed would mean that it needs retranslation.

goobertron: OK, sorry about that.  @dmorley, see Jonne's advice above.

jhass (approved):

PR #8035, opened 2019-06-23
Add a more detailed modal when reporting a post or a comment

Flaburgan:
As a moderator,
I want to receive report only for posts and comments which violate the terms of use
To easily focus on the important ones

Many reports received on pods I moderate are a people whining about others "look how bad they talk to me". As a consequence of that flow, important reports are sometimes lost in the middle of this. As an attempt to reduce "not real report", I implemented a real modal (instead of the previous javascript alert) which tell to the users to report only content which violates the terms of use, and remind them that they can block the other user if they want to. Also, the reason is required but not filled, which force them to fill something.

Here is how it currently looks like:

report

PR #8305, opened 2021-10-07
Fix jasmine tests

Flaburgan:
So, because we were struggling with Jasmine in https://github.com/diaspora/diaspora/pull/8035, I decided to have a look at it separately.

When running Jasmine locally, I saw that some tests change the URL in the address bar of the browser, and that ends up blocking the tests suite. So I decided to run every tests one by one, and here is what I found:

- [x] http://localhost:8888/?spec=app.collections.Contacts -> Failing because not cleaned correctly
- [ ] http://localhost:8888/?spec=app.collections.Notifications -> cleaning problem, but I don't understand how to fix. I asked on [stackoverflow](https://stackoverflow.com/questions/69487743/re-initializing-a-backbone-collection-doesnt-unbind-events)
- [ ] http://localhost:8888/?spec=app.pages.GettingStarted -> ends on http://localhost:8888/getting_started
- [x] http://localhost:8888/?spec=app.views.SinglePostInteractions -> textarea not displayed, because user not logged in
- [ ] http://localhost:8888/?spec=app.views.StreamShortcuts -> All red, app.page is undefined
- [ ] http://localhost:8888/?spec=app.views.ConversationsInbox -> ends on http://localhost:8888/conversations
- [ ] http://localhost:8888/?spec=app.views.Help -> visits all /help/ urls and ends on one (depending of the test order)
- [ ] http://localhost:8888/?spec=app.views.NotificationDropdown -> All red, TypeError, either this.collection is undefined or this.view is undefined
- [ ] http://localhost:8888/?spec=app.views.Publisher -> Asks for leaving the page
- [ ] http://localhost:8888/?spec=app.views.StreamPost -> All red, TypeError, either this.collection is undefined or this.view is undefined
- [ ] http://localhost:8888/?spec=app.views.Stream -> All red, TypeError, either this.collection is undefined or this.view is undefined
- [ ] http://localhost:8888/?spec=app.Router -> Some red, change the URL

As you can see, there still are many things to check, so this is a work in progress. The most important fixes would be to avoid the tests to change the URL to avoid being stuck. It's done explicitly in `app.views.Help` but I don't have time to check right now.


ragesoss:
@Flaburgan here are a few changes I found that I think fix some of the individually run tests.

* https://github.com/ragesoss/diaspora/commit/ab0d91d962f93450e618ad6a3d81546948c2e36f
* https://github.com/ragesoss/diaspora/commit/8490e85a2818b69e92343eea4af74d76787a00e9
* https://github.com/ragesoss/diaspora/commit/df81ca5c7cd74374c107d03e3f29c1a256fe5605
* https://github.com/ragesoss/diaspora/commit/14b0443e3432e015a4429a338fcef95986e2a1c3

My strategy was to remove tests in the order they are run, to discover when one test relies on an earlier test to set up the expected environment. The same fixes might be applicable to other files.

ragesoss:
Here's another tidbit that might help: https://github.com/ragesoss/diaspora/commit/faef372cbee8b675d1e380a81e4aa23f01269e25

Flaburgan:
@ragesoss thank you very much for your work in https://github.com/diaspora/diaspora/pull/8333 and https://github.com/diaspora/diaspora/pull/8331. Can you please confirm that everything valuable in this PR is somehow integrated in your two PRs and that I can close this one?

ragesoss:
@Flaburgan I think so, yes.

PR #8310, opened 2021-10-27
Use old person private key if relayable author migrated away

SuperTux88:
We only store signatures for relayables if the author is external, but if the author becomes external through a migration, the signature is missing. Lets just use the old persons private key to still be able to generate a signature for the export.

This fixes the issue found here: https://diasp.org/posts/4cbeca400483013acc58002590c0bfb8

We will have a similar issue on import, where the signature doesn't match the authors diaspora-ID, because the signature was created with an old user, and then the author changed through a migration, so the signature maybe matches a previous account of the author. But that's a different issue and needs to be handled in the importer.

### DIFF
--- a/app/serializers/federation_entity_serializer.rb
+++ b/app/serializers/federation_entity_serializer.rb
@@ -5,11 +5,26 @@
 # are used as for federation messages generation.
 class FederationEntitySerializer < ActiveModel::Serializer
   include SerializerPostProcessing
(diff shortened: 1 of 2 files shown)


skyschub (approved):

SuperTux88:
Merged as eb977dc25a45cf44db8e8ff51a84507ec2f76c54

PR #8312, opened 2021-10-28
Bump faraday to 0.17.4, diaspora_federation to 0.2.7

Flaburgan:
Now that @SuperTux88 released the version 0.2.7 of the diaspora_federation gem, I'm able to bump faraday to 0.17.4 and with that support Ruby 2.7.

The changelog is available [here](https://github.com/lostisland/faraday/blob/main/CHANGELOG.md).
Note that if we want to see warnings about the incompatibility with Faraday 1.0, we will need to run with the parameter `FARADAY_DEPRECATE=warn`. I didn't do that as the moment, as I saw that the federation gem was requiring < 1.0, but as 2.0 is already almost here (currently alpha), I will probably add this while developing to see if there are any.

diaspora_federation 0.2.7 is compatible with current diaspora\* codebase as far as I know so I did nothing special regarding this update. Is it fine @SuperTux88 ?


SuperTux88 (approved):

SuperTux88:
Merged as 8a3c9a24639d59dc8140199f0f0f5c5f9e129cff

Thanks for the PR @Flaburgan :cookie:

PR #8313, opened 2021-10-28
Add a max-width to user name in the header, fixes #7676

### LINKED ISSUE #7676: Navbar too high on small screens in combination with long username
When the username is long and the width of the screen is small the navbar is higher than normal. This wouldn't be much of a problem if it wasn't blocking the view of the main page. When composing a message the buttons on top (preview and others) are not visible.

screenshot

Flaburgan:
300px is small enough, if the screen width get small the header becomes responsive so the user name is not displayed anymore.

### DIFF
--- a/app/assets/stylesheets/header.scss
+++ b/app/assets/stylesheets/header.scss
@@ -119,6 +119,19 @@
     }
   }
 
+  .dropdown-toggle {
+    align-items: center;
+    display: flex;
+
+    .user-name {
+      margin-right: 3px;
+      max-width: 250px;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
+  }
+
   .user-menu-dropdown {
     padding: 0;
   }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/header.scss
```diff
@@ -119,6 +119,19 @@
     }
   }
 
+  .dropdown-toggle {
+    align-items: center;
+    display: flex;
+
+    .user-name {
+      display: inline-block;
```
SuperTux88, line 127: I think this isn't needed, at least my browser (firefox and chrome) is telling me, that it's using `display: block` anyway and just ignores it.

SuperTux88 (changes requested):
Looks good otherwise

SuperTux88:
Oh, I also think 300px is too much, it works with English, but still breaks with some other languages (German for example). So maybe just reduce it to 250px so we have some more space for languages with longer texts in the header?

Flaburgan:
Done. I also added a small margin between the name and the caret.

SuperTux88 (approved):

SuperTux88:
Merged as 37f081959c8ea1aad7bd23481a34d8ca5815bb13

Thanks for the fix :cookie:

PR #8316, opened 2021-10-30
Refactor hovercards, fix #8315

Flaburgan:
Now process the template with handlebars syntax instead of by JS. That fixes #8315 and should allow me to handle https://github.com/diaspora/diaspora/issues/6542 and https://github.com/diaspora/diaspora/issues/7469 after.

PR #8317, opened 2021-10-31
Add sharing status to hovercards

### LINKED ISSUE #6542: show if contact is following back in hovercards
mockup:
hovercard
##

Flaburgan:
This is based on https://github.com/diaspora/diaspora/pull/8316 please merge it first.

Fixes https://github.com/diaspora/diaspora/issues/6542

In the same time, I also fixed the distorted "non-sharing" circle which was looking like this before:
Capture d’écran de 2021-10-31 15-33-06

### DIFF
--- a/app/assets/javascripts/app/views/hovercard_view.js
+++ b/app/assets/javascripts/app/views/hovercard_view.js
@@ -107,11 +107,12 @@ app.views.Hovercard = app.views.Base.extend({
       if( !person || person.length === 0 ) {
         throw new Error("received data is not a person object");
       }
-
-      if (app.currentUser.authenticated()) {
-        self.aspectMembershipDropdown = new app.views.AspectMembership({person: new app.models.Person(person)});
-      }
+      var personModel = new app.models.Person(person);
+      person.is_sharing = personModel.isSharing();
       self.person = person;
+      if (app.currentUser.authenticated()) {
+        self.aspectMembershipDropdown = new app.views.AspectMembership({person: personModel});
+      }
       self.render();
 
       if( !self.showMe ) {
--- a/app/assets/javascripts/app/views/hovercard_view.js
+++ b/app/assets/javascripts/app/views/hovercard_view.js
@@ -19,8 +19,13 @@ app.views.Hovercard = app.views.Base.extend({
 
     this.showMe = false;
     this.parent = null;  // current 'hovercardable' element that caused HC to appear
-
     this.active = true;
+  },
+
+  presenter: function() {
+    return _.extend({}, this.defaultPresenter(), {
+      person: this.person
+    });
   },
 
   postRenderTemplate: function() {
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/app/views/hovercard_view.js
```diff
@@ -102,14 +107,14 @@ app.views.Hovercard = app.views.Base.extend({
       if( !person || person.length === 0 ) {
         throw new Error("received data is not a person object");
       }
-
+      var personModel = new app.models.Person(person);
+      person.is_sharing = personModel.isSharing();
```
diaspora-code-review, line 111: Identifier 'is_sharing' is not in camel case.

app/assets/stylesheets/people.scss
```diff
@@ -45,3 +45,11 @@
     .btn-danger { margin-top: 9px; }
   }
 }
+#sharing_message {
+  &.entypo-check {
+    color: darken($brand-success, 20%);
+  }
```
diaspora-code-review, line 51: Rule declaration should be followed by an empty line

app/assets/stylesheets/people.scss
```diff
@@ -45,3 +45,11 @@
     .btn-danger { margin-top: 9px; }
   }
 }
+#sharing_message {
```
diaspora-code-review, line 48: - Avoid using id selectors
- Selector `sharing_message` should be written in lowercase with hyphens

app/assets/javascripts/app/views/hovercard_view.js
Flaburgan: The handlebar helper is using this variable name, I don't think it's worth to refactor this.

Flaburgan:
I added some tests, and everything is green \o/
This is ready for review, but please merge https://github.com/diaspora/diaspora/pull/8316 first.

app/assets/stylesheets/people.scss
```diff
@@ -45,3 +47,13 @@
     .btn-danger { margin-top: 9px; }
   }
 }
+
+#sharing_message {
+  &.entypo-check {
+    color: darken($brand-success, 20%);
```
SuperTux88, line 53: You missed the override for this for the dark theme.

https://github.com/diaspora/diaspora/blob/9fa032e4c91710df072291a26b5344429fd69d65/app/assets/stylesheets/color_themes/_color_theme_override_dark.scss#L58-L60

You moved this so it also selects the sharing state on the hovercard ... but you didn't change the override on the dark theme, which still only selects the profile, so the dark hovercard looks like this now:

image

SuperTux88 (changes requested):

app/assets/stylesheets/people.scss
msph31: good work supertux88
(btw it saddens me to see "mara" as test subject here):
> image

Flaburgan: Nice catch! We should have a checklist on the wiki about what to test while changing the UI: user not logged in, dark theme, mobile... Those are common things which tend to be forgotten.

Flaburgan: Fixed. This looks weird to me though, we should only need to set different variables to make the themes work. But this is another topic.

SuperTux88 (approved):

SuperTux88:
Merged as c67fc4e0f77cd687c314184c33c08b5e3f2594ed

Thx for this feature :cookie:

PR #8320, opened 2021-11-07
Show local posts in sidebar, if enabled

### LINKED ISSUE #8319: "Show local posts for user roles" do not show the link in the UI
With #8232 a useful function was added. It will local users with a given role (Admin. moderators or all) to see just all local posts. 

But in the UI the call was used not correctly so that nobody can see the extra link in the sidebar. 
Current sidebar
current_sidebar_if_enabled

expected sidebar
expected_sidebar_if_enabled

This setting can be set in the Diaspora.toml with the 'enable_local_posts_stream' setting.

tclaus:
The Link "Local Posts" was not shown in the UI when enabled in UI.

fixes #8319

### DIFF
--- a/lib/configuration_methods.rb
+++ b/lib/configuration_methods.rb
@@ -52,9 +52,9 @@ module Configuration
 
     def local_posts_stream?(user)
       return true if settings.enable_local_posts_stream == "admins" &&
-                     Role.is_admin?(user)
+                      user.admin?
       return true if settings.enable_local_posts_stream == "moderators" &&
-                     (Role.moderator?(user) || Role.is_admin?(user))
+                      user.moderator?
 
       settings.enable_local_posts_stream == "everyone"
     end
--- a/spec/lib/configuration_methods_spec.rb
+++ b/spec/lib/configuration_methods_spec.rb
@@ -129,11 +129,11 @@ describe Configuration::Methods do
   end
 
   describe "#has_local_posts_stream" do
-    let!(:moderator) { create(:person) }
-    let!(:admin) { create(:person) }
+    let!(:moderator) { create(:user) }
+    let!(:admin) { create(:user) }
     before do
-      moderator.roles.create(name: "moderator")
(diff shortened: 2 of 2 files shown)


lib/configuration_methods.rb
```diff
@@ -50,11 +50,11 @@ def show_service?(service, user)
...
     end
 
-    def local_posts_stream?(user)
+    def local_posts_stream?(person)
       return true if settings.enable_local_posts_stream == "admins" &&
-                     Role.is_admin?(user)
+                     Role.is_admin?(person)
       return true if settings.enable_local_posts_stream == "moderators" &&
-                     (Role.moderator?(user) || Role.is_admin?(user))
+                     (Role.moderator?(person) || Role.is_admin?(person))
```
SuperTux88, line 57: `Role.is_admin?` isn't needed here, since `Role.moderator?` includes admins.

But also, since this `local_posts_stream?` is only relevant for the user and not for persons, you can keep `user` as parameter, and then use `user.moderator?` here (which also includes admins) and `user.admin?` above.

SuperTux88 (changes requested):

SuperTux88:
Oh, and by the way, if you find an issue and directly fix it, you don't need to open an issue first, you can just directly do a PR and describe the problem there.

lib/configuration_methods.rb
tclaus: Ok. By the way.. is there a reason the "role" is linked to the person rather than the to local user?

SuperTux88: Roles are also used for community spotlights, which can be people from other pods which don't have a user object.

tclaus: Hm, isn't this also not available in the UI? 
But OK for now, thanks.

SuperTux88 (approved):

SuperTux88:
Merged to `develop`, thanks for the fix :cookie:

PR #8324, opened 2021-11-23
Extend short links to full pod links

### LINKED ISSUE #8223: cross-pod link handling improvement
hello, i suggest a feature, i hope is very easy to implement:

diaspora has two kind of links: 

* pod.tld/posts/<shortnumber> (i think only digits)

* pod.tld/posts/<long generated hash> - this one contains digits and letters, i guess, is generated like cookie, we can be pretty sure it's unique.

that is why cross-pod links are based on this second kind of URL that represents the post.

so the cross-pod links are written in a way `[desc](/posts/<long_generated_hash>)

then the diaspora handles them as cross-pod, shows you the link with your pod domain.

the improvement i suggest is very small: when encountering the diaspora link of second kind, with unique post identifier, independently of the pod, show it as cross-pod link. i. e.  replace the domain with the current pod domain.

thank you!

tclaus:
Fixes #8223 

If a short link was given in the text
and it starts with /posts/ff1122...
It will be changed to the full pod address with
hostname and port.

This enables copying the link or bookmark it.

### DIFF
--- a/app/assets/javascripts/app/helpers/text_formatter.js
+++ b/app/assets/javascripts/app/helpers/text_formatter.js
@@ -30,6 +30,8 @@
     md.use(inlinePlugin, "link_new_window_and_missing_http", "link_open", function (tokens, idx) {
       tokens[idx].attrs.forEach(function(attribute, index, array) {
         if( attribute[0] === "href" ) {
+          let host = window.location.protocol + "//" + window.location.host;
+          array[index][1] = attribute[1].replace(/^\/posts\//, host + "/posts/");
           array[index][1] = attribute[1].replace(/^www\./, "http://www.");
         }
       });


tclaus:
A short quick win.

Flaburgan:
Does that change anything? Links like `/posts/ID` are automatically converted to `https://www.pod.tld/posts/ID` by the browser when you copy then, aren't they?
To automatically convert ID to UUID would help more I think.

tclaus:
>  Links like `/posts/ID` are automatically converted to `https://www.pod.tld/posts/ID` by the browser when you copy then

Cool - I did not know this copy - feature. Is this standard in all browser? 

This PR changes the <a> tag in the browser to a full qualified name. 

But I see my misunderstanding - the request was to change pod.tld./posts/ff11dd... links to the domain of the current running pod. That means, not to jump out of the current pod to an external link.

SuperTux88:
Yes, this doesn't change anything, because the browsers already do that automatically.

What we actually need is when somebody need is when somebody writes a post and copy-pastes an URL with the `https://<pod-domain>/posts/<id|guid>` format into it, it gets replaced with `diaspora://<author>/post/<guid>` URL (documentation: https://diaspora.github.io/diaspora_federation/federation/diaspora_scheme.html), and federated that way, so the receiving pod can fetch the linked post and show it as local URL (that part is already working)

tclaus:
The exact opposite was requested than done in this PR. The issue was to  /remove/ the domain if a full url with a hash is given.

PR #8326, opened 2021-11-30
adjusted method name from portuguese to english

carlosdanielpohlod:


SuperTux88:
Thanks for your PR, but "registrar" is actually already an english word and the correct term here, so I'm closing this.

PR #8327, opened 2021-12-03
Remove host from URL if post-guid exist locally

tclaus:
Fixes 8223
It removes the host from any URL that points to a post which exists on this Pod.

### DIFF
--- a/lib/stream/base.rb
+++ b/lib/stream/base.rb
@@ -2,6 +2,7 @@
 
 class Stream::Base
   TYPES_OF_POST_IN_STREAM = ['StatusMessage', 'Reshare']
+  HTTP_MATCHER = %r{(http|https)://([\w_+]+(?:\.[\w_-]+)*:\d{3,4})?(/posts/(\w+))}.freeze
 
   attr_accessor :max_time, :order, :user, :publisher
 
@@ -35,6 +36,7 @@ class Stream::Base
   def stream_posts
     self.posts.for_a_stream(max_time, order, self.user).tap do |posts|
       like_posts_for_stream!(posts) #some sql person could probably do this with joins.
+      normalize_internal_post_links(posts)
     end
   end
 
@@ -85,6 +87,15 @@ class Stream::Base
 
     posts.each do |post|
       post.user_like = like_hash[post.id]
+    end
+  end
+
+  def normalize_internal_post_links(posts)
+    posts.each do |post|
+      post.text.match(HTTP_MATCHER) do |m|
+        link_post_id = m[4]
+        post.text.gsub!(HTTP_MATCHER, "/posts/#{link_post_id}") if Post.exists?(guid: link_post_id)
+      end
     end
   end
 


SuperTux88:
Again, it would be a lot better to do this on the sending side, and when somebody pastes a link to a post (with ID or GUID) in the editor, replace it with a `diaspora://<author>/post/<guid>` URL. Yes, this here is a small improvement, but it's only a workaround, and it only works in the stream (not on single post views), only works if the post is known, only works for URLs with GUIDs and needs to check if the posts exist every time the stream loads.

The `diaspora://` URL are the long term solution for this problem and support is already in diaspora (and even friendica) on the receiving side for a long time (since 0.7.1.0, so it's save to start using it on the sending side without things to break, see https://github.com/diaspora/diaspora_federation/issues/75). `diaspora://` URLs are fetched when the post isn't known yet (so also works for unknown Posts), and it only needs to be processed once when the post is received, and it works on the stream, on the single post view and even in comments and private messages. All that's missing is replacing the URLs in the editor when an URL to a post is pasted and with a little request to the server you could also replace local IDs from the sending pods with the GUID, so it also works when a user just pastes a URL with an ID. So that would make that feature usable for a lot more people without them needing to know that they need to use URLs with GUID or doing anything special when sending a post/comment (People who already know about the current GUID workaround usually also know that they should leave out the domain, so they don't need this patch anyway).

tclaus:
Sorry - but the diaspora - handler is a fail. 
In my Safari (Intel Mac) it won't register in any way. On my Linux machine (Firefox), there is something 'registered'. Do I enter a -for-me valid handler, I get a "address not recognized". Do this handler only works inside posts?

For me - this won't work. 
Its just another technical barrier to "register" this link. It did not work for me. Should I use and create this link manually? Whats on computers outside of this technical understanding? Public ones, users who just dropped by for a moment? Windows User? (Can they register/install anything?) 

By the way: Is it a Diaspora :// - link or a web+diaspora:// - link? This is not clear in any documentation, and not clear for any user. (Both do not work if I register them and created a post with such a link as content) 

There might be a reason to use this kind of links in inter process communication if a client is installed on a device, but not as a tool for a user. 


In the meantime, for me it's a good step in the right direction to remove the hostname from a url if the post is known locally. It should be OK to then save the post again without the hostname.


If this protocol is a internal protocol a user should never enter manually - the registration is a barrier. Then this PR still solves a couple of problems for existing posts. It works as long the post is on the local pod. It will not hurt, if not.

SuperTux88:
You are mixing two different (but also kinda similar) things here:

* `diaspora://` URLs are on a federation level. The user shouldn't need to do anything for them, not register a handler, and they are completely handled automatically by the federation layer (the only part that is missing is something that converts normal Links to `diaspora://` URLs when sending a new post/comment/message. When that is done everything is already working on the receiving side without any user needing to do anything manually and they will just show as normal local links in the stream and comments and everywhere. You can already test this manually, if you create such link manually and send a post/comment and the post you link to is public (but not known yet on the receiving pod) it will even be fetched and then shown as normal local link on the receiving side. But the end goal is that the user doesn't need to know and/or do anything, it's all done for the user.

* `web+diaspora://` links are something a user needs to handle manually, they are an extra feature which idea came up after the `diaspora://` links were created. For them to work the user needs to enable the handler in the settings (on Firefox this works, but I can't test safari since I don't own a mac), and then you can use `web+diaspora://` links from outside diaspora, so you can for example link to a diaspora post from your blog, and people (who enabled the handler) can click the `web+diaspora://` link which then is handled by the handler and brings the user to the post on their pod. But this is not on a federation level and still needs users to do something, but that's not what I was talking about in my last comment.

> Then this PR still solves a couple of problems for existing posts. It works as long the post is on the local pod.

Yeah, I said this PR is a small improvement of the current situation, but it's still very limited and still just a workaround. The `diaspora://` URLs are the long-term solution for the problem, but the automatic conversion to `diaspora://` URLs is still missing on the sending side because back then we needed to wait until enough pods have 0.7.1.0 or newer, because older pods just would show the `diaspora://` URLs to the user directly which the users then can't do anything with it. But now years later we can do the last missing part because all pods should be new enough.

tclaus:
Thanks for the explanation. 
So, the user still enters a link to a post, this is then transcoded into a diaspora- protocol, federated across other pods and then translated back to a real http protocol? 

The transcoding part is still missing, right?

SuperTux88:
> The transcoding part is still missing, right?

Yes, the last step missing is when a users posts a link to a post (either with an local ID or GUID), that that link is converted to a `diaspora://` URL and then federated that way. The receiving part is all done so `diaspora://` URLs in posts/comments/messages are shown as `https://` links to the user already.

Flaburgan:
To sum up
As a user logged in diaspora-fr.org
If I'm writing a post with content `https://diaspora-fr.org/posts/7375716` (local id of a post by hq)
Then the content of the post should be changed to `diaspora://hq@pod.diaspora.software/post/969a86a02a9b013ae0af52540086c3e0`

What about pasting a link from another pod? If I paste `https://pod.diaspora.software/posts/969a86a02a9b013ae0af52540086c3e0` should diaspora-fr.org recognize it is a link to a diaspora* post even if not local, and also change it?

SuperTux88:
The first step is to just replace links to posts on the same domain. In theory we can detect if it's a link with GUID on another pod (and we know the post exists), but then also what if the user wanted to specially link to that pod?

tclaus:
Fla has found a good point. 
What if, a user changes its name or migrates to another pod? Links should be rewritten or at least checked against the migration log. 
I am still wondering where the numeric Ids come from - and why. 
There is a strong feeling in me, that we shouldn't mess up with links inside the posts.

SuperTux88:
> What if, a user changes its name or migrates to another pod? Links should be rewritten or at least checked against the migration log.

The author in the `diaspora://` links is currently only relevant when receiving, so it would only be problematic when you send a post with a link to a post of a user which is at exactly this time migrating to a new pod ... but even then it wouldn't break because the old pod would just redirect them to the new pod.

> I am still wondering where the numeric Ids come from - and why.

The numeric IDs is what diaspora uses by default for local links.

>  There is a strong feeling in me, that we shouldn't mess up with links inside the posts.

What exactly do you mean with that?

tclaus:
The source here won't solve the initial problem, so I close this for now.

PR #8330, opened 2021-12-19
Bump apparition to a commit to avoid warnings while running cucumber

Flaburgan:
Looks like apparition is now a dead project, at least they don't do any release but we want this fix so I'm bumping to a commit. See  https://github.com/twalpole/apparition/issues/81

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -274,7 +274,8 @@ group :test do
 
   # Cucumber (integration tests)
 
-  gem "apparition",       "0.6.0"
+  # Switch apparition back to a release when https://github.com/twalpole/apparition/issues/81 will be fixed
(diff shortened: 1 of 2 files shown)


Flaburgan:
Hm. looks like the CI doesn't look this.

tclaus:
What's about to switch to currently supported web driver cuprite? 
https://github.com/rubycdp/cuprite

SuperTux88:
> Shouldn't we use a currently supported driver cuprite?

cuprite also isn't super active, there hasn't been a new release since over a year and we would have the same problem as with apparition because the current release of cuprite doesn't support dragging because the feature (https://github.com/rubycdp/cuprite/pull/176) isn't released yet. So at the moment we wouldn't have any advantage by switching. But whoever releases a new version first wins.

SuperTux88:
Superseded by #8418

PR #8338, opened 2022-02-15
Add yalphabetize for yaml normalisation

samrjenkins:
This PR introduces [yalphabetize](https://github.com/samrjenkins/yalphabetize) to the project. I would love to hear your thoughts/feedback.

It provides us with the ability to quickly normalise/alphabetise YAML files such as locales and to maintain this normalisation into the future.

Advantages of alphabetising YAML include:
- easier to locate specific keys in large files
- easier to spot duplicate keys
- easier to compare equivalent locale YAMLs for different languages

Yalphabetize can also be integrated into CI to ensure that future PRs maintain the alphabetisation of locale files.

I am slightly unsure how to handle alphabetisation of each language because the contribution guidelines state that the translations are maintained via a 3rd party. Is it ok to normalise the 8 locales I have touched in this PR?

### DIFF
--- a/lib/tasks/yalphabetize.rake
+++ b/lib/tasks/yalphabetize.rake
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'yalphabetize'
+
 namespace :yalphabetize do
   task run: :environment do
     Yalphabetize::Yalphabetizer.call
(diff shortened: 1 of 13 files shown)


lib/tasks/yalphabetize.rake
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+namespace :yalphabetize do
+  task :run do
```
diaspora-code-review, line 4: Rails/RakeEnvironment: Include `:environment` task as a dependency for all Rake tasks.

PR #8343, opened 2022-03-05
Don't suggest to retry exports on failure.

skyschub:
Retrying is completely pointless in 99 of 100 cases.


SuperTux88 (approved):
LGTM :cookie:

SuperTux88:
Merged as 7611391f9f2cf431665be6d7ea36ed13b6c96df8

PR #8325, opened 2021-11-26
Links in open_id views should be disabled

tclaus:
Fixes #8168

During open_id auth process (for enabling external Apps) in the login view and grants_list all diaspora links are enabled and would mess up the login process if a user tries these out.

PR #8345, opened 2022-03-29
Set Name from client in auth dialog

tclaus:
In Auth Dialog the name should be prefilled from client.


jhass (approved):

SuperTux88:
@tclaus so you took this from #8325, but there you also changed what `prefilled_username` does (using `login_hint` instead of `username`), so I guess this also needs to be copied to this PR?

tclaus:
@SuperTux88 Great, that you noticed this. I have added the missing line to This PR.

SuperTux88 (approved):
Looks good now :cookie:

PR #8351, opened 2022-04-27
Do not allow the user to mass assign their own password alongside other parameters

jhass:
Much thanks to Breno Vitório (@brenu) for the report!


skyschub (approved):
❤️

PR #8361, opened 2022-07-03
Remove entypo-rails dependency and add font to vendor/assets

SuperTux88:
`entypo-rails` doesn't support newer rails versions


skyschub (approved):

SuperTux88:
Merged as 5f48cedb84cc61533c223f10e3cb417643ac3c7d

PR #8363, opened 2022-07-06
API: update Search endpoint to be aware of ignored users

cmrd-senya:
While I was using insporation I noticed that when I was looking up posts by the tag name it showed up posts by the people I block. It didn't happen in the same case for the web app.

So in this PR I fixed that.

@tclaus looks like I changed the same files as you did in https://github.com/diaspora/diaspora/pull/8284. It means whatever gets shipped first will create conflicts for the other. I'll try to review your PRs, especially the one I referenced.

### DIFF
--- a/spec/integration/api/search_controller_spec.rb
+++ b/spec/integration/api/search_controller_spec.rb
@@ -460,6 +460,41 @@ describe Api::V1::SearchController do
       )
       expect(response.status).to eq(401)
     end
+
+    context 'when the author is blocked' do
+      before do
+        FactoryBot.create(
+          :block,
+          user: auth_public_only_read_only.user,
+          person: eve.person
+        )
+        FactoryBot.create(
+          :block,
+          user: auth.user,
+          person: eve.person
+        )
+      end
+
+      it "hides the blocked author's posts in with public only scopes" do
+        get(
+          "/api/v1/search/posts",
+          params: {tag: "tag2", access_token: access_token_public_only_read_only}
+        )
+        expect(response.status).to eq(200)
+        posts = response_body_data(response)
+        expect(posts.length).to eq(1)
+      end
+
+      it "hides the blocked author's posts in with default scopes" do
+        get(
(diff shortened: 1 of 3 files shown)


spec/integration/api/search_controller_spec.rb
```diff
@@ -460,6 +460,41 @@ def add_contact(*aspects)
...
+      before do
+        FactoryBot.create(
+          :block,
+          user: auth_public_only_read_only.user,
+          person: eve.person
+        )
+        FactoryBot.create(
+          :block,
+          user: auth.user,
+          person: eve.person
```
Flaburgan, line 474: Shouldn't you use someone else than `eve` here?

tclaus (approved):
Looks good to me.
@jhass  What do you think?

spec/integration/api/search_controller_spec.rb
```diff
@@ -460,6 +460,41 @@ def add_contact(*aspects)
...
+      end
+
+      it "hides the blocked author's posts in with public only scopes" do
+        get(
+          "/api/v1/search/posts",
+          params: {tag: "tag2", access_token: access_token_public_only_read_only}
+        )
+        expect(response.status).to eq(200)
+        posts = response_body_data(response)
+        expect(posts.length).to eq(1)
```
SuperTux88, line 485: I think it would be a good idea to actually check if the one post that gets returned here is the one that we actually expected, so there can't be a bug where there is only one post returned but it's the one from eve?

But on the other hand, all the other specs here also only check the number of posts, but I also think it would be a good idea to check more there too (for example if only public posts should be returned, check if all the posts that got returned are actually public).

SuperTux88 (commented):
This looks good to me, but I still have one little comment about the tests. Since it also affects all the other tests and isn't something you did new in this PR, I'm still interested in what you think about it and if you think it's worth changing or if you want to keep it the way it is.

salil-khanna:
lgtm!

PR #8364, opened 2022-07-12
Use yarn for assets

cmrd-senya:
Replace `rails-assets` usage with `yarn`. List front-end dependencies in `package.json` instead of `Gemfile`.


bin/yarn
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env ruby
```
cmrd-senya, line 1: The presence of this file automatically adds `yarn install` to the asset precompilation: https://github.com/rails/rails/blob/v5.1.2/railties/lib/rails/tasks/yarn.rake#L9-L11

bin/yarn
```diff
@@ -0,0 +1,11 @@
...
+APP_ROOT = File.expand_path('..', __dir__)
+Dir.chdir(APP_ROOT) do
+  begin
+    exec "yarnpkg", *ARGV
+  rescue Errno::ENOENT
+    $stderr.puts "Yarn executable was not detected in the system."
+    $stderr.puts "Download Yarn at https://yarnpkg.com/en/docs/install"
+    exit 1
```
cmrd-senya, line 9: I thought of adding `npm install yarn` here, but it turns out it's impossible to install `yarn` without installing the rest of dependencies via `npm`. This is a default message, I think this behavior is good enough.

package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "diaspora",
+  "version": "0.7.17.0",
+  "repository": "git@github.com:diaspora/diaspora.git",
+  "author": "diaspora* team",
+  "license": "AGPL-3.0",
+  "dependencies": {
+    "jquery.textchange": "gist:424774",
```
cmrd-senya, line 8: This gist seems to be the first original source of this jquery extension

lib/tasks/assets.rake
```diff
@@ -34,3 +34,5 @@ namespace :assets do
     Rake::Task["assets:non_digest_assets"].invoke
   end
 end
+
+Rake::Task["assets:generate_error_pages"].enhance [ "yarn:install" ]
```
cmrd-senya, line 38: CI runs `assets:generate_error_pages` without running assets precompilation, so adding `yarn install` here as a dependency too

app/assets/stylesheets/gallery.scss
```diff
@@ -14,6 +14,10 @@ $margin: 15px;
       top: $margin;
     }
 
+    .slide-loading {
+      background-image: image-url("blueimp-gallery/img/loading.gif");
```
cmrd-senya, line 18: So this sets correct loading path for this image, overriding this: https://github.com/blueimp/Gallery/blob/6075850b76e511cf013ee18c238ab7ddc756a131/css/blueimp-gallery.css#L85

There must be a better way of doing this, but I could get it now

package.json
SuperTux88: I don't like adding a dependency from gist in production. I don't really understand what happened to this as there was something with a version number before (so a release?) and now there is only a gist left?

But also, I don't think we need this dependency anymore at all. The only place that I can find where this is used is in `charcount.js` which is the counter for when posting to services with limited length (like twitter for example). And we can probably just replace it with the native `input` event. And getting rid of dependencies is always something I like, especially dependencies that were only used at one place and can be replaced with very little code.

SuperTux88 (changes requested):
Thanks for the PR, looks mostly good (and good to see you back :smiley:)

Also, little side-note, it's probably also a good idea to update the javascript dependencies, which should be easier now, as we don't need to manually upload them to our own gem server anymore. But that can be done in a separate PR, I just wanted to mention it here, since I expect some of the dependencies to be pretty outdated. While I updated the ruby-dependencies regularly, I mostly ignored the frontend stuff because of the whole rails-assets stuff, but it should be easier now.

package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "diaspora",
+  "version": "0.7.17.0",
+  "repository": "git@github.com:diaspora/diaspora.git",
+  "author": "diaspora* team",
+  "license": "AGPL-3.0",
```
denschub, line 6: Can you add a

```
"private": true
```

as well? That's generally good practice for things that should not end up on npm.

package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "diaspora",
+  "version": "0.7.17.0",
```
SuperTux88, line 3: Oh, I just noticed the version here, is this needed for what we use this for? Is this something we also need to update for each release? Or can we leave this blank?

(as this will be merged to `develop` only, as it adds a new requirement for having yarn installed, it should probably be 0.7.99.0 if it's really needed to have a version here)

denschub: `name` and `version` are the only required fields, and [it needs to be parsable by `node-semver` ](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#version). But since this will never be published to npm, there is on harm in setting this to `0.0.0`.

SuperTux88: "If you don't plan to publish your package, the name and version fields are optional." :thinking: so maybe it's not required in our case? But also setting it to `0.0.0` sounds good to me.

cmrd-senya:
@SuperTux88 @denschub ready for rereview

SuperTux88 (approved):
Looks good to me now :cookie:

skyschub (approved):

PR #8365, opened 2022-07-13
Remove markerb dependency and refactor mails

SuperTux88:
This is to prepare for upgrading to rails 6, because markerb doesn't work with rails 6 anymore, and it looks like the project is so dead, that not even the github repo exists anymore. Since I didn't find a replacement, I removed the dependency. But removing it
means adding separate templates for text and html mails, this sadly means some code duplication, but also luckily the mail templates aren't that complicated, so the duplication is very minimal, so I think it's fine. On the plus side, since we have different templates for text and html, I made it so the html mails actually contain the rendered post, not just the same "just convert the markdown to text". So the html-mails are more useful now, while the text mails are the same as before.

And since I was working on the mails and testing everything anyway, I also fixed some other issues related to mails.


skyschub (approved):

SuperTux88:
Merged as 41633fcea9f6faa2e3848b0151b021db3bb14054, f76a8d7debadb47bc21aec4cac8a696a0265425f, 6cea355636b7fd8a4d1570807f77762eaccb23e0, a9ae84f3a6b269247c68c172548b186bb643bf60, 3c02a1f067b718e2d1806b65eef5bc9d9f446757, dd3bc39c97e6b6d17646c8cbf33e13c252e0c2e4 and a5e5c7f3782e83d527b9b28a0e8f46fb1f83389b

PR #8366, opened 2022-07-17
Upgrade to rails 6.1 and ruby 2.7

SuperTux88:
Rails 5.2 and ruby 2.6 are both end-of-life. So this upgrades to supported versions again. I also upgraded to sprockets 4 and updated a few other gems that were not possible before (blocked by old rails/ruby version).


Flaburgan:
Thank you for this much needed work!
I'm surprised you're targeting a minor diaspora* version for a big change like Rails upgrade, you feel confident enough about it?

SuperTux88:
We already did a ruby upgrade in an earlier minor update (0.7.14.0), and since ruby 2.6 is EOL, and rails 5.2 doesn't support ruby 2.7, we don't have much choice than doing it. But I kept the changes so it doesn't contain breaking changes (the rest of the rails stuff can be done on `develop` only for the next major). So it should be fine for a minor update and it allows podmins to install a ruby version again which still gets updates. So the only breaking change is dropping ruby 2.5 (which is EOL since even longer).

SuperTux88:
I also updated bundler to bundler 2 now. This is maybe also a small breaking change, as podmins also need to update bundler ... but we have `gem install bundler` in our update-guide anyway (we just need to remove the `1.17.3` version number when releasing 0.7.18.0), and if podmins update to ruby 2.7 anyway (which is recommended), it's easier to install the current bundler instead of the old one, so I think that's fine.

.rubocop.yml
```diff
@@ -57,6 +57,11 @@ Layout/SpaceAroundEqualsInParameterDefault:
 # are needed.
 Style/StringLiterals:
   EnforcedStyle: double_quotes
+  Exclude:
+    - "config/application.rb"
+    - "config/boot.rb"
+    - "config/environment.rb"
+    - "config/environments/*.rb"
```
denschub, line 64: I assume you excluded those files to allow for single-quotes in the config files to keep the diff to the Rails-original sources minimal? If so, a comment stating as much probably would be a good idea, because I was about to write a comment asking why you flipped the quotes in those files. :)

SuperTux88: Yes, that's exactly the reason, because when updating rails it also generates new versions of these files, and I wanted to keep them as close to the original as possible to see what actually changed ... but yes, I'll add a comment here :+1:

skyschub (approved):
This is kinda huge, and my review quality probably reflects that. But I didn't spot anything bad, so let's ship it!

skyschub (approved):

SuperTux88:
Merged as many commits to `next-minor` and 4edaebb94f10bb3ab616398ff89c27820dfaef9d, e7e34a8c24ad6527660e308592117e0a6bd03176 and 03d2001cf2ee0bc93d5e413b30387382ec13e1e9 to `develop` only.

PR #8368, opened 2022-07-20
Upgrade diaspora federation (and some preparations for ruby 3)

SuperTux88:
There are still some issues with ruby 3, but I created #8369 to track them so we can fix ruby 3 support in the future.


skyschub (approved):
💯 (with dots inbetween)

PR #8371, opened 2022-07-21
Don't eager load active storage in production

SuperTux88:
We don't use active storage, but eager loading it tries to read the `config/storage.yml`, which doesn't exist, because we don't need it.

replacement for #8370 :smiley:

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -15,6 +15,11 @@ Rails.application.configure do
   config.eager_load = true
   config.eager_load_paths += %W[#{config.root}/lib]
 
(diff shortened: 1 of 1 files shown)


skyschub (approved):
🚒

SuperTux88:
Merged as ee68da7eebc8d01a9510328a0f64abca692f1552

Thanks for the quick review @denschub :cookie: and sorry for the big fire, I didn't expect this to break only in production.

denschub:
> I didn't expect this to break only in production.

Well, I mean, me neither, otherwise I would have updated tomorrow... ;D

PR #8372, opened 2022-07-21
Disable export_concurrent to prevent segfault during precompile

SuperTux88:
See https://github.com/sass/sassc-ruby/issues/207


skyschub (approved):
🧯

SuperTux88:
Merged as 3cf84c838f58316a574b8ab69ba81a2d214c8fb2

Thanks for the quick review @denschub :cookie:

PR #8373, opened 2022-07-21
Add dummy param to receive workers

SuperTux88:
it looks like something still schedules jobs with an additional parameter which breaks incoming federation. Lets add a dummy parameter to till be able to consume the jobs until I know what's going wrong here.

If I understand this correctly the jobs should be scheduled without this parameter (line 93 and 100):

https://github.com/diaspora/diaspora/blob/975afe03bbc9406de5998f91f32aae9ba58a54b4/config/initializers/diaspora_federation.rb#L92-L102

### DIFF
--- a/app/workers/receive_private.rb
+++ b/app/workers/receive_private.rb
@@ -6,7 +6,7 @@
 
 module Workers
   class ReceivePrivate < ReceiveBase
-    def perform(user_id, data)
+    def perform(user_id, data, _dummy=nil)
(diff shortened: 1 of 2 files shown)


skyschub (approved):

skyschub (changes requested):
actually, no. (we're talking on irc, just blocking this PR)

SuperTux88:
OK, this time @denschub outsmarted me ... it's a different issue

PR #8374, opened 2022-07-21
Disable the default CSRF protection.

skyschub:
This was added [in Rails 5.2 defaults](https://github.com/rails/rails/commit/ec4a836919c021c0a5cf9ebeebb4db5e02104a55), but we upgraded from 5.1 defaults to 6.1, so we didn't notice until now.


SuperTux88 (approved):
:cookie:

PR #7721, opened 2018-02-21
Prevent duplicate mention notifications when the post is received twice

SuperTux88:

PR #8375, opened 2022-07-21
Add in-app notification configuration

### LINKED ISSUE #7686: User setting: enable/disable notifications in-app as well as by email

cmrd-senya:
fix #7686 

This PR implements in-app notification configuration feature - ability to turn off in-app notifications per event similar to email. See the comments for screenshots.

### Original comment (outdated)

This PR is rather trivial... I don't know if we  are okay with moving things in tiny steps. Tell me if we're not.

Our database schema for user notification preferences needs to be extended if we want to support another type of notifications, i.e. in app notifications.

Currently the presence of the user preference record means that the email notification is disabled while absence means it is enabled.

If we want to add in-app notifications and if we want to make them configurable independently from email notifications then we need to support 4 states, not just 2 like now.

Therefore I added 2 new columns -- `email_enabled` and `in_app_enabled`.

The default states of the columns correspond current behavior. This is done for backward compatibility. In app notifications are enabled because they are always enabled currently.

Email notifications are disabled because the presence of the record means they are disabled.

Having `null: false` and the default value set will "back fill" the existing records to correct values in the databases during the migration or lazily depending on the DB type and version.

In the next step (which I imagine to be a separate PR) we can update the existing code to check for `email_enabled` field value instead of the presence of the record. This will prepare us for the scenarios where the record exists but `email_enabled` is true.

We can also rename `email_type` column to `notification_type` sometime later.


db/migrate/20220721190153_add_email_enabled_to_user_preferences.rb
```diff
@@ -0,0 +1,6 @@
+class AddEmailEnabledToUserPreferences < ActiveRecord::Migration[6.1]
```
diaspora-code-review, line 1: Style/FrozenStringLiteralComment: Missing frozen string literal comment.

SuperTux88 (approved):
Looks good to me so far. But I'm not really sure if it's worth already merging single commits that don't do anything useful yet (well, they do something useful for the future, but merging them now doesn't have any benefit). It only fills the commit log with merge commits and later makes it more difficult to find connected commits (so other steps for the same feature, when they were spread out as single merges).

What I'm absolutely fine with (or even prefer) is, doing small steps in a PR, so having smaller commits with a clear commit message, to split a branch into multiple steps, that makes the reviewing a lot easier. And you are also welcome to ask for feedback or a review whenever you want, even when the feature isn't completely done yet.

There are some scenarios when it's also worth to already merge a PR before the feature is completely done, for example:
  * when it already does something useful for the user, so users can profit from it and don't need to wait for the full feature
  * when there are some cases that would be useful to have them merged so another feature can also build on the changes
  * or when the PR becomes bigger and there is risk of having conflicts with other branches (so similar to above, if other features are working in the same area)
  * or maybe something else that doesn't come to my mind at the moment, this list isn't complete

But none of this is the case for this PR yet, so I would just continue pushing your next steps to this PR and ask for feedback whenever you want.

Small question regarding your plan for the next steps: are you planning to insert a `true`/`true` entry for every notification type and user, or do you want to still only inserting stuff when the user changes the defaults, but assume `true`/`true` if there is no entry?

cmrd-senya:
> Small question regarding your plan for the next steps: are you planning to insert a true/true entry for every notification type and user, or do you want to still only inserting stuff when the user changes the defaults, but assume true/true if there is no entry?

I was thinking about the latter - assuming true/true when there is no entry

app/models/user.rb
```diff
@@ -180,20 +180,18 @@ def send_reset_password_instructions
...
-      else
-        block = user_preferences.find_by(email_type: key)
-        if block
-          block.destroy
-        end
-      end
+      email_enabled = pref_hash[key] == "false"
+      user_preferences
+        .find_or_create_by(email_type: key)
+        .update(email_enabled: email_enabled)
```
cmrd-senya, line 194: @SuperTux88 quick implementation question, while I haven't gone too far.

So the way I've written it here, once the user submits email configuration we create all possible records for user preferences per notification type. 

So even if I as a user uncheck just one checkbox and save, I'll still have 9 records created, the rest 8 of which match the default value. So the records are absent until the moment when the user first interacts with the email settings and after that they are created and stay in the DB forever.

Alternatively I could make the behavior here a bit more sophisticated and only create the records when the behavior is different from the default one.

In both cases if the user has never interacted with these settings they don't have user preferences records and the absence means the default behavior. But should we intentionally reduce the number of records by not creating the ones that match the default?

I like to keep the code simpler and keep how it works here -- create records for the user even if they match the default. 

But what do you think?

SuperTux88: Yes, that sounds good to me. I don't think the size of the table should become a problem (and if it does, we can still easily clean it up again in the future by just deleting all the rows that match the default).

Also I like that with the way you have done it now, if the user didn't touch the settings, it stays the default, but as soon as the user changes something, we know that they intentionally wanted to keep the others enabled, so it's fine to store them, even if it's the current default. If we would change the default in the future, then it would only change for the users who didn't touch any settings, but all users who intentionally set their preferences keep it the way they set it.

cmrd-senya: Sounds good,  thanks! There are a few tests that set expectations around number of records so I wanted to clarify that before updating it everywhere.

cmrd-senya:
image

Okay, we're almost there. It's left to update the specs.

SuperTux88:
Maybe add an explanation or a header at the top of the two columns to explain what they mean ... because at the moment somebody could also understand it as they receive a private message instead of a mail? :thinking:

cmrd-senya:
> Maybe add an explanation or a header at the top of the two columns to explain what they mean ... because at the moment somebody could also understand it as they receive a private message instead of a mail? thinking

Yes, I thought about that too. I can add a header. Maybe also some kind of tooltip could be helpful?

SuperTux88:
Yes, a tooltip would also work, but tooltips only work on desktop, not on mobile with touch as you can't hover over things.

app/views/users/_edit.haml
```diff
@@ -122,61 +122,180 @@
       .row
         .col-md-12
           %h3
-            = t(".receive_email_notifications")
```
cmrd-senya, line 125: Should we clean up existing translations for this string? Or it will be done automatically when we sync with the translation platform?

SuperTux88: You only need to update the English translations, other languages are managed by webtranslateit.

cmrd-senya:
Added hearer 
image

cmrd-senya:
@SuperTux88 I think it can be reviewed now. Let me know if I should add some additional specs.

SuperTux88 (changes requested):
As I also mentioned in some comments, but TL;DR: I would also refactor how this whole thing works some more and also change some naming, as everything is touched now anyway, so I think that's a good moment to clean some stuff up that was already confusing before and now is even more confusing.
* rename `user_preferences`/`email_preferences` to `notification_settings`
* also rename/remove some mentions of `email` in variables where it now also means in-app notifications
* fix that the booleans are the wrong way around when submitting the form, `true` should always mean enabled, `false` should always mean disabled.

Again, most of this isn't because of stuff you changed now, it was already confusing in the past, I think it's just a good opportunity to make it less confusing as most of the code is touched now anyway.

Also, the way you implemented it now, it doesn't send mails anymore if you disable in-app notifications. But emails should still be sent if they are enabled. So you also need to refactor how this works, so you are able to send emails without persisting an in-app notification to the database.

But overall I think this goes into the right direction :+1:

app/models/user.rb
```diff
@@ -180,20 +180,18 @@ def send_reset_password_instructions
 
   def update_user_preferences(pref_hash)
     if self.disable_mail
-      UserPreference::VALID_EMAIL_TYPES.each{|x| self.user_preferences.find_or_create_by(email_type: x)}
+      UserPreference::VALID_EMAIL_TYPES.each do |type|
+        user_preferences.find_or_create_by(email_type: type).update(email_enabled: false)
+      end
```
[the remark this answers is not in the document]
cmrd-senya: Yes, it seemed to me as well that it is not being used. I'll remove it.

cmrd-senya:
Thanks for the feedback, indeed, I didn't realize that sending email notifications depends on the Notification record presence. I can see how I can do that but it'll require a bit more refactoring. I'll work on it.

cmrd-senya:
@SuperTux88 I'm thinking about moving these `self.notify` methods out of the models:

https://github.com/diaspora/diaspora/blob/e82690963dc55d10491b8da6b4fc0b6f397bea6d/app/models/notifications/comment_on_post.rb#L15

I'd like to put the code to the "service" layer, possibly creating a new service object for it. There is already `NotificationService` but it will be too much to put everything there so I'd rather create another one. Could be something like `NotifyUserService` or `SendNotificationService`. I like service names to be more like verbs than like nouns. `NotificationService#notify` method could also be moved to that new service too. Basically `NotificationService` is doing a few different things right now which don't need to be kept in one class, I think that a service should implement one action, not a few different like index, create and update at once.

How does it sound? Would it be okay?

SuperTux88:
Sounds good to me so far. I don't mind too much if one service does multiple things about notifications, or if there are different services for it. But if it otherwise would become too big, it's probably better to split them up so it's clearer to read what belong together and also maybe easier to test. As the current `notify` method doesn't add a lot to the `NotificationService` at the moment, so it was fine to have it in there, but if it now gets more logic to differ between mail and in-app, it probably makes sense to split stuff up a bit. But I don't know what exactly is needed yet, but I think you'll figure that out as you go :+1:

app/services/notifications/mentioned_in_comment_service.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Notifications
+  class MentionedInCommentService
+    def self.model
+      Notifications::MentionedInComment
+    end
+
+    def self.notify(mentionable, recipient_user_ids)
```
diaspora-code-review, line 9: Metrics/AbcSize: Assignment Branch Condition size for notify is too high. [<6, 27, 6> 28.3/20]

cmrd-senya:
@SuperTux88 so after some thought I went with the following solution: I moved all `self.notify` methods from the notification models to respective services which I put to `app/services/notifications/*_service.rb`. Then I change the `self.notify` methods in the services the way that the email sending code wasn't dependent on existing notification record instance. How do you like that?

I didn't update the specs yet, thought better to discuss if the refactoring looks okay.

I'm also thinking of moving `User#mail` method somewhere to the services from the user model.

app/services/notifications/also_commented_service.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Notifications
+  class AlsoCommentedService
+    def self.notify(comment, _)
```
diaspora-code-review, line 5: Metrics/AbcSize: Assignment Branch Condition size for notify is too high. [<4, 20, 2> 20.49/20]

app/services/notifications/mentioned_in_comment_service.rb
```diff
@@ -0,0 +1,46 @@
...
```
diaspora-code-review, line 9: - Metrics/AbcSize: Assignment Branch Condition size for notify is too high. [<6, 28, 6> 29.26/20]
- Metrics/MethodLength: Method has too many lines. [25/20]

app/services/notifications/mentioned_in_post_service.rb
```diff
@@ -0,0 +1,39 @@
...
+      actor = mentionable.author
+      relevant_mentions = filter_mentions(
+        mentionable.mentions.local.where.not(person: actor),
+        mentionable,
+        recipient_user_ids
+      )
+
+      relevant_mentions.each do |mention|
+        recipient = mention.person.owner
+        next if model.exists?(recipient: recipient, target: mention)
```
SuperTux88, line 19: This relies on the notification entity existing, which isn't the case if in-app notifications are disabled. This was a workaround for when posts were received twice, no duplicate notifications were created (https://github.com/diaspora/diaspora/pull/7721). I don't remember all the details and reasons, but `Diaspora::Federation::Receive.status_message` still processes a received post even, if it already exists, and because of that it also calls `ReceiveLocal` again.

I don't know/remember why a post should be received again if it already exists, maybe that was a workaround for when a post was created incomplete because of the scenario described [here](https://discourse.diasporafoundation.org/t/federation-issue-trigger-again/4648/4?u=supertux88) where some data for posts was missing, and receiving them again properly through the federation, would add this missing data. But if that gets fixed so it fetches the posts using federation code, it doesn't need to be received a second time, as it was already complete the first time. That would then also solve the problem of duplicate notifications, without needing the check for them here (as I don't know how otherwise this should be solved for not sending duplicate mails).

SuperTux88 (changes requested):
I think overall this goes into the right direction, but there are still some issues with handling in-app and mails differently and some places still relying on the notification objects and some other minor comments.

cmrd-senya:
Thanks for the detailed review @SuperTux88! Gonna work through this :+1:

app/services/notifications/mentioned_in_post_service.rb
cmrd-senya: @SuperTux88 is it enough to do the proper fetching using federation code? It seems quite easy.

https://github.com/diaspora/diaspora/commit/880b2585ffbb231951eacec73df5a460db959517

Didn't look at the tests yet.

cmrd-senya: (Probably worth adding exception handling too)

SuperTux88: > is it enough to do the proper fetching using federation code? It seems quite easy.

Yes, that's theoretically all that's needed, but the `type` should be the lowercase, snake_case version of the entity name ([doc](https://github.com/diaspora/diaspora_federation/blob/develop/lib/diaspora_federation/federation/fetcher.rb#L9)).

But you can even improve that a little bit more by just using `:post` as type (as it could also be a reshare) and you can remove the `check_type` in `validate`, so that way we can even fetch reshares (you could also use the type from the json, but then you would need to convert that to the snake_case version, so just always use `:post` works and is easier) :)

But yes, basically that, it's fairly easy :+1:

SuperTux88: And the exception should be `DiasporaFederation::Federation::Fetcher::NotFetchable`.

EnginnerDog:
I think I just saw a unicorn in this PR.

Flaburgan:
I have regularly users complaining about not being able to disable notifications, especially the birthday one. I tried to rebase this MR but it has conflicts, and also it does not take into account the new "Likes on comment" notification. I don't know if you are still around @cmrd-senya or if someone else should take over?

Flaburgan:
@cmrd-senya I'm taking over this work, can I open PR on your fork that you would merge so they are reflected here?

cmrd-senya:
@Flaburgan sure, go ahead

Flaburgan:
OK I did the rebase here: https://github.com/cmrd-senya/diaspora/pull/1 the easiest would be for you to reset your branch with mine because rebase means history rewrite. Thank you! <3

app/models/user.rb
```diff
@@ -180,20 +180,19 @@ def send_reset_password_instructions
...
-      if pref_hash[key] == 'true'
-        self.user_preferences.find_or_create_by(email_type: key)
-      else
-        block = user_preferences.find_by(email_type: key)
-        if block
-          block.destroy
-        end
-      end
+      email_enabled = pref_hash[key]["mail"] == "false"
+      in_app_enabled = pref_hash[key]["in_app"] == "false"
```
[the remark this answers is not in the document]
Flaburgan: The reason why it is like this is because a boolean false `unchecked_value` makes Rails omit the hidden field, so unchecking would submit nothing. A fix would be to use strings instead of boolean so we return "true" or "false". Would that work for you @SuperTux88 ?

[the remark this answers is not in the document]
Flaburgan: This is what I am suggesting: https://github.com/cmrd-senya/diaspora/pull/2/changes/d35f2cc93d7d8b8292309c33bc584c89deace5d1

Flaburgan:
I don't understand, when I run `script/diaspora-dev pronto` locally it does not tell me anything. Is there something to be configured?

PR #8376, opened 2022-07-21
Improve setup and docker scripts

SuperTux88:


skyschub (approved):
Me like. I wonder if we should also drop MySQL support from this, as we're pretty much actively unsupporting it...

Flaburgan:
> Me like. I wonder if we should also drop MySQL support from this, as we're pretty much actively unsupporting it...

+1 to this, we should ditch totally MySQL support at some point so we could start with that.

denschub:
> we should ditch totally MySQL support at some point so we could start with that.

@SuperTux88 pinged me on IRC abou that, and we decided that it's probably best to keep the support for MySQL in place for now, and remove everything MySQL-related together at some point when we decide to drop it. It doesn't do any harm in the dev-docker setup, and if we have to debug something weird, keeping it in place can help us. So let's keep it for now. :)

(Sorry, I forgot to transfer this information from a private chat to the public!)

SuperTux88:
@Flaburgan yes, I'm also +1 for removing MySQL support, but (as denschub said) as long as we still support it, it's best to also keep it supported in the docker development script. Because if we need to debug/fix something MySQL related, then it's easier for us if we have an easy way to do that, instead of needing to install a full MySQL setup on our local dev environments.

PR #8377, opened 2022-07-22
Don't use host-meta in connection tester anymore

SuperTux88:
This route was removed from the federation protocol and doesn't exist anymore, so checking for it doesn't make any sense.

But lets check if a server responds to `/.well-known/nodeinfo` instead. All other software which supports the diaspora protocol should have this endpoint by now. Parsing/validating nodeinfo is still handled gracefully.

### DIFF
--- a/lib/connection_tester.rb
+++ b/lib/connection_tester.rb
@@ -1,4 +1,3 @@
-
 # frozen_string_literal: true
 
 class ConnectionTester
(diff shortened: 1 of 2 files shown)


skyschub (approved):

SuperTux88:
Merged as e984fa7d91a287bc0882cc3db2f451158035243b

Thanks for the quick review @denschub :cookie:

PR #8379, opened 2022-07-23
Add NodeInfo 2.1 and parse the newest available version

SuperTux88:
Add NodeInfo 2.1 with URL to the repo and homepage.

Also add parsing of version 2.0 and 2.1. This allows us to parse NodeInfo from other software which can't have a valid 1.0, but has a valid 2.x. It always chooses the newest version both servers support.

I also added the software version directly to the table of there weren't any errors, this allows to see easier which versions are used at the moment. It would also be cool to have a toggle to hide all pods which are offline for more than 14 days (which is also when we stop federating to that pod), probably with inactive pods hidden by default, so the list is less overloaded with offline pods ... but my frontend knowledge isn't good enough to do that myself at the moment.

### DIFF
--- a/app/assets/templates/pod_table_entry_tpl.jst.hbs
+++ b/app/assets/templates/pod_table_entry_tpl.jst.hbs
@@ -1,5 +1,5 @@
 
-<td class="ssl-status"">
+<td class="ssl-status">
   {{#if ssl}}
     <i title="{{t 'admin.pods.ssl_enabled'}}" class="entypo-check">
   {{else}}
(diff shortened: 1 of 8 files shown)


jhass (approved):

SuperTux88:
I decided to also merge this back to `next-minor` (because why not), so merged as 49ba740b45c57e4ec745736c84226dc55b18d10b, 01f8f55dbb86cb4d141e5ef8d87176b528f1f36b and 4efc4dabf87affe91e53085a5043f5cec0457c3c

PR #8383, opened 2022-07-24
Prepare the backend for generating a list of active pods

tclaus:
I've heard that pods that are offline longer than 14 days should be filtered out from admin view.

### DIFF
--- a/app/assets/javascripts/app/pages/admin_pods.js
+++ b/app/assets/javascripts/app/pages/admin_pods.js
@@ -29,6 +29,25 @@ app.pages.AdminPods = app.views.Base.extend({
 
   _showMessages: function() {
     var msgs = document.createDocumentFragment();
+    if (gon.totalCount && gon.totalCount > 0) {
+      let totalPods = $("<div class='alert alert-info' role='alert' />")
+        .append(Diaspora.I18n.t("admin.pods.total", {count: gon.totalCount}));
+      if (gon.activeCount) {
+        if (gon.activeCount === 0) {
+          totalPods
+            .append(" " + Diaspora.I18n.t("admin.pods.none_active"));
+        }
+        if (gon.activeCount === gon.totalCount) {
+          totalPods
+            .append(" " + Diaspora.I18n.t("admin.pods.all_active"));
+        } else {
+          totalPods
+            .append(" " + Diaspora.I18n.t("admin.pods.active", {count: gon.activeCount}));
+        }
+      }
+      msgs.appendChild(totalPods[0]);
+    }
+
     if( gon.uncheckedCount && gon.uncheckedCount > 0 ) {
       var unchecked = $("<div class='alert alert-info' role='alert' />")
         .append(Diaspora.I18n.t("admin.pods.unchecked", {count: gon.uncheckedCount}));
--- a/app/models/pod.rb
+++ b/app/models/pod.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class Pod < ApplicationRecord
+  # a pod is active if it is online or was online less than 14 days ago
+  ACTIVE_DAYS = 14.days
+
   enum status: %i(
     unchecked
     no_errors
@@ -39,6 +42,10 @@ class Pod < ApplicationRecord
     where(arel_table[:status].gt(Pod.statuses[:no_errors])).where.not(status: Pod.statuses[:version_failed])
   }
 
+  scope :active, -> {
+    where(["offline_since is null or offline_since > ?", DateTime.now.utc - ACTIVE_DAYS])
+  }
+
   validate :not_own_pod
 
   class << self
@@ -73,9 +80,9 @@ class Pod < ApplicationRecord
     Pod.offline_statuses.include?(Pod.statuses[status])
   end
 
-  # a pod is active if it is online or was online less than 14 days ago
+  # a pod is active if it is online or was online recently
   def active?
-    !offline? || offline_since.try {|date| date > DateTime.now.utc - 14.days }
+    !offline? || offline_since.try {|date| date > DateTime.now.utc - ACTIVE_DAYS }
   end
 
   def to_s
--- a/config/locales/javascript/javascript.en.yml
+++ b/config/locales/javascript/javascript.en.yml
@@ -70,9 +70,17 @@ en:
           other: "<%= count %>ms"
         unknown: "unknown"
         not_available: "not available"
+        total:
+          one: "There is only one known pod."
+          other: "There are <%= count %> known pods in total."
         unchecked:
           one: "There is still one pod that hasn't been checked at all."
           other: "There are still <%= count %> pods that haven't been checked at all."
+        active:
+          one: "One pod was active recently."
+          other: "<%= count %> pods were active recently."
+        none_active: "None of them were active recently."
+        all_active: "All of them were active recently."
         version_failed:
           one: "There is one pod that has no version (old pod, no NodeInfo)."
           other: "There are <%= count %> pods that have no version (old pods, no NodeInfo)."
--- a/app/controllers/admin/pods_controller.rb
+++ b/app/controllers/admin/pods_controller.rb
@@ -14,7 +14,8 @@ module Admin
           gon.unchecked_count = Pod.unchecked.count
           gon.version_failed_count = Pod.version_failed.count
           gon.error_count = Pod.check_failed.count
-
+          gon.active_count = Pod.active.count
+          gon.total_count = Pod.count
           render "admins/pods"
         end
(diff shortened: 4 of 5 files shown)


app/models/pod.rb
```diff
@@ -41,6 +41,10 @@ class Pod < ApplicationRecord
 
   validate :not_own_pod
 
+  scope :possible_available , -> {
```
diaspora-code-review, line 44: Layout/SpaceBeforeComma: Space found before comma.

app/models/pod.rb
```diff
@@ -41,6 +41,10 @@ class Pod < ApplicationRecord
...
+    where(["offline_since is null or offline_since > ?", Time.now - 14.days])
```
diaspora-code-review, line 45: Rails/TimeZone: Do not use `Time.now` without zone. Use one of `Time.zone.now`, `Time.current`, `Time.now.in_time_zone`, `Time.now.utc`, `Time.now.getlocal`, `Time.now.xmlschema`, `Time.now.iso8601`, `Time.now.jisx0301`, `Time.now.rfc3339`, `Time.now.httpdate`, `Time.now.to_i`, `Time.now.to_f` instead.

SuperTux88 (changes requested):
What I meant in #8379 was that I want to optionally filter them in the frontend (but I would need to look more into how the frontend works to do that myself). I don't just want to hide them forever, I want to have them hidden by default (because for most scenarios and for most podmins that's what makes the most sense), but I also want to be able to view them with an button, for example when I want to check what happened to an old pod and since when it's gone.

Maybe it would also be great to show two numbers at the top instead of only "The connection test returned an error for xxx pods." show how many are marked as offline, but still marked as active and still tried to federate to them, and how many are dead and not federated anymore.

app/models/pod.rb
SuperTux88: I probably would just have used `DateTime.now.utc - 14.days` as it's already used here (but that's irrelevant now anyway, as I wouldn't filter them in the backend, so this isn't needed):

https://github.com/diaspora/diaspora/blob/d9c72edb520a12d70f9c17329a743f5ea6b81e78/app/models/pod.rb#L81-L83

By the way, the naming we used for this until now was "active" (so pods who are down for less than 14 days still count as active), and you can probably just use this method that already exists and add it to the `PodPresenter` so the frontend can access the flag, so you don't need to compare timestamps in the frontend and it uses the exact same logic as in the backend. That way if we ever change the numbers of days, we still only have this one place and don't need to change it at dfifferent places.

app/assets/javascripts/app/pages/admin_pods.js
```diff
@@ -29,6 +29,16 @@ app.pages.AdminPods = app.views.Base.extend({
 
   _showMessages: function() {
     var msgs = document.createDocumentFragment();
+    if ( gon.totalCount && gon.totalCount > 0) {
```
diaspora-code-review, line 32: There should be no spaces inside this paren.

app/assets/javascripts/app/pages/admin_pods.js
```diff
@@ -29,6 +29,16 @@ app.pages.AdminPods = app.views.Base.extend({
...
+      let total_pods = $("<div class='alert alert-info' role='alert' />")
```
diaspora-code-review, line 33: Identifier 'total_pods' is not in camel case.

app/assets/javascripts/app/pages/admin_pods.js
```diff
@@ -29,6 +29,16 @@ app.pages.AdminPods = app.views.Base.extend({
...
+        .append(Diaspora.I18n.t("admin.pods.total", {count: gon.totalCount}));
+      msgs.appendChild(total_pods[0]);
```
diaspora-code-review, line 35: Identifier 'total_pods' is not in camel case.

app/assets/javascripts/app/pages/admin_pods.js
```diff
@@ -29,6 +29,16 @@ app.pages.AdminPods = app.views.Base.extend({
...
+    }
+    if ( gon.activeCount && gon.activeCount > 0) {
```
diaspora-code-review, line 37: There should be no spaces inside this paren.

app/models/pod.rb
```diff
@@ -39,6 +41,10 @@ class Pod < ApplicationRecord
     where(arel_table[:status].gt(Pod.statuses[:no_errors])).where.not(status: Pod.statuses[:version_failed])
   }
 
+  scope :active , -> {
```
diaspora-code-review, line 44: Layout/SpaceBeforeComma: Space found before comma.

tclaus:
It now just adds total and active counts to the pods list. 
Bildschirmfoto 2022-07-28 um 23 24 30

Any further front end features can now be done in another PR.

app/assets/javascripts/app/pages/admin_pods.js
```diff
@@ -29,6 +29,16 @@ app.pages.AdminPods = app.views.Base.extend({
...
   _showMessages: function() {
     var msgs = document.createDocumentFragment();
+    if (gon.totalCount && gon.totalCount > 0) {
+      let totalPods = $("<div class='alert alert-info' role='alert' />")
+        .append(Diaspora.I18n.t("admin.pods.total", {count: gon.totalCount}));
+      msgs.appendChild(totalPods[0]);
+    }
+    if (gon.activeCount && gon.activeCount > 0) {
+      let active = $("<div class='alert alert-info' role='alert' />")
+        .append(Diaspora.I18n.t("admin.pods.active", {count: gon.activeCount}));
```
SuperTux88, line 39: I think there are already a lot of boxes here, I would combine these two new sentences into the same box and maybe change the wording a little. And if there are `totalCount` there should always also be `activeCount`, shouldn't it?

So how about "x pods were active recently" for the second part, and that can go into the same box after the total count? I think it's also fine to show if there were 0 active pods recently, so no need to check `activeCount > 0`.

image

app/models/pod.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class Pod < ApplicationRecord
+  ONLINE_DAYS = 14.days
```
SuperTux88, line 4: I think I would name this `ACTIVE_DAYS` or something else with "active".

Also, I think I would move the comment from above the `active?` method to here, to explain what this is used for, as this is the common place for where this is used now and also where the number "14" lives now. Or maybe also keep a small comment with "... was online recently" at the `active?` method? :thinking:

SuperTux88 (changes requested):

tclaus:
Done.
Bildschirmfoto 2022-07-30 um 23 36 42

config/locales/javascript/javascript.en.yml
```diff
@@ -70,9 +70,17 @@ en:
...
+          one: "There is only one known pod."
+          other: "There are <%= count %> known pods in total."
         unchecked:
           one: "There is still one pod that hasn't been checked at all."
           other: "There are still <%= count %> pods that haven't been checked at all."
+        active:
+          one: "One pod is active."
+          other: "<%= count %> pods were active recently."
+        none_active: "None of them are active."
+        all_active: "All of them are active"
```
SuperTux88, line 83: This is missing the `.` at the end of the sentence.

SuperTux88 (changes requested):
Looks good to me, only one small last thing. Also can you squash the review changes to the initial commit please?

config/locales/javascript/javascript.en.yml
SuperTux88: Oh, I also just noticed that these are in different times, so the `other` uses past with "were" while the others use "are" in present. I'd probably use "... were active recently" in all of the variants.

tclaus:
Hope, I have not overseen anything else..

config/locales/javascript/javascript.en.yml
```diff
@@ -70,9 +70,17 @@ en:
...
+          one: "There is only one known pod."
+          other: "There are <%= count %> known pods in total."
         unchecked:
           one: "There is still one pod that hasn't been checked at all."
           other: "There are still <%= count %> pods that haven't been checked at all."
+        active:
+          one: "One pod was active."
+          other: "<%= count %> pods were active recently."
+        none_active: "None of them were active."
+        all_active: "All of them were active."
```
SuperTux88, line 83: The other times still miss the "recently". I mean all pods in the list were active at some point in time, but maybe not recently, so that's the important thing here.

tclaus: Oh.

SuperTux88 (approved):
Looks good to me now :cookie:

PR #8387, opened 2022-08-26
Make inline code inside links show the link color.

### LINKED ISSUE #8386: Linked inline code blocks are impossible to visually distinguish
This has been bugging me for a while now... Have a look at the following comment:

Screenshot 2022-08-26 at 19 26 05

Two of the three inline code blocks are a link. Can you spot which? No, no you can't. 😆 

I'll submit a PR in a bit.

skyschub:
Fixes #8386.

The linked issue has a screenshot of The Before. Here is what it looks with this PR in light mode:

Screenshot 2022-08-26 at 19 45 29

and dark mode:

Screenshot 2022-08-26 at 19 45 50

You can't actually link around or inside multi-line code blocks, so this is enough.

### DIFF
--- a/app/assets/stylesheets/code.scss
+++ b/app/assets/stylesheets/code.scss
@@ -5,3 +5,12 @@ pre {
   white-space: pre;
   code { white-space: pre; }
 }
+
+// For inline-code inside links, let's force the color to the default link
+// color to make them be recognizable as links.
+p a {
+  code,
+  pre {
+    color: $link-color;
+  }
+}


app/assets/stylesheets/code.scss
```diff
@@ -5,3 +5,12 @@ pre {
   white-space: pre;
   code { white-space: pre; }
 }
+
+// For inline-code inside links, let's force the color to the default link
+// color to make them be recognizable as links.
+p a {
+  code,
+  pre {
```
SuperTux88, line 13: With `pre` it would even work for multi-line code blocks. If you only want inline-code blocks, then `code` would be enough, but I guess it doesn't hurt? Or do we actually use `pre` somewhere else than in multi-line code blocks?

SuperTux88 (approved):
LGTM :cookie:

app/assets/stylesheets/code.scss
denschub: This does **not** actually work:

```md
[<pre>foo
bar</pre>](https://example.com)
```

It looks like it works, but it doesn't. The markdown parser will transform that into


```html
<pre><a href="https://example.com" target="_blank" rel="noopener noreferrer">foo<br>
bar</a></pre>
```

So there isn't actually a `<pre>` **inside** a link. My motivation for adding `pre` as well is to adequately override where the original color comes from (Bootstrap, actually). We don't do pre-inside-a, it would be invalid, but Bootstrap supports it, I overrode that. And, well, because an extra selector adds virtually no cost to anything. :)

PR #8389, opened 2022-09-03
Update next major to 1.0.0

SuperTux88:
As [it is official now](https://blog.diasporafoundation.org/74-diaspora-s-10-years-in-community-hands) that the next major release will be 1.0, I changed the changelog to that version. I also added an "unreleased" note next to it, as I have seen similar things in other projects and I like it, as it makes clear that these changes aren't in any official release yet.

I also changed the deprecation warning for the `diaspora.yml` to 2.0, as this will be the next major release now, where support for the old config will be removed.

I also changed the milestone on GitHub to [1.0.0](https://github.com/diaspora/diaspora/milestone/35)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,4 +1,4 @@
-# 0.8.0.0
+# 1.0.0 (unreleased)
 
 ## New configuration file!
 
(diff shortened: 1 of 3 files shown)


jhass (approved):

skyschub (approved):
I know Jonne already approved this, but... I WANT, TOO!

goobertron:
Thanks!

SuperTux88:
I wasn't sure about changing the development version number, but was thinking about using a `-dev` suffix. And after @denschub asked me exactly this in a private conversation, I also changed the current development version to `1.0.0-dev` to reflect the 1.0 change also in the develop version.

This means, future minor release development versions (`next-minor` branch, after the 1.0 release) will have `1.1.0-dev`, and after the 1.0 release the `develop` branch will go to `2.0.0-dev`

denschub:
Yeah, I like that, and I think the intend behind a `2.0.0-dev` is even clearer than a `1.99.0`, which feels kinda arbitrary. So 👍 for that.

PR #8390, opened 2022-09-04
Improve public posts fetch on account search

cmrd-senya:
TO DO:
- [x] Fix tests

### DIFF
--- a/lib/diaspora/fetcher/public.rb
+++ b/lib/diaspora/fetcher/public.rb
@@ -110,7 +110,7 @@ module Diaspora; module Fetcher; class Public
           post["guid"]
         )
       rescue DiasporaFederation::Federation::Fetcher::NotFetchable => e
-        logger.debug e.message
+        logger.warn e.message
       end
       set_fetch_status Public::Status_Processed
     end
--- a/lib/diaspora/fetcher/public.rb
+++ b/lib/diaspora/fetcher/public.rb
@@ -128,9 +128,8 @@ module Diaspora; module Fetcher; class Public
     # @see check_existing
     # @see check_author
     # @see check_public
-    # @see check_type
     def validate post
-      check_existing(post) && check_author(post) && check_public(post) && check_type(post)
+      check_existing(post) && check_author(post) && check_public(post)
     end
 
     # hopefully there is no post with the same guid somewhere already...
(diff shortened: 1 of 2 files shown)


lib/diaspora/fetcher/public.rb
```diff
@@ -104,16 +104,13 @@ def process_posts
...
-          created_at:            ActiveSupport::TimeZone.new("UTC").parse(post["created_at"]).to_datetime,
+        DiasporaFederation::Federation::Fetcher.fetch_public(
+          @person.diaspora_handle,
+          :post,
+          post["guid"]
         )
-        entry.save
-
+      rescue DiasporaFederation::Federation::Fetcher::NotFetchable => e
+        logger.debug e.message
```
SuperTux88, line 113: As this should basically never happen (as if the post was just there before, it should always be fetchable, unless it really gets deleted in exactly this moment), I'm not sure if it should be logged as warning (as in production we don't have debug logging enabled, and it might be interesting to see if it happens) ... on the other hand, if it basically should never happen, it might as well be fine if it just gets silently dropped in production? :man_shrugging:

SuperTux88 (commented):
Looks good so far :+1: not approving yet as you wrote something about fixing tests.

lib/diaspora/fetcher/public.rb
cmrd-senya: I think it might happen in case of some network instability, or for example if the pod is overloaded, or if there is a bug and for some reason fetching the post responds `500`. So I think it might be, though it's unlikely. So I agree it's worth making it more visible -- it shouldn't be very noisy but it could be interesting to see when this happens and how othen.

cmrd-senya:
> not approving yet as you wrote something about fixing tests.

Yeah, a few specs need to be fixed, I'm not gonna do it right now as it's getting late but I want to get back to it sometime this week.

cmrd-senya:
@SuperTux88 tests updated, ready for review

spec/lib/diaspora/fetcher/public_spec.rb
```diff
@@ -100,7 +134,7 @@
       before do
         Timecop.freeze
         @now = DateTime.now.utc
-        @data = JSON.parse(@fixture).select {|item| item["post_type"] == "StatusMessage" }
+        @data = fixture_data.select {|item| item["post_type"] == "StatusMessage" }
```
SuperTux88, line 137: This shouldn't be filtered to only `StatusMessage` anymore, as we also fetch/import reshares now.

So both the `it "applies the date from JSON to the record"` and the `it "applies now to interacted_at on the record"` tests below can directly access `fixture_data` instead of `@data`, and instead of `StatusMessage.find_by` use `Post.find_by`.

The `it "copied the text correctly"` still only applies for `StatusMessage`, as reshares don't have a `text`, but then you can just replace `@data` with `fixture_data.select {|item| item["post_type"] == "StatusMessage" }` only in that test locally (and maybe rename it to `it "copied the text of status messages correctly"`.

And then you don't need `@data` anymore in the `before` block here.

SuperTux88 (changes requested):
Only one small thing in the test, but otherwise looks good

cmrd-senya:
@SuperTux88 updated

SuperTux88 (approved):
Looks good now :cookie:

PR #8395, opened 2022-09-10
Fixes for two date-related spec breakages

### LINKED ISSUE #8394: `check_birthday_spec` is broken again...
... but it's different this time, not [the fun leap-day bug we had in 2016](https://overengineer.dev/blog/2016/03/05/the-leap-day-the-profiles-died.html)!

Turns out that our test suite will fail on September 9th, because, well, the test profile has its birthday today!

https://github.com/diaspora/diaspora/blob/1c064e016f4d1d1657778fcc739df8cd25919d76/spec/workers/check_birthday_spec.rb#L10

We really should make this date dynamically, to always set it to something valid that is.. well.. not today.

(I'm adding the "weird" label, just because it's fun.)

skyschub:
Two fixes,

1. an adjustment to `queue_users_for_removal_spec.rb` that doesn't check exact dates, but allows for a bit of tolerance (1 day). This is to accord for the timespan where the spec runs before a DST shift, but the "30 days" in the future is already after the DST switch. In those cases, the difference has been 30 days + 1 hour, and thus the test fails. Checking with 1 day of tolerance is what we do in `pod_spec.rb`, so it seems appropriate.
2. A change to `check_birthday_spec.rb` to no longer use the 9th of September as a fixed birthday, causing the specs to fail on that specific date. This fixes #8394.

### DIFF
--- a/spec/workers/check_birthday_spec.rb
+++ b/spec/workers/check_birthday_spec.rb
@@ -6,13 +6,8 @@ describe Workers::CheckBirthday do
   let(:contact2) { eve.contact_for(bob.person) }
 
   before do
-    Timecop.freeze(Time.zone.local(1999, 9, 9))
-    birthday_profile.update(birthday: "1990-09-09")
+    birthday_profile.update(birthday: Time.zone.now)
     allow(Notifications::ContactsBirthday).to receive(:notify)
-  end
-
-  after do
-    Timecop.return
   end
 
   it "calls notify method for the birthday person's contacts" do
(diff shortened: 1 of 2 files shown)


denschub:
Okay, actually, this needs more work. The better solution here is to completely remove time-travel from the birthday spec, and just assume that the profile always has its birthday *today*, and *not today* in the test that tests the opposite. Unfortunately, this means a bit of change in the tests expectations. It currently expects that no Birthday notification will be fired. However, the other test profiles that are generated into the fixtures have a birthday as well, which is the day the fixtures have been generated...

The solution here is to refactor the tests to explicitly check for birthday notifications from the profile in question. I'll change this soon.

denschub:
I have refactored the entire `check_birthday_spec` to completely remove the need for time travel.

This required some more verbose spec expectations to make sure that we only expect/not-expect notifications for the actual contacts in question, since it's very possible that other users generated by the fixtures have their birthday at the same time (their birthday is the date the fixtures have been generated)...

SuperTux88 (approved):
LGTM, thanks for fixing these :cookie:

PR #8396, opened 2022-09-19
Fix docker development setup after switch to puma

SuperTux88:
* #8392 removed `single_process_mode`, which means it now requires a redis instance for development setups. This wasn't the case yet in our docker setup, so it failed to start because it didn't find redis. I added a redis container to our docker-compose setup.
* #8392 also changed the default `listen` configuration for development to only listen on localhost. While this is a good default for normal development setups to not expose a port, it also broke our docker setup because the port now couldn't be forwarded outside of the docker container anymore. So I change this back to the old default when used with docker.
* I also exposed the `docker-compose` command outside of the `diaspora-dev` script, so I'm able to run normal `docker-compose` commands directly without needing to manually set the correct environment variable that are required by our `docker-compose.yml`.

For the configurations for `redis` and `listen`, I overwrite the values set in the `diaspora.toml` as environment variables in the `docker-compose.yml`, as otherwise everybody would need to ensure that the configure the correct settings themselves in their `diaspora.toml`. That way it ensures that you can't break the docker-setup with a wrong setting, but can still set different settings in the `diaspora.toml` when running without docker.

### DIFF
--- a/docker/develop/docker-compose.yml
+++ b/docker/develop/docker-compose.yml
@@ -24,6 +24,7 @@ services:
       - ${DIASPORA_DOCKER_PORT:-3000}:3000
     environment:
       - ENVIRONMENT_REDIS=redis://redis
+      - SERVER_LISTEN=tcp://0.0.0.0:3000
     depends_on:
       - "${DIASPORA_DOCKER_DB}"
       - redis
--- a/docker/develop/docker-compose.yml
+++ b/docker/develop/docker-compose.yml
@@ -1,6 +1,7 @@
 version: "3.4"
 
 volumes:
+  redis_data:
   postgresql_data:
   mysql_data:
   dia_data_tmp:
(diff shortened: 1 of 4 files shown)


docker/develop/docker-compose.yml
```diff
@@ -21,8 +22,18 @@ services:
...
+      - SERVER_LISTEN=tcp://0.0.0.0:3000
     depends_on:
       - "${DIASPORA_DOCKER_DB}"
+      - redis
+
+  redis:
+    image: redis:7
+    command: redis-server --save 60 1 --loglevel warning
+    volumes:
+      - redis_data:/data
```
denschub, line 36: Technically this isn't needed, as we don't care too much if the data in Redis doesn't survive a restart for dev - but it doesn't matter. :)

skyschub (approved):
I ran a few tests and it seems to work fine. Thanks for cleaning up my mess.. 🙈

docker/develop/docker-compose.yml
SuperTux88: Yes and no, I thought about that. You are right, with the `single_process_mode` we didn't keep anything before, but when doing federation tests (where it also was required to disable `single_process_mode` before) or something, it might be useful to keep the state (and retries or scheduled jobs), so I thought it doesn't hurt.

denschub: Yeah, that's reasonable!

Issue #7975, opened 2019-02-01
Cannot tag users that have usernames that start with an underscore


Dartan:
I have a user on my pod with a username that starts with an underscore (2 of them actually). Because of this it's impossible to tag them in anything. Their name does popup in the list when you type @__ to search for them and it looks like it's adding the tag properly but when you post/comment the tag isn't working.

jakehell:
@Dartan I am not able to reproduce locally. I tried tagging a user with one and two underscores both in a comment and a post.

SuperTux88:
Oh, maybe that was already fixed with #7788 in the `develop` branch? It sounds similar as #7640 (which was fixed by that), but this time it's the username and not the display name containing the underscores. Can you check if you can reproduce with `master`-branch, but not with `develop` (which you already did)?

jakehell:
@SuperTux88 I switched to the master branch and recompiled and am still unable to reproduce.

SuperTux88:
I fixed the issue for the mobile UI in #8398, for the desktop UI it should be fixed in [`markdown-it-diaspora-mention`](https://github.com/diaspora/markdown-it-diaspora-mention).

PR #8398, opened 2022-10-05
Escape mentions before parsing message with markdown for mobile UI

SuperTux88:
Usernames that contained underscores were parsed by markdown first. This broke the diaspora IDs and also added weird html at places where it wasn't needed. Escaping them before sending the message through the markdown parser fixes this issue.

As underscores are the only allowed character that can be used for markdown that is also allowed inside a diaspora ID, this escaping can be kept pretty simple.

Initially I also added `escape_mentions_for_markdown` in the `plain_text_for_json`, which then gets used for the desktop UI. This kind-of fixes the issue also there, but I then decided to revert it again after thinking more about it. I think this is better fixed in the markdown-it parser for the desktop UI directly (so probably with the [`markdown-it-diaspora-mention`](https://github.com/diaspora/markdown-it-diaspora-mention) plugin?). Here are my thoughts about it:
* While fixing it the same way as for mobile worked, it didn't work for the preview. Handling it in `markdown-it-diaspora-mention` would also fix it in the preview.
* Adding it to `plain_text_for_json` would also escape it for the API. While it might be cool that not all clients need to handle that themselves, I don't think it's a good thing to do and this was the main reason I reverted it again. I initially thought, that we could use the escaping in the backend for now, until we have a proper solution in the frontend, but when we then remove the backend-escaping again, that would be a "breaking change" for the API too. Also the escaping only works, if you parse the post through markdown FIRST, before handling the mentions, the other way around the escaping would break the mentions. And I don't know if all clients using the API are doing it in that order. So I think it's better for all clients to handle this themselves, and either also escape it themselves, or do whatever is needed to parse it properly.

Related to #7975 (fixes it for mobile only)

### DIFF
--- a/lib/diaspora/mentionable.rb
+++ b/lib/diaspora/mentionable.rb
@@ -71,6 +71,16 @@ module Diaspora::Mentionable
     }
   end
 
+  # Escapes special chars in mentions to not be parsed as markdown
+  #
+  # @param [String] text containing mentions
+  # @return [String] escaped message
+  def self.escape_for_markdown(msg_text)
+    msg_text.to_s.gsub(REGEX) {|match_str|
+      match_str.gsub("_", "\\_")
+    }
+  end
+
   private_class_method def self.find_or_fetch_person_by_identifier(identifier)
     Person.find_or_fetch_by_identifier(identifier) if Validation::Rule::DiasporaId.new.valid_value?(identifier)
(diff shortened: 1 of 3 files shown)


skyschub (approved):
Sorry, took some extra time here because I wanted to check if maybe just skipping parsing for the mention sequence is feasible, but this looks like an even bigger pain, so this is fine.

PR #8399, opened 2022-10-06
Add pull command to diaspora-dev script

SuperTux88:
Before the images were only pulled once and then never updated which lead to really outdated images and OS dependencies. Now all images (including the base image for the diaspora container when it gets built) are pulled when running `setup`. So the idea is to run the `setup` command from time to time to bring everything up to date again.

I also updated to use the latest postgres 10 image. This is currently still supported, but we probably should think about upgrading to newer postgres versions. But as https://github.com/docker-library/postgres/issues/37 is still not fixed, this isn't that easy. I don't know if we could just upgrade to a newer versions and tell devs to start a fresh database (so nothing needs to be migrated), or we we need to provide an upgrade path somehow?

Also, the used mariadb version isn't even supported anymore, but there I know even less about how to handle upgrades (so I didn't touch that for now). Apparently it also requires manually running a command. But as we only have this for the rare case that we need to debug something with it, it might be fine to just force that to a new version and create a fresh database when we need to debug something?

### DIFF
--- a/docker/develop/docker-compose.yml
+++ b/docker/develop/docker-compose.yml
@@ -23,8 +23,8 @@ services:
     ports:
       - ${DIASPORA_DOCKER_PORT:-3000}:3000
     environment:
-      - ENVIRONMENT_REDIS=redis://redis
-      - SERVER_LISTEN=tcp://0.0.0.0:3000
+      ENVIRONMENT_REDIS: redis://redis
+      SERVER_LISTEN: tcp://0.0.0.0:3000
     depends_on:
       - "${DIASPORA_DOCKER_DB}"
       - redis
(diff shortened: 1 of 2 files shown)


docker/develop/docker-compose.yml
```diff
@@ -36,7 +36,7 @@ services:
       - redis_data:/data
 
   postgresql:
-    image: postgres:10.3
+    image: postgres:10-bullseye
```
denschub, line 39: Is there any reason not to use 15?

skyschub (approved):

docker/develop/docker-compose.yml
SuperTux88: I initially didn't want to do a major upgrade (see initial PR comment), but after discussing in IRC we decided, that it doesn't matter for a development environment, as the database probably gets wiped regularly anyway and there shouldn't be any important data.

So you need to delete the volume and let it recreate, the easiest way is probably running `script/diaspora-dev docker-compose down -v` followed by `script/diaspora-dev setup` (there are other ways which selectively only delete the postgres volume, or you can try to migrate/upgrade the data, but I leave that to the devs if they want to do that).

PR #8400, opened 2022-10-07
Add redirect to mobile-only photo URL when in desktop UI

### LINKED ISSUE #8352: Gallery links don't work on Mozilla Firefox
# Versions
Diaspora* 0.7.15.0 / 0.7.16.0 (I tried on 2 different pods)
Mozilla Firefox for Arch Linux 100.0 (64-bit)
# Problem
I use Firefox, and a friend of mine uses Safari. Both of us aren't logged in.

- My friend sent me this link https://iviv.hu/people/bfe281001b5a8561/photos/127981 and it shows a blank page for me.
- They tried sharing it from another pod https://pod.dapor.net/people/b29871b0fdf4013827ac10bf487fe7cf/photos/1163493, but it still had the same result.
- Then, we found that this link https://iviv.hu/posts/609670 works.
- Later they tried opening the gallery link on their firefox, and it also didn't work.

I didn't try on a pod on latest version (0.8.0.0) but I don't see anything in the changelog implying that this was fixed.

SuperTux88:
This URL is only used in the mobile UI, but when somebody then copies the link and sends it to somebody on the desktop UI, they don't see anything. So lets just redirect to the post containing the photo, so there is at least something to show.

If there is no linked post, just redirect to the image instead.

Fixes #8352

### DIFF
--- a/app/controllers/photos_controller.rb
+++ b/app/controllers/photos_controller.rb
@@ -13,9 +13,17 @@ class PhotosController < ApplicationController
                current_user.photos_from(Person.find_by(guid: params[:person_id])).where(id: params[:id]).first
              else
                Photo.where(id: params[:id], public: true).first
-    end
+             end
 
     raise ActiveRecord::RecordNotFound unless @photo
+
+    respond_to do |format|
+      format.html {
+        post = @photo.status_message
+        redirect_to post ? post_path(post) : @photo.url
+      }
+      format.mobile { render "photos/show" }
+    end
   end
 
   def index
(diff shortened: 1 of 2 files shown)


skyschub (approved):
*Technically* this is a feature addition and needs a test. But I think it's fine without, it makes sense.

SuperTux88:
> Technically this is a feature addition and needs a test.

Sure, I added even _two_ tests :+1:

denschub:
Uhhh. 🍪🍪

PR #8403, opened 2022-10-31
Cleanup duplicate pods

SuperTux88:
There were already attempts to remove duplicates in the past ... but the unique index doesn't work with `NULL` ports, which is used for when a default port is used (80 or 443) ... which is almost every pod. PostgreSQL >= 15 would now have an index that also handles `NULL` values correctly (with `NULLS NOT DISTINCT`), but not all podmins have PostgreSQL 15 yet and there is also still MySQL which has the same problem.

So I now use `-1` as value for when a default port is used, as this also works with unique indexes everywhere, and I didn't find a better solution that works with MySQL. We maybe can revert that back to `nil` once we drop MySQL (or even drop PostgreSQL < 15). I thought about using the actual port (80/443), but then we have the problem that it a pod could exist with both of them, that's why I rejected this idea again.

There were also duplicates because the hostname wasn't normalized to lowercase, so the same hostname with mixed case could exist multiple times, so I also fixed that as a drive-by.

If pods use a non-default port I also added the port in the overview, so two pods with different pods don't look like duplicates.

### DIFF
--- a/app/assets/javascripts/app/views/pod_entry_view.js
+++ b/app/assets/javascripts/app/views/pod_entry_view.js
@@ -27,6 +27,7 @@ app.views.PodEntry = app.views.Base.extend({
   presenter: function() {
     return _.extend({}, this.defaultPresenter(), {
       /* jshint camelcase: false */
+      hasPort: (this.model.get("port") >= 0),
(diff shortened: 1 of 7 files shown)


skyschub (approved):
Seems like a reasonable workaround to me.

Issue #8118, opened 2020-05-14
No scrolling after closing the photo viewer


haheute:
Hello, 

Sometimes after viewing a photo in the photo viewer, the scrolling of the page is disabled.
I hope I found the right steps to reproduce:

* open the [photos page](https://diaspora.schoenf.de/people/ec8d8a8064eb0137510f101b0ee4c30c/photos) (I did with chromium in an incognito window) 
* scroll until you saw the loading, spinning images below the photos. (trying to load more, I guess)
* click a photo, so that the viewer opens up
* close the viewer with a click on the "X" in the top right corner
* try to scroll. body should still have `overflow:hidden` (no scrollbar)

(tried on https://diaspora.schoenf.de with code 0.7.13.0 with chromium Version 81.0.4044.138 (Offizieller Build) Arch Linux (64-Bit))

Flaburgan:
I reproduce this issue on framasphere.org running last stable. Thanks for reporting it! I see if I am able to fix it but I'm not very active on diaspora at the moment so no promise.

Flaburgan:
Okay, I found the (beginning) of the problem.
When opening the gallery, Blueimp stores the overflow of `<body>` and then set it to hidden:

```
this.bodyOverflowStyle = document.body.style.overflow
document.body.style.overflow = 'hidden'
```

Then when the user closes the gallery, it restores it:
```
document.body.style.overflow = this.bodyOverflowStyle
```

This works well on the stream page, and even on the user page in the posts tab. The bug on the images is simple: the Blueimp gallery is loaded **several times**. So the first time, it stores `""` (the empty string) in `bodyOverflowStyle`. But then the second time, `document.body.style.overflow` is set to `hidden`. So, blueimp stores `hidden` in `bodyOverflowStyle`, and restores it to `hidden` when we quit.

Now, I need to find why blueimp is executed several time on that page **only after we loaded more images**.

PR #8404, opened 2022-10-31
Do not recreate blueimp each time you're scrolling in the photos page

Flaburgan:
Fxes #8118


SuperTux88 (approved):
Thanks for this fix :cookie:

PR #8406, opened 2022-10-31
Do not show the service to the user if it is disabled, fixes #6184

### LINKED ISSUE #6184: Users are able to select services in the publisher that are no longer supported
Hi,

I linked facebook a long time ago with diaspora*. I would like to unlink it, but it´s not possible and I cannot see it in https://joindiaspora.com/services to "disconnect it".
##

Flaburgan:
I was wondering if we should disconnect all users if we detect that the podmin changed the config to disable the service, but first of all config change isn't easy to detect, and second I think that if the podmin then re-enables it it would be a shame to ask all users to reconnect.

### DIFF
--- a/app/views/publisher/_publisher.html.haml
+++ b/app/views/publisher/_publisher.html.haml
@@ -57,7 +57,8 @@
           .btn-toolbar.pull-right#publisher-service-icons
             - if current_user.services
               - current_user.services.each do |service|
-                = service_button(service)
+                - if AppConfig.configured_services.map(&:to_s).include? service.provider
+                  = service_button(service)
             .btn.btn-link.question_mark{title: t("shared.public_explain.manage"),
                            data: {toggle: "modal", target: "#publicExplainModal"}}
               %i.entypo-cog
--- a/app/views/publisher/_publisher.html.haml
+++ b/app/views/publisher/_publisher.html.haml
@@ -55,10 +55,9 @@
             %button.btn.btn-group.btn-primary#submit= t("shared.publisher.share")
 
           .btn-toolbar.pull-right#publisher-service-icons
-            - if current_user.services
-              - current_user.services.each do |service|
-                - if AppConfig.configured_services.map(&:to_s).include? service.provider
-                  = service_button(service)
+            - available_services.each do |service|
+              = service_button(service)
+
             .btn.btn-link.question_mark{title: t("shared.public_explain.manage"),
                            data: {toggle: "modal", target: "#publicExplainModal"}}
               %i.entypo-cog
--- a/app/helpers/publisher_helper.rb
+++ b/app/helpers/publisher_helper.rb
@@ -5,6 +5,10 @@
 #   the COPYRIGHT file.
 
 module PublisherHelper
+  def available_services
+    current_user.services.select {|service| AppConfig.configured_services.map(&:to_s).include? service.provider }
+  end
+
   def service_button(service)
     provider_title = I18n.t("services.index.share_to", provider: service.provider.titleize)
     content_tag :div,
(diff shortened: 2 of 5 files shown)


denschub:
This breaks Jasmine, there are specs that have expectations around available services that are no longer true with this PR.

skyschub (changes requested):
(see previous comment)

spec/javascripts/app/views/publisher_view_spec.js
```diff
@@ -317,6 +317,7 @@ describe("app.views.Publisher", function() {
 
   context("services", function(){
     beforeEach( function(){
+      gon.appConfig = {configured_services: ["twitter", "tumblr"]};
```
diaspora-code-review, line 320: Identifier 'configured_services' is not in camel case.

Flaburgan:
Yep, it should be fixed now. I had to switch to another computer so it took time for `diaspora-dev setup` to run.
Should I also add a test to test the code I wrote? Should it be rspec or cucumber?

Flaburgan:
OK so I'm confused @SuperTux88 said this to me on IRC:

>the jasmine spec is probably broken because it generates the templates with different settings ... the change you made in `publisher_view_spec.js` doesn't affect `_publisher.html.haml` so you need to change the configured services in the rspec test that generates the template that is used for the failing jasmine test ... and to test that locally you need to re-generate the fixtures locally, because you probably still have them from before your change

But actually locally the jasmine tests are green without adding anything else that what I did here. What you are saying is that I should also modify `publisher_spec.rb`?

SuperTux88:
> But actually locally the jasmine tests are green without adding anything else that what I did here. What you are saying is that I should also modify `publisher_spec.rb`?

When I wrote that I didn't check which fixture is used and where it's generated. I assumed that you didn't re-generate your fixtures locally, so you are still using fixtures that were generated before your change in `_publisher.html.haml`. And the change you did to `publisher_view_spec.js` isn't needed and doesn't change anything, as it doesn't affect the fixture (locally your tests are probably green with or without that change there? But if you re-generate fixtures with `bin/rake tests:generate_fixtures`, then I assume that your tests aren't green anymore).

So in the line below the line you added, you see it loads `aspects_index_services`, and that one is generated in `aspects_spec.rb`, so you need to change that so the fixture still contains the services, and then jasmine will work again.

Flaburgan:
The related tests are fixed, the current failure in Rspec is unrelated, `develop` is also failing on that test. So this should be ready for review.

skyschub (approved):
Ideally, this would have a test with a service that's disabled globally but enabled in the user settings, and then confirming that the button isn't showing up. Writing a test for the actual behavior change here, that is.

But I feel adventurous today, so I'm personally not gonna block on that.

app/views/publisher/_publisher.html.haml
```diff
@@ -57,7 +57,8 @@
           .btn-toolbar.pull-right#publisher-service-icons
             - if current_user.services
```
SuperTux88, line 58: Sidenote, as you didn't change it here, but I don't think this `if` is needed, as if there are no services, the list would just be empty, so there is nothing to loop over anyway?

SuperTux88 (changes requested):

SuperTux88 (commented):
> However, it is hard to test as I can't enable services for my user locally (if you can tell me how to do that).

Just create a dummy service with dummy data on the user manually. You won't be able to post to it, but you can test if the buttons show:

```ruby
user = User.find_by(username: "test")
user.services << Services::Tumblr.new(nickname: "nick", access_token: "token", access_secret: "secret", uid: "uid", provider: "Tumblr")
user.services << Services::Wordpress.new(nickname: "nick", access_token: "token", access_secret: "secret", uid: "uid", provider: "Wordpress")
```

SuperTux88:
I removed your last commit, and added my own with the helper that just filters the services and also added a test for that helper.

This now just filters the services for both UIs (so fixes the bug for both of them), but keeps the design otherwise as is.

skyschub (approved):

Flaburgan:
Thank you for finishing this

PR #8409, opened 2022-11-13
Add app smart banner to web site when using an iOS device

tclaus:
As Inspiration App looms on the horizon, an app smart banner would help iOS user to find the app in the store and to inform them that there is an app. 

A app smart banner is shown in the header of the web site if it is open on an iOS device.
The user is then informed that an app exists for diaspora and can be redirected directly to the store.
When a user ever discarded the message its never shown again. Has the user the app already installed, the "install" button changes to "open app". 

It helps spreading the information that there is an official app for diaspora.

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -6,6 +6,7 @@
 %html{lang: I18n.locale.to_s, dir: (rtl? ? "rtl" : "ltr")}
   %head{prefix: og_prefix}
     %meta{name: "viewport", content: "width=device-width, initial-scale=1"}/
+    %meta{name: "apple-itunes-app", content: "app-id=1538074832, app-argument=#{uri_with_username}"}
 
     - content_for :javascript do
       = javascript_include_tag :main
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -13,6 +13,14 @@ module ApplicationHelper
     AppConfig.version.number
   end
 
+  def uri_with_username
+    if user_signed_in?
+      AppConfig.pod_uri + "?username=#{current_user.username}"
+    else
+      AppConfig.pod_uri
+    end
+  end
+
   def changelog_url
     return AppConfig.settings.changelog_url.get if AppConfig.settings.changelog_url.present?
 
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -165,4 +165,34 @@ describe ApplicationHelper, :type => :helper do
       expect(pod_version).to match "0.0.1.0"
     end
   end
+
+  describe "#uri_with_username" do
+    attr_reader :current_user
+
+    before do
+      @current_user = alice
+      def user_signed_in?
+        true
+      end
+    end
+
+    it "displays the pod uri and username if logged in" do
+      allow(AppConfig).to receive(:pod_uri) { "https://diaspora.social" }
+      expect(uri_with_username).to match "https://diaspora.social?username=alice"
+    end
+  end
+
+  describe "#uri_with_username without logged in user" do
+    before do
+      @current_user = alice
+      def user_signed_in?
+        false
+      end
+    end
+
+    it "displays the pod uri" do
+      allow(AppConfig).to receive(:pod_uri) { "https://diaspora.social" }
+      expect(uri_with_username).to match "https://diaspora.social"
+    end
+  end
 end


app/views/layouts/application.html.haml
```diff
@@ -6,6 +6,7 @@
 %html{lang: I18n.locale.to_s, dir: (rtl? ? "rtl" : "ltr")}
   %head{prefix: og_prefix}
     %meta{name: "viewport", content: "width=device-width, initial-scale=1"}/
+    %meta{name: "apple-itunes-app", content:"app-id=1538074832"}
```
diaspora-code-review, line 9: Layout/SpaceAfterColon: Space missing after colon.

SuperTux88 (changes requested):
I didn't know something like that is possible, I think it's a great idea. I looked at [the apple documentation](https://developer.apple.com/documentation/webkit/promoting_apps_with_smart_app_banners) of this feature and it shows that you can also add an `app-argument` parameter, wouldn't it make sense to also provide the pod URL there (or maybe even a link to the current post? not sure about that), so that when one clicks the banner they get already the correct pod selected so they don't need to enter their URL. This probably also needs to be supported by the app, but I assume that we can also already add the parameter now even if the app doesn't know what to do with it yet and it just ignores it?

Also, it looks like you messed up the commits, as you have 4 commits for a single line change, I guess you have pulled (and merged) instead of force-pushing after a rebase. So you need to rebase/squash again and then force-push.

jhass:
There's no "pod selection" UI in the app, so prefilling the pod doesn't really exist, but when the user is signed in we can certainly prefill the username.

tclaus:
Unser it as @SuperTux88 it won't hurt and enables the app for future tasks?

app/views/layouts/application.html.haml
```diff
@@ -6,6 +6,7 @@
 %html{lang: I18n.locale.to_s, dir: (rtl? ? "rtl" : "ltr")}
   %head{prefix: og_prefix}
     %meta{name: "viewport", content: "width=device-width, initial-scale=1"}/
+    %meta{name: "apple-itunes-app", content: "app-id=1538074832, app-argument=#{AppConfig.pod_uri}"}
```
SuperTux88, line 9: So @jhass said that the app doesn't use the URL to select the pod, but it needs the username (I assume in form of the diaspora ID? Or how else does the app know which pod to contact?), so I think we should also add the username part here (as parameter `?username=<username>`, because the apple documentation says this needs to be an URL?), so the app can parse the things it needs to directly redirect the user to the correct pod.

I must confess that I never used the app yet (I'm just more a desktop person), so I don't know exactly what the app asks for, but I just thought it would be cool to provide it with the needed information to make it easier for the user to log in the first time.

goobertron:
It would certainly be good to think about how to co-ordinate this functionality with the app. Even if the app ignores it presently, it might be good to build in a method of pre-selecting the pod, and that could then be match by some code in the next iteration of the app. Or perhaps leave that functionality until the app can do something with it?

tclaus:
Makes even more sense. Actually I am not sure how to extract the parameter in the iOS part and transmit it to the flutter part, but it is good to have this data anyway.

app/views/layouts/application.html.haml
jhass: Yes the app only asks for the diaspora ID

tclaus:
> > Makes even more sense. Actually I am not sure how to extract the parameter in the iOS part and transmit it to the flutter part, but it is good to have this data anyway.
> 
> A bit OT here, but probably something in the vein of the share request or authorization handlers: https://github.com/jhass/insporation/blob/main/ios/Runner/AppAuthHandler.swift -> https://github.com/jhass/insporation/blob/2c99b88bd2f527cf61a09cb2bb6ae68e0e0de38a/lib/app.dart#L63-L70

I'll check.. sometime :/

app/views/layouts/application.html.haml
tclaus: @SuperTux88 I already added the username (aka diaspora ID) to the banner. Can you check?

SuperTux88 (approved):
Looks good to me, I still don't know if it would be useful to include the full URL to a post, if the user is on a single-post-page, so the app can also open the same post again. But also that's again something the app needs to support, and I'm not sure if that's actually the goal of this or if it's only to tell the users that the app exists. So if we want to add that feature (and it's also supported by the app), we can always add the path to a post later.

PR #8410, opened 2022-12-05
Allow DOCKER_HOST env var for diaspora-dev docker setup

SuperTux88:
This allows to use rootless podman (which doesn't require sudo) instead of docker.


skyschub (approved):
Nice.

PR #8417, opened 2023-06-04
Always strip exif data and drop user setting for it

SuperTux88:
Some imagemagick-versions (I tested Ubuntu 22.04 and debian bullseye) always loose exif data when converting from jpg to webp. So this made our CI fail now, but even if it wasn't failing before, some pods always had and have versions which might loose the information anyway. So having a setting to keep exif information is kinda pointless, if we can't guarantee that the information isn't lost. Also, diaspora isn't a photo sharing platform and we don't display exif information anywhere, so I think we should just always strip exif data (which was already the default before), as we don't need them.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -31,18 +31,6 @@ class UsersController < ApplicationController
     render :edit
   end
 
-  def update_privacy_settings
-    privacy_params = params.fetch(:user).permit(:strip_exif)
-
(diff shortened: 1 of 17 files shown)


denschub:
> exif is still mentioned in `schema.rb`

That's a local file that reflects *your* database state - it's not in this repo.

skyschub (approved):
I'm amazed at our test suite's ability to randomly fail. Thanks for the fix.

PR #8418, opened 2023-06-05
Switch from apparition to cuprite driver

SuperTux88:
Apparition isn't really maintained anymore and it looks like cuprite won the race of releasing something first (see: https://github.com/diaspora/diaspora/pull/8330#issuecomment-1186580512).

Supersedes #8330


skyschub (approved):
It's still green, so it's hard to object. Looks good. :)

PR #8419, opened 2023-06-05
Use podman to build when installed and configured

SuperTux88:
I had problems when trying to build with podman with docker-compose 2.x, but it works when just calling the `podman build` command directly instead.

### DIFF
--- a/script/diaspora-dev
+++ b/script/diaspora-dev
@@ -235,8 +235,15 @@ dia_get_db() {
 
 dia_build() {
   if [ $# -gt 0 ] && [ "$1" == "--no-cache" ]; then nocache="--no-cache"; fi
+
   # Build the diaspora Docker container (diaspora:dev-latest)
-  dia_docker_compose build --pull $nocache diaspora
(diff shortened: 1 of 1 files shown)


script/diaspora-dev
```diff
@@ -235,8 +235,15 @@ dia_get_db() {
 
 dia_build() {
   if [ $# -gt 0 ] && [ "$1" == "--no-cache" ]; then nocache="--no-cache"; fi
+
   # Build the diaspora Docker container (diaspora:dev-latest)
-  dia_docker_compose build --pull $nocache diaspora
+  if type podman &> /dev/null && [[ "${DOCKER_HOST}" == *"podman"* ]]; then
```
denschub, line 240: I think `command -v` might be more common than `type`, but no objections here, just a note.

skyschub (approved):

script/diaspora-dev
SuperTux88: Sure. I don't really mind, I just picked one that worked for me at another place, but I switched it to `command -v`. :+1:

Issue #8369, opened 2022-07-21
Support ruby 3


SuperTux88:
I already fixed some issues to support ruby 3 in #8368, but there are still some issues that don't work yet with ruby 3.0.x:

* [x] [i18n-inflector-rails](https://rubygems.org/gems/i18n-inflector-rails) isn't compatible with ruby 3, and the last release is from 2013. But there has been some activity in the [Repo](https://github.com/siefca/i18n-inflector-rails), and it looks like the issue with ruby 3 [was fixed](https://github.com/siefca/i18n-inflector-rails/pull/10) ... but there hasn't been a release yet. And the fix also changed the requirement to be ruby `>= 3`, so we can't support ruby 2.7 at the same time (which means as soon as we would ruby 3 support with it, we need to drop 2.7, so we can't have a release that supports 2.7 AND 3). There are multiple options now:
  * Wait until a new version is released, update to that, and drop ruby 2.7
  * Find a different solution to achieve the same, which maybe already works with ruby 3 (and also 2.7)
  * Drop the dependency, as far as I can tell it's only used in 8 translation strings (mostly in `pl`)
* [x] there is a problem with [carrierwave](https://rubygems.org/gems/carrierwave) and `mini_magick`. Basically the feature added in #7924 doesn't work in ruby 3: ``convert: unrecognized gravity type `{:combine_options=>{:unsharp=>"1.5x1+0.7+0.02"}}' @ error/convert.c/ConvertImageCommand/1762.``, because it adds the `combine_options` to the `gravity` parameter. As far as I can tell there isn't a fix for that yet.
* [x] With ruby 3.1 the tests don't even start, and I haven't looked closer into that. Lets first focus on getting ruby 3.0 in a working state.

Neustradamus:
@SuperTux88: Any news on it?

Ruby 2.7.x is EOL.

SuperTux88:
The `develop` branch now support ruby 3.0 and 3.1, and 2.7 got dropped.

PR #8420, opened 2023-06-08
Remove i18n-inflector-rails dependency

SuperTux88:
There are no new releases anymore and the current version isn't compatible with ruby 3.x, see: https://github.com/siefca/i18n-inflector-rails/pull/10 and https://github.com/siefca/i18n-inflector-rails/pull/11

As this feature wasn't really used a lot (Icelandic didn't even setup inflections properly), it's probably not worth fighting for it, so lets just drop it.

Related to #8369


denschub:
I feel like leaving a comment here just for the after-world if this issue resolves itself. This is too funny to let go.

The tests for this PR are currently red:

Screenshot 2023-06-09 at 03 44 47

Why, you ask? Well:

Screenshot 2023-06-09 at 03 45 30

Because our test suite somehow decided that it now wants to run in French instead of English. Why does it do that? We have absolutely no idea.

Flaburgan:
> Because our test suite somehow decided that it now wants to run in French instead of English. Why does it do that? We have absolutely no idea.

We run the world :joy:

app/controllers/application_controller.rb
```diff
@@ -91,9 +88,11 @@ def set_diaspora_header
   def set_locale
     if user_signed_in?
       I18n.locale = current_user.language
+      puts "logged in locale: #{locale}"
```
diaspora-code-review, line 91: Rails/Output: Do not write to stdout. Use Rails's logger if you want to log.

app/controllers/application_controller.rb
```diff
@@ -91,9 +88,11 @@ def set_diaspora_header
...
     else
       locale = http_accept_language.language_region_compatible_from AVAILABLE_LANGUAGE_CODES
       locale ||= DEFAULT_LANGUAGE
+      puts "locale: #{locale}"
```
diaspora-code-review, line 95: Rails/Output: Do not write to stdout. Use Rails's logger if you want to log.

SuperTux88:
> Because our test suite somehow decided that it now wants to run in French instead of English. Why does it do that? We have absolutely no idea.

OK, as fun as this sounds, it was actually very hard to debug and find out what's going on (especially because whatever I tried, I wasn't able to reproduce it locally). But I finally found the reason ... it turns out that after the locale was set to French once, it modified the English default translations, and after that point English was translated to French in-memory ... so lets just not do that anymore ...

I have no idea how exactly that worked before, maybe something with the inflector prevented that from happening? But I already debugged enough today, so I'm just happy it works now.

jhass (approved):

denschub:
Amazing investigative work, @SuperTux88. 🙌

PR #8421, opened 2023-06-09
Cleanup unicode emojis from email headers

SuperTux88:
Some email providers (for example gmail) block emails if they have emojis in the from header, as they could be confused with UI elements. So the easy solution is to just filter all emojis from the name.

The normal `\p{Emoji}` selector also matches normal numbers, because of the emoji-version of numbers (1️⃣), but the `\p{Emoji_Presentation}` then doesn't match colored emojis anymore (❄️), so we need a mix of both to find all emojis

### DIFF
--- a/app/mailers/notification_mailers/base.rb
+++ b/app/mailers/notification_mailers/base.rb
@@ -36,13 +36,21 @@ module NotificationMailers
 
     def default_headers
       from_name = AppConfig.settings.pod_name
-      from_name += " (#{@sender.profile.full_name.empty? ? @sender.username : @sender.name})" if @sender.present?
+      from_name += " (#{person_name(@sender)})" if @sender.present?
 
       {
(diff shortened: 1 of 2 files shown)


Flaburgan:
But do we want to keep support for emoji numbers? We can simply remove them too, or replace them with real font numbers?

Abeanna (approved):
approve

jhass (approved):

SuperTux88:
> But do we want to keep support for emoji numbers? We can simply remove them too, or replace them with real font numbers?

I don't think it's worth replacing anything, just removing them should be "good enough" here (that's what I did). Also, if we know which emojis gmail is blocking, we could only remove these and leave the others, but I didn't find a complete list (I only learned about this because some mails from my pod got rejected), and also maintaining a list (as it maybe needs to be updated) is also a lot more work than just dropping all of them.

PR #8425, opened 2023-06-11
Use YAML.unsafe_load_file when available in bundler helper

jhass:
fixes #8424

following https://github.com/rails/rails/blob/f642505479b06fa20d9f717d4770dad4d9ad5893/activesupport/lib/active_support/configuration_file.rb#L21-L32

Issue #8424, opened 2023-06-11
Loading `database.yml` fails in `config/bundler_helper.rb` fails with Ruby 3.1


denschub:
Using Ruby 3.1 with its internal Psych 4.x YAML parser, starting diaspora\* fails with

<details>
<summary>Unable to load application: Psych::BadAlias: Unknown alias: postgresql</summary>

```
! Unable to load application: Psych::BadAlias: Unknown alias: postgresql
/usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:432:in `visit_Psych_Nodes_Alias': Unknown alias: postgresql (Psych::BadAlias)
        from /usr/lib/ruby/3.0.0/psych/visitors/visitor.rb:30:in `visit'
        from /usr/lib/ruby/3.0.0/psych/visitors/visitor.rb:6:in `accept'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:35:in `accept'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:347:in `block in revive_hash'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:345:in `each'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:345:in `each_slice'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:345:in `revive_hash'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:169:in `visit_Psych_Nodes_Mapping'
        from /usr/lib/ruby/3.0.0/psych/visitors/visitor.rb:30:in `visit'
        from /usr/lib/ruby/3.0.0/psych/visitors/visitor.rb:6:in `accept'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:35:in `accept'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:347:in `block in revive_hash'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:345:in `each'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:345:in `each_slice'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:345:in `revive_hash'
        from /usr/lib/ruby/3.0.0/psych/visitors/to_ruby.rb:169:in `visit_Psych_Nodes_Mapping'
[...]

PR #8426, opened 2023-06-12
Add ruby 3.1, drop ruby 2.7

SuperTux88:
So that fixes the rest of the things required to support ruby 3.1. And I also dropped 2.7 instead (it's EOL anyway, so I don't think there is a need to support it any longer?). And as we don't have 2.7 anymore, I also changed to workaround from #8425 to be simpler, as mentioned in https://github.com/diaspora/diaspora/issues/8424#issuecomment-1586172231


skyschub (approved):
Full support. I added a section to the changelog yesterday, we can just edit that to say that 2.7 is no longer supported. This might trip up some folks running `develop`, as up until literally yesterday, 2.7 was the only supported version, but it was overdue.

SuperTux88:
> This might trip up some folks running `develop`, as up until literally yesterday, 2.7 was the only supported version, but it was overdue.

I hope people running `develop` are paying a bit more attention to the changelog, but some people are maybe also already waiting for newer ruby versions as 2.7 is already EOL.

PR #8436, opened 2023-10-16
Allow CORS requests to nodeinfo endpoint

palant:
Web applications meant to work with different Fediverse applications need to download `/.well-known/nodeinfo` and `/nodeinfo/2.0` in order to determine what software they are dealing with. Diaspora currently doesn’t allow this to happen on the client side, without CORS headers third-party web pages aren’t allowed to do it. I’ve hit this issue myself when implementing “Share to Fediverse” functionality for my website.

Adding CORS headers is trivial and solves this issue. With the node info being public information, there is no reason why third-party websites shouldn’t have access to it.


skyschub (approved):
Fair enough. Thanks for the PR!

PR #8439, opened 2023-11-24
add API routes for comment likes

jhass:


tclaus (approved):
LGTM

app/controllers/api/v1/likes_controller.rb
```diff
@@ -66,9 +80,67 @@ def post_service
...
+        comment ? true : false
+      rescue ActiveRecord::RecordNotFound
+        false
+      end
+
+      def comment_is_for_post(post_guid, comment_guid)
+        comments = comment_service.find_for_post(post_guid)
+        comment = comments.find {|comment| comment[:guid] == comment_guid }
+        comment ? true : false
+      end
```
SuperTux88, line 143: The `comment_exists` is kinda redundant, if the comment doesn't exist, it also doesn't exist "for the post", so only checking `comment_is_for_post` is enough.

Also in `comment_is_for_post` you can use `.exists?` directly on an SQL level, instead of looping through all comments with `find`.

Like this:
```suggestion
      def comment_and_post_validate(post_guid, comment_guid)
        if comment_is_for_post(post_guid, comment_guid)
          true
        else
          render_error 404, "Comment not found for the given post"
          false
        end
      end

      def comment_is_for_post(post_guid, comment_guid)
        comments = comment_service.find_for_post(post_guid)
        comments.exists?(guid: comment_guid)
      end
```

This way the whole validation is just a single sql-query, instead of 2 queries and looping through all comments in ruby.

SuperTux88 (changes requested):
Looks mostly good, only one little comment for when checking if a comment exists and belongs to the correct post.

skyschub (approved):
🚢

PR #8442, opened 2024-02-11
Mark Likes on comment notifications as read when visiting a post

Flaburgan:
So this is fixing the fact that "Like on comment" notification is not marked as read when one visits a post.

Initially, I was thinking about marking it read only when the actual comment is visited, like:  
image

But then I saw in the code that everything is marked as read when a post is visited (mentions in comment too for example) so I kept this consistent.

I need help with rubocop: 
- it asks me to not have a if if it's multilines, but then tell me I should not have a block with if and use a safe guard instead
- it says that I should not use update_all as this skips validation, but that's what is used in the same file above for mentions, and when I [search online](https://stackoverflow.com/questions/39644646/running-validations-when-using-update-all) the other solution  is apparently much slower (create the objects in memory, do one SQL query per update...) What do you think?

### DIFF
--- a/app/services/post_service.rb
+++ b/app/services/post_service.rb
@@ -104,9 +104,8 @@ class PostService
   end
 
   def mark_like_on_comment_notifications_read(post_id)
-    comment_ids = Comment.where(commentable_id: post_id)
-
-    Notification.where(recipient_id: user.id, target_type: "Comment", target_id: comment_ids, unread: true)
-                .update_all(unread: false) if comment_ids.any?
+    Notification.where(recipient_id: user.id, target_type: "Comment",
+                       target_id: Comment.where(commentable_id: post_id, author_id: user.person.id),
+                       unread: true).update_all(unread: false) # rubocop:disable Rails/SkipsModelValidations
   end
 end
--- a/app/services/post_service.rb
+++ b/app/services/post_service.rb
@@ -33,6 +33,7 @@ class PostService
     return unless user
     mark_comment_reshare_like_notifications_read(post_id)
     mark_mention_notifications_read(post_id)
+    mark_like_on_comment_notifications_read(post_id)
   end
 
   def destroy(post_id, private_allowed=true)
@@ -100,5 +101,12 @@ class PostService
     Mention
       .joins("INNER JOIN comments ON mentions_container_id = comments.id AND mentions_container_type = 'Comment'")
       .where(comments: {commentable_id: post_id, commentable_type: "Post"})
+  end
+
+  def mark_like_on_comment_notifications_read(post_id)
+    comment_ids = Comment.where(commentable_id: post_id)
+
+    Notification.where(recipient_id: user.id, target_type: "Comment", target_id: comment_ids, unread: true)
+                .update_all(unread: false) if comment_ids.any?
   end
 end
--- a/spec/services/post_service_spec.rb
+++ b/spec/services/post_service_spec.rb
@@ -168,6 +168,50 @@ describe PostService do
       expect_any_instance_of(PostService).not_to receive(:mark_mention_notifications_read).with(post.id)
       PostService.new.mark_user_notifications(post.id)
     end
+
+    context "for comments" do
+      let(:comment) { post.comments.create(author: alice.person, text: "comment") }
+
+      it "marks a corresponding notifications as read" do
+        FactoryBot.create(:notification, recipient: alice, target: comment, unread: true)
+        FactoryBot.create(:notification, recipient: alice, target: comment, unread: true)
+
+        expect {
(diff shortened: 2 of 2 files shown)


denschub:
While it's easy, it might not be the best approach. Just looping over all comments in a given thread is kinda bad, as there could be an unlimited number of comments to iterate on. You're fetching a potentially unlimited number of comments, and you're passing that into the lookup for the notification. This is bound to be slow.

A better solution here is to adjust the notification itself to include a link that contains a get parameter containing the actual comment id. That way, you can fetch only notifications for that comment, which will be much quicker (and constant). It'll also limit the number of notifications you need to update to always exactly one, removing any other performance problem.

Of course, that means that if a user has multiple of their comments liked in the same thread, it would only mark one notification as read, leaving the other one as unread. I'm not sure if this would be a bug or a feature, I don't actually know what our users would expect to happen in this case. Ultimately, I'd go with your initial implementation and only mark one notification as read. If we get complaints about this, we can move the the other (slower) impl.

SuperTux88:
Yes, looping through all comments every time a post gets loaded is a really bad idea. We have some posts with a lot of comments, where loading the comments is already slow (which happens async at the moment, so at least the post loads fast). But now this would also slow down the loading of the post itself.

> Of course, that means that if a user has multiple of their comments liked in the same thread, it would only mark one notification as read, leaving the other one as unread. I'm not sure if this would be a bug or a feature, I don't actually know what our users would expect to happen in this case.

I think it's fine to only mark the notification as read when looking at the specific comment (so clicking the notification). It's the same as it works for posts, where it only gets marked as read when you visit the specific post, but when you scroll over the post on a stream, and maybe read it, it doesn't mark it as read. So I would even consider marking all notifications of comments on the same post (which would happen with the loop over all comments method) as a "bug", given our current behavior for posts.

denschub:
> given our current behavior for posts.

The problem is that our current impls are inconsistent, too. If you get mentioned two times in the same comment thread, accessing the post will mark both "you have been mentioned" notifications as read. Performance-wise, it's a bit less severe [as we're joining over the comments in the database](https://github.com/diaspora/diaspora/blob/71e6f2074069249f79894a7e03d0c7bffa457194/app/services/post_service.rb#L88-L104), but we could to the same, here, too.

Flaburgan:
Just to be sure, as I'm not familiar with ActiveRecord. My actual code is not looping, is it? `comment_ids = Comment.where(commentable_id: post_id)` gets actually converted to a single SELECT in SQL, isn't it? 
And same for `Notification.where(recipient_id: user.id, target_type: "Comment", target_id: comment_ids, unread: true).update_all(unread: false)` it's actually only one UPDATE statement?

About the behavior, at the moment when a user visits a post, every notification linked to that post is marked as read: mention in the post, comment on the post, mention in a comment on the post, like on the post. That's why as said, to be consistent, I also added "like on a comment on the post" to this list. To me in the end it makes sense to assume that if users visited a post, they read  the comments thread.

SuperTux88:
> If you get mentioned two times in the same comment thread, accessing the post will mark both "you have been mentioned" notifications as read.

Oh, you are right. But it's also kinda consistent, as the mentioned notification replaces/counts as the normal "commented" notification (so you don't get one of these if you also get mentioned, no double notification). And the normal "commented" notification get marked as read with the post. if you would want it completely consistent you would need to send 2 notifications, and attach the mention notification to the comment, which then can be marked as read only with this specific comment. But I think that would be worse UX.

> My actual code is not looping, is it?

Ok, it's not "looping", but it's still loading all the comments on a post.

> About the behavior, at the moment when a user visits a post, every notification linked to that post is marked as read: mention in the post, comment on the post, mention in a comment on the post, like on the post

The problem with likes on comments is, it's linked to the comment, not to the post, that's why at the moment it doesn't get marked as read.

And as mentioned above, the "mention in a comment on the post" has kind of a special role, as it's a two in one notification and also replaces the "comment on the post" in that case, that's why it gets marked as read with the post.

SuperTux88:
I changed the code to let the database do all the work in a single query, so it's at least a bit more optimized. There is no need to load all the comments to ruby first. And I also added some tests.

skyschub (approved):
We talked about this on IRC a bit, and as opposed to the original approach, this is

- passing the IDs of possible affected comments directly into the outer query (here, it's doing a `target_id IN (..subselect..)`)
- it's also filtering the list of comment candidates to only comments written by the current user, as yo don't get like notifications for someone else's comments.

It's hard to tell how fast or slow this really is, as it's hard to test in isolation. We did test this on a post with 200 comments and the performance isn't a concern. This is a simple implementation, that matches what the user probably expects, so let's ship this. If this causes issue in the future, we can work on this again. :)

PR #8444, opened 2024-04-06
Adding filter-buttons in pod list view (#21)

tclaus:
* Filter buttons for Pod view
On Pods with large lists of pods - the admin pod list is mixed with available and invalid pods. 
This PR adds a simple filter for 'active' and 'invalid' pods. 

Bildschirmfoto 2024-04-06 um 08 44 08

### DIFF
--- a/app/assets/javascripts/app/pages/admin_pods.js
+++ b/app/assets/javascripts/app/pages/admin_pods.js
@@ -4,10 +4,48 @@ app.pages.AdminPods = app.views.Base.extend({
   templateName: "pod_table",
 
   tooltipSelector: "th i",
+  events: {
+    "click #show_all_pods": "showAllPods",
+    "click #show_active_pods": "showActivePods",
+    "click #show_invalid_pods": "showInvalidPods"
+  },
 
   initialize: function() {
     this.pods = new app.collections.Pods(app.parsePreload("pods"));
     this.rows = []; // contains the table row views
+    this.podfilter = "active";
+  },
+
+  showAllPods: function() {
+    this.podfilter = "";
+    this.postRenderTemplate();
+    this.$("#show_all_pods").addClass("active");
+    this.$("#show_active_pods").removeClass("active");
+    this.$("#show_invalid_pods").removeClass("active");
+  },
+
+  showActivePods: function() {
+    this.podfilter = "active";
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/app/pages/admin_pods.js
```diff
@@ -4,10 +4,48 @@ app.pages.AdminPods = app.views.Base.extend({
...
     this.rows = []; // contains the table row views
+    this.podfilter = "active";
+  },
+
+  showAllPods: function() {
+    this.podfilter = "";
+    this.postRenderTemplate();
+    this.$("#show_all_pods").addClass("active");
+    this.$("#show_active_pods").removeClass("active");
+    this.$("#show_invalid_pods").removeClass("active");
```
SuperTux88, line 24: I don't know if there is a better solution than manually adding and removing the `active` class everywhere? Because that doesn't scale that well if there are are more buttons (as it looks like you plan to add one for blocked pods in the future).

I don't know if there is a possibility to group these buttons somehow so only one can be active at once, no matter how many buttons exist (like radio buttons in forms?). Maybe just remove it from all buttons and just add it back to the one that was clicked?

app/assets/javascripts/app/pages/admin_pods.js
```diff
@@ -16,11 +54,17 @@ app.pages.AdminPods = app.views.Base.extend({
...
     var fragment = document.createDocumentFragment();
+    this.$("tbody").empty();
+
     this.pods.each(function(pod) {
-      self.rows.push(new app.views.PodEntry({
-        parent: fragment,
-        model: pod
-      }).render());
+      if (self.podfilter === "" ||
+        self.podfilter === "active" && pod.get("status") === "no_errors" ||
```
SuperTux88, line 61: Maybe change this one to check the `active` field instead? This would then include all pods which are being handled as "active" in the backend (all pods to which is being federated). This also includes pods with warnings (but which are still online and working), and pods which only have errors since less than 14 days (so maybe only a temporary downtime?).

If you want to filter for "no error" then the label on the button text should probably say something like "no error", because "active" means something different.

SuperTux88 (changes requested):
I wasn't the one who was requested for review, but since there was no review yet I still left some feedback ;)

PR #8448, opened 2024-05-30
Update dependencies and ruby version

SuperTux88:
It's time to do another round of dependency update.

I also switched to the latest ruby versions (3.2 and 3.3), as on local tests everything seems to work (except a little workaround that was needed, but in my opinion still better than being stuck on 3.1).


config/boot.rb
```diff
@@ -6,3 +6,11 @@
 
 # Load configuration early
 require_relative "load_config"
+
+# Ruby 3.2 removed the `exists?` alias from `File`, but at least the `logging-rails` gem still uses it.
+# This is only a workaround, and we need a different long-term solution. It looks like the `logging-rails` gem
+# is not maintained anymore, so we maybe need to find a replacement. But since this is the only thing preventing
+# us from upgrading to Ruby >= 3.2, we can just create our own alias for now.
+class << File
```
denschub, line 14: I wonder if we should have a dedicated `monkey_patches.rb` file just to make this a bit more discoverable. But I'm fine with this as-is.

skyschub (approved):
<3

config/boot.rb
SuperTux88: Yeah, I was wondering that as well, but decided against it (as I hope we don't need a lot of monkey patches and a dedicated file would maybe more encourage it? :man_shrugging:). But if we need more, we could create one.

denschub: Sounds good, let's keep it as-is :)

SuperTux88:
Since there was just a new rails bugfix release today, why not include that as well :joy: But since that was basically the first thing I update, I needed to rebase ALL commits :rofl:

skyschub (approved):

PR #8452, opened 2024-08-30
docker-compose.yml - bump mariadb version

grooverdan:
Noted its not the primary development environment, but 10.9 is out of support so just keep the the LTS tag.

MARIADB_AUTO_UPGRADE=1 facilitates in place upgrades.

### DIFF
--- a/docker/develop/docker-compose.yml
+++ b/docker/develop/docker-compose.yml
@@ -45,10 +45,11 @@ services:
       POSTGRES_PASSWORD: postgres
 
   mysql:
-    image: mariadb:10.9
+    image: mariadb:lts
     ports:
       - 53306:3306
     volumes:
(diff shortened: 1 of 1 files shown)


SuperTux88 (changes requested):
The upgrade is fine, and if it auto-upgrades the data, even better (but since this is only for development, it would be also fine if you need to start with a fresh database every now and then). :+1: 

But renaming the container won't work, as it's expected to be named `mysql` by the scripts and the other configs.

SuperTux88 (approved):

SuperTux88:
Merged as 53b9e3cbf30bc9bdabb235569711913fe1bc4d35, thanks for the fix :cookie:

grooverdan:
Easy. Thanks for the merge and current MariaDB support.

PR #8454, opened 2024-10-30
Let users report comments below their own posts.

skyschub:
A user just raised this, as they were trying to report spam below their own post and couldn't. I think the motivation behind the current logic here was to hide the report button since the user can just delete the comment, but there's value to reporting spam (for example to make sure a spammer is deleted globally).

The current CSS already handles this nicely:

Screenshot 2024-10-30 at 22 16 54

We don't have behavior testing coverage for the comment view AFAICT, so I didn't add any. We [do have unit tests for the `canRemove()` function](https://github.com/diaspora/diaspora/blob/37096d9c8805b4c6f0ea397fff8176d928a526ad/spec/javascripts/app/views/comment_view_spec.js#L49-L77), but to be somewhat direct: they're kinda pointless, they're effectively saying "a function that has been called has indeed been called". I don't see a huge point in writing unit tests for a function that is composing two other functions with test coverage - but of course I'm happy to add some. :)

I also removed the `{{#if loggedIn}}` gate. Both `canRemove()` and `canReport()` already do check if the user is authenticated, so that's not really needed and just adds more noise.

### DIFF
--- a/app/assets/javascripts/app/views/comment_view.js
+++ b/app/assets/javascripts/app/views/comment_view.js
@@ -27,6 +27,7 @@ app.views.Comment = app.views.Content.extend({
   presenter: function() {
     return _.extend(this.defaultPresenter(), {
       canRemove: this.canRemove(),
+      canReport: this.canReport(),
       text: app.helpers.textFormatter(this.model.get("text"), this.model.get("mentioned_people")),
       likesCount: this.model.attributes.likesCount,
       userLike: this.model.interactions.userLike()
@@ -43,6 +44,10 @@ app.views.Comment = app.views.Content.extend({
 
   canRemove: function() {
     return app.currentUser.authenticated() && (this.ownComment() || this.postOwner());
+  },
+
+  canReport: function() {
+    return app.currentUser.authenticated() && !this.ownComment();
   },
 
   toggleLike: function(evt) {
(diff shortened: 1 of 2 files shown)


SuperTux88:
I don't know if this is really that useful. If people want the spam to be removed, they should just delete it (which they already can). If they want to report it, they would need to keep the comment, until the podmin saw the report and deleted the spam globally, otherwise the podmin just sees an empty report and doesn't know what was reported, because it's already gone. This isn't clear by the UI, so I expect most people will then just report, and immediately delete the comment, which doesn't help anybody, expect that the podmin gets a notification that something was reported, but can't do anything. I too had somebody report a user via mail who spammed under their posts, but the comment was also already gone, so I couldn't see it anymore. But, because it was an email, luckily they also mentioned the diaspora-ID of the spammer in the mail, so I could then look that user up and see that it's indeed a spammer and delete the whole user. But the current report feature doesn't give this functionality.

So instead of allowing people to report comments on their own posts, it would probably be more useful to be able to report other accounts, so you have a report about the whole account, which you can still handle, even if one specific comment was already deleted (but for a spammer there are probably more). And even better would be, to allow to actually report other accounts to the podmin of that account, so that can actually be closed at the source, but that's probably way out of scope here.

But I don't think this change in it's current form is that useful, as it will just generate a bunch of reports about then already immediately deleted comments, as the UI doesn't say, that they would need to keep the comment after they reported it, so the podmin can actually see it.

PR #8455, opened 2024-11-24
Basic changes for italian translations

faghy:
Fixed errors in the locales folder that caused missing elements in the app display when the user was logged in, by setting DEFAULT_LANGUAGE = "it". Translated javascript.it.yml in the javascript folder and added the missing necessary strings in it.yml inside the diaspora folder

### DIFF
--- a/config/locales/diaspora/it.yml
+++ b/config/locales/diaspora/it.yml
@@ -488,6 +488,7 @@ it:
       settings: Impostazioni
       logout: Esci
       code: Codice
+      toggle_navigation: "Attiva/disattiva navigazione"
     application:
       powered_by: POWERED BY DIASPORA*
       whats_new: Novità?
@@ -984,3 +985,4 @@ it:
     registrations: Registrazioni
     enabled: Disponibile
     disabled: Non disponibile
+
--- a/config/locales/javascript/javascript.it.yml
+++ b/config/locales/javascript/javascript.it.yml
@@ -3,204 +3,345 @@
 #   the COPYRIGHT file.
 
 
-
 it:
   javascripts:
-    timeago:
-      prefixAgo: ''
-      suffixAgo: fa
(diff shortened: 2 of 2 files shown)


denschub:
Hi! Thanks for your willingness to contribute. Unfortunately, we cannot accept this PR.

[As documented here](https://wiki.diasporafoundation.org/Contribute_translations), we use an external tool, webtranslateit, for translations. We do this for a couple of different reasons, like having discussions about specific strings, an easier-to-use history, as well as allowing people without technical skills to translate.

Sadly, pushing your updates to WebTranslateIt would be quite a big pain, so I have to ask you to enter your translations there. We regularly pull those updates, so your changes will automatically be included in a future release.

faghy:
ok, thanks, please know that if someone sets the DEFAULT_LANGUAGE to Italian in initializers/locale.rb with the current files the application does not load the header and menus when the user is logged in, so it is a application bug, in fact our newly created https://diaspo.it pod is currently the only one in Italian language: https://diaspora.fediverse.observer/diaspo.it

denschub:
This is hardly a bug, as you're not supposed to change anything outside the configuration files we designed. diaspora\* automatically detects the user's browser language and sets that as the profile language, and if you don't touch with our implementation code, diaspora\* will just fall back to the English strings whenever something isn't translated to Italian.

faghy:
ok, what I don't understand is why the algorithm of fediverse.observer where you report the pod did not detect our pod in Italian, but in English. After changing DEFAULT_LANGUAGE (or something else that I don't know) it correctly detected the Italian language. I am aware that the correct value is EN and the translations are loaded automatically based on the user's language and where they do not exist the fallback is inserted, but we liked to emphasize to the users that our pod is for Italians and to make known the social and the philosophy of our country. In any case, thanks for your availability.

Issue #8196, opened 2021-01-08
Feature request: All notifications connected to a user are removed when that user is blocked


SunyataZero:
### Background

I (and others) have been having problems with getting a lot of spam-mentions from Ivan (who creates a lot of different user accounts), he will @mention me perhaps 50 times in comments and this will be my procedure:
* blocking him
* going to "notifications" and then "mentioned in comment" and clicking "mark all shown as read"

### Suggestion

Why not remove (alternatively: mark as read) all notifications connected to the user that we block? This would speed up dealing with spam messages

***

Gratitude: Thank you to everyone who has contributed to diaspora, i've made friends using your/our platform that i wouldn't have met otherwise. Just yesterday i talked over jitsi to one person that i've chatted with on diaspora. It's wonderful to make new friends like this. (And the platform has some really interesting people)

zlaxy:
1. A bunch of political propaganda accounts operates many years in the Diaspora*. They use fake girl profiles to promote mainly neoliberal agendas as well as [SEO promotion of pornographic or prostitution accounts](https://diasp.org/u/mary23z). Here the details of one of the main of these botfarm accounts:
https://sysad.org/posts/1311982
Remarkably, this popular account tried to recruit me but blocked me from commenting after i tried to [provided evidence of Ilya's murder at the request of another commenter](https://despora.de/posts/11066755#48887550a0bb0137b51900155d000111). That is, the first blocking against me by this bot farm was the attempt to hide [evidence of Ilya's murder](https://sysad.org/posts/1389054).
Here is a description of other fake accounts run by this organisation:
https://sysad.org/posts/1412989
Two years ago, bunch of these accounts were engaged in [well-paid promotion](https://www.insidephilanthropy.com/fundraising-for-climate-change) of a neocolonial agenda. I have [personally suffered from this type of propaganda](https://sysad.org/posts/1280269), and i also have many personal friends from Africa [who suffer from this type](https://www.euronews.com/2019/07/27/europe-s-electronic-waste-ends-up-at-this-toxic-landfill-in-ghana) of [Western propaganda](https://www.irishtimes.com/opinion/cobalt-mining-shows-clean-energy-revolution-comes-at-a-price-1.3717629). So i have left critical comments on the posts of these propagandists, they have deleted such comments and used blockers to hide their motives. But one time, they made a blunder and left insults against me from 4 of their puppet accounts within 5 minutes:
[...]

PR #8456, opened 2024-12-03
Implement issue #8196: Leave on read all notifications involving a blocked person

nat-sch:
Hi, hope everyone is doing well!

This PR would implement the feature requested in #8196.

I went with the proposed alternative of marking the notifications as read instead of deleting them, because I think that's more trackable in case the notifications contained content that might be of interest to keep, such as proof of harassment, and also because I feel deleting them would be more unexpected for the user.

I didn't use a cuke since I didn't wanna add such a slow and high level test for this comparatively low impact feature, but I can add it if we deem it necessary.

I'll be waiting to hear your thoughts! 😄

### DIFF
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -56,6 +56,17 @@ class NotificationService
     notification_types(object).each {|type| type.notify(object, recipient_user_ids) }
   end
 
+  def read_all_involving(person)
+    Notification
+      .for(@user)
+      .joins(:notification_actors)
+      .where(notification_actors: {person: person})
+      .find_each do |note|
+        note.unread = false
+        note.save
+      end
+  end
+
   private
 
   def notification_types(object)
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -60,7 +60,10 @@ class NotificationService
     Notification
       .for(@user)
       .joins(:notification_actors)
-      .where(notification_actors: {person: person})
+      .where(
+        notification_actors: {person: person},
+        unread:              true
+      )
       .find_each do |note|
         note.unread = false
         note.save
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -56,13 +56,18 @@ class NotificationService
     notification_types(object).each {|type| type.notify(object, recipient_user_ids) }
   end
 
-  def read_all_involving(person)
+  def read_all_only_involving(person)
+    one_actor_notes = Notification
+                      .joins(:notification_actors)
+                      .group("notifications.id")
+                      .having("COUNT(notification_actors.notification_id) = 1")
     Notification
       .for(@user)
       .joins(:notification_actors)
       .where(
         notification_actors: {person: person},
-        unread:              true
+        unread:              true,
+        id:                  one_actor_notes
       )
       .find_each do |note|
         note.unread = false
--- a/app/services/notification_service.rb
+++ b/app/services/notification_service.rb
@@ -57,21 +57,14 @@ class NotificationService
   end
 
   def read_all_only_involving(person)
-    one_actor_notes = Notification
-                      .joins(:notification_actors)
-                      .group("notifications.id")
-                      .having("COUNT(notification_actors.notification_id) = 1")
     Notification
-      .for(@user)
       .joins(:notification_actors)
-      .where(
-        notification_actors: {person: person},
-        unread:              true,
-        id:                  one_actor_notes
-      )
+      .where(recipient_id: @user.id, unread: true)
+      .group("notifications.id")
+      .having("COUNT(notification_actors.notification_id) = 1")
+      .having("MIN(notification_actors.person_id) = ?", person.id)
       .find_each do |note|
-        note.unread = false
-        note.save
+        note.update!(unread: false)
       end
   end
 
--- a/app/assets/javascripts/app/models/person.js
+++ b/app/assets/javascripts/app/models/person.js
@@ -39,7 +39,10 @@ app.models.Person = Backbone.Model.extend({
 
     // return the jqXHR with Promise interface
     return block.save()
-      .done(function() { app.events.trigger('person:block:'+self.id); });
+      .done(function() {
+        app.events.trigger("person:block:" + self.id);
+        app.notificationsCollection.fetch();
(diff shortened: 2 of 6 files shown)


nat-sch:
I see that the job `(3.2, postgresql, other)` failed. I can run the failing command as `RAILS_ENV=test bin/rake jasmine:ci` on my machine, with Ruby 3.2.6 and over Postgres 14.13, without it failing. I get `954 specs, 0 failures, 4 pending specs`, like in the sucessful Jasmine jobs. I don't know why it failed on the CI 😞

denschub:
The test failure is a random intermittent, it looks like. I just re-triggered the specific suite.

nat-sch:
Thank you!! It looks like it was, now all checks have passed! 🚀

skyschub (approved):
Okay, first: thanks for your PR, and welcome to this project. It's rare that the first time contribution of anyone a) passes all lints, b) passes all tests, c) even introduces new tests. Kudos. :)

This looks fine to me, especially the client side. I'll hold off on @SuperTux88 in case he has some comments about the Ruby side, but it looks good to me.

app/services/notification_service.rb
```diff
@@ -56,6 +56,17 @@ def notify(object, recipient_user_ids)
...
 
+  def read_all_involving(person)
+    Notification
+      .for(@user)
+      .joins(:notification_actors)
+      .where(notification_actors: {person: person})
+      .find_each do |note|
+        note.unread = false
+        note.save
+      end
```
SuperTux88, line 67: I'm wondering if it would be worth also filtering only still unread notifications? Or instead of changing each entity separately, maybe use `update_all(unread: false)`? This would skip the vailidation of the entity (as it directly calls the database and skips the rails logic) and pronto/rubocop will be unhappy about it. But there is no validation based on the read-state, if it was valid before, it's also still valid after being marked as read, so I think in this case it should be fine to ignore this rule.

However, I think the current implementation is fine for now for the current intended use-case, where people probably don't have a lot of already read notifications for a person they want to block. And also they probably don't have a lot of notifications overall yet for the person they want to block (at least not as many as it would make a notable performance difference).

SuperTux88 (approved):
Thanks for the PR. This looks good to me. I had one comment that came to my mind while reading the code, but it's more of a note/question and I think it's fine for now as it is. :+1:

app/services/notification_service.rb
nat-sch: Ahh, insightful, thank you! I went with filtering the already read notes in the `where`.

I don't like sending a query that updates a row with exactly the same data, I didn't realize that Rails' optimizations didn't reach that far. Also, thinking about the use case, I'm not sure, but think that it will probably happen sometimes that a user wants to block another user with whom they have have been interacting for a while.

About the validations, I like the idea of not skipping them right now, because I want to make it easy to extend them in the future (therefore reducing the amount of places where they can be skipped).

SuperTux88: > after more research, maybe Rails doesn't send a query if the record didn't change upon save, but it will run the validations.

I'm also not 100% sure if rails really sends out the update or what it maybe automatically skips, but it will at least run something, and for a loop with potentially a lot of entries, I at least wanted to think about.

> About the validations, I like the idea of not skipping them right now, because I want to make it easy to extend them in the future (therefore reducing the amount of places where they can be skipped).

Yeah, I also thought about if it could be a problem for future validations, but I couldn't think of any potential useful validation where a notification could be valid if it's unread, but then invalid if it's read. But I'm fine with leaving it with the loop and update each notification separately. The `update_all` was only an idea that would enable it to do it with just one query. But with the read notifications filtered out, there shouldn't be a lot of notifications left in most cases, so this is fine. :+1:

app/services/notification_service.rb
```diff
@@ -56,6 +56,20 @@ def notify(object, recipient_user_ids)
     notification_types(object).each {|type| type.notify(object, recipient_user_ids) }
   end
 
+  def read_all_involving(person)
+    Notification
+      .for(@user)
+      .joins(:notification_actors)
+      .where(
+        notification_actors: {person: person},
```
SuperTux88, line 64: While thinking about this a bit more, another potential problem came to my mind. If multiple people comment on the same post, we aggregate the notifications (that's why it can have multiple actors). So there could be a notification "Person A, Person B and Person C commented on your post", and I think with the current logic, if I block Person B, it would mark this notification as read, and I might miss the comments of Person A and C.

As a user, my expected behavior would be, that it only marks the notification as read, where the blocked user is the only actor.

However this maybe also is only an edge case, when I decide to block somebody who commented on a lot of posts where I also have a lot of notifications from other users I didn't read yet. So maybe it's fine to ignore that in the first version and improve later. But if you find an easy way to select only notifications where there is only one actor (the blocked person), that would be cool to add.

nat-sch: I agree, thank you! Also thank you for the explanation on the purpose of storing notification actors!

I implemented only fetching notifications that have exclusively the blocked person as their actor. Let me know what you think!

nat-sch: Hi @SuperTux88, I hope you are doing well. I wanted to kindly follow up on this when you have a moment. Let me know if there's anything else I can do to move it forward! Thank you for all your help so far, and thank you in advance!

SuperTux88 (approved):
Approved from my side, but since I also did some changes now, I requested another review from @denschub for my commit.

skyschub (approved):
No objection, looks good to me. Thank you so much, both! 🧡

SuperTux88:
Merged as 5ac83952d686ce97042521d8be3d5d3c1f970c69 5fd1199d2035612ec4ce38a797b877ba64bfcca9 7586cda5e0951dbd19b3ade877ab0b37e8e82543 9f31159ffd45100299df620d7280f02cc9c7ec6f 1e7b844c2ea068c37745e7f0e1ad629a2a4a2a8c c661c9ea0d535b059f455d70a6269af8ea1b784f f44dd680958f88143120fa66c310f891fa3de704 59ff286c6408c055411781d0a01b3b2e60122407 and d5b03e0220b8a32dfd60231615a21367ae69e75f

And thanks @nat-sch for the PR and sorry it took so long to review and merge.

PR #8457, opened 2024-12-09
Fix target blank diaspora links - Issue 7907

SEF-101:
Fixes the issue of diaspora links opening in a new tab as mentioned in Issue 7907.

This was not tested on mobile but was user-tested on an Ubuntu VM. This is for a class project


Flaburgan:
Hello @yusef800, thank you for your contribution.
If you want to test the mobile version of diaspora*, you can switch to it using the main menu: first reduce your browser width to simulate a phone, then tap on the burger menu, and tap on "Switch to touch-optimized mode".

PR #8458, opened 2024-12-09
Add UID and GID to Dockerfile to simplify building

kyle678:


denschub:
This PR makes no sense. The variables are set by the `script/diaspora-dev` script, which is the only way this image is meant to be used. 

If this is school work, I have to warn you that you will get blocked from this project - and many other projects - if you use this as a place to submit minimal-effort PRs.

PR #8460, opened 2025-01-10
Allow ImageMagick to use 1GiB of disk caches

skyschub:
This matches the ["websafe" upstream preset](https://github.com/ImageMagick/ImageMagick/blob/c7d070ec654573ddd9bc0356909a99fd1a7b07bd/config/policy-websafe.xml#L58), and our current limit is too low for high-noise images (like a scene full of snow).

### DIFF
--- a/config/imagemagick/policy.xml
+++ b/config/imagemagick/policy.xml
@@ -1,6 +1,6 @@
 <policymap>
   <policy domain="resource" name="time" value="30" />
-  <policy domain="resource" name="disk" value="256MiB"/>
+  <policy domain="resource" name="disk" value="1GiB"/>
   <policy domain="resource" name="map" value="256MiB"/>
   <policy domain="resource" name="memory" value="256MiB"/>
 


SuperTux88 (approved):
LGTM :+1: :cookie:

denschub:
I'm not even gonna mention this in the release notes, as I don't think this was a frequent enough issue to cause problems (otherwise, we'd have known before today).

SuperTux88:
I'd still mention it, it's a change, and  it affects production installations (not sure how many users would have reported that to their podmins or created a bugreport themselves, as it wasn't easy to debug). I'm otherwise fine with no changelog entry for small stuff that only affects CI or tests.

denschub:
> I'd still mention it, it's a change, and it affects production installations 

Okay, fine. :)

And for context for everyone else: I got a support request from a Geraspora user today who was unable to upload a specific image, and @SuperTux88 and me had a very fun, and very confusing, debugging session to figure this out. I originally guesstimated the limits, so this was kinda my fault. :)

PR #8461, opened 2025-01-12
Upload migration files from profile settings and start processing

tclaus:
This is a new PR which is rebased on latest dev and a squash of the fixing commits.

### DIFF
--- a/app/views/users/_edit.haml
+++ b/app/views/users/_edit.haml
@@ -198,8 +198,11 @@
     %hr
 
     .row
-      .col-md-6.account-data
+      .col-md-12
         %h3= t(".export_data")
+        %p= t(".export_data_subline")
+    .row
+      .col-md-6.account-data
         %h4= t("profile")
         .form-group
           - if current_user.exporting
@@ -215,6 +218,7 @@
             = link_to t(".request_export"), export_profile_user_path, method: :post,
               class: "btn btn-default"
 
+      .col-md-6
         %h4= t("javascripts.profile.photos")
         .form-group
           - if current_user.exporting_photos
@@ -229,7 +233,20 @@
             = link_to t(".request_export_photos"), export_photos_user_path, method: :post,
               class: "btn btn-default"
 
-      .col-md-6
+    .row
+      .col-md-12
+        %h3
+          = t(".import.headline", accountname: current_user.diaspora_handle)
+        %p
+          = t(".import.body")
+          %strong
+            = t(".import.warning")
+        .form-group
+          .btn.btn-primary{id: "import_account", data: {toggle: "modal", target: "#importAccountModal"}}
+            = t(".import.account")
+          = render "import_account_modal"
+    .row
+      .col-md-12
         %h3
           = t(".close_account_text")
         .form-group
--- a/app/views/users/_edit.haml
+++ b/app/views/users/_edit.haml
@@ -242,8 +242,11 @@
           %strong
             = t(".import.warning")
         .form-group
-          .btn.btn-primary{id: "import_account", data: {toggle: "modal", target: "#importAccountModal"}}
-            = t(".import.account")
+          - if current_user.account_migration_pending?
+            .import-in-progress= t(".import_in_progress")
+          - else
+            .btn.btn-primary{id: "import_account", data: {toggle: "modal", target: "#importAccountModal"}}
+              = t(".import.account")
           = render "import_account_modal"
     .row
       .col-md-12
--- a/config/locales/diaspora/en.yml
+++ b/config/locales/diaspora/en.yml
@@ -287,11 +287,10 @@ en:
   comments:
     create:
       error: "Failed to comment."
+      fail: "Comment creation has failed"
     new_comment:
       comment: "Comment"
       commenting: "Commenting..."
-    create:
-      fail: "Comment creation has failed"
     destroy:
       success: "Comment %{id} has been successfully deleted"
       fail: "Comment deletion has failed"
@@ -1318,13 +1317,22 @@ en:
       request_export: "Request my profile data"
       request_export_update: "Refresh my profile data"
       export_data: "Export data"
+      export_data_subline: "If you want to migrate this account to another pod, start by exporting your data and photos using the buttons below."
       export_in_progress: "We are currently processing your data. Please check back in a few moments."
       last_exported_html: "(Last updated %{timeago})"
       download_export_photos: "Download my photos"
       request_export_photos: "Request my photos"
       request_export_photos_update: "Refresh my photos"
       export_photos_in_progress: "We are currently processing your photos. Please check back in a few moments."
-
+      
+      import:
+        headline: "Import another account into %{accountname}"
+        body: "diaspora* allows you to migrate from one pod to another.
+               To do this, first export your data and photos from the pod you want to move from. Then, import those data and photo archives into this account.
+               Your old account on the other pod will be deleted once this process has completed."
+        warning: "There is no way back; that account can’t be restored if you change your mind."
+        account: "Import another account..."
+      
       close_account:
         dont_go: "Hey, please don’t go!"
         make_diaspora_better: "We’d love you to stay and help us make diaspora* better instead of leaving. If you really do want to leave, however, here’s what will happen next:"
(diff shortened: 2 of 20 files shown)


config/locales/diaspora/en.yml
```diff
@@ -1318,13 +1317,22 @@ en:
...
       export_photos_in_progress: "We are currently processing your photos. Please check back in a few moments."
-
+      
+      import:
+        headline: "Import another account into %{accountname}"
+        body: "diaspora* allows you to migrate from one pod to another.
+               To do this, first export your data and photos from the pod you want to move from. Then, import those data and photo archives into this account.
+               Your old account on the other pod will be deleted once this process has completed."
+        warning: "There is no way back; that account can’t be restored if you change your mind."
+        account: "Import another account..."
```
Flaburgan, line 1334: I think we should remove the `...` here.

config/locales/diaspora/en.yml
```diff
@@ -1339,6 +1347,28 @@ en:
...
+    
+    import_modal:
+      title: "Import another account"
+      subtext: "You are about to import and merge another account into %{accountname} <br>
+                Here is what will happen:"
+      list1: "Your profile (name, bio, birthday, gender, etc.) will be transferred from the imported account"
+      list2: "Your settings (theme, default post visibility, etc.) will be transferred from the imported account"
+      list3: "All your aspects and contacts from that account will be added to this account"
+      list4: "All posts and comments from that account will be added to this account"
+      closing_text: "The imported account and all its linked data will be deleted from the previous pod."
```
Flaburgan, line 1359: I guess an archive can be imported only once? Or what would happen if I import the same archive to two pods at the same time? The old pod will answer to the second import as if the user does not exist I assume? Maybe we should mention it here?

app/views/users/_edit.haml
```diff
@@ -229,7 +233,20 @@
...
+    .row
+      .col-md-12
+        %h3
+          = t(".import.headline", accountname: current_user.diaspora_handle)
+        %p
+          = t(".import.body")
+          %strong
+            = t(".import.warning")
+        .form-group
+          .btn.btn-primary{id: "import_account", data: {toggle: "modal", target: "#importAccountModal"}}
```
Flaburgan, line 245: I think after validating the modal and while the import is ongoing, this blue button "import another account" should be removed and replaced by a sentence "Account xxx@ccc is being imported, please come back in a few minutes" like we are doing for the export.

Flaburgan (changes requested):
Hey, so I just tried this locally, and it worked great, although there were no comments in my posts, apart from mine. Is this because I did the import offline, and they are supposed to be pulled? But then if the post is private or the old pod offline, this won't work. I guess it's "normal" to not include in the archive the comments made by others (although we could discuss that I guess), but then I think we should inform the user.
I also made a few suggestions to improve the UX. I'm going to read the code now.

config/locales/diaspora/en.yml
tclaus: What do you recommend? "Be careful", "After the import you can't start the process another time" ? Any ideas?

app/views/users/_edit.haml
tclaus: Button is now invisible for users with a pending / ongoing import.

Flaburgan:
Hello!
The button being disabled does not work locally, it's because I don't have sidekiq there right?

Can you also enable the message on the registration page by removing this if? https://github.com/diaspora/diaspora/blob/develop/app/views/registrations/_new.erb#L57

Thanks!

tclaus:
@Flaburgan It should work, when a migration is completed. (This worked locally) BTW I have disabled the If to show the message on registration page.

Flaburgan:
I just deployed this branch to diaspora-fr and tried to import my diasp.org account but it did not work. I submitted the dialog as I did locally, but the Import button is still available in my settings. I  don't know if the jobs was actually scheduled in sidekiq, I just realized that despite setting `log = log/sidekiq.log` in diaspora.toml, there actually is no log file for sidekiq there, only the production one.

Flaburgan:
Another remark: after the import, I have been spammed by email notifications of all my imported contacts, that they have started sharing with me. I think we don't need to have a notification about this, as the user is triggering it.

tclaus:
> I just deployed this branch to diaspora-fr and tried to import my diasp.org account but it did not work. I submitted the dialog as I did locally, and the green message "Import is scheduled" but the Import button is still available in my settings. I don't know if the jobs was actually scheduled in sidekiq, I just realized that despite setting `log = log/sidekiq.log` in diaspora.toml, there actually is no log file for sidekiq there, only the production one.
> 
> Edit: maybe no sidekiq log file is due to the fact that I start it with systemd. I checked them with journalctl but no mention of `import` there.
> 
For the records: $: `sudo journalctl -u diaspora-sidekiq` or `sudo journalctl -u sidekiq` shows the logs

Flaburgan:
@tclaus any chance you can have a look at the "Import" button not disappearing when closing th dialog? I wish I can tell people to come to diaspora-fr.org before diasp.org is closed.

tclaus:
To disable the button I have made some deeper investigation - first, the dialog is submitted - here Javascript may do the job to disable the button immediately,
Import file is uploaded and validation stars - here an enum (new column) in the user-table may indicate an ongoing validation check (status: “Import Validation”)
Then a job is triggering (status: “Import job pending”)
Then the job could set a “Import running” and at least “import completed” or “import failed”.

I think on a new column, with some statuses, to indicate that an import has been started and its current status.
IMHO this kind of status does not exist currently?

What did you think? Is this a way we can go?

Flaburgan:
>first, the dialog is submitted - here Javascript may do the job to disable the button immediately,

I think when you submit the form you should do a POST redirect GET. At the moment if the users press F5 after the dialog is closed, it is re-submitting the form, that should not happen. To reload the page entirely will avoid this problem, and allow us to display the correct status instead of the button.

Then about how to get this information, how is this done for the export buttons? Because everything is working nicely for those, so I think we should do the same.

Flaburgan:
So I just checked how it is done for exporting, and indeed we have two booleans on the user table, one [exporting](https://github.com/diaspora/diaspora/blob/develop/db/migrate/0000_create_schema.rb#L613) for the data and one [exporting_photos](https://github.com/diaspora/diaspora/blob/develop/db/migrate/0000_create_schema.rb#L617) for the pictures.
They are then [set to true when the export starts](https://github.com/diaspora/diaspora/blob/develop/app/models/user.rb#L315) and [back to false](https://github.com/diaspora/diaspora/blob/develop/app/models/user.rb#L325) when it's done. So I guess we should do the same for the import.
Will you have time to have a look in the coming days, or should I try (even if I never modified the schema or anything similar).

tclaus:
@Flaburgan 
> So I just checked how it is done for exporting, and indeed we have two booleans on the user table, one [exporting](https://github.com/diaspora/diaspora/blob/develop/db/migrate/0000_create_schema.rb#L613) for the data and one [exporting_photos](https://github.com/diaspora/diaspora/blob/develop/db/migrate/0000_create_schema.rb#L617) for the pictures. They are then [set to true when the export starts](https://github.com/diaspora/diaspora/blob/develop/app/models/user.rb#L315) and [back to false](https://github.com/diaspora/diaspora/blob/develop/app/models/user.rb#L325) when it's done. So I guess we should do the same for the import. Will you have time to have a look in the coming days, or should I try (even if I never modified the schema or anything similar).

Sounds OK for me - let's include one boolean for importing user data and one for photos. The details are already stored in the table 'account_migrations'. 

This evening I have some time to implement this.

db/migrate/20250327181326_add_importing_data_field_to_user.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddImportingDataFieldToUser < ActiveRecord::Migration[6.1]
+  def change
```
Flaburgan, line 4: This is failing:
```
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

wrong number of arguments (given 3, expected 2)
/home/andupara/.rvm/gems/ruby-3.1.4/gems/activerecord-6.1.7.8/lib/active_record/connection_adapters/abstract/schema_definitions.rb:603:in `column'
/home/andupara/diaspora/diaspora/db/migrate/20250327181326_add_importing_data_field_to_user.rb:6:in `block in change'
```

Flaburgan:
@tclaus hey, it looks like with your last commits the import is not launched anymore: I don't have the green confirmation popping up, and it does not look like it starts. Why did you do the POST with JS, I think it was working fine the way it was before, but in the rails controller at the end you should simply redirect to the user settings, instead of returning it directly.

tclaus:
> @tclaus hey, it looks like with your last commits the import is not launched anymore: I don't have the green confirmation popping up, and it does not look like it starts. Why did you do the POST with JS, I think it was working fine the way it was before, but in the rails controller at the end you should simply redirect to the user settings, instead of returning it directly.

Hi @Flaburgan , I thought the POST with JS in combinati0on was the only way to hide the dialog after uploading, but I was wrong. 
Changed this back and enabled the flash-message again. (Was a simple parameter in the flash statement that prevented the flash in the UI - I fixed this. 
I also fixed another failing test.

Also without a visible flag message the import should have been startet (I can see in the debugger a start of the import job every time)

PR #8464, opened 2025-04-13
Add reported user diaspora ID in /report view

Flaburgan:
I have a massive amount of spam on d-fr, and as I want to not only delete one post or comment but all of them and also block the user, I am using [Jhass's script](https://gist.github.com/jhass/9104029) to get rid of them.

However, that means that I have to copy paste the diaspora ID in the script so each time, I open the post, mouse over the author, select the diaspora ID and copy it, and I have to do it again for the next comment without knowing if it will be the same author.

This super smaller PR simply adds the diaspora handle next to the reported content, so I can know if it's the same user, and copy it without navigating anywhere else:

image

I put it on a new line so that a triple click will select the full diaspora ID immediately.


app/views/report/_reports.haml
```diff
@@ -21,6 +21,10 @@
               = report.text
         .panel-body
           .content
+            %strong
+              = t("report.author_label")
+            %p
+              = report.reported_author.diaspora_handle
```
denschub, line 27: `Reported_author` can be `nil` when the reported item has been deleted. This can happen without admin action, for example if the post owner deletes a spam-comment that has been reported, which would make this throw a NoMethodError.

`.try()` or `&.` with a fallback-label of "Item deleted"or something would be a good-enough solution here.

skyschub (changes requested):
Good idea.

I've also recently wondered if we should add the reported contents to the report email, which would persist after the content has been deleted (which is useful if a user deleted spam comments below their own comment). Do you have an opinion? :)

app/views/report/_reports.haml
SuperTux88: The `if` at line 11 ensures that the item is available here, so this should be fine as is. Otherwise it goes to the `else` after line 38 and doesn't show the author anymore.

denschub: Oh.

SuperTux88:
> I've also recently wondered if we should add the reported contents to the report email, which would persist after the content has been deleted (which is useful if a user deleted spam comments below their own comment). Do you have an opinion? :)

I wondered the same. I think the mail is missing additional content for privacy reasons (but I could be wrong?). However I don't know if the author alone would be enough, because if I only see the mail after the post/comment was already deleted, I wouldn't know why it was reported and why it was deleted. Yes, at the moment spam is likely, but could also be other reasons, and I personally wouldn't be comfortable banning a user just because of a report without any additional context. So the mail would need to contain also the content of the post/comment, and if it's spam, I'm not sure if it's a good idea to send spam content in mails, even if it's only to admins/moderators. Ok, there is also the reason, which is included in the mail, but that could also be wrong or not useful?

denschub:
I'd honestly be happy having the author + content + reporter + reason in the email. I know we don't dispatch private post/conversation over email for privacy reasons, but for report actions, having the context accessible outweighs those concerns, *especially* since this only applies to admin-emails anyway, and if an admin's email account is compromised, there are bigger concerns to be worried about.

The alternative solution would be to persist the content and some author information in the report itself, even if the post is deleted, and then truncating that report once an action has been taken. However, I'm not a fan of duplicating data, plus this would be a lot more effort than just throwing a bunch of fields into an email, which could be done relatively quickly.

Flaburgan:
I agree that privacy is not a good reason not to include the content in the email, however, I am worried that my e-mail provider would flag my pod as a spammer if we include the content, because the content is spam and so it will be recognized as such by my email provider. I don't want to harm the email reputation of my pod.

denschub:
You're already dispatching emails with the full content as part of the comment notification, since spammers tend to spam public posts.

denschub:
But anyway, I'm not blocking on this, this could be a follow-up discussion. If you fix the tests, this is good.

SuperTux88:
> You're already dispatching emails with the full content as part of the comment notification, since spammers tend to spam public posts.

Oh yeah, you're right :man_facepalming: :disappointed:

features/step_definitions/report_steps.rb
```diff
@@ -8,16 +8,18 @@
   step %(I should see "Reporting content" within "#reportModal")
 end
 
-And /^I should see a report by "([^"]*)" with reason "([^"]*)" on post "([^"]*)"$/ do |reporter, reason, content|
-  test_report("Post", User.find_by(email: reporter), reason, content)
+And /^I should see a report by "([^"]*)" with reason "([^"]*)" on post "([^"]*)" by "([^"]*)"$/ do |reporter, reason, content, author|
```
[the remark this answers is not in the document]
Flaburgan: I am not sure the way I broke the line is the expected way, please confirm.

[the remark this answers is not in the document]
denschub: damn, if only someone would invent a way to run `pronto` locally, eh? 😜

skyschub (approved):

denschub:
Merged as 452f46a548b773b9f31679e9eb68d72a5a2dbea5

Issue #8472, opened 2026-02-24
Comments fail to federate from Friendica instances


xanni:
This may be related to an error caused when Friendica reports support for the bluesky protocol, which is not in Diaspora's JSON Schema for the handshake and causes the handshake to fail.

PR #8473, opened 2026-02-24
Add bluesky to the JSON schema

xanni:
It appears that adding "bluesky" to the JSON Schema allows my Diaspora* pod to resume federating with Friendica pods.  Addresses issue #8472

### DIFF
--- a/vendor/nodeinfo/schemas/2.1.json
+++ b/vendor/nodeinfo/schemas/2.1.json
@@ -83,6 +83,7 @@
           "items": {
             "enum": [
               "atom1.0",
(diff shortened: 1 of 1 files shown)


denschub:
Yeah, this needs to be changed "upstream" in the nodeinfo repo. I put that in quotes because jhass is a diaspora\* maintainer, so... it's just half-upstream, but still.

Related, I'm not Jonne so I can't say for sure, but I'd expect this change to get rejected. The protocol's name is `atproto`, not `bluesky`, so Friendica made an oopsie. They should probably change that. :/

PR #8474, opened 2026-04-01
Fix file copy path for diaspora.toml setup

tclaus:


skyschub (approved):
Ah this is an unfortunate mistake. Good catch!

PR #8476, opened 2026-04-06
Update docker to use debian trixie as base image

SuperTux88:
Also get rid of the docker-compose warning about the version attribute.

And finally, fix `script/server` again, as that was missed in #8475 where the redis gem was removed because sidekiq now uses the redis-client gem.


skyschub (approved):
Yikes, sorry. :) Thank you!

🍪

PR #8478, opened 2026-04-10
Add parallel tests to diaspora-dev script

SuperTux88:
I was annoyed that tests run so slow locally, so I extended `diaspora-dev` script to allow to run all tests in parallel. This halves the test execution time for me and makes it easier to run tests locally more often.

The database config already had two integration databases configured, so I'm just using these to run cucumber (and I split desktop and mobile cukes, so they can also run in parallel). The `setup-tests` initializes all 3 databases now and runs migrations everywhere. It also uses separate redis databases for each test-job.

Rspec still uses the default test database and default redis connection.

By default it prints all logs from all tests in parallel. This isn't that useful when everything is green (besides that you see that tests are still running), but I think it's still useful, as you see immediately when something red shows up (especially if you broke everything and 50% of the tests are red, I don't need to wait for everything to complete). But there is a `--silent` option which only shows the current status and if they passed. The output of each test-job gets forwarded into a logfile, if you want to check it out separately.

You can still run the tests separately either without docker or with the diaspora-dev script like before. So this is just an additional thing you can use if you want to run all tests.

Also I fixed some other test-related things I found.

### DIFF
--- a/features/step_definitions/hovercard_steps.rb
+++ b/features/step_definitions/hovercard_steps.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 When(/^I activate the first hovercard$/) do
-  page.execute_script("$('.hovercardable').first().trigger('mouseenter');")
+  find(".hovercardable", match: :first).hover
 end
 
 Then(/^I should see a hovercard$/) do
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -58,6 +58,8 @@ Capybara.default_max_wait_time = 30
 ActionController::Base.allow_rescue = false
 
 DatabaseCleaner.strategy = :truncation
+# Ensure we always start from a clean DB state, even after interrupted runs.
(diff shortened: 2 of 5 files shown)


denschub (approved):
Smart, I like it. The test runner script feels a bit overcomplicated, it could be a lot simpler if you'd boil down on visuals, but since this is already done now, I'm not gonna complain about it. Thanks! 🍪

SuperTux88:
Yes, I wanted it to look fancy and might have overengineered it a bit :smile:

denschub:
Merged as d0fc726e5b078f24ca4190868c8e26818e8a0b5a 4e88c72f4d62b5f94ebd2eb0e4e79d72a27a77b9 90903b8938ebbec031a32001e5a5fafb54994b49 7eb7b48494cd8497b9923fedd7065446a6d9e0a6 aaand bb8db6e63dc0b1a4b31b9872eca819aca2409948. 🍪

Issue #7494, opened 2017-07-17
Automatically Cross-Post to Mastadon


ImSeth:
In the "Services" section of settings you allow posts to be automatically cross-posted to other social media like Twitter. Would you please add the ability to automatically cross-post to [Mastadon](https://mastodon.social/about). They are an open source social network similar to Diaspora.
https://github.com/tootsuite/mastodon

SuperTux88:
I don't know if we could add mastadon as service, because it would have different API endpoints (twitter has only one and podmins can add an API-key for twitter. For mastodon tootsuite/mastodon#780 or #7412 would probably be the better way. But mastodon wants to go the ActivityPub way and they don't want to add the diaspora protocol (which is already working and implemented by many other networks).

Flaburgan:
I think the best scenario "the networks are able to natively talk to each others" is not going to happen in a near future, so to be able to publish to mastodon would be nice. We can assume (like we did for the other services) that to be able to link to one account per user is enough. So I would like for example to link fla@diaspora-fr.org with fla@mastodon.social to be able to post there from diaspora*.
We can use the official mastodon API for that: https://github.com/tootsuite/mastodon-api

@gargron can you please point us to an application using the mastodon API somewhere? We need to be able to link to a Mastodon account located on any mastodon server. How do we deal with that?

Flaburgan:
@Gargron Up? Where should I go if I want some support for the mastodon API?

Gargron:
Oh sorry I did not see this thread. [This](https://github.com/tootsuite/mastodon-bridge) makes use of the mastodon-api gem, though only barely. There is also [this](https://github.com/renatolond/mastodon-twitter-poster)

DeadSuperHero:
I'm [currently playing around with this](https://github.com/DeadSuperHero/diaspora/commits/mastodon-crosspost), but so far have not made much real progress.

The only documentation I can currently find on adding new services is this [woefully outdated](https://github.com/diaspora/diaspora/wiki/Adding-your-favorite-service-to-Diaspora/3fc2c17fbc094970fae624f228407d2d6a40c830) document on the old Diaspora wiki.

__A few other caveats:__

1. Currently, the social sharing buttons use the Entypo font. Unfortunately, this font does not include the Mastodon logo (and neither does FontAwesome). Should I just use an image for this particular toggle?

[...]

PR #8479, opened 2026-04-10
Cleanup various stuff

SuperTux88:
This is mostly a cleanup PR of various old and deprecated stuff and some leftover stuff from a long time ago that should have already been removed years ago. :sweat_smile: 

This is in preparation for rails 7 upgrade to make that easier, but I already create this cleanup PR, as this already works with the current rails version.

The biggest changes are probably removing the twitter integration and google analytics, both of which don't work anymore since multiple years ... and since there were no bugreports, I assume nobody was using it anyway.

See the separate commits for the rest of the changes.


Gemfile
```diff
@@ -179,6 +179,8 @@ gem "multi_json", "1.15.0"
 # Logging
 
 gem "logging-rails", "0.6.0", require: "logging/rails"
+gem "syslog", "0.4.0"
```
denschub, line 182: I'm kinda surprised Pronto doesn't complain, but it looks like we usually column-align gems. I don't actually like that, so I'm not gonna complain about it, but I want to leave a comment so you can consider if you want to column-align or not.

db/migrate/20260409000000_remove_twitter_service.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class RemoveTwitterService < ActiveRecord::Migration[6.1]
+  class Service < ActiveRecord::Base
+  end
+
+  def change
+    reversible do |change|
```
denschub, line 8: Wouldn't it be better to treat this is a non-reversible migration to not even pretend we can undo it? Deleting services is lossy, and we can't recover from it except for restoring backups.

denschub (approved):
This all looks good to me, thanks for working on this. 🍪 

As we've discussed in chats, [I pushed a poll about usage of the remaining services](https://pod.diaspora.software/posts/9fafb1201722013f47f152540086c3e0). It's too early to draw conclusions, but it looks promising.

db/migrate/20260409000000_remove_twitter_service.rb
SuperTux88: Hmm, I don't know about that. 🤔 Yes, reversing it won't restore the data, but it would still allow to downgrade and restore the functionality (by adding the removed column below again, but it also won't restore the content of the column).

We also did the same thing when we removed the facebook service: https://github.com/diaspora/diaspora/blob/develop/db/migrate/20181004003638_remove_facebook.rb

denschub: Ah. Well. Uh. Okay, let's keep it, then. :)

denschub:
The poll now has 55 responses, 4 responders said they use the services. That's ~7%, and while that's honestly more than I anticipated, I still don't have huge concerns removing the services entirely to get rid of the cruft. How do you feel?

Flaburgan:
I wanted to add Mastodon to the services at some point, but not sure I will actually do it. https://github.com/diaspora/diaspora/issues/7494

denschub:
Setting aside the fact that this is unlikely to ever happen, it would be a massive refactor anyway as you couldn't hard-code some credentials in the pod configs, you'd need dynamic client registrations for whatever node the user wants to be on. That's effectively incompatible with the current architecture, so there's no harm in removing that code.

db/migrate/20260409000000_remove_twitter_service.rb
```diff
@@ -0,0 +1,14 @@
...
+      change.up { Service.where(type: "Services::Twitter").delete_all }
+    end
+
+    remove_column :posts, :tweet_id, :string
```
tclaus: Shouldn't the removed column be added in the UNKNOWN_PROPERTIES_NAMES in lib/diaspora/federation/receive.rb ?

tclaus (changes requested):

db/migrate/20260409000000_remove_twitter_service.rb
SuperTux88: No, that is for stuff that is federated (already known to the protocol library), but doesn't have a place in the database yet. `tweet_id` isn't known to the federation library and also not federated in any way.

tclaus: Maybe that its my own custom environment, but in a test, when receiving public posts, I run into this: 

```
PG::UndefinedColumn: ERROR: column "tweet_id" of relation "posts" does not exist LINE 1: ...bed_cache_id", "reshares_count", "interacted_at", "tweet_id"... ^

lib/diaspora/federation/receive.rb:213:in `block (2 levels) in status_message'
lib/diaspora/federation/receive.rb:208:in `block in status_message'
lib/diaspora/federation/receive.rb:343:in `try_load_existing_guid'
lib/diaspora/federation/receive.rb:200:in `status_message'
lib/diaspora/federation/receive.rb:9:in `public_send'
lib/diaspora/federation/receive.rb:9:in `perform'
config/initializers/diaspora_federation.rb:118:in `block (3 levels) in '
app/workers/receive_public.rb:11:in `block in perform'
app/workers/receive_base.rb:11:in `filter_errors_for_retry'
app/workers/receive_public.rb:10:in `perform'
```

who sends the column?

skyschub: Did you forget to restart the workers after applying the database migration?

tclaus: Possible, that I restarted the worker - But I won't give my live for this observation. If it works somewhere else, Its possible a problem on my side.

SuperTux88:
After the poll @denschub created, we decided to also remove the other services and all the other code related to services. And after a quick chat with @denschub we also decided to remove pubsubhubbub, this was never really advertised or used, I doubt anybody knows that it exists or how it even would work.

I decided to ignore a few pronto complaints for now, because they were already existing before and fixing them would mean bigger refactorings, which I don't want to do now (this PR is already big enough).

I also decided to make the migration irreversible now, as this was easier than recreating the whole service table again on the down-migration (similar to removing the chat).

There is one thing, I'm not really sure about: Previously there was a cog icon in the publisher, where the services where located:
image
There were informations about the connected services on the popup that opened there.

I replaced the cog with a question mark now, and the popup now only has a sentence about public posts and the atom feed remaining:
image
image

I don't know if we still need this popup (while the info about public posts is true, I doubt that anybody finds that)? And the atom feed doesn't exist anywhere else, but I also don't think users find that there? Maybe the atom feed would be better placed somewhere on the profile page? Maybe one of these two places?
image
(we have other icons there already on other peoples profiles, but nothing on the own profile)

But I'm also fine with merging the popup as is.

db/migrate/20260409000000_remove_twitter_service.rb
SuperTux88: This error can only happen, if you didn't restart sidekiq after running the migration, because sidekiq still thinks the column exists, but the DB disagrees. And while the stacktrace looks like it has something to do with federation, this happens purely on diaspora side with the diaspora entities, the federation entities don't know anything about a tweet_id and the diaspora entities now also don't have it anymore after you restarted everything.

If you already deployed this branch somewhere: You need to rollback the old migration (re-creating the column) before updating to the new version of the branch, because I removed the old migration, replacing it with one removing all the services (which now is irreversible).

skyschub:
> Maybe the atom feed would be better placed somewhere on the profile page?

We do correctly place a `<link rel="alternate">` on the profile, which, to me, feels like the right thing to do. No users would manually consume the atom feed, and all feed readers should have the ability to discover them that way, so I think that's good as-is. 

I'd be fine with removing the popup entirely, I think we have been pretty consistent on the meaning of "public".

db/migrate/20260409000000_remove_twitter_service.rb
skyschub: > If you already deployed this branch somewhere

Also.. just don't do that?! Unmerged PR are unmerged for a reason, and it's generally a really bad idea to deploy them anywhere but your local dev env where data loss is not a concern.

tclaus: Thanks for the warning - I followed the changes in the commits and reworked them (kind of cherry-pick) one by one in my code. Good to see some cleanup here.  Good work. 
"the dead of Diaspora is exaggerated.."

SuperTux88:
OK, I removed the popup. My feed-reader (thunderbird) doesn't auto-detect the correct feed-url from the profile url, but I think it's still fine for the context of this PR. The popup only showed my own feed url, if I'd want to subscribe to someones feed, I'd need their feed, not mine. Also the link was probably so hidden, that nobody found it.

I still think a visible feed-icon with the direct feed-url would be useful on the profile page ... but that can also be done in a separate PR, so keep this only for cleanup stuff. Also removing the modal fixed most of the pronto complaints :laughing:

Flaburgan:
>I still think a visible feed-icon with the direct feed-url would be useful on the profile page

I definitely agree with this, I even created an issue about it years ago. We can use it to track it instead of this cleaning MR. https://github.com/diaspora/diaspora/issues/4370

skyschub (approved):
delta r+ for everything I reviewed since the last time the stack was updated.

PR #8481, opened 2026-05-01
Migrate to Zeitwerk autoloader

SuperTux88:
This is the next step on the road to rails 7, as rails 7 requires the Zeitwerk autoloader.


skyschub:
+1,293 -1,336

oh my, oh dear.

SuperTux88:
Yes, that's why I'm splitting the whole thing into multiple PRs :innocent:

skyschub (approved):
Okay, sorry for the delay here. Reviewing this PR made me realize how far my backend development knowledge went... off the Rails. ... okay I'm sorry.

Anyway, this is kinda hard to review. The changes all make logical sense since most of the changes are just a giant "move into a different namespace" kind of changes. Evidently the code works since the tests are green. I did some tests locally and everything appears to be fine, too. I did scroll [over the official guide](https://guides.rubyonrails.org/v7.0/classic_to_zeitwerk_howto.html) and didn't notice anything that would hurt us. Resource usage seems fine too - if anything a bit lower - even in production mode with eager loading, so we're also not loading the entire universe into memory.

I think we should merge this to get us unstuck and ... hope nothing unexpected breaks in production?

Either way, thanks so much for your work here. 🧡

PR #8482, opened 2026-07-14
Use docker compose v2 in dev script

Flaburgan:
This PR drops the support for v1, is this acceptable or do we want to support both with something like 
```
  local compose=(docker compose)
  if ! docker compose version >/dev/null 2>&1; then
    compose=(docker-compose)
  fi
   # Check permissions of docker socket and use sudo if needed
   if [ -r "${docker_socket}" ] && [ -w "${docker_socket}" ]; then
    "${compose[@]}" "$@"
```


skyschub (approved):
I see no reason for a fallback, `docker compose` has been a command for... a while. I don't think this needs a changelog entry, so I'll just merge this as-is.

PR #8483, opened 2026-07-21
Migrated from Sidekiq to GoodJob

anibha88:


skyschub:
You should talk to projects first before submitting an architectural change like this. 

Setting aside the general Slop-vibes I get from this - this PR is a complete non-starter anyway. diaspora\* supports MySQL deployments, and even if we don't recommend new pods to use it, we have not removed MySQL/MariaDB support, we have no plans to do so in the near future, and we have even less of a plan on how to migrate existing pods in a reliable and predictable manner.

PR #8485, opened 2026-07-21
migrate from Sidekiq to GoodJob

anibha88:


skyschub:

If you open another Pull Request without even as much as acknowledging my initial comment, I will block you from the entire org permanently. Stop spamming.

PR #8486, opened 2026-07-21
migrate from Sidekiq to GoodJob

anibha88:


skyschub:

If you open another Pull Request without even as much as acknowledging my initial comment, I will block you from the entire org permanently. Stop spamming.

skyschub:
Ah, actually, no. I'll block you right away! I was willing to give you some slack for maybe being naive and having a good excuse for opening FOUR pull request without even saying hi, but in this PR, you left traces of your Claude-shit in. Your slop-contribution is not welcome here. Your account will be blocked permanently in the entire org, if you try to circumvent that block, I will report as much to GitHub directly.

PR #8488, opened 2026-08-10
Do not warn users they may lose data after submitting a comment and navigating

Flaburgan:
This one was annoying me for a very long time, I'm happy I finally took the time to fix it.

Step to reproduce:
- In the stream or in the SPV, comment on a post
- Navigate to a different page
- An alert is displayed saying you may lose data if you quit

The comment form was keeping the dirty flag from the are-you-sure jquery plugin. This PR properly cleans it when we receive a 200 from the server.


tclaus:
I tried to reproduce the issue, but when I am in the edit mode of a comment, the message appears. After I submit a comment, no warning appears when I navigate to a different page. 
Did I overlook anything? 
(Absolut fresh install of diaspora in a new folder, no left over JS or compiled assets) 

https://github.com/user-attachments/assets/b961e966-e79a-405f-a2d8-340ea2039717

Flaburgan:
I reproduce it very often on diaspora-fr, I also reproduced it locally but I don't right now. I think it is racy?

tclaus (approved):
Although I was unable to reproduce the original issue, the PR does add a clean-up action and cleans up the fields.

